PR #97, opened 2014-09-29
Key space isolation extension to ISubscriber

mwikstrom:
I think there should be a `WithKeyPrefix` extension for `ISubscriber` too. Here's one (with tests).

The proposed implementation keeps track of mapped handlers so that they can be used for unsubscribe invocations. 

It also keeps track of all subscriptions so that they can be unsubscribed from when calling `UnsubscribeAll`. With this approach the wrapper will do one `UnsubscribeAsync` per channel. 

As an alternative, it could do a prefixed `PUNSUBSCRIBE`. However, that might not be what the user wants, and it would require the extension to be more tightly integrated with the underlying `RedisSubscriber`, right?

Also: I made `ConnectionMultiplexer.CreateMultiplexer` internal so that it can be used from unit tests.


odinserj:
Hi @mgravell, what is the status of this PR? There is a confusion when we have to use keyspace isolation feature together with pub/sub – the `Publish` method considers the prefix, while `Subscribe` method is not. We should be very careful when dealing with subscriptions, adding prefix for each channel for each `ISubscriber` method call:

```csharp
Subscriber.Subscribe($"{_prefix}queue:{queue}:events", handler);
```

Looks like this PR is still pending due to custom subscription handler tracking. I can make a new PR without any additional features, just implement the `WithKeyPrefix` extension method for the `ISubscriber` interface (and some refactoring if needed).

I've tried to do implement it in my own project, but faced with a lot of internal methods, including all the methods that relate to prefix in the `RedisKey` class. So the good implementation should be here, in the official repository.

mgravell:
We've paused on this for ages. I don't think it is a generally useful and usable tweak. The earlier key part was great, and is fully merged and supported and used - but I don't think this bit has the same use case. I'm closing the PR for now, as there's no way it will physically merge due to drift, and I'm not sure we need it. I appreciate the effort and contribution.

PR #117, opened 2014-10-29
Added ability to write to a non-readonly slave

stepaside:


mgravell:
It feels like this should check for `config get slave-read-only`; I might try to add that later, but since `config get` is not always available, it is not unreasonable to allow it to be set explicitly too

mgravell:
Much obliged; thanks; sorry for delay

PR #127, opened 2014-11-17
Azure stale connection detection

### LINKED ISSUE #124: Azure timeouts/disconnection detection
I was able to isolate and reproduce strange StackExchange.Redis client behavior specific to Windows Azure Cloud services and Azure Redis Cache.

It seems that starting or stopping VM in the cloud service causes instant physical disconnection of all clients running on other VMs in the same cloud service. The problem is that it takes up to a minute to StackExchange.Redis to detect the disconnection. And all operations fails with timeout until the client will even try to reconnect.

Pattern seems to be the following:
1. Starting or stopping VM causes connection drop.
2. Client keeps writing to the network stream and getting no error on write until internal socket buffer will overflow. No data is read from the server, socket are still treated as connected (as there is no way to detect TCP disconnection until the write fail).
3. Requests fails with sync timeout. If a number of get requests >> put requests, that it takes up to a minute to fill up the socket write buffer (as it takes a number of "get key" to fill 8k buffer)
4. Socket finally detects buffer overflow and connection drop and throws `An existing connection was forcibly closed by the remote host` on either Read or Write. Client quickly reconnects to the server.

The exact re-connection message on my test system is:
    
    SocketFailure on 191.238.18.17:6379/Subscription, 
    input-buffer: 0, outstanding: 22, last-read: 79s ago,  <--- more than 1 min
    last-write: 1s ago, keep-alive: 60s, 
    pending: 0, state: ConnectedEstablished, <--- already reconnected
    last-heartbeat: 0s ago, last-mbeat: 0s ago, global: 0s ago

There is no complex sample code, just

    {
        SyncTimeout = 2000,
              
        AbortOnConnectFail = false,
        EndPoints = {{ RedisHost, RedisPort }},
        Password = "pass"
    };

    ConnectionMultiplexer redis = ConnectionMultiplexer.Connect(options); // 
    redis.ConnectionFailed += redis_ConnectionFailed; // print e.FailureType and e.Exception
    redis.GetDatabase(1).Ping();
    Console.ReadLine();

On a live system it results in 20-30 sec cache unavailability and a lot of timeouts logged.
Is there any way to force fast reconnect in that case? 

Tracking last-read - last-write diff and forcing disconnection when it is over sync timeout is the only idea I got so far. Not sure if it will be compatible with pub/sub.

PashaPash:
Fixed #124 - added check to detect azure-specific disconnection type. 
I used a SyncTimeout as a read-write threshold. Maybe this check needs a separate timeout settings?

### DIFF
--- a/StackExchange.Redis.Tests/ConnectingFailDetection.cs
+++ b/StackExchange.Redis.Tests/ConnectingFailDetection.cs
@@ -8,7 +8,7 @@ namespace StackExchange.Redis.Tests
     public class ConnectingFailDetection : TestBase
     {
 #if DEBUG
-        [TestCase]
+        [Test]
         public void FastNoticesFailOnConnectingSync()
         {
             try
@@ -34,16 +34,15 @@ namespace StackExchange.Redis.Tests
 
                     Assert.IsTrue(muxer.IsConnected);
                 }
-
-                ClearAmbientFailures();
             }
             finally 
             {
                 SocketManager.ConnectCompletionType = CompletionType.Any;
+                ClearAmbientFailures();
             }
         }
 
-        [TestCase]
+        [Test]
         public void ConnectsWhenBeginConnectCompletesSynchronously()
         {
             try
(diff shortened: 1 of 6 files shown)


mgravell:
Thanks

PR #129, opened 2014-11-26
Fixed connect timeout handling in ConnectionMultiplexer.Connect

PashaPash:
- fixed hardcoded completion type in `SocketManager.BeginConnect` (forced Sync completion type in #113  by mistake). Both successful and failed connections should now be handled as fast as in 1.0.333.
  - connect timeout exception type changed to `RedisConnectionException` (for consistency with non/timeout connection errors) .
  - forced async execution of `ReconfigureAsync` call from Connect (although `ReconfigureAsync` returns Task is wasn't executed asynchronously because `Connect` was not marked as async. So `task.Wait` call in Connect was always executed after the `ReconfigureAsync` completion, making that timeout ineffective).

### DIFF
--- a/StackExchange.Redis.Tests/ConnectToUnexistingHost.cs
+++ b/StackExchange.Redis.Tests/ConnectToUnexistingHost.cs
@@ -1,0 +1,52 @@
+﻿using NUnit.Framework;
+using System;
+using System.Diagnostics;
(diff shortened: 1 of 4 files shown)


mgravell:
Much obliged; sorry for delay

Issue #111, opened 2014-10-22
HashSet with RedisValue.EmptyString do not work as ""


YoungjaeKim:
When we do `HashSet`,

``` CSharp
HashEntry[] hashEntries =
{
    new HashEntry("Id", 1),
    new HashEntry("Email", ""),
};
RedisDatabase.HashSet("Key1", hashEntries);
```

In line 4, 
- **working**: `new HashEntry("Email", ""),`
- **ArgumentNullException**: `new HashEntry("Email", RedisValue.EmptyString),`

Why do they work differently?

Seems like `RedisValue.EmptyString` is treated as `null` value as we see the stacktrace.
Related stacktrace is as below.

>    Location: StackExchange.Redis.RedisValue.AssertNotNull() File c:\TeamCity\buildAgent\work\58bc9a6df18a3782\StackExchange.Redis\StackExchange\Redis\RedisValue.cs: line 237
>    Location: StackExchange.Redis.RedisDatabase.GetHashSetMessage(RedisKey key, HashEntry[] hashFields, CommandFlags flags) File c:\TeamCity\buildAgent\work\58bc9a6df18a3782\StackExchange.Redis\StackExchange\Redis\RedisDatabase.cs: line 1567
>    Location: StackExchange.Redis.RedisDatabase.HashSet(RedisKey key, HashEntry[] hashFields, CommandFlags flags) File c:\TeamCity\buildAgent\work\58bc9a6df18a3782\StackExchange.Redis\StackExchange\Redis\RedisDatabase.cs: line 216

NickCraver:
I'm not sure if this is intentional or not, but *if* we change the behavior, it'd need to be in a major version release - relating to #528.

PR #133, opened 2014-12-07
Issue #111: HashSet with RedisValue.EmptyString do not work as ""

luberg:
Simple reorder of static field init order fixes this issue.


mgravell:
Wow; static re-order; that is embarrassing for me! Thanks, and sorry for delay

PR #135, opened 2014-12-10
StackExchange.Redis to consider separately Response Array as either NULL  or an Empty Array

deepakverma:
When multiple clients are trying to modify the same key in a transaction, Redis server returns null response *-1\r\n causing it to throw an “Unexpected response to EXEC: MultiBulk: 0” exception during transaction.ExecuteAsync()."
The fix is to consider *-1\r\n as a null array instead of empty array.


mgravell:
Again, I'm very reluctant to merge this without understanding the offending scenario, and ideally: a test that shows the problem. Can you clarify? I'm not saying it isn't valid or correct: just that it warrants a demonstrable test scenario. I don't have a repro.

cschertzglympse:
I dont understand why anyone would say that dont have a repro for this.   The repro is in the initial issue that was filed.   I copied that code and was able to reproduce without a problem.  Doing this because we are hitting the same issue and wanted to see if the pull request would solve the problem.   Seems to fix this issue

mgravell:
Merged; thanks

PR #142, opened 2015-01-01
expose Prepend and Append methods on RedisKey, so we can avoid some string concats in common cases

kevin-montrose:
We've got some code that routinely concats an arbitrary string to a fixed prefix.  It'd be nice to not have to allocate a whole new `string` (or `byte[]`, or whatever) when we're just gonna turn this into a `RedisKey` and be done with it.

This commit exposes the already existing [`WithPrefix`](https://github.com/StackExchange/StackExchange.Redis/blob/d994b7ad57a76586487cd2d460d55cb850498448/StackExchange.Redis/StackExchange/Redis/RedisKey.cs#L253) method on `RedisKey` in separate `Prepend` and `Append` wrapper methods.

### DIFF
--- a/StackExchange.Redis.Tests/Keys.cs
+++ b/StackExchange.Redis.Tests/Keys.cs
@@ -66,5 +66,40 @@ namespace StackExchange.Redis.Tests
 
             }
         }
+
+        [Test]
+        public void PrependAppend()
+        {
+            {
+                // simple
+                RedisKey key = "world";
+                var ret = key.Prepend("hello");
+                Assert.AreEqual("helloworld", (string)ret);
+            }
+
+            {
+                RedisKey key1 = "world";
+                RedisKey key2 = Encoding.UTF8.GetBytes("hello");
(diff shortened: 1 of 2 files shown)


mgravell:
Are you aware that there is a mechanism for obtaining a database connection
with an automatic inbuilt key prefix? It then does the entire thing without
combines. Would that be useful? Or is this related but complimentary?

mgravell:
I should note: the prefix is on the IDatabase, not the multiplexer - so you
can have multiple different prefixes (or none) on the same underlying
connection.

kevin-montrose:
@mgravell I consider manipulating the key a little clearer but I *could* make the existing stuff work I suppose...  Might be a tad confusing, especially if we start storing more kinds of data against some of our (at the moment) pretty much homogeneous redis instances, but I can isolate it for now.

I still like this PR, but it's your call.

mgravell:
Nah, you're grand; thanks

PR #145, opened 2015-01-09
Adding an environ var (SERedis_ClientCertStorageFlags) for configuring which cert store to check.

billycc:


mgravell:
Thanks; good stuff

PR #149, opened 2015-01-23
KeyRename return boolean

shiweichuan:
it is ok for When.Always but it must be boolean when using When.NotExists, so change to ResultProcessor.Boolean


mgravell:
The problem is: it doesn't, as far as I can tell. It returns +OK or -ERR no such key

Can you be more specific with an example of where this would help?

shiweichuan:
http://redis.io/commands/rename
http://redis.io/commands/renamenx
The return value is different for rename and renamenx.
rename returns +OK or -ERR but renameex returns 1 if key was renamed to newkey or 0 if newkey already exists.

mgravell:
Ah, gotcha; sorry, my bad - I misread; yes, that looks good, then - thanks

manual merge ("We can’t automatically merge this pull request.") - working on it

mgravell:
Git was very unhappy with the merge (100% changed, regardless of how I treated line-endings), so I took your branch "forward" to master and reapplied - retained your ownership, at least!

Issue #20, opened 2014-04-17
NotImplementedException under Mono on OSX when connecting


returnString:
<pre>at System.Net.EndPoint.Serialize () [0x00000] in /private/tmp/source/bockbuild-mono-3.2.6/profiles/mono-mac-xamarin/build-root/mono-3.2.6/mcs/class/System/System.Net/EndPoint.cs:51
  at System.Net.Sockets.Socket.BeginConnect (System.Net.EndPoint end_point, System.AsyncCallback callback, System.Object state) [0x000fa] in /private/tmp/source/bockbuild-mono-3.2.6/profiles/mono-mac-xamarin/build-root/mono-3.2.6/mcs/class/System/System.Net.Sockets/Socket_2_1.cs:1499 
  at StackExchange.Redis.SocketManager.BeginConnect (System.Net.EndPoint endpoint, ISocketCallback callback) [0x00000] in <filename unknown>:0 
  at StackExchange.Redis.PhysicalConnection..ctor (StackExchange.Redis.PhysicalBridge bridge) [0x00000] in <filename unknown>:0 
  at StackExchange.Redis.PhysicalBridge.GetConnection () [0x00000] in <filename unknown>:0</pre>

Got this when calling `var multiplexer = ConnectionMultiplexer.Connect("localhost");`

_mono --version
Mono JIT compiler version 3.2.6 ((no/9b58377 Thu Jan 16 17:49:56 EST 2014)_

Is this tested under Mono and therefore something is wrong on my end, or does it need a patch to work? :)

olviko:
*The latter can be avoided; the former is more problematic - but ResolveDns already exists and fixes it.*
ResolveDns=true also caches server IP addresses, which can change in certain failure scenarios.

PR #155, opened 2015-01-30
A fix for https://github.com/StackExchange/StackExchange.Redis/issues/20

olviko:
Fixing NotImplementedException under Mono while preserving dynamic host name resolution.


mgravell:
Thanks, looks good

PR #156, opened 2015-01-30
Prevent EVAL to SCRIPT conversion if it is not supported 

bgertsberg:
StackExchange.Redis "optimizes" all Lua scripts by preloading them and executing by hash. This fix allows to execute Lua scripts as is through the EVAL when SCRIPT command is not supported. Without this fix you can't run any Lua script with twemproxy.


mgravell:
Sounds good, thanks; it looks like CommandMap already knows that twemproxy doesn't allow `script`, so this should work great. Thannks.

PR #170, opened 2015-02-27
Named script parameters squashed

kevin-montrose:
This is a squashed version (+ 1 merge commit to catch up to master) of the [named-script-parameters](https://github.com/kevin-montrose/StackExchange.Redis/tree/named-script-parameters) branch.  It's a big one.

Adds the [`LuaScript` and `LoadedLuaScript` public classes](https://github.com/kevin-montrose/StackExchange.Redis/blob/named-script-parameters-squashed/StackExchange.Redis/StackExchange/Redis/LuaScript.cs), as well as a number of convenience methods to [`IServer`](https://github.com/kevin-montrose/StackExchange.Redis/blob/named-script-parameters-squashed/StackExchange.Redis/StackExchange/Redis/IServer.cs#L320), [`IDatabase`](https://github.com/kevin-montrose/StackExchange.Redis/blob/named-script-parameters-squashed/StackExchange.Redis/StackExchange/Redis/IDatabase.cs#L474), and [`IDatabaseAsync`](https://github.com/kevin-montrose/StackExchange.Redis/blob/named-script-parameters-squashed/StackExchange.Redis/StackExchange/Redis/IDatabaseAsync.cs#L446).

The gist of usage:
```
  IDatabase db = ...
  var prepared = LuaScript.Prepare("redis.call('set', @key, @value)");
  prepared.Evaluate(db, new { key= (RedisKey)"my-key", value=123 });
  // alternatively: db.ScriptEvaluate(prepared, new { key= (RedisKey)"my-key", value=123 });
```

The Lua script actually sent to Redis in the above example is:
```
redis.call('set', ARGV[1], ARGV[2])
```

While the generated Lua script will never **use** the KEYS variable, the properties on the parameters object which are of type `RedisKey` are extracted and sent up as the `KEYS` collection.

`LuaScript` objects are cached, `LoadedLuaScript` objects are not.  When `LuaScript` are collected, they should remove themselves from cache; there is also an explicit [`PurgeCache()` method.](https://github.com/kevin-montrose/StackExchange.Redis/blob/named-script-parameters-squashed/StackExchange.Redis/StackExchange/Redis/LuaScript.cs#L78).


mgravell:
Thanks Monty

JasonPunyon:
YAY!

mgravell:
@Jason this should already be available on the SE internal repo; I will
push to the public feed once we've dogfooded it

PR #184, opened 2015-04-03
Add ThreadPool statistics to Timeout message to help debug IOCP issues

JonCole:
When clients have a burst of IO operations, the thread pool will scale slowly after it hits the minIoThreads.  This can lead to timeouts on clients that are not related to any issues on the network or on the server.  Logging these details will help identify such issues without having to use a debugger to see the state of the thread pool.


mgravell:
Sounds excellent. Reviewing

mgravell:
Thanks - looks great; much obliged

JonCole:
Awesome!  Thanks for merging so quickly.  When do you expect this to be pushed to NuGet.org?

PR #196, opened 2015-04-22
Added a way to specify the default database in the connection string

freakingawesome:
I believe there is value in being able to specify the default database ID within the connection string. It falls back to zero if left unspecified.

### DIFF
--- a/StackExchange.Redis.Tests/Issues/DefaultDatabase.cs
+++ b/StackExchange.Redis.Tests/Issues/DefaultDatabase.cs
@@ -1,0 +1,63 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using NUnit.Framework;
+using StackExchange.Redis;
+
+namespace StackExchange.Redis.Tests.Issues
+{
+	[TestFixture]
+	public class DefaultDatabase : TestBase
+	{
+		[Test]
+		public void UnspecifiedDbId_ReturnsNull()
+		{
+			var config = ConfigurationOptions.Parse("localhost");
+			Assert.IsNull(config.DefaultDatabase);
+		}
+
+		[Test]
+		public void SpecifiedDbId_ReturnsExpected()
+		{
+			var config = ConfigurationOptions.Parse("localhost,defaultDatabase=3");
+			Assert.AreEqual(3, config.DefaultDatabase);
+		}
+
+        [Test]
+        public void ConfigurationOptions_UnspecifiedDefaultDb()
+        {
+            var log = new StringWriter();
+            try
+            {
+                using (var conn = ConnectionMultiplexer.Connect(string.Format("{0}:{1}", PrimaryServer, PrimaryPort), log)) {
+                    var db = conn.GetDatabase();
+                    Assert.AreEqual(0, db.Database);
+                }
+            }
+            finally
+            {
+                Console.WriteLine(log);
+            }
+        }
+
+        [Test]
+        public void ConfigurationOptions_SpecifiedDefaultDb()
+        {
+            var log = new StringWriter();
+            try
+            {
+                using (var conn = ConnectionMultiplexer.Connect(string.Format("{0}:{1},defaultDatabase=3", PrimaryServer, PrimaryPort), log)) {
+                    var db = conn.GetDatabase();
+                    Assert.AreEqual(3, db.Database);
+                }
+            }
+            finally
+            {
+                Console.WriteLine(log);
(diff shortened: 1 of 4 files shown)


freakingawesome:
Altering the signature of `GetDatabase()` is of course, less than ideal. If that doesn't fly, perhaps we could add an additional method, `GetDefaultDatabase()` as shown here: https://gist.github.com/freakingawesome/59d53f27cb5cd038d0f9

mgravell:
Changing a method signature like that is a non-starter - we can't do that: it would break compatibility and break builds with `MethodMissingException`. Frankly, I'm not sure why this would be necessary - can't you just put this logic in where-ever calls `GetDatabase()` ?

freakingawesome:
While we can change databases already in the GetDatabase() call in code, I was hoping to make it part of the connection string to avoid having to specify the database elsewhere in code and configuration.

My case is this: I'm splitting my instance into two Redis databases: one for cache data that I can wipe clean using `FLUSHDB` and another for session data that won't be affected by the cache `FLUSHDB`. My dependency injection model would be a little cleaner if I could just have two different connection strings rather than also having to inject a DatabaseID property.

Yea, my first pull request changing `GetDatabase()` was shortsighted because of backwards compatibility. The alternative, adding a [`GetDefaultDatabase()`](https://gist.github.com/freakingawesome/59d53f27cb5cd038d0f9) method, would fix that but seems likely to cause confusion in the long run. Perhaps in a far-future version that already introduces breaking changes we can include this?

MichaCo:
The `ConfigurationOptions` are meant to configure the connection to the Redis server. You can/should use the same connection and same multiplexer if you use one Redis server, right? 
And in addition, opening the database is separated from the initial connect for a good reason, so that you reuse the connection.

Your suggestion would actually work against that pattern. So that's why I personally don't like the idea to add configuration options which are not related to the connection, to that `ConfigurationOptions`...

In your scenario you would create two Multiplexer instances just to use the default database of both configurations, right? That would be odd I'd say.

Can't you just add some configuration to your code which lets you configure the database the multiplexer should use and inject that via your DI model?

freakingawesome:
Yes, two different connection strings would imply the use of two different multiplexers. I'm also using the Microsoft Session State and Output Cache providers, which themselves (I'm assuming) spin up their own multiplexer, so that's where I got the idea of going against canon.

> Can't you just add some configuration to your code which lets you configure the database the multiplexer should use and inject that via your DI model?

Certainly, we could, but coming from my past experiences with connection strings, it seemed odd that changing a default database wasn't an option, especially when configuring against those Microsoft providers. I guess I just like that kind of configuration all in one place.

I still like the idea of being able to override the magic number of zero from within a connection string, but I understand the arguments against it too.

mgravell:
Interestingly, *changing the default value* (but not changing the parameter
type) is not (IIRC) a breaking change - it would, however, not take effect
until code referencing the library was recompiled. One option might be to
change the default database to -1, and special-case this to mean: use the
default database, which itself defaults to zero.

Thoughts?

freakingawesome:
D'oh. Good call. I've updated the PR to reflect this: 4d94274f4fcb98de03d4090ddcf14d5333b2d8c0

Now the method signature remains unchanged. The default value of `db` is now -1.

mgravell:
K; I'll need to check the DLL compatibility, but I'm 95%+ sure that is OK -
and I don't have anything the principle.

clstopher:
Just thinking about this (no code written/testing done), but it seems like changing the default value of the parameter is a slightly breaking change.  With no recompile of the calling assembly, the behavior will remain the same but now a connection string parameter will be ignored (because the calling code is still passing zero).  After recompiling the calling code, a value in the connection string will suddenly start to be used.

This seems like it could also cause some minor problems (or at least a touch of confusion) in cases where someone is using a package that depends on StackExchange.Redis while also using StackExchange.Redis in their own project.

I'm not sure if this is a realistic concern or not (depends on how many packages depend on StackExchange.Redis and also how often people call the method with default parameters, etc).  Unfortunately I don't have a good solution to offer as an alternative -- My first thought would be to keep the default parameter value at 0 and use the connection string if it is, but that could result in an inaccessible "0" database (if a connection string includes a different value, then there is no way to get there since explicitly passing 0 will use the connection string value).

mgravell:
That is pretty much the case whenever a configuration option is added or
tweaked in some way, if the "ignore unknown configuration options" flag is
set (which is isn't by default). Without that flag, it will raise a loud
error.

I understand your concern, but this isn't a problem IMO.

mgravell:
Thank you; merged. I really like the idea here.

PR #209, opened 2015-05-21
Add validation to StringGet and StringGetAsync

qcz:
`StringGet` and `StringGetAsync` does not check that the passed `keys` array is null or empty.
Empty arrays cause RedisServerException (ERR wrong number of arguments for 'mget' command), while null arrays cause NullReferenceException.

This PR fixes these problems.


mgravell:
Thanks; all looks good

PR #216, opened 2015-05-26
Faster byte array comparison

breyed:
This algorithm runs about 40% faster than the one it replaces.

### DIFF
--- a/StackExchange.Redis/StackExchange/Redis/RedisValue.cs
+++ b/StackExchange.Redis/StackExchange/Redis/RedisValue.cs
@@ -163,26 +163,20 @@ namespace StackExchange.Redis
 
         internal static unsafe bool Equals(byte[] x, byte[] y)
         {
-            if ((object)x == (object)y) return true; // ref equals
-            if (x == null || y == null) return false;
-            int len = x.Length;
-            if (len != y.Length) return false;
-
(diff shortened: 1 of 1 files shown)


breyed:
This is a edit I made directly from the GitHub website. I don't know why the diff is showing massive changes. The only real change is to replace Equals with a different algorithm, copied directly from [this StackOverflow answer](http://stackoverflow.com/a/8808245/145173), with a couple variable name changes to match the original method parameter names. I tried the edit twice and got the same diffing problem each time.

mgravell:
Github is showing 100% change there (presumably line endings). Can you
perhaps give a description of the main changes? Is it mainly the `unsafe
bool Equals(byte[] x, byte[] y)` method?

mgravell:
Ah, sorry - your second post arrived while I was reading. Thanks - reading
now.

mgravell:
This seems to change the logic for comparing two null arrays, which were previously considered equal - and in doing so it also omits the (essentially free) "the same array compared to itself" scenario. The latter is a nice-to-have, but the former needs fixing, IMO. I want to run my own speed comparison, out of curiosity; but yes, it is a nice way of handling those pesky last few bytes after the 64-bit comparable chunks.

mgravell:
Concern not addressed in a year; closing

PR #218, opened 2015-05-27
Default SslHost in clustering mode

hrishi18pathak:
In clustering mode currently, ClusterConfiguration for every server endpoint is built using the output of "CLUSTER NODES" command which consists of IP addresses of the nodes inside the cluster (SetClusterConfiguration and UpdateClusterRange methods)

However, when we try to authenticate the client to the ssl endpoint, inside SocketMode ISocketCallback.Connected, using the ssl.AuthenticateAsClient method it throws an exception
"The remote certificate is invalid according to the validation procedure” since we are trying to authenticate using the IP address as opposed to the Dns Host name.
An existing workaround this problem is to configure "SslHost" explictly inside the connection string or specify it inside ConfigurationOptions.
However if the users of the client do not configure this explicitly, it is a reasonable fallback behavior to infer this information from the specified server endpoints. The fallback behavior will be as follows:
If all the endpoints inside the configuration are dns endpoints and have identical hostnames, the SslHost is the dns host name for one for one of these endpoints.

### DIFF
--- a/StackExchange.Redis.Tests/SSL.cs
+++ b/StackExchange.Redis.Tests/SSL.cs
@@ -271,5 +271,24 @@ namespace StackExchange.Redis.Tests
 
         }
 
+        [Test]
+        public void SSLHostInferredFromEndpoints() {
+            var options = new ConfigurationOptions() {
+                EndPoints = { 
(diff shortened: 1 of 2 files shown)


mgravell:
Thanks

PR #222, opened 2015-06-01
Update Configuration.md

sporty81:
Suggested edit to your doc to show how to connect to Azure with the password


mgravell:
Thanks, appreciated

MichaCo:
Just a comment on that: When you configure something like this for Azure, it is advisable to use app settings and use the Azure portal to set them. This way you do not have to store passwords and that kind of security relevant information in your source repository. 
Imagine you maintain your site on github. You most likely do not want to have the passwords publicly available ;)

Just for reference - A while ago I wrote [an article which also mentions how to setup those settings](http://mconrad.azurewebsites.net/Documentation/Index/cachemanager_backed_todo_web_app#hosting).

PR #234, opened 2015-07-13
Add ClientName to output for timeout messages

JonCole:


mgravell:
pushed in nuget 481 build

PR #236, opened 2015-07-14
Updating cluster range for a give node iff the node endpoint is compatible with the user specified configuration

hrishi18pathak:
Issue: If a user has specified endpoints in the multiplexer configuration using
 Dns endpoints, Stackexchange.redis returns the ConnectionMultiplexer object successfully 
 after connecting to all the nodes in the connection string (using their host names). 
 Also, before returning the multiplexer object it updates the mapping of nodes to their clusterslots 
 and this mapping is of the format “DNS hostname:port” : “clusterSlotNumber”.
 However at the same time, the client also issues “CLUSTER NODES” command to each of the nodes in the above list  (to determine master and slave configuration) and re-updates the cluster mapping with the output of the clusternodes command  which now is going to contain IP addresses (redis-server itself does not understand host names).  So the cluster mapping is now updated to the format: “IP Address” : “clusterSlotNumber”. If the StackExchange.Redis has not been able to connect to all the nodes using their IP addresses by the time the first command (GET,SET)  is issued to the cluster, it results in failure to connect to that particular node resulting in the “MOVED ERROR” 
 (since it tries to hit a random node in the list subsequently)

Resolution: Do not update cluster range if the node configuration is incompatible with the multiplexer  endpoints configuration.

### DIFF
--- a/StackExchange.Redis/StackExchange/Redis/ConfigurationOptions.cs
+++ b/StackExchange.Redis/StackExchange/Redis/ConfigurationOptions.cs
@@ -391,6 +391,10 @@ namespace StackExchange.Redis
             return false;
         }
 
+        internal bool HasAllDnsEndPoints() {
+            return !endpoints.Any(ep => !(ep is DnsEndPoint));
+        }
+
 #pragma warning disable 1998 // NET40 is sync, not async, currently
         internal async Task ResolveEndPointsAsync(ConnectionMultiplexer multiplexer, TextWriter log)
         {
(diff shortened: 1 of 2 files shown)


hrishi18pathak:
I have modified this pull request to make this fix more comprehensive. The fix: Add an endpoint to the 'serversnapshot' of the multiplexer if its compatible with the user specified configuration!

debbiswal:
Hi @hrishi18pathak,
I am still getting the 'MOVED' exception after taking the latest code from your branch.
In my scenario 7001,7002,7003 are  masters and 7004(master=7001),7005(master=7002),7006(master=7001) are slaves in cluster. Then I shutdown 7001 and 7004 is automatically promoted to master. Then I started 7001.So now 7001 is slave of 7004.
Below I have attached the screenshot of 'cluster nodes' command in Linux box.

cluster

ConnectionMultiplexer in wach window :
cluster1

PR #243, opened 2015-07-23
Redis command interception feature and basic IoC to implement further extensibility points

mfidemraizer:
Hi,

I've been digging around on how to implement Redis value compression and decompression in the application layer and I've found that, while it might work fine, it would be great if I could implement this at library level.

IMHO, I believe features like compression fit better at library level because I can leave existing code *as is* and it encourages a good separation of concerns, because an app or service code doesn't need to know if the data is being compressed or not in the data layer.

Thus, I cloned *StackExchange.Redis* repository and tried to implement compression and I didn't find a good point where I could implement compression until I realized that these points are there already, but **why I need to mess up *StackExchange.Redis* with compression if I can implement a generic extensibility point to externally alter Redis command execution pipeline?**.

And then I thought about implementing **interception**, and I designed a very simplistic interface called `IRedisCommandHandler` which can provide code to execute before a command is about to be executed and after it has been already executed. Even more: **why I don't implement a way that allow third-party extensions (`IRedisCommandHandler` implementations) gain the ability to alter what's going to be written to and received from Redis?**

So this is what I did. 

Since I believe that command handlers may be the first library component to allow *extensibility*, I thought that the best way to implement this should be using *inversion of control*, and then I implemented a very basic inversion of control container centralized in the `RedisServiceFactory` class, which provides service registration and unregistration plus multiple implementation retrieval to some given service interface.

Furthermore, `RedisServiceFactory` only allows supported services by StackExchange.Redis (for now, *command handlers*).

In summary, now you can provide *n* `IRedisCommandHandler` implementations that will be executing during every Redis command request/response. I've also provided a white list system to activate specific Redis commands for interception.

At the end of the day, I need to merge my contribution with official branch because I want to provide a `IRedisCommandHandler` implementation to compress/decompress Redis values outside the library, but be able to *do stuff* inside the library externally without affecting the *single responsibility principle* that should follow a library like *StackExchange.Redis*: it's a Redis .NET driver, and it shouldn't provide things like compression as part of its codebase, but it should provide a way to soft-plug a compressor as part of its execution pipeline **Why not?** ;)

Finally, I've provided tests, XML doc and markdown documentation to my additions. *Anyway, I don't know if you want specific tests to prove that command handlers won't produce bugs when they're integrated in an actual Redis command execution, we can discuss this point... For now, I've implemented unit tests rather than integration tests. If you check the code you'll find that command interception is very simple and I doubt it could produce errors (I've already executed all tests and test run showed no errors related to command interception...)*

Thanks in advance, Marc, and I hope my contribution fits your standards, and if you want any modification, improvement or whatever prior to get it merged, I'm absolutely open-minded to discuss  them!


mfidemraizer:
I'm going to pull request again because tests were failing.

Issue #137, opened 2014-12-11
Subscriptions to slaves not allowed


piccit:
In `Subscription.SubscribeToServer`, `multiplexer.SelectServer` is using `DemandMaster`. (also present in `RedisSubscriber.IsConnected`)

redis itself replicates publishes from masters to slaves and allows for subscriptions to be made on the slave. Therefore, it should be a valid use case for StackExchange.Redis to place a subscription against a slave.

Unless I'm missing something that would explain the `DemandMaster`, I'm thinking it should be `PreferMaster` instead?

Thanks

mckenzieg1:
An even better solution would be to have `Subscription.SubscribeToServer` use the supplied `CommandFlags` for server selection as well as for the command itself. This allows the client to specify the desired selection mode.

I am about to submit a PR with this fix.

piccit:
Same as issue #321 which has some discussion on. Not sure if there is a "merge issue" option or if someone wants to close one to the other

NickCraver:
Merging with #321 since it has more info - we'll indeed have to address the reconnect issue.

PR #254, opened 2015-08-10
Added support for subscribe-to-slave...

mckenzieg1:
...by making Subscription.SubscribeToServer() use passed-in CommandFlags for server selection. [Fixes StackExchange/StackExchange.Redis#137]

Included a pair of unit tests that highlight the underlying issue and the broken behavior (subscription to slave-only configurations don't work at all).


NickCraver:
IMO this makes sense - can you please rebase the changes?

mckenzieg1:
@NickCraver - sure. This one got caught in limbo between Marc Gravell's and my back-to-back vacations. I have since handed off Redis-related stuff to another developer at my company - I will let him know that if we fix up the PR we can probably get it accepted soon.

NickCraver:
See discussion in #321 - we'll have to address the tracking a well, so a bit more complicated that it initially appears here.

mckenzieg1:
@NickCraver - understood. We have started using 'subscribe to slave' more extensively in our system recently (with our fork of the library including my change). I haven't been working on that aspect lately - I will ask my colleague who is the current expert what he has been seeing and what he thinks about re-connection scenarios specifically. If he comes up with some more fixes we will submit them.

NickCraver:
@mckenzieg1 I need to cleanup here for a VS 2017 migration. I'm not ignoring this issue - we'll track it in #321, but I need to get PRs either merged or cleared for this migration and we need to think through this much deeper and address re-subscribes...so the final PR will look very different anyway.

Note: there's also a related async issue for callers off the subscription that may or may not have a related solution.

mckenzieg1:
@NickCraver Understood - no problem.

Issue #255, opened 2015-08-10
ThreadPriority.AboveNormal for SocketManager reader/writer threads should be optional


mckenzieg1:
The 'reader' and 'writer' threads used by the `SocketManager` are created with `ThreadPriority.AboveNormal`. Depending on the client's threading model and the relative priority of Redis operations vs. other client code, this may not be desirable for all clients. A configuration option would allow clients to override this if they chose.

NickCraver:
Fixed via #256

PR #256, opened 2015-08-10
Added configuration option "HighPrioritySocketThreads"…

mckenzieg1:
… (default 'true' = current behavior)

Closes issue StackExchange/StackExchange.Redis#255


mckenzieg1:
I considered adding the new config option to the table in the Configuration.md documentation file, but I don't think it fits the "more common options" label - I expect this one to be used only by those who know they need it. (My team needs it, which is why I added it.)

mgravell:
This is fine except for the optional constructor parameter which is a breaking change (it changes the signature in a way that existing IL cannot work with). I'll fix that - need to manually merge anyway.

Issue #248, opened 2015-07-27
StackExchange.Redis.RedisServerException -  MOVED 


holidaycottages:
@mgravell have set up a redis cluster and using your code example and set up the end points and trying to add to it and getting the following error:

StackExchange.Redis.RedisServerException -  MOVED 2852 127.0.0.1:7001

Seems like we are connected as a single instance and not as a cluster with the "-c" parameter, how does one do this?

vlavellauruit:
@mgravell, how do you handle the MOVED exception in your code? We downloaded the code and looking through it we couldn't understand how the message is redirected to the new master. We thought that the logic that handle this is not being called on the communication protocol.
Could you clarify a little bit how this worked?

mgravell:
This should be handled internally. However, I have not had chance to
validate against recent cluster builds, so it is possible a cog has fallen
out. I'm on vacation at the moment, but when I get back I'll try to set up
a vCurrent cluster and check.

debbiswal:
Hi..
I am also facing the same issue.
Did u get any chance to look into this issue

vlavellauruit:
@debbiswal, please check the new version of master. @leosvelperez did a pull request that solved a part of the problem. 
Keep in mind that MOVED exception is the way that the server notified the change on the slots. You could still getting some exceptions, and should handle that in your code. 
Regards

debbiswal:
Thanks @vlavellauruit .. I will check the new code and update u.

debbiswal:
Hi @mgravell , @vlavellauruit , @leosvelperez,

With slight change in code for 2nd try(in my previous post ) , I have managed to do the set /get.
What I have done is , If I am getting the 'MOVED' error , I am opening a new connection with specified port(from exception) to get the value. I am not sure whether this approach is the correct one and what is the overhead in production env. Pls suggest .
Below is the sample code :
 try
                {
                    value = db.StringGet(key);
                    Console.WriteLine("1st try:GET " + key + " = " + value);
                }

                catch (Exception ex)
                {
                    Console.WriteLine("1st try:Unable to get " + ex.Message);
                    try
                    {
                        ConnectionMultiplexer con = ConnectionMultiplexer.Connect("127.0.0.1:" + ex.Message.Split(':')[1]);
                        IDatabase db2 = con.GetDatabase();
                        value = db2.StringGet(key);
                        Console.WriteLine("2nd try:GET " + key + " = " + value);
                        con.Close();
                        con.Dispose();
[...]

PR #260, opened 2015-08-13
#248 Exchanging operands in the ternary operators of server selection in TryResend in ServerSelectionStrategy

leosvelperez:
The client was selecting in a wrong way the server endpoint to resend the command, therefore it was failing and propagating the MOVED exception. 
I have noticed that although it was rearranging well the map of servers and hashSlots, if the flag was CommandFlags.DemandMaster and server.IsSelectable(command) returned true then it was selecting null instead selecting the server endpoint itself which was fulfilling the condition needed. Something similar was happening with the other flags.
By exchanging those operands the client is able to select correctly the server to which it will resend the command issued and there is no more MOVED exceptions.


mgravell:
Awesome; well spotted, thanks

leosvelperez:
Glad to help :)

PR #264, opened 2015-08-17
Fixed starvation issue when connecting synchronously.

bcuff:
If the worker pool is too busy to handle await
callbacks it can make it very difficult for a connect
operation to complete without timing out. This change
forces the await callbacks to run on the thread that
invokes the Connect operation, ensuring that a thread
is always available while connecting synchronously.

### DIFF
--- a/StackExchange.Redis/StackExchange/Redis/ConnectionMultiplexer.cs
+++ b/StackExchange.Redis/StackExchange/Redis/ConnectionMultiplexer.cs
@@ -37,13 +37,16 @@ namespace StackExchange.Redis
             return task;
         }
 
+        [ThreadStatic]
+        public static bool ContinueOnCapturedContext;
+
         public static ConfiguredTaskAwaitable ForAwait(this Task task)
         {
-            return task.ConfigureAwait(false);
+            return task.ConfigureAwait(ContinueOnCapturedContext);
         }
         public static ConfiguredTaskAwaitable<T> ForAwait<T>(this Task<T> task)
         {
-            return task.ConfigureAwait(false);
+            return task.ConfigureAwait(ContinueOnCapturedContext);
         }
     }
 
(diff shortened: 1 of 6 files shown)


StackExchange.Redis/StackExchange/Redis/ConnectionMultiplexer.cs
```diff
@@ -37,13 +37,16 @@ public static Task<T> ObserveErrors<T>(this Task<T> task)
             return task;
         }
 
+        [ThreadStatic]
+        public static bool ContinueOnCapturedContext;
```
mgravell: I think what this PR aims to do is great, but I'm dubious about this as the chosen implementation; it seems... unusual, to me - it seems that this should be something we can drill down into the appropriate methods, rather than relying on thread-static variables, which are dangerous on a good day, and downright confusing when `async` and the TPL are involved. Any specific reason you preferred this option?

bcuff: I used this technique to avoid having to re-factory the existing code too much. `SynchronizationContext.Current` is also thread-local so it seems to make sense to me. I suppose ForAwait can change to something like

```csharp
public static ConfiguredTaskAwaitable ForAwait(this Task task)
{
    var useContext = SynchronizationContext.Current.GetType() == typeof(SychronousSynchronizationContext);
    return task.ConfigureAwait(useContext);
}

Alternatively we could pass true/false to every async method.

How do you feel about using a custom SyncrhonizationContext implementation in general? I chose this method because it required the least amount of code change and allowed the use of async/await to continue.

mgravell: I'm not sure what you mean re a custom sync-context; the sync-context
relates to the calling infrastructure : we shouldn't be messing with it at
all, unless we're genuinely providing the threading model (which: we aren't)

bcuff: In this case I am providing an alternative threading model. The await callbacks are queued to a local queue that the thread calling connect services until completion.

NickCraver:
Closing this out to cleanup - as Marc notes above, changing the entire threading model isn't a road we really want to go down here.

bcuff:
understandable. thanks for taking the time to look at least.

PR #277, opened 2015-09-11
Change property type IRedisAsync.Multiplexer to IConnectionMultiplexer

Cybermaxs:
A few weeks ago IConnectionMultiplexer was added to facilitate unit testing.
The return type of IRedisAsync.Multiplexer is not this interface but ConnectionMultiplexer.

For example it's impossible to test the connection status (mydb.Multiplexer.IsConnected) from the IDatabase/IDatabaseAsync point of view.


NickCraver:
This would be a breaking change, since a number of methods available on `ConnectionMultiplexer` are not on the interface. At the very least, this change would be reserved for a major version increment.However, since versioning the interface is incredibly hard (read: impossible), I think we probably want to stick with the class directly here.

I'm not indifferent to the problem proposed here - that's very real, I'm just not sure this is the best approach for a fix.

Issue #275, opened 2015-09-10
StringSet with When.Always returning false (under load?)


kylestittleburg:
We've just been ramping up some Redis usage in an ASP.Net website using the StackExchange.Redis client.  During load tests I've been seeing `StringSet` return false.  

```
_redisDb.StringSet(key, bytes, ttl)
```

**What exactly does this response mean in the case of a When.Always string set?**  

Trying to narrow things done further, during last night's automated perf runs I setup 3 of our API nodes with different SE.Redis versions:
* Node1: nuget release 1.0.450
* Node2: nuget release 1.0.481 (lastest on nuget)
* Node3: local release build from current github code.

I saw many `StringSet` failures with both nuget releases (30k to 50k failures per node), but Node3 did not experience any failures -- so maybe this issue has been fixed? 

This morning I built Verbose StackExchange.Redis DLL from commit aa4e6f697a3e2289c1badd4158c2f03d1f137887, guessing that might be around the time of the the latest nuget release, but have yet to reproduce the failure.

FWIW, Redis is setup as a 3 master, 3 slave cluster, version 3.0.3, and does not appear to be overloaded by any means during the load tests (only doing about 1k ops/sec total among all the Redis nodes), was able to fire up our little test utility that also uses StackExchange.Redis and push it beyond 20k ops/sec while the webapp was seeing StringSet = false errors.

kylestittleburg:
Spent some time trying to get some logs/repro code, unfortunately the only case I've been able to repro is when I purposely have a bad node in the connection string, I usually get a StringSet=False on write appears to be a successful write.

```
        [Test]
        public void SetStringInvalidNode()
        {
            var value = Guid.NewGuid().ToString();
            int threadCount = 10;

            int minWorker, minIOC;
            ThreadPool.GetMinThreads(out minWorker, out minIOC);
            ThreadPool.SetMinThreads(threadCount, minIOC);

            using (var conn = ConnectionMultiplexer.Connect("invalid:6379,10.1.14.120:6379,10.1.14.118:6379,abortConnect=false"))
            {
                var db = conn.GetDatabase();

                Parallel.For(0, threadCount, i =>
                {
                    for (int j = 0; j < 1; j++)
                    {
                        Assert.IsTrue(db.StringSet(Guid.NewGuid().ToString(), value, TimeSpan.FromSeconds(10)));    
                    }
                });

            }
        }
```

During a run of this test, this key `StringSet` returned *false*:
```
[...]

Issue #237, opened 2015-07-20
OnDisconnected corrupts state of ResultBox<bool>.store when PING on top of the message queue


bencican:
Hi Mark,

I was chasing down a race today where seemingly put transaction Commits returned false even though these went through just fine from redis server logs.

Debugging further into it PhysicalBridge.OnDisconnect turned out to be the root cause as when a PING message is on the top of the queue it will get completed twice. I marked them in the code snippet below but basically what happens is that it gets completed the first time because it's a ping... and then second time as part of ChangeState(State.Disconnected) -> AbortUnsent().

This is consistent with the scenario I was encountering where we had a transaction on a disconnected multiplexer which failed and subsequent transaction on a connected multiplexer which succeeded but returned result as though it failed too.

```c#
    internal void OnDisconnected(ConnectionFailureType failureType, PhysicalConnection connection, out bool isCurrent, out State oldState)
    {
        Trace("OnDisconnected");

        // if the next thing in the pipe is a PING, we can tell it that we failed (this really helps spot doomed connects)
        // note that for simplicity we haven't removed it from the queue; that's OK
        int count;
        var ping = queue.PeekPing(out count);
        if (ping != null)
        {
            Trace("Marking PING as failed (queue length: " + count + ")");
            ping.Fail(failureType, null);

            // First completion <--------------------------------------------
            CompleteSyncOrAsync(ping);
        }
        oldState = default(State); // only defined when isCurrent = true
        if (isCurrent = (physical == connection))
        {
            Trace("Bridge noting disconnect from active connection" + (isDisposed ? " (disposed)" : ""));

            // Second completion <--------------------------------------------
            oldState = ChangeState(State.Disconnected);
            physical = null;

            if (!isDisposed && Interlocked.Increment(ref failConnectCount) == 1)
            {
                GetConnection(null); // try to connect immediately
            }
        }
```

Two completions cause the PING message to be completed twice which in turn completes the underlying ResultBox<bool> twice adding a duplicate to the ResultBox<bool>.store and thus corrupting the state of the store and upcoming messages based on how unlucky you get since now a single ResultBox<bool> will be potentially updated completed from more than one message callback at any given time.

Please let me know if there's any more info needed. I don't have a clean repro atm but I think the issue is easy enough to follow and understand.

nmehlei:
[...]

Issue #205, opened 2015-05-11
KeyExists for existing key returned false during high utilization


nmehlei:
It seems to be possible (albeit rarely) that KeyExists returns an incorrect value either during connection problems or high utilization. 

In our case that occured yesterday, utilization was pretty high, which resulted in connection interruptions and KeyExists failed a few times with "No connection is available to service this operation" and when it finally succeeded it returned false which lead to heavy problems inside our application. 

Redis logs showed no problems (except client connects and disconnects etc.), especially no data loss, and our application has no way to remove this key (only change it), so I can pretty much rule out our side. Additionally, there are other open issues related to "No connection is available to service this operation", which makes it even more plausible.

nmehlei:
The same situation happened again last night. Two attempts for KeyExists invocation resulted in

StackExchange.Redis.RedisConnectionException: No connection is available to service this operation: EXISTS my:key:name

The third attempt returned with "false" although the Key does exist. Like last time, this happened during relatively high utilization which seems to be prerequisite for this issue to occur. There are no write operations for this key anywhere in the application, no way for Redis to delete it itself and Redis did not restart during this time, so I can guarantee that the key did have the correct value during this time.

NickCraver:
Fixed in #279 :)

PR #279, opened 2015-09-14
Ensure Message.resultBox is nulled after successful async TryComplete

kylestittleburg:
Attempting to address #237, #275, #205.  I believe @bencican called out the root cause in #237, which is the multiple completions of same the PING message in `PhysicalBridge.OnDisconnected`.  The `PhysicalBridge` code is a bit too unfamiliar to me and it felt safer/more future proof to guard against this within `Message.TryComplete`.  

#275 has a sample unit test that I was using to check this fix, results looked good.  Not sure if we should lock the `resultBox` in case of concurrent calls to `Message.TryComplete`.

### DIFF
--- a/StackExchange.Redis/StackExchange/Redis/Message.cs
+++ b/StackExchange.Redis/StackExchange/Redis/Message.cs
@@ -501,6 +501,12 @@ namespace StackExchange.Redis
             if (resultBox != null)
             {
                 var ret = resultBox.TryComplete(isAsync);
+
+                if (ret && isAsync)
+                {
+                    resultBox = null; // in async mode TryComplete will have unwrapped and recycled resultBox; ensure we no longer reference it via this message
+                }
+
                 if (performance != null)
                 {
                     performance.SetCompleted();


kylestittleburg:
@mgravell Any feedback on this PR?  We've been maintaining a separate build of SE.Redis with this change but I'm hoping to switch back to the standard release Nuget packages

mgravell:
Ooh, that's interesting. Somehow this one escaped me - I was probably drowning in core-clr stuff at the time. I've checked the code, and all looks legit. Merging. I am very curious how many problems this little snafu might fix.

TheCloudlessSky:
@mgravell When do you plan on releasing a new version with this fix?

mgravell:
Good question. I'll see if I can find time this weekend.

TheCloudlessSky:
Thanks!

mgravell:
1.1.553 is released to nuget but hidden; I've left it hidden while I
dogfood etc.

On 15 January 2016 at 17:29, Adrian Phinney <notifications@github.com>
wrote:

> Thanks!
>
> —
> <https://github.com/StackExchange/StackExchange.Redis/pull/279#issuecomment-172025865>
> .
>



-- 
Regards,

Marc

Issue #287, opened 2015-10-04
SortedSetRangeByScoreWithScores issue with positive infinity values


thepirat000:
Create a SortedSet with an item with +Infinity score and try to get its score with SortedSetRangeByScoreWithScores.
The score returned by the mentioned method is NaN instead of PositiveInfinity.

Last assert will fail:
```C#
var cnn = ConnectionMultiplexer.Connect("...");
var db = cnn.GetDatabase();
db.SortedSetAdd("key", "first", double.NegativeInfinity);
db.SortedSetAdd("key", "last", double.PositiveInfinity);
var result = db.SortedSetRangeByScoreWithScores("key", double.NegativeInfinity, double.PositiveInfinity);
Assert.AreEqual(double.NegativeInfinity, result[0].Score);
Assert.AreEqual(double.PositiveInfinity, result[1].Score);  // <-- This assert fails
```

![capture](https://cloud.githubusercontent.com/assets/9836380/10266674/b744a54c-6a34-11e5-9405-755dc9f0e4d2.PNG)

This is the result via redis-cli:
![capture2](https://cloud.githubusercontent.com/assets/9836380/10266731/614f5836-6a38-11e5-84b9-b6e4609acab2.PNG)

NickCraver:
Resolved via #288

PR #288, opened 2015-10-12
Handle "inf" as well as "+inf"

thepirat000:
Redis (v2.8 and v3) returns "inf" for positive infinity values, but here a "+inf" is expected. So changed the code to work with both: "+inf" as well as "inf".
See Issue #287: https://github.com/StackExchange/StackExchange.Redis/issues/287


mgravell:
Looks great, thanks

PR #302, opened 2015-11-11
Add Transaction Conditions for List Index

AlphaGremlin:
I came across the need to run a transaction only if the first item in a list is an expected value. As suggested in the documentation, I've added some extra conditions for transactions to handle list values.

This adds conditions for whether a list index exists, whether it's equal to an expected value, and the reverse (index does not exist, or is not a specified value). Index can be zero-based (from the start of the list) or negative (from the end of the list).

Includes unit tests to verify the various combinations. Tested on Windows 7, all passed successfully.

### DIFF
--- a/StackExchange.Redis.Tests/Transactions.cs
+++ b/StackExchange.Redis.Tests/Transactions.cs
@@ -208,6 +208,96 @@ namespace StackExchange.Redis.Tests
         }
 
         [Test]
+        [TestCase(false, false, true)]
+        [TestCase(false, true, false)]
+        [TestCase(true, false, false)]
+        [TestCase(true, true, true)]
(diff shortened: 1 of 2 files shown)


mgravell:
Looks fine; I think there's  a minor optimization that could be made in terms of `LLEN` for the "exists" case, but that's probably a nice-to-have. In general, I'd suggest that Lua scripting is probably the route forwards in place of transactions, but ... this  works. Thanks.

AlphaGremlin:
No worries. Good point on LLEN instead of LINDEX for the exists case; no point retrieving the value if we just need to make sure the list is long enough.

I considered Lua scripting for what I'm doing, but I'm wary of the "no other script or Redis command will be executed while a script is being executed" bit in the documentation. Unless I'm reading it wrong that sounds like a global lock.

PR #303, opened 2015-11-16
Fixed IServer.Hang command in RedisServer. Database number is not nee…

regiuskornel:
Database number is not needed for CLIENT PAUSE.


mgravell:
Thanks; makes sense

Issue #202, opened 2015-05-06
Core 5


billpratt:
Will you be implementing a Core 5 build for this library? I'd like to use it in a Core 5 project I'm working on.

Thanks!

mgravell:
Dnx? Yes, certainly. I'm working through projects: dapper is done,
protobuf-net is "kinda working ish"; SE.Redis is next on my list, but I'll
need to think about which socket polling methods to use - the regular .net
library does some things that won't be available in CoreCLR

mgravell:
Timescales: hopefully before Monday - possibly sooner.

billpratt:
![b4dfjkyciaaifz_](https://cloud.githubusercontent.com/assets/2847527/7520700/922790ce-f4b6-11e4-947f-e5c44323d598.jpg)

mgravell:
Merry Christmas! Shitter was full!

niemyjski:
+1

TPoise:
Any update on a CoreCLR build?  Looking forward as well to using this...

stebet:
Looking forward to this as well :) I'd love to help if possible.

jmlothian:
I would absolutely love to have this as well.  Unfortunately, I'm not sure the CoreCLR is ready for it.  I put together a project myself just to get a general idea of the changes and for the features StackExchange.Redis uses, Core 5 is about 95% there. (And if anyone has looked closer, please correct me.  This was just a cursory scan of the compile results and the CoreFX github.)

A bit of the remaining stuff are easy fixes for things like some missing interfaces and array methods, and some things more complex with sockets like @mgravell mentioned.  

But the final 2% are things like... System.Threading.Thread missing a large number of functions or working in different ways (there's no Thread.Priority!?).  

For now, it looks like Thread and Threadpool are both somewhat incomplete and the CoreFX repo doesn't have a good public roadmap (that I've found anyhow), so its difficult to say when/if these features will be added.  I wouldn't want to go tinkering with the threading optimizations just to have to change things back later when features get added though.  

Unless someone has some insider information or is already planning a different approach anyhow.

niemyjski:
We'd use this and are working towards getting our library to use core 5 as
well. We have a sole dependency on this library and would love if there was
support for it.

Thanks
-Blake Niemyjski

ArieJones:
[...]

PR #304, opened 2015-11-16
Add support for Dnx Core [WIP]

jeremymeng:
This is for #202. We are still working on a few remaining items.  But I feel it's better to send this out and get feedback earlier.

Open issues
- [ ] CI builds failed because NUnit 3 is not supported in AppVeyor yet. Need workaround.
    - ~~The commands at
      https://gist.github.com/jeremymeng/06e881f9981acbd8b504 seem
      enough to have `nunit-console.exe` in `%PATH%`.~~  NUnit 3
      should work now (https://github.com/appveyor/ci/issues/498).
- [ ] Performance issues
    - [ ] Currently `BufferedStream` isn't available in .Net Core.  This caused > 10x slowdown. it is being added back (tracked by https://github.com/dotnet/corefx/issues/1793).  Once that's available (`1.0.0-rc2`) we can re-enable `BufferedStream` for `DNXCORE50`.
    - [ ] Remaining ~2x slowdown is under investigation.
    - [ ] When running `MassiveBulkOpsSync(bool preserveOrder, int threads)`, increasing number of threads doesn't help much on the .Net Core port, not like the original code.
- [ ] `Stream.BeginRead()` and `Stream.EndRead()` are not supported in .Net Core.  Which of the following options is better?
    1. I've added extension methods to simulate the pattern.  I didn't want to make big changes so I went with extension methods.    
    2. Another option is to rewrite the code using the async api Stream.ReadAsnc(). @deepakverma had an implementation at [his previous port](https://github.com/deepakverma/StackExchange.Redis/blob/client_for_k/StackExchange.Redis_NET50/src/StackExchange.Redis_NET50/StackExchange/Redis/PhysicalConnection.cs#L587).  Is this a concern for the .Net 40 projects?
- [ ] Re-visit `Type.GetTypeCode(Type)` after moving to RC.
- [x] `PLAT_SAFE_CONTINUATIONS` doesn't seem to be available on Core. **Edit** it is available.  I was using 4.0.0 version of `System.Thread.Tasks`.  Fixed.
- [ ] Move to RC. There are some breaking changes between ASP .NET beta8 and RC.
    - https://github.com/aspnet/Announcements/issues/98
- [ ] The dnx test project is a console-based project because it's the only supported way at this point.  NUnit will be adding a dnx console runner and VS integration.
- [ ] `Moq` based tests are all passing using a private build of package that supports .Net Core, but currently disabled for dnx test project.  It can be re-enabled by defining `FEATURE_MOQ` in `project.json`.
- [x] `BookSleeve` related tests were disabled.
    - It looks like there are already similar tests in the StackExchange.Redis.Tests project.
- [ ] We were not able to run the following tests as they require special setup.
    - SSDB
    - SSL
    - VPNTest
    
@mgravell 
CC @rustd @deepakverma @rogerbestmsft


mgravell:
Wow, that's huge. And very much appreciated. I'm going to have to dedicate
some time to reading through this lot. Great to hear that BufferedStream
should be back. I need to review the changes to comment on the rest.

Kudos

stebet:
Wow. This is awesome! If there are any issues that are up for grabs, please tag them as such. I'd love to help :)

ForkBug:
Thank you for the hard working!

StackExchange.Redis/StackExchange.Redis.csproj
```diff
@@ -83,6 +84,7 @@
     <Compile Include="StackExchange\Redis\HashEntry.cs" />
     <Compile Include="StackExchange\Redis\IConnectionMultiplexer.cs" />
     <Compile Include="StackExchange\Redis\InternalErrorEventArgs.cs" />
+    <Compile Include="StackExchange\Redis\InternalRegexCompiledOption.cs" />
```
mgravell: This is actually a really cute solution to the compiled regex problem; I had a similar problem in a few other libs, and just settled for 5.2 vs 5.4 + dnx451 builds, but I'm very tempted to steal this, if you don't mind. Kudos.

mgravell:
Review complete; a couple of big ones I'd like to hear more about, but the biggest is the introduction of TPL into the socket-read - I really really want to understand what that is about, because on the surface that seems incredibly bad.

Much more minor thoughts, summarized (I may have missed one):

- the odd change of case sensitivity
- the ConvertAll would be preferred to right-size allocate (sample code provided)
- would be really nice if this was rc1 compatible rather than beta-8
- the tests don't seem to run for me at all
- the allocations around BeginRead - which is tied into the major socket-read issue

StackExchange.Redis/StackExchange.Redis.csproj
jeremymeng: Credits to @sharwell https://github.com/dotnet/corefx/issues/340#issuecomment-120749951

jeremymeng:
> the odd change of case sensitivity

Fixed

> the ConvertAll would be preferred to right-size allocate (sample code provided)

Fixed

> would be really nice if this was rc1 compatible rather than beta-8

I have been working on it.  Other than bumping versions, the only issue is that in RC1 `Socket.BeginConnect()` is gone as well.  So the situation is similar to `BeginRead()/EndRead()`:  I am not confident enough to rewrite the code using `ConnectAsync()` while keeping the same behavior (exception, logging, etc.)

Maybe you could rewrite `master` branch to use Async Api's instead? Then this port would get them for free.

> the tests don't seem to run for me at all

Were you able to run the project, but got a crash?  I think it might a NUnit bug where some exception are unhandled.   Due to the performance issue, the `BasicOpsTests` could throw `TimeoutException` which will crash dnx. I was able to run that test after modifying timeout limits.

Another test issue is that some `AggregateException` from the Async Api were not handled thus show up in the ambient exceptions during `TearDown()`.

> the text encoding chunk

I noticed that and even attempted to fix it but it didn't do anything good.  Everything looks normal in VS.  Maybe I should use another editor, or adjust my git `crlf` option?  I have it set to `core.autocrlf=true` now.  Will give it another try.


> PLAT_SAFE_CONTINUATIONS just triggers TaskCreationOptions.RunContinuationsAsynchronously, which is declared - does it not work?

Fixed.

> BufferedStream - I don't mind writing a version of that; shouldn't be too tricky especially as it is write only

It's coming soon...

> I would wager $5 that the async-over-sync-over-async is a good part of the rest of the slowdown (that and the extra allocations); any takers?

I have seen the Task overhead in perfview traces.  I have not captured a trace with the `Stream.ReadAsync()` version. I think without the `BufferedStream` fix this overhead is just masked by the regression so it might make sense to investigate later.

mgravell:
I have merged this into the branch on SE.Redis: https://github.com/StackExchange/StackExchange.Redis/tree/jeremymeng-netcore - and updated to the latest bits (and made the project files more manageable, etc). There is a problem with the Socket.Connect API - looks like this changed between beta8 and rc1; I will try to take a look

mgravell:
(happy to do an alpha in the current state, even though there are outstanding problems, but we need to get that Connect API working)

mgravell:
Note: I've also changed project.json to target net40, net45, net46 and dotnet5.5; I would *like* to try to target down-level netstandard, but there are a few problems there, too; 5.4 fails because of `Socket.Available`, for example

mgravell:
I've added something that *probably* fixes the connect problem. There is still something **very** wrong about the `BeginRead` code; that *really* needs to be refactored - we cannot keep (past alpha) something that does async over sync over async. Good enough for alpha, but that's all.

rustd:
Thank you Marc. How do you want to track the remaining work?

mgravell:
I think I've fixed the StreamExtensions (BeginRead etc) code - would appreciate some eyeballs. I've also enabled direct dnx451 comparison in the basic test, and the terrible news is: it is 50-100 times slower. I'm hoping a lot of this is `BufferedStream` !

The current code seems to at least pass the most basic of operations. I'm happier to issue an alpha with the BeginRead stuff changed, which I will do now; however, I have **not** done extensive testing (I'm up to my neck in 20 different things, plus home life is ... complicated and busy).

My biggest concern right now is the performance. That is going to take some digging.

mgravell:
Issued:

https://www.nuget.org/packages/StackExchange.Redis/1.1.0-alpha1
https://www.nuget.org/packages/StackExchange.Redis.StrongName/1.1.0-alpha1

Now that we have something there, we should be able to do a **much** faster turnaround if we need to do an alpha2-alpha347, etc

jeremymeng:
Thanks a lot Marc!  Should we close this PR and do additional work against the other branch?

mgravell:
I was hoping it would do that when I pulled it into a branch, but I guess the tools only do that for master. I've changed the name of the branch to `CoreCLR` for convenience, but it is all there. I need to go through the PR again and create a new issue with a checklist of things we need to look at.

Thanks for all your work making this happen. There *is* still quite a bit of work needed, but it is great to have a release - you've saved me a lot of time and frustration.

mgravell:
Note: I've introduced `BufferedOutputStream` for the core-clr build, which fixes the performance problem, but there seems to be an intermittent glitch that causes problems connecting.

Performance on dnx451:

After 10000: 10000
(with continuation, preserve order)
MassiveBulkOpsAsync: Time for 10000 ops: 27ms; ops/s: 367595.703541417
After 10000: 10000
(no continuation, preserve order)
MassiveBulkOpsAsync: Time for 10000 ops: 10ms; ops/s: 930249.86511377
After 10000: 10000
(with continuation, any order)
MassiveBulkOpsAsync: Time for 10000 ops: 20ms; ops/s: 482516.03159515
After 10000: 10000
(no continuation, any order)
MassiveBulkOpsAsync: Time for 10000 ops: 10ms; ops/s: 942649.221371743

Performance on dnxcore50

After 10000: 10000
(with continuation, preserve order)
MassiveBulkOpsAsync: Time for 10000 ops: 46ms; ops/s: 214035.140289333
After 10000: 10000
(no continuation, preserve order)
MassiveBulkOpsAsync: Time for 10000 ops: 10ms; ops/s: 934169.103291078
After 10000: 10000
(with continuation, any order)
MassiveBulkOpsAsync: Time for 10000 ops: 17ms; ops/s: 563783.664932092
After 10000: 10000
(no continuation, any order)
MassiveBulkOpsAsync: Time for 10000 ops: 31ms; ops/s: 314012.164831266

This is such a massive performance difference that I'm going to issue alpha2

jeremymeng:
Wonderful news!  Sounds like the perf regression is mostly due to lacking `BufferedStream`, and task overhead in my implementation.
I just submit a PR for test change to ignore exceptions that doesn't seem to be caught and ignored by the async implementation.  Could you please take a look?

mgravell:
Am at a family weekend, and have a PDA course on Monday. Will look Tuesday.

noahvans:
Please note that dnx doesn't support Sockets.IOControl() method for Linux\Unix now.

code is here:
https://github.com/dotnet/corefx/blob/master/src/System.Net.Sockets/src/System/Net/Sockets/SocketPal.Unix.cs#L970

Since StackExchange.Redis.SockectManager.SetFastLoopbackOption method is still using it now, so we cannot make these code run on Linux/Unix Platform to connect to any redis server.

mgravell:
K, I'll see if we can try/catch around that. The full .net version actually
does OS sniffing on that line, but the OS sniffing it uses: *also* isn't
available in Core CLR. Such fun!

jeremymeng:
@mgravell @noahvans 
I submitted a PR to add the try-catch around socket.IOControl() call.  Please review when you get a chance.  Thanks!

jeremymeng:
Since the commits have been merged into CoreClr branch.  I am closing this PR.  Open work items are copied to https://github.com/StackExchange/StackExchange.Redis/issues/328.

PR #309, opened 2015-11-20
Update summary comments for SortedSetRangeByScore and SortedSetRangeByScoreWithScores

mattmazzola:
The comments for these two functions was incorrectly stating that start and stop were 0-based indexes for fetching a fixed number of items. ZRANGEBYSCORE actually takes a min and max value which create a range which an unknown amount of members may qualify.

I think the method arguments should be changed to min / max to be more semantic and match actual Redis documentation, but this would require changes to implementation and might be breaking.


mgravell:
954 additions, 954 deletions not shown
916 additions, 916 deletions not shown

eugh; was this a CRLF change as well? I can't tell what has changed here...

mgravell:
Managed to convince the tools to show me the changes; thanks

PR #311, opened 2015-11-26
Add resilience to non ideal cluster topology in CLUSTER NODES response parsing

antoinecellerier:
When hosting a redis cluster deployment on cloud environments where machines get reimaged or reassigned frequently, the CLUSTER NODES output might contain nodes with no endpoint address or multiple nodes assigned to the same endpoint address. These commits add resilience to this state by ignore nodes with no address and adding resilience to duplicate endpoints.


mgravell:
Looks good, thanks

PR #319, opened 2015-12-11
Ignore certain unobserved Task exceptions in test teardown

jeremymeng:
Test-only change.  Some exceptions are caught and ignored in synchronous version. On CoreClr with the introduction of NetworkStream.ReadAsync(), they are no longer caught.  This change ignores them so the tests are not marked as failures due to unobserved ambient exceptions.

@mgravell


mgravell:
Cheers

PR #329, opened 2015-12-31
GetTypeCode() extension method bug fix

jeremymeng:
This fixes the non-CoreClr version of the GetTypeCode() extension method.


mgravell:
Ouch. Nasty recursion bug. Thanks.

PR #331, opened 2016-01-06
Fix cluster configuration regressions from 6b317395ace23c4b1018de638de586061547e22b

antoinecellerier:
Unconfigured cluster nodes might report themselves as endpoint ":{port}",
hence the origin fallback value to make sure that we can address them


mgravell:
Thanks

PR #334, opened 2016-01-09
Correct the bool operator comment

johnnycardy:
The comment was indicating the reverse of what `operator bool` is doing.


NickCraver:
Good catch, thanks!

PR #338, opened 2016-01-18
Using RuntimeInformation.IsOSPlatform for core clr builds

MichaCo:
Added System.Runtime.InteropServices.RuntimeInformation package to use the RuntimeInformation for core clr.


MichaCo:
@mgravell Marc, during some testing I found that the `PlaformNotSupported` exception issue was already fixed by some PR via try/catch. There is actually a new package available, `System.Runtime.InteropServices.RuntimeInformation` to at least figure out on which platform the code is running on.

NickCraver:
Can you undo all the line-ending changes in project.json here? The change surface looks much larger than intended and can't readily review.

MichaCo:
@NickCraver yup, just saw that... New PC / VS install, didn't disable the auto formatting ~~

NickCraver:
Will leave this one for @mgravell - I'm honestly learning towards a `try`/`catch` here instead of the dependency, unless it has applications in more spots (which it very well might).

MichaCo:
@NickCraver that's fine, up to you ;)
I bet there will be some use for getting the client working on mono :innocent:

NickCraver:
@MichaCo yep - I don't want to close this either, let's see how CoreCLR goes with RC2, the decision on Mono support and 4.5.1 vs future isn't clear yet. Labelling this to look at with all CoreCLR bits.

NickCraver:
Well this is really only a drop in the bucket isn't it? Welcome to the world of .Net core I suppose :)

I agree this will be useful for the Mono bits as well; pulling it in.

MichaCo:
yup ;) can't wait for RC2... 
(currently trying to get my stuff building again..., but they just change too much right now I suppose. Lots of fun)

PR #343, opened 2016-01-22
added diagnostic info to the timeoutexcpetion  if client is running on ServerGC 

deepakverma:
added diagnostic info to the timeoutexcpetion  if client is running on ServerGC


deepakverma:
I am closing this for now to figure out why GitHub is saying +2,105  −2,103
May be the line feed in incorrect here. Will resubmit it.

PR #344, opened 2016-01-22
proposing to change back SE.REdis synctimeout to be 1 second 

deepakverma:
ResponseTimeout is set as default to synctimeout. A higher responseTime would have adverse side effect on quickly reconnecting a mux in case of a network blip.


mgravell:
This came from https://github.com/StackExchange/StackExchange.Redis/commit/d694d922e3b86fbfd4c484a6dbb32f23eb9a9c0d - good spot, thanks

Issue #42, opened 2014-05-09
RedisConnectionException at each first Connect(), while redis-py and hiredis work fine


tw-bert:
Hi Marc,

Version 1.0.289.0 (nuget, .net 4.0).

When I leave my workstation alone for five minutes, after calling Connect() I get a RedisConnectionException the first time. Immediately retrying does work with no delay at a 90% success rate. No info is written to the supplied TextWriter.

 

    var oCli = ConnectionMultiplexer.Connect("srv-flux-01:14144", log);

Complete exception message:

    An unhandled exception of type 'StackExchange.Redis.RedisConnectionException' occurred in StackExchange.Redis.dll
    Additional information: It was not possible to connect to the redis server(s); to create a disconnected multiplexer, disable AbortOnConnectFail

I connect from Win7 Pro X64 to Redis 2.8.9 (x64) on a Debian machine on our LAN (same subnet). No hardware firewalls in between, only Windows firewall.
Redis connections from redis-py or RedisDesktopManager(which uses hiredis) never fail, and always connect within milliseconds.

I use a simple Windows Forms project, VS2013 Express, for testing.
I use (and need) the .NET Framework 4 Client Profile.

Any ideas?

Kind regards, TW

ship2moon:
I dont know writing it here will solve our poblem. Please add a "+1" to the same error:
"It was not possible to connect to the redis server(s); to create a disconnected multiplexer, disable AbortOnConnectFail. UnableToResolvePhysicalConnection on PING at StackExchange.Redis.ConnectionMultiplexer"

kmacmcfarlane:
In my case, the call to Connect(...) fails if it is inside an async delegate. I found this by trying to connect from within Nito's AsyncContext.Run(async () => {}) method. Hope that helps someone else :-)

turowicz:
I'm getting that error again. The interesting thing is that some of my peer devs are working on the exact same setup and they do not experience this.

Any ideas?

venkatperi:
+1

Stackexchange.Redis: 1.0.316.0 (via nuget)
Mono 3.12.0 ((detached/a813491)

```
StackExchange.Redis.RedisConnectionException: It was not possible to connect to the redis server(s); to create a disconnected multiplexer, disable AbortOnConnectFail. UnableToResolvePhysicalConnection on PING
[...]

PR #353, opened 2016-02-04
Remove indirection to mitigate a presumed deadlock. Fixes StackExchange/StackExchange.Redis#42.

rjmooney:
So I ran into this issue as well. After reading through some of the comments and taking a look at the code, I noticed an unnecessary indirection in `ConnectionMultiPlexer.ConnectImpl` which appears to be causing a deadlock:

```
// note that task has timeouts internally, so it might take *just over* the regular timeout
// wrap into task to force async execution
var task = Factory.StartNew(() => muxer.ReconfigureAsync(true, false, log, null, "connect").Result);
```

What's happening here is the code is spawning an async task which waits _synchronously_ on an async task result. This is bad juju. What it appears is happening next is the code blocks the context thread waiting for the wrapper task created with `Factory.StartNew` to return:

```
if (!task.Wait(muxer.SyncConnectTimeout(true)))
```

When the continuation from `ReconfigureAsync` returns, it waits for the context thread to become available. Since the context thread is blocked waiting for the continuation, it results in deadlock. Stephen Cleary has an [excellent description](http://blog.stephencleary.com/2012/07/dont-block-on-async-code.html) of the pattern on his blog.

The fix, I believe, is simple: spawn the `ReconfigureAsync` method directly:

```
// note that task has timeouts internally, so it might take *just over* the regular timeout
var task = muxer.ReconfigureAsync(true, false, log, null, "connect");

if (!task.Wait(muxer.SyncConnectTimeout(true)))
```

If I had to guess, I'd say `ReconfigureAsync` was once a synchronous method that was converted to async without updating the `ConnectImpl` implementation to remove the async wrapping. 

The above is conjecture at this point, however in my limited testing it appears to solve the problem.

### DIFF
--- a/StackExchange.Redis/StackExchange/Redis/ConnectionMultiplexer.cs
+++ b/StackExchange.Redis/StackExchange/Redis/ConnectionMultiplexer.cs
@@ -813,8 +813,7 @@ namespace StackExchange.Redis
                 var muxer = multiplexerFactory();
                 killMe = muxer;
                 // note that task has timeouts internally, so it might take *just over* the regular timeout
-                // wrap into task to force async execution
-                var task = Factory.StartNew(() => muxer.ReconfigureAsync(true, false, log, null, "connect").Result);
+                var task = muxer.ReconfigureAsync(true, false, log, null, "connect");
 
                 if (!task.Wait(muxer.SyncConnectTimeout(true)))
                 {


NickCraver:
This problem originates from https://github.com/StackExchange/StackExchange.Redis/commit/aeedada2e69e98862f84d80511bf791c69b6ea9c which was via PR https://github.com/StackExchange/StackExchange.Redis/pull/129 to fix a timeout race.

I agree that fix was bad though, because of the deadlock creation - just adding context since we need to fix both issues.

mgravell:
Yeah, there is definitely something odd here. The `TaskFactory` approach seems an arbitrary and likely unreliable way of fixing it, especially as there were a couple of missing `ForAwait()` calls (now added, but that isn't the point). Thanks; will see how it plays.

NickCraver:
@rjmooney heads up, new package is on NuGet with this fix: https://www.nuget.org/packages/StackExchange.Redis/1.1.572-alpha

rjmooney:
I agree with @NickCraver that both the initial connection issue #42 and the timeout issue in #129 need to be resolved. This patch resolves the former. However in synchronous mode the connection will use the default socket connect timeout instead of the `ConnectionOptions` timeout, causing the unexpected behavior described in #129. This causes the `ConnectToUnexistingHostFailsWithinTimeout` synchronous unit test to fail.

The culprit is the call to `AsyncWaitHandle.WaitOne()` in `RunWithCompletionType` which doesn't specify a timeout. My limited insight into the design of the library notwithstanding, making the current connection context available to downstream methods such that`ConnectionOptions` can be referenced would allow `RunWithCompletionType` to specify a timeout to `WaitOne`. I think this is the ideal solution, although it will require refactoring.

In the mean time, rewrapping the call to `ConnectAsync`'s `ReconfigureAsync` in a `Task.Run` *async Task* for `CompletionType.Sync` invocations should resolve the issue despite the smell.

PashaPash:
@rjmooney thanks for the fix! Actually, both test and `Factory.StartNew` wrapper were added only to cover `CompletionType.Sync` removal in `SocketManager`. 

`RunWithCompletionType` with `CompletionType.Sync` is a test-only mode for #113 unit tests, so it is not necessary to support timeout in it.

It was used in #129 only as a way to put something long-running that will totally ignore all timeout settings inside of `ReconfigureAsync`. It was basically a way to check what will happen if `ReconfigureAsync` will ignore `ConnectTimeout` (as it actually does) and will not timeout internally in a reasonable time. So `RunWithCompletionType` should not support timeout directly.

There is a comment on that in the code:

    // note that task has timeouts internally, so it might take *just over* the regular timeout

If that `*just over*`case is not a critical problem for anyone, then it is better to leave it as is, and just remove `[TestCase(CompletionType.Sync)]` from `ConnectToUnexistingHostFailsWithinTimeout`.

PR #363, opened 2016-02-25
Capture client system CPU percentage in timeout error messages

JonCole:
Capture client system CPU percentage in timeout error messages when possible to help debug timeouts.


mgravell:
Thanks

PR #365, opened 2016-03-04
updated to coreclr version 1.0.0-rc1-update1

GeertVL:
It is a minor update but so it is bound to the latest version of the coreclr

### DIFF
--- a/global.json
+++ b/global.json
@@ -1,6 +1,6 @@
 {
   "sdk": {
-    "version": "1.0.0-rc1-final",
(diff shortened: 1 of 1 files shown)


mgravell:
Merged, thanks. I was aware of update1, but I was hoping rc2 would drop by now ;p It hasn't, so: agree

GeertVL:
Could you give me some more work I could work on. I like this repo and want to contribute.

PR #370, opened 2016-03-10
reworded redis pub/sub doc to be more readable

willharrison:
I wasn't sure if `too` was supposed to be `tool` and was a typo, or if `too` was deliberate.  Regardless, it sounded weird to me and I think this is clearer.

### DIFF
--- a/Docs/Basics.md
+++ b/Docs/Basics.md
@@ -69,7 +69,7 @@ The entire range of [redis database commands](http://redis.io/commands) covering
 Using redis pub/sub
 ----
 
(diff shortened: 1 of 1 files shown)


mgravell:
It was intended as written; I don't mind rephrasing it if it is unclear, but I kinda don't like the repetition of "tool" so soon, from a language thing. Possibly the question of "what phrasing sounds natural?" is impacted by region - I'm in the UK.

willharrison:
I'm not a fan of using "tool" in repetition, either.  I just wasn't sure what else to put.  My only issue with the current phrasing was that I had to reread it a few times, not sure if "too" was meant to be "tool" and missing an "L".  Not that the meaning would change vastly either way.

mgravell:
"This is also..." maybe?

willharrison:
That sounds fine to me.  I have updated the commit.

PR #375, opened 2016-03-16
Set default version to 3.0 for Azure

deepakverma:
With default version of 2.0, for a large number of connection to a Redis endpoint features.InfoSections returns false and  causes full info command to be executed during initiation of a multiplexer. 
info command from high number of connections trying to connect,  can be expensive for redis server.

Making default version to 3.0 so that 'info replication' and 'info server' commands are leveraged and improves performance.

### DIFF
--- a/MigratedBookSleeveTestSuite/Config.cs
+++ b/MigratedBookSleeveTestSuite/Config.cs
@@ -188,45 +188,59 @@ namespace Tests
         }
 
         [Test]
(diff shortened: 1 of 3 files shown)


mgravell:
This will be fine. Don't have my 2FA, but will merge when at PC.

deepakverma:
Cool, thank you.

PR #378, opened 2016-03-21
Fix overflow in parsing "-9223372036854775808"

veblush:
`RedisValue.TryParseInt64` method throws an overflow exception while parsing `-9223372036854775808` which is a minimum value of long. This PR fixes this bug.


mgravell:
Looks good, thanks for the fix.

PR #379, opened 2016-03-25
Default ClientName to Roleinstance ID for a web/worker role 

deepakverma:
helps to identify where a client connection is coming from, further helps to identify any connection related problems.

### DIFF
--- a/StackExchange.Redis.Tests/Config.cs
+++ b/StackExchange.Redis.Tests/Config.cs
@@ -125,6 +125,28 @@ namespace StackExchange.Redis.Tests
         }
 
         [Test]
+        public void DefaultClientName()
+        {
+            using (var muxer = Create(allowAdmin: true))
+            {
+                Assert.AreEqual(Environment.MachineName, muxer.ClientName);
+                var conn = muxer.GetDatabase();
+                conn.Ping();
+#if DEBUG
+                var name = GetServer(muxer).ClientGetName();
+                Assert.AreEqual(Environment.MachineName, name);
+#endif
+            }
+        }
+
+        [Test]
+        public void TryGetAzureRoleInstanceIdNoThrow()
+        {
+            ConfigurationOptions config = new ConfigurationOptions();
+            Assert.IsNull(config.TryGetAzureRoleInstanceIdNoThrow());
+        }
+
+        [Test]
         public void ReadConfigWithConfigDisabled()
         {
             using (var muxer = Create(allowAdmin: true, disabledCommands: new[] { "config", "info" }))
--- a/StackExchange.Redis/StackExchange/Redis/ConfigurationOptions.cs
+++ b/StackExchange.Redis/StackExchange/Redis/ConfigurationOptions.cs
@@ -5,7 +5,6 @@ using System.IO;
 using System.Linq;
 using System.Net;
 using System.Net.Security;
-using System.Reflection;
 using System.Text;
 using System.Threading.Tasks;
 
@@ -162,7 +161,7 @@ namespace StackExchange.Redis
         /// <summary>
         /// The client name to use for all connections
         /// </summary>
-        public string ClientName { get { return clientName ?? (clientName = (AzureRoleInstanceId != null ? AzureRoleInstanceId : Environment.GetEnvironmentVariable("ComputerName"))); } set { clientName = value; } }
+        public string ClientName { get { return clientName; } set { clientName = value; } }
 
         /// <summary>
         /// The number of times to repeat the initial connect cycle if no servers respond promptly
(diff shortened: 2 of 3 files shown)


mgravell:
I greatly appreciate that you didn't add a dependency for this - that was
my initial concern when I went to read the change set. However, unless I'm
mistaken, this could get run every time? (no caching). Also, I'm not at a
PC, but I thought we already had some defaults around the client- name,
specifically: the machine name. It seems that it would be desirable to site
these at the same place? I might also simply be misremembering...

In principle, I'm not opposed. It seems a shame that this isn't available
from an environment variable, though. Can I just double-check that this
*isn't* in an environment variable? It kinda feels like where I'd expect to
find it...

deepakverma:
The code actually runs it only once while initializing the static field storing the roleinstanceid. I have updated the pull request with the logic of setting it to machine name at the same place.
I re-checked the environment variables on a worker role and it's unfortunate that there is no environment variable that has a human readable role instance id there :(.

Roleinstance id would give something like "WorkerRole1_IN_0", "WorkerRole1_IN_1"  for a role instance in a cloud service.

following is output from one of my azure worker role:
ALLUSERSPROFILE=D:\ProgramData
APPDATA=D:\Users\azureuser\AppData\Roaming
CommonProgramFiles=D:\Program Files\Common Files
CommonProgramFiles(x86)=D:\Program Files (x86)\Common Files
CommonProgramW6432=D:\Program Files\Common Files
COMPUTERNAME=RD00155DF879C4
ComSpec=D:\Windows\system32\cmd.exe
FP_NO_HOST_CHECK=NO
HOMEDRIVE=D:
HOMEPATH=\Users\azureuser
LOCALAPPDATA=D:\Users\azureuser\AppData\Local
LOGONSERVER=\\RD00155DF879C4
NUMBER_OF_PROCESSORS=4
OS=Windows_NT
Path=D:\Windows\system32;D:\Windows;D:\Windows\System32\Wbem;D:\Windows\System32
\WindowsPowerShell\v1.0\;D:\Program Files\Microsoft SQL Server\120\Tools\Binn\;D
:\Program Files\Microsoft\Web Platform Installer\
PATHEXT=.COM;.EXE;.BAT;.CMD;.VBS;.VBE;.JS;.JSE;.WSF;.WSH;.MSC
PROCESSOR_ARCHITECTURE=AMD64
PROCESSOR_IDENTIFIER=AMD64 Family 16 Model 8 Stepping 1, AuthenticAMD
PROCESSOR_LEVEL=16
PROCESSOR_REVISION=0801
ProgramData=D:\ProgramData
ProgramFiles=D:\Program Files
ProgramFiles(x86)=D:\Program Files (x86)
ProgramW6432=D:\Program Files
PROMPT=$P$G
PSModulePath=D:\Windows\system32\WindowsPowerShell\v1.0\Modules\;d:\Program File
s\Microsoft Security Client\MpProvider\;D:\Program Files\WindowsPowerShell\Modul
es\;D:\Program Files (x86)\Microsoft SDKs\Azure\PowerShell\ResourceManager\Azure
ResourceManager\;D:\Program Files (x86)\Microsoft SDKs\Azure\PowerShell\ServiceM
anagement\
PUBLIC=D:\Users\Public
SESSIONNAME=RDP-Tcp#31
SystemDrive=D:
SystemRoot=D:\Windows
TEMP=D:\Users\AZUREU~1\AppData\Local\Temp\2
TMP=D:\Users\AZUREU~1\AppData\Local\Temp\2
USERDOMAIN=RD00155DF879C4
USERDOMAIN_ROAMINGPROFILE=RD00155DF879C4
USERNAME=azureuser
USERPROFILE=D:\Users\azureuser
VS140COMNTOOLS=C:\Program Files (x86)\Microsoft Visual Studio 14.0\Common7\Tools
\
windir=D:\Windows

mgravell:
Fair enough. I misread the chain re running multiple times. My bad. Will
look at the code on a PC :)

mgravell:
K, I've had chance to look at it more properly. The intent is fine, but I don't think the current positioning is right; by putting it into `ConfigurationOptions`, you change how the parse/format works - suddenly configuration options starts writing a client name even when one wasn't specified.  The right place to put this change is `ConnectionMultiplexer`; you *simplified* the line, but IMO you should have made it more complicated. I suspect the correct implementation is:

    public string ClientName => configuration.ClientName ?? TryBlahAzureBlah() ?? Environment.GetEnvironmentVariable("ComputerName");

I'm happy to change that during the merge, or you can change it at include it in the PR. Any preference?

mgravell:
Or perhaps more efficiently:

    public string ClientName => configuration.ClientName ?? GetDefaultClientName();
    private static string defaultClientName;
    private static string GetDefaultClientName() {
        if(defaultClientName == null)
        {
            defaultClientName = TryBlahAzureBlah() ?? Environment.GetEnvironmentVariable("ComputerName");
        }
        return defaultClientName;
    }

deepakverma:
Yeah, Good point regarding how it would work with parse/format.  I will update my pull request. thx for reviewing it.

mgravell:
Actually, it *might* have not impacted tostring *unless* someone explicitly
looked at ClientName, but either way, it is an odd and unexpected
side-effect in that location.

On 25 March 2016 at 23:31, dverma <notifications@github.com> wrote:

> Yeah, Good point regarding how it would work with parse/format. I will
> update my pull request. thx for reviewing it.
>
> —
> You are receiving this because you commented.
> <https://github.com/StackExchange/StackExchange.Redis/pull/379#issuecomment-201585132>
>



-- 
Regards,

Marc

deepakverma:
yeah, I agree. I have updated the pull request.

PR #381, opened 2016-03-30
Add geo commands

wjdavis5:
I have implemented 

- GEOADD
- GEODIST
- GEOHASH
- GEOPOS
- GEORADIUS

Still needs

- GEORADIUSBYMEMBER

Thus far, have only implemented the Sync. versions of the methods, and have not yet implemented unit tests.


StackExchange.Redis/StackExchange/Redis/GeoEntry.cs
```diff
@@ -0,0 +1,144 @@
...
+using System.ComponentModel;
+
+namespace StackExchange.Redis
+{
+
+    /// <summary>
+    /// Describes a GeoEntry element with the corresponding value
+    /// GeoEntries are stored in redis as SortedSetEntries
+    /// </summary>
+    public struct GeoEntry : IEquatable<GeoEntry>, IEquatable<SortedSetEntry>, IComparable, IComparable<GeoEntry>, IComparable<SortedSetEntry>
```
mgravell: (writing as I work down the commit, so initial thoughts) - this looks a lot like `SortedSetEntry`; but I can get behind having a separate struct if it has a valid different semantic meaning; does it actually need to be equatable / comparable to `SortedSetEntry` though? is there an API overlap that this is catering for?

StackExchange.Redis/StackExchange/Redis/GeoEntry.cs
```diff
@@ -0,0 +1,144 @@
...
+    {
+        internal readonly RedisValue element;
```
mgravell: haven't checked what uses these yet, but do they need to be `internal`? `private` is a good default - if not, it would be good to state why not

StackExchange.Redis/StackExchange/Redis/GeoUnit.cs
```diff
@@ -0,0 +1,27 @@
...
+{
+    /// <summary>
+    /// Units associated with Geo Commands
+    /// </summary>
+    public enum GeoUnit : short
+    {
+        /// <summary>
+        /// Meters
+        /// </summary>
+        m,
```
mgravell: Part of me is shouting that this should be `Meters`, `Kilometers`, `Miles`, etc... thoughts?

StackExchange.Redis/StackExchange/Redis/IDatabase.cs
```diff
@@ -45,6 +45,44 @@ public interface IDatabase : IRedis, IDatabaseAsync
         RedisValue DebugObject(RedisKey key, CommandFlags flags = CommandFlags.None);
 
         /// <summary>
+        /// Add the specified member to the set stored at key. Specified members that are already a member of this set are ignored. If key does not exist, a new set is created before adding the specified members.
+        /// </summary>
+        /// <returns>True if the specified member was not already present in the set, else False</returns>
+        /// <remarks>http://redis.io/commands/sadd</remarks>
+        bool GeoAdd(RedisKey key, RedisValue value0, RedisValue value1, RedisValue value2, CommandFlags flags = CommandFlags.None);
```
mgravell: is there a meaning to `value0` etc here? looking at http://redis.io/commands/geoadd, this is `longitude`, `latitude`, `member` - is that right? and if so: longitude and latitude should be `double`, not `RedisValue`, no? Part of me expected this to take a `GeoEntry`, but that seems to not be the case, which confused me.

StackExchange.Redis/StackExchange/Redis/GeoEntry.cs
```diff
@@ -0,0 +1,144 @@
...
+
+        /// <summary>
+        /// The unique element stored in the sorted set
+        /// </summary>
+        public RedisValue Element => element;
+
+        /// <summary>
+        /// The score against the element
+        /// </summary>
+        public double Score => score;
```
mgravell: I genuinely don't know, but: what does this *mean* in the case of a `GeoEntry`? what is the "score"?

StackExchange.Redis/StackExchange/Redis/IDatabase.cs
```diff
@@ -45,6 +45,44 @@ public interface IDatabase : IRedis, IDatabaseAsync
...
+        /// <returns>the number of elements that were added to the set, not including all the elements already present into the set.</returns>
+        /// <remarks>http://redis.io/commands/geoadd</remarks>
+        long GeoAdd(RedisKey key, RedisValue[] values, CommandFlags flags = CommandFlags.None);
+
+        /// <summary>
+        /// Return the distance between two members in the geospatial index represented by the sorted set.
+        /// </summary>
+        /// <returns>The command returns the distance as a double (represented as a string) in the specified unit, or NULL if one or both the elements are missing.</returns>
+        /// <remarks>http://redis.io/commands/geodist</remarks>
+        RedisValue GeoDist(RedisKey key, RedisValue value0,RedisValue value1,GeoUnit geoUnit = GeoUnit.m, CommandFlags flags = CommandFlags.None);
```
mgravell: API calls them `member1`, `member2` - it is probably worth being very clear on terminology

StackExchange.Redis/StackExchange/Redis/IDatabase.cs
```diff
@@ -45,6 +45,44 @@ public interface IDatabase : IRedis, IDatabaseAsync
...
+
+
+        
+        /// <summary>
+        /// Return valid Geohash strings representing the position of one or more elements in a sorted set value representing a geospatial index (where elements were added using GEOADD).
+        /// </summary>
+        /// <returns>The command returns an array where each element is the Geohash corresponding to each member name passed as argument to the command.</returns>
+        /// <remarks>http://redis.io/commands/geohash</remarks>
+        RedisValue[] GeoHash(RedisKey key, RedisValue[] value, CommandFlags flags = CommandFlags.None);
```
mgravell: Reading the API, I think this should actually be returning `string[]`; geo-hash is always base-32 using alpha-numerics - there is no obvious requirement for it to behave as `RedisValue`

StackExchange.Redis/StackExchange/Redis/IDatabase.cs
mgravell: it feels like this should return either `double` or `double?` - I don't see that this needs to operate as a `RedisValue`. Perhaps there is a lot of confusion here with `RedisValue` - it isn't intended to represent *all values* - rather, it suffices for those things where redis will treat interchangeably a `string`, a `byte[]` and a number. That applies for *members*, and for *string values* (in the `GET` / `SET` redis sense) - but not here.

StackExchange.Redis/StackExchange/Redis/KeyspaceIsolation/DatabaseWrapper.cs
```diff
@@ -27,6 +27,33 @@ public RedisValue DebugObject(RedisKey key, CommandFlags flags = CommandFlags.No
...
+        {
+            return Inner.GeoHash(key, value, flags);
+        }
+
+        public RedisValue[,] GeoPos(RedisKey key, RedisValue[] value, CommandFlags flags = CommandFlags.None)
+        {
+            return Inner.GeoPos(key, value, flags);
+        }
+
         public double HashDecrement(RedisKey key, RedisValue hashField, double value, CommandFlags flags = CommandFlags.None)
```
mgravell: (note: also `GEORADIUS` and `GEORADIUSBYMEMBER`, but I'm not saying that it is your job to do those, and I can also relate if the intent is to check the implementation etc)

StackExchange.Redis/StackExchange/Redis/KeyspaceIsolation/DatabaseWrapper.cs
```diff
@@ -27,6 +27,33 @@ public RedisValue DebugObject(RedisKey key, CommandFlags flags = CommandFlags.No
             return Inner.DebugObject(ToInner(key), flags);
         }
 
+          public long GeoAdd(RedisKey key, RedisValue[] values, CommandFlags flags = CommandFlags.None)
+        {
+            return Inner.GeoAdd(ToInner(key), values, flags);
+        }
+
+        public RedisValue GeoDist(RedisKey key, RedisValue value0, RedisValue value1, GeoUnit geoUnit = GeoUnit.m,
```
mgravell: Naming: I've tended to use fully expanded names for clarity; `Geo` is fine, but this should probably be `GeoDistance`

StackExchange.Redis/StackExchange/Redis/KeyspaceIsolation/DatabaseWrapper.cs
```diff
@@ -27,6 +27,33 @@ public RedisValue DebugObject(RedisKey key, CommandFlags flags = CommandFlags.No
...
+        {
+            return Inner.GeoAdd(ToInner(key), value0,value1,value2, flags);
+        }
+
+        public RedisValue[] GeoHash(RedisKey key, RedisValue[] value, CommandFlags flags = CommandFlags.None)
+        {
+            return Inner.GeoHash(key, value, flags);
+        }
+
+        public RedisValue[,] GeoPos(RedisKey key, RedisValue[] value, CommandFlags flags = CommandFlags.None)
```
mgravell: Probably `GeoPosition`. The return type is intriguing. It kinda wants to be a `Coordinate[]` or `GeoPosition[]`, where that in turn is essentially a long/lat tuple. Alternatively, `double[,]` is possibly viable, but feels ... unsatisfactory. Whatever it is, it should probably mirror what happens in `Add`?

mgravell: Probably `GeoPosition`. The return type is intriguing. It kinda wants to be a `Coordinate[]` or `GeoPosition[]`, where that in turn is essentially a long/lat tuple. Alternatively, `double[,]` is possibly viable, but feels ... unsatisfactory. Whatever it is, it should probably mirror what happens in `Add`?

mgravell: Probably `GeoPosition`. The return type is intriguing. It kinda wants to be a `Coordinate[]` or `GeoPosition[]`, where that in turn is essentially a long/lat tuple. Alternatively, `double[,]` is possibly viable, but feels ... unsatisfactory. Whatever it is, it should probably mirror what happens in `Add`?

StackExchange.Redis/StackExchange/Redis/GeoEntry.cs
```diff
@@ -0,0 +1,144 @@
+
```
mgravell: Fundamental question: what, outside of `GeoEntry.cs` uses `GeoEntry`? None of the APIs do...

StackExchange.Redis/StackExchange/Redis/RedisCommand.cs
```diff
@@ -38,6 +38,13 @@ enum RedisCommand
         FLUSHALL,
         FLUSHDB,
 
+        GEOADD,
```
mgravell: note to self: this would need to be checked in the "can I run on a slave" method; can't remember whether this is whitelist or blacklist.

StackExchange.Redis/StackExchange/Redis/RedisDatabase.cs
```diff
@@ -45,6 +46,37 @@ public RedisValue DebugObject(RedisKey key, CommandFlags flags = CommandFlags.No
...
 
+        public bool GeoAdd(RedisKey key, RedisValue value0,RedisValue value1,RedisValue value2,CommandFlags flags = CommandFlags.None)
+        {
+            var msg = Message.Create(Database, flags, RedisCommand.GEOADD, key, value0,value1,value2);
+            return ExecuteSync(msg, ResultProcessor.Boolean);
+        }
+
+        public long GeoAdd(RedisKey key, RedisValue[] values, CommandFlags flags = CommandFlags.None)
+        {
+            if(values.Length%3!=0) throw new Exception("Each value must provide latitude, longitude and name");
```
mgravell: Yeah, seeing it here: I definitely hate this way of expressing longitude and latitude

mgravell: probably want a null check here against `values`, if preserved

mgravell: what type of `Exception`? probably an `ArgumentException` specifying the `nameof(values)`

StackExchange.Redis/StackExchange/Redis/RedisDatabase.cs
```diff
@@ -45,6 +46,37 @@ public RedisValue DebugObject(RedisKey key, CommandFlags flags = CommandFlags.No
...
+            var msg = Message.Create(Database,flags, RedisCommand.GEOADD, key, values);
+            return ExecuteSync(msg, ResultProcessor.Int64);
+        }
+
+        public RedisValue GeoDist(RedisKey key, RedisValue value0, RedisValue value1, GeoUnit geoUnit = GeoUnit.m,CommandFlags flags = CommandFlags.None)
+        {
+            var msg = Message.Create(Database, flags, RedisCommand.GEODIST, key,value0,value1,geoUnit.ToString());
```
mgravell: if the enum names change, this will need a map somewhere - probably just a `private static RedisValue[] geoUnitNames = {"m", ...};` - but in either case: what if it isn't one of the defined values? (enums are not checked)

StackExchange.Redis/StackExchange/Redis/GeoUnit.cs
```diff
@@ -0,0 +1,27 @@
+using System;
+
+namespace StackExchange.Redis
+{
+    /// <summary>
+    /// Units associated with Geo Commands
+    /// </summary>
+    public enum GeoUnit : short
```
mgravell: no real reason to make this ` : short` IMO; it is never stored etc

StackExchange.Redis/StackExchange/Redis/KeyspaceIsolation/DatabaseWrapper.cs
mgravell: (note: also the async API surface, but: definitely worth getting the sync surface nailed before duplicating everything)

mgravell:
Awesome! I genuinely appreciate someone taking the time to start whacking this together. I feel like I've been a bit brutal in the review, but I think there is a lot of stuff that needs attention - kinda hard to fix it "later" if we merge it "now"; I think it mostly boils down to:

- naming
- types - in particular, `RedisValue` has a *quite specific purpose*, that doesn't match any of longitude, latitude or distance
- how we should express co-ordinates as a result, and as input alongside a member

I couldn't see what the `struct` does.

But: definitely has potential! All thoughts and feedback welcome.

StackExchange.Redis/StackExchange/Redis/GeoEntry.cs
wjdavis5: The whole Geo concept in redis is really just a Sorted Set with some new methods that operate on them. My thought was it should be presented this way. My initial thought was the GeoEntry should inherit from SortedSetEntry, but it was represented as a struct , and I didnt want to change that w/o your input.

StackExchange.Redis/StackExchange/Redis/GeoUnit.cs
wjdavis5: Redis yelled at me when I changed it from this. It required the lowercase abbriviation for these. Perhaps that will change when it goes GA.

StackExchange.Redis/StackExchange/Redis/IDatabase.cs
wjdavis5: TBH I was trying to mimic the SortSet methods. They also used the RedisValu, so I did too. I'm not currently sure that the GeoEntry is even needed.

StackExchange.Redis/StackExchange/Redis/GeoEntry.cs
wjdavis5: Again here, just copying the semantics from the SortedSetEntry. Each time you perform GEOADD in redis it is only adding an item to a SortedSet. It has a valid score just like a SortedSetEntry.

StackExchange.Redis/StackExchange/Redis/IDatabase.cs
wjdavis5: I was going to use string, b/c redis returns a double represented by a string, and let the end user decide how to cast it. I used RedisValue b/c it was convenient to use as a string. But, I'm good with making it a double.

StackExchange.Redis/StackExchange/Redis/IDatabase.cs
wjdavis5: Here I was just following the semantics of the SortedSet methods again.
They used both SortedSetEntry[] and RedisValue[] for returns:
```
 public RedisValue[] SortedSetRangeByRank
  public SortedSetEntry[] SortedSetRangeByRankWithScores
```

I'm fine with returning string[] though too. Making it return RedisValue was easier and it was already coded that way in several places.

StackExchange.Redis/StackExchange/Redis/KeyspaceIsolation/DatabaseWrapper.cs
wjdavis5: I'll probably knock out the remaining too members after I get the unit tests for the existing ones finished, and after some refactor work to get things fixed up.

StackExchange.Redis/StackExchange/Redis/KeyspaceIsolation/DatabaseWrapper.cs
wjdavis5: I'm fine with that, was just using the redis command names there.

StackExchange.Redis/StackExchange/Redis/KeyspaceIsolation/DatabaseWrapper.cs
wjdavis5: Makes sense to create a type with a fitting name.

StackExchange.Redis/StackExchange/Redis/RedisCommand.cs
wjdavis5: this is on my radar to check

StackExchange.Redis/StackExchange/Redis/RedisDatabase.cs
wjdavis5: Agree this is not a good long term way to express lat / lon - and I agree ArgumentException is appropriate.

wjdavis5:
Hey - thank you for taking time to take a look at this. I opened this PR a little premature hoping for some dialog. I wanted to get your input so I could understand some of the intentions / semantics of the existing code better.

RedisValue was very simple to implement quickly. Which is probably why I've abused it so. Simple was good at first so I could have a chance to test the Redis functionality for a sprint planning week which ends Friday. I'm totally on board with some significant refactor, and was planning on it after getting some feedback from you.

I had added some types in - GeoEntry- with the intention of implementing them in the methods. But realized I could get these few methods functioning in a couple hours time if I used RedsValue.

As for the coordinates, we can create a simple object to represent them if it fits better with the current semantics. 

I'll be doing some functionality / performance testing on redis using these methods the next couple of days. Once thats wrapped up I'll be able to knock alot of this out.

Again - thanks for taking the time to look and I appreciate the feedback!

StackExchange.Redis/StackExchange/Redis/GeoUnit.cs
mgravell: Yes, it would need a corresponding tweak to the message generation; but the name of the enum here should **not** have to match what redis understands.

StackExchange.Redis/StackExchange/Redis/GeoEntry.cs
mgravell: @wjdavis5  but *where is this used?*

StackExchange.Redis/StackExchange/Redis/IDatabase.cs
mgravell: In the case of `SortedSetAdd`, only the `member` (i.e. the location name in the case of `Geo`) is a `RedisValue`; the `score` is a `double`:

    public bool SortedSetAdd(RedisKey key, RedisValue member, double score,  ...

In the case of `Geo`, the `longitude` and `latitude` and kinda like the `score`, so they should a: be `double` and b: be named clearly.

mgravell: @wjdavis5  but a `GeoEntry` only has one `double` currently; would need 2?

StackExchange.Redis/StackExchange/Redis/GeoEntry.cs
mgravell: @wjdavis5 indeed, but the meaning of that value is opaque and not really exposed; the outside caller talks in terms of `longitude` and `latitude`

StackExchange.Redis/StackExchange/Redis/IDatabase.cs
mgravell: Fine; the trick, though, is what to do for the null case; `double.NaN` maybe? or maybe it is clearer to use `double?`; see the `GEODIST Sicily Foo Bar` example on http://redis.io/commands/geodist to see what I mean

StackExchange.Redis/StackExchange/Redis/IDatabase.cs
mgravell: There is no comparable method to `GeoHash` on `SortedSet...`; geo-hash is a very specific thing, and is *not* the same as "get me the members". It is not comparable to `SortedSetRangeByRank` etc

StackExchange.Redis/StackExchange/Redis/KeyspaceIsolation/DatabaseWrapper.cs
mgravell: Note: if `GeoEntry` has lat, long and member: we could return `GeoEntry` for that. Even if redis doesn't actually return the names, we can fill in the blanks from the inputs

mgravell:
Absolutely - fully understand it is a work in progress, and glad to have a chance to chime in sooner rather than later. Let me know if I'm talking nonsense at any point - that does happen.

I think the `GeoEntry` does have a place when we are talking about the triad of lat, long and member - useful for the multi-add and pos scenarios. There is an interesting question as to whether we should also include an internally-settable only `Distance` which would allow us to also use it for `GeoRadius` etc, which return a member and *optionally* (and separately) the coords and distance.

But: `GeoEntry` in that scenario has very little relationship to `SortedSetEntry`, and we should remove the interchange etc.

MichaCo:
Few comments regarding the `GeoAdd` API
In this scenario, the method expects x and y and a name. Coordinates are usually double values and the name is always a string I guess. So, why not use those types right away?

I suggest to change the signature of GeoAdd to 

    bool GeoAdd(RedisKey key, double longitude, double latitude, string name, CommandFlags flags = CommandFlags.None)

Also, you might want to add client side out of bounds checks for longitude and latitude to prevent errors early.

For the multi add, I'd either use a `ICollection<Tupple<double, double, string>>` or a list of new/dedicated POCO object (preferred) to store the values instead of just an array of something.

Same applies to any other method where you pass in either the name or coordinates.

mgravell:
I doubt it is safe to assume all members are strings. The advantage of `RedisValue` *in that case* is that it silently works with any value that could be a regular redis member. Similarly, you could say "a redis key is a string" - not quite true for the same reasons, hence `RedisKey`. So: I do **not** agree - at least without testing it  - that the member should be a `string`. Totally agree on `longitude` and `latitude`, though.

`Tuple<double, double, string>` would be a very bad choice, IMO; tuples should *almost never* be preferred over named contextual members, etc. That is probably where `GeoEntry` comes in. As for collection vs array vs list vs interface: the API uses arrays elsewhere, and for designed reasons (not just chance); I would like to keep that consistent.

mgravell:
Random observation: I think it might actual be a good idea to add `GeoRemove`. This is actually just the same as `ZREM` / `SortedSetRemove` - and could almost certainly just be implemented as `return SortedSetRemove(...);` - but I don't think we should force people to know that `Geo*` is implemented as sorted-set. So:

        public bool SortedSetRemove(RedisKey key, RedisValue member, CommandFlags flags = CommandFlags.None)
        {
            return SortedSetRemove(key, member, flags);
        }

MichaCo:
> I doubt it is safe to assume all members are strings

Yeah, agree. Testing will show I guess ;) So for name it will be RedisValue? Parameter name should be `name` though, I guess?

    bool GeoAdd(RedisKey key, double longitude, double latitude, RedisValue name, CommandFlags flags = CommandFlags.None)

> Tuple would be a very bad choice

Totally agree, hating Tuples... I'd prefer a POCO, too

mgravell:
No, the parameter name is `member`: http://redis.io/commands/geoadd

In the general case, when talking about sorted sets, etc, whether it is a *name* is assuming something that needn't be the case. In examples it tends to be a name, *because that is easy to illustrate*. In reality, though, it could be arbitrary payloads that aren't "names".

I *assume* this continues to be the case with geo. In which case: no, it isn't a "name"; it is a "member".

StackExchange.Redis/StackExchange/Redis/GeoEntry.cs
wjdavis5: RIght now - now where. But it'll be swapped in for RedisValue.

StackExchange.Redis/StackExchange/Redis/IDatabase.cs
wjdavis5: This is one area where using RedisValue kinda made sense, b/c it could be either a double, or a string or nil.

mgravell: Which line (and specifically: which usage) are you referring to for "This is one area where using RedisValue kinda made sense" ? So I can understand it in context?

mgravell: Assuming you mean the return value:

> The command returns the distance as a double (represented as a string) in the specified unit, or NULL if one or both the elements are missing.

The reason it is represented as a string is simply: the redis protocol *does not have* a binary representation (I've asked Salvatore if we can get one, but: no). So: it is always a `double` or a nil value; the question there in my mind is `double?` vs `double` with `double.NaN` for nil.

MichaCo: `double?` is more clear to the user of the API. That you might have to check for double.NaN would need extra documentation (expecting the user reads it :imp: )

mgravell: > double? is more clear to the user of the API.

I'd agree with that. Sounds like a plan.

On 31 March 2016 at 13:13, MichaC <notifications@github.com> wrote:

> In StackExchange.Redis/StackExchange/Redis/IDatabase.cs
> <https://github.com/StackExchange/StackExchange.Redis/pull/381#discussion_r58042940>
> :
>
> > +        /// <remarks>http://redis.io/commands/sadd</remarks>
> > +        bool GeoAdd(RedisKey key, RedisValue value0, RedisValue value1, RedisValue value2, CommandFlags flags = CommandFlags.None);
> > +
> > +        /// <summary>
> > +        /// Add the specified members to the set stored at key. Specified members that are already a member of this set are ignored. If key does not exist, a new set is created before adding the specified members.
> > +        /// </summary>
> > +        /// <returns>the number of elements that were added to the set, not including all the elements already present into the set.</returns>
> > +        /// <remarks>http://redis.io/commands/geoadd</remarks>
> > +        long GeoAdd(RedisKey key, RedisValue[] values, CommandFlags flags = CommandFlags.None);
> > +
> > +        /// <summary>
> > +        /// Return the distance between two members in the geospatial index represented by the sorted set.
> > +        /// </summary>
> > +        /// <returns>The command returns the distance as a double (represented as a string) in the specified unit, or NULL if one or both the elements are missing.</returns>
> > +        /// <remarks>http://redis.io/commands/geodist</remarks>
> > +        RedisValue GeoDist(RedisKey key, RedisValue value0,RedisValue value1,GeoUnit geoUnit = GeoUnit.m, CommandFlags flags = CommandFlags.None);
>
> double? is more clear to the user of the API. That you might have to
> check for double.NaN would need extra documentation (expecting the user
> reads it [image: :imp:] )
>
> —
> You are receiving this because you commented.
> <https://github.com/StackExchange/StackExchange.Redis/pull/381/files/ff4ed675461ccbe6b59ad4bd8279d6edc721a60a#r58042940>
>



-- 
Regards,

Marc

StackExchange.Redis/StackExchange/Redis/GeoEntry.cs
wjdavis5: Score is valid b/c in the sorted set the geohash is set to the score But I think we'd be best removing score here, and if for some reason they want to access the score the sorted set commands could be used.

StackExchange.Redis/StackExchange/Redis/GeoUnit.cs
wjdavis5: The enum names of been updated, and a ```private string[]``` member has been implemented to translate to the redis accepted value.

wjdavis5:
This has been updated to implement several of the aforementioned suggestions / comments. Still need to:

- Implement the two missing members
- Implement the Async Methods
- Test coverage.

mgravell:
https://www.nuget.org/packages/StackExchange.Redis/1.2.0-alpha1

PR #383, opened 2016-03-31
Sentinel GET-MASTER-ADDR-BY-NAME throws exception when service name i…

aaronmell:
…s not being watched by sentinel.


mgravell:
Thanks; looks good

PR #384, opened 2016-04-01
Fixed build error caused by TryGetAzureRoleInstanceIdNoThrow test

DeRauk:
Looks like this was missed in PR #379 when moving the _TryGetAzureRoleInstanceIdNoThrow_ method from the _ConfigurationOptions_ class to the _ConnectionMultiplexer_ class.

### DIFF
--- a/StackExchange.Redis.Tests/Config.cs
+++ b/StackExchange.Redis.Tests/Config.cs
@@ -138,14 +138,7 @@ namespace StackExchange.Redis.Tests
 #endif
             }
         }
-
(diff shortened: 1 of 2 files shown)


deepakverma:
thanks, I had missed to include the updated one in the checkin. @mgravell  is it possible for CI to catch this?

mgravell:
You'd think so. I need to look at why the CI is screwed :(

mgravell:
Thanks - much appreciated

PR #385, opened 2016-04-04
Length based conditions for Transactions

awr:
For enabling length checks (==, >, <) against a variety of types.

These map to:
String -> STRLEN
List -> LLEN
Set -> SCARD
SortedSet -> ZCARD
Hash -> HLEN

### DIFF
--- a/StackExchange.Redis.Tests/Transactions.cs
+++ b/StackExchange.Redis.Tests/Transactions.cs
@@ -346,14 +346,17 @@ namespace StackExchange.Redis.Tests
                     case ComparisonType.Equal:
                         expectSuccess = valueLength == length;
                         condition = Condition.StringLengthEqual(key2, length);
+                        Assert.That(condition.ToString(), Contains.Substring("String length == " + length));
                         break;
                     case ComparisonType.GreaterThan:
                         expectSuccess = valueLength > length;
                         condition = Condition.StringLengthGreaterThan(key2, length);
+                        Assert.That(condition.ToString(), Contains.Substring("String length > " + length));
                         break;
                     case ComparisonType.LessThan:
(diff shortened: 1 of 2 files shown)


mgravell:
I like. Looks fine - not at PC, will merge later (I like to see it in an IDE too), but: appears fine at first glance. I'm curious: was this something that you had a genuine need for?

awr:
Thanks!

Yeah, but only the hash or set ones -- I added the others more for
completeness than anything else.

My use case revolves around a global semaphore, where the
length/cardinality was the # of slots. (I started with a set, then moved to
a hash for easier operational visibility.)

The point was to be able to have a known key with a collection, and
randomized values, each with a TTL. The randomized values are represented
by separate keys, but are also added to the set as long as the cardinality
is < the number o slots available on the semaphore. A separate process
deals with cleanup (timed to x minutes, but also evented, subscribed to
expired events on the randomized keys).

Originally I was using a lua script, but ran into the issue where
replication of the lua script doesn't work as I naively expected: the
script is run separately on the slaves, so the result of which randomized
key gets added is highly dependent on the ordering of the script runs.
Practically I was seeing some strange behavior with scripts. With the
changes coming in 3.2, this wouldn't have been an issue (I think).

On Mon, Apr 4, 2016 at 12:15 PM Marc Gravell <notifications@github.com>
wrote:

> I like. Looks fine - not at PC, will merge later (I like to see it in an
> IDE too), but: appears fine at first glance. I'm curious: was this
> something that you had a genuine need for?
>
> —
> You are receiving this because you authored the thread.
> <https://github.com/StackExchange/StackExchange.Redis/pull/385#issuecomment-205371928>
>
-- 
Andrew Robinson

PR #398, opened 2016-04-25
Add support for the ZCOUNT command

ricardopieper:
Hello!

This is my first pull request ever, so I'm sorry if I'm doing something wrong. We are creating a wrapper around StackExchange.Redis and we missed the ZCOUNT method. Although we don't really use it, I figured I could add the method myself, am I doing this correctly? 

I added the SortedSetCount and SortedSetCountAsync to the related interfaces and implementations, but I was wondering if I really need to add the method, or if there is a reason for not having the ZCOUNT method (and other methods as well). It seems that there are more missing methods. Anyway, here is the pull request :smile:

### DIFF
--- a/StackExchange.Redis.Tests/SortedSets.cs
+++ b/StackExchange.Redis.Tests/SortedSets.cs
@@ -1,0 +1,35 @@
+﻿using NUnit.Framework;
+
+namespace StackExchange.Redis.Tests
+{
+    [TestFixture]
+    public class SortedSets : TestBase
+    {
+        [Test]
+        public void ZCount()
+        {
+            using (var conn = Create())
+            {
+                var server = GetServer(conn);
+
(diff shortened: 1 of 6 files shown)


NickCraver:
Forgive me if this is a stupid question, but `SortedSetLength` and `SortedSetLengthAsync` already exist...what's the goal here?

ricardopieper:
Ahh damn, you're right. I thought that the ZCOUNT message on GetSortedSetLengthMessage was being used to do something unrelated, although there aren't many other "unrelated" possibilities. I didn't see that I could pass the range to the SortedSetLength method. I didn't read the source carefully.

Note to self: If it's a friday night, just wait until monday.

mgravell:
Awesome pull request, but not needed (function already exists)

PR #404, opened 2016-05-12
Add geo commands

wjdavis5:
Wanted to start this conversation fresh. 3.2 is now released with support of these commands. I am still missing the impl. of one method *GeoRadisuByMember*.

I plan on getting to that but its been slow since my need for this has been delayed.

Please offer up any comments. 

Thanks - Will.


StackExchange.Redis.Tests/TestBase.cs
```diff
@@ -135,7 +135,7 @@ public void Teardown()
         }
 
         protected const int PrimaryPort = 6379, SlavePort = 6380, SecurePort = 6381;
-        protected const string PrimaryServer = "127.0.0.1", SecurePassword = "changeme", PrimaryPortString = "6379", SlavePortString = "6380", SecurePortString = "6381";
+        protected const string PrimaryServer = "192.168.10.120", SecurePassword = "changeme", PrimaryPortString = "6379", SlavePortString = "6380", SecurePortString = "6381";
```
mgravell: we may need to think of more dev-friendly way of specifying the remote IP... that works on multiple machines, etc

StackExchange.Redis/StackExchange/Redis/GeoEntry.cs
```diff
@@ -0,0 +1,284 @@
...
+    public enum GeoRadiusOptions
+    {
+        /// <summary>
+        /// No Options
+        /// </summary>
+        None = 0,
+        /// <summary>
+        /// Redis will return the coordinates of any results.
+        /// </summary>
+        WITHCOORD = 1,
```
mgravell: I would prefer C# idiomatic naming and casing here; it is absolutely not required that the enum name is *anything at all* like the redis instruction. Elsewhere in the library, I've gone for C# idiomatic over redis protocol names, and simply mapped between them.

StackExchange.Redis/StackExchange/Redis/GeoEntry.cs
```diff
@@ -0,0 +1,284 @@
...
+        /// </summary>
+        WITHHASH = 4
+    }
+
+    /// <summary>
+    /// The result of a GeoRadius command.
+    /// </summary>
+    public struct GeoRadiusResult
+    {
+        internal readonly RedisValue member;
```
mgravell: why internal?

StackExchange.Redis/StackExchange/Redis/GeoEntry.cs
```diff
@@ -0,0 +1,284 @@
...
+        internal readonly GeoRadius geoRadius;
+        internal readonly double? distanceFromCenter;
+        internal readonly long? geoHash;
+        internal readonly GeoPosition? geoPosition;
+
+        /// <summary>
+        /// The matched member.
+        /// </summary>
+        public RedisValue Member => member;
```
mgravell: I wonder if all of these should just be `{get;}` (and remove the fields); note that I haven't read end-to-end yet, so I don't have much context of what this type is doing yet

StackExchange.Redis/StackExchange/Redis/GeoEntry.cs
```diff
@@ -0,0 +1,284 @@
...
+        }
+
+    }
+
+    /// <summary>
+    /// Represents a GeoRadius command and its options.
+    /// </summary>
+    public class GeoRadius
+    {
+        internal readonly GeoUnit geoUnit;
```
mgravell: ditto `internal` and `{get;}`

StackExchange.Redis/StackExchange/Redis/GeoEntry.cs
```diff
@@ -0,0 +1,284 @@
...
+        }
+
+    }
+
+    /// <summary>
+    /// Describes the longitude and latitude of a GeoEntry
+    /// </summary>
+    public struct GeoPosition
+    {
+        internal readonly double longitude;
```
mgravell: ditto... (tritto?)

StackExchange.Redis/StackExchange/Redis/GeoEntry.cs
```diff
@@ -0,0 +1,284 @@
...
+        internal readonly double radius;
+
+        /// <summary>
+        /// The Radius size of this GeoRadius command.
+        /// </summary>
+        public double Radius => radius;
+        /// <summary>
+        /// The center point to base the search.
+        /// </summary>
+        public GeoPosition GeoPosition => geoPosition;
```
mgravell: but what *is* the position? is this the center? if so: `Center` or `Origin` might be a better name.

StackExchange.Redis/StackExchange/Redis/GeoEntry.cs
```diff
@@ -0,0 +1,284 @@
...
+        internal readonly GeoPosition? geoPosition;
+
+        /// <summary>
+        /// The matched member.
+        /// </summary>
+        public RedisValue Member => member;
+        /// <summary>
+        /// The original GeoRadius command
+        /// </summary>
+        public GeoRadius GeoRadius => geoRadius;
```
mgravell: if this is the command, I wonder if `Command` is a clearer name

StackExchange.Redis/StackExchange/Redis/GeoEntry.cs
```diff
@@ -0,0 +1,284 @@
...
+        /// </summary>
+        public GeoPosition GeoPosition => geoPosition;
+        /// <summary>
+        /// The key to use.
+        /// </summary>
+        public RedisKey Key => key;
+        /// <summary>
+        /// The unit to return distance measurments in.
+        /// </summary>
+        public GeoUnit GeoUnit => geoUnit;
```
mgravell: name = `Unit` surely? the fact that it happens to *be* a `GeoUnit` is separate

StackExchange.Redis/StackExchange/Redis/GeoEntry.cs
```diff
@@ -0,0 +1,284 @@
...
+        public override string ToString()
+        {
+            return string.Format("{0} {1}", longitude, latitude);
+        }
+        /// <summary>
+        /// See Object.GetHashCode()
+        /// </summary>
+        public override int GetHashCode()
+        {
+            return longitude.GetHashCode() ^ latitude.GetHashCode();
```
mgravell: minor, but `^` tends to have problems with diagonals; would prefer a different implementation, but OTOH, I guess diagonals aren't going to come up *that* often...

StackExchange.Redis/StackExchange/Redis/GeoEntry.cs
```diff
@@ -0,0 +1,284 @@
...
+            this.geoRadiusOptions = geoRadiusOptions;
+            this.maxReturnCount = maxReturnCount;
+        }
+
+    }
+
+    /// <summary>
+    /// Describes the longitude and latitude of a GeoEntry
+    /// </summary>
+    public struct GeoPosition
```
mgravell: it looks like this should `: IEquatable<GeoPosition>`

StackExchange.Redis/StackExchange/Redis/GeoEntry.cs
```diff
@@ -0,0 +1,284 @@
...
+        public override bool Equals(object obj)
+        {
+            return obj is GeoPosition && Equals((GeoPosition)obj);
+        }
+        /// <summary>
+        /// Compares two values for equality
+        /// </summary>
+        public bool Equals(GeoPosition value)
+        {
+            return Equals(longitude, value.Longitude) && Equals(latitude, value.latitude);
```
mgravell: unless I'm missing something (reading in github, not the IDE), this is a boxing operation (4 boxes). Very inefficient. Should be just:

    return this.longitude == value.longitude && this.latitude == value.latitude;

Also; it may be desirable to only define the *implementation* of equality once; so `Equals` could also be written:

    return this == other;

(using the `==` operator below)

StackExchange.Redis/StackExchange/Redis/GeoEntry.cs
```diff
@@ -0,0 +1,284 @@
...
+        }
+        /// <summary>
+        /// Compares two values for equality
+        /// </summary>
+        public static bool operator ==(GeoPosition x, GeoPosition y)
+        {
+            return Equals(x.longitude, y.longitude) && Equals(x.latitude, y.latitude);
```
mgravell: ditto boxing and "write once"

StackExchange.Redis/StackExchange/Redis/GeoEntry.cs
```diff
@@ -0,0 +1,284 @@
...
+        }
+        /// <summary>
+        /// Compares two values for non-equality
+        /// </summary>
+        public static bool operator !=(GeoPosition x, GeoPosition y)
+        {
+            return !Equals(x.longitude, y.longitude) || !Equals(x.latitude, y.latitude);
```
mgravell: tritto boxing

StackExchange.Redis/StackExchange/Redis/GeoEntry.cs
```diff
@@ -0,0 +1,284 @@
...
+        }
+    }
+
+    /// <summary>
+    /// Describes a GeoEntry element with the corresponding value
+    /// GeoEntries are stored in redis as SortedSetEntries
+    /// </summary>
+    public struct GeoEntry : IEquatable<GeoEntry>
+    {
+        internal readonly RedisValue member;
```
mgravell: why `internal`?

StackExchange.Redis/StackExchange/Redis/GeoEntry.cs
```diff
@@ -0,0 +1,284 @@
...
+        /// </summary>
+        public GeoPosition GeoPosition => geoPos;
+       
+
+        /// <summary>
+        /// See Object.ToString()
+        /// </summary>
+        public override string ToString()
+       {
+           return string.Format("{0} {1} {2}", geoPos.Latitude, geoPos.Latitude, member);
```
mgravell: might be clearer to do something that looks a: co-ordinates, and b: key-value; for example `return $"({Longitude},{Latitude})={member}";` - also, note you use `Latitude` twice

StackExchange.Redis/StackExchange/Redis/GeoEntry.cs
```diff
@@ -0,0 +1,284 @@
...
+        {
+            return !Equals(x.longitude, y.longitude) || !Equals(x.latitude, y.latitude);
+        }
+    }
+
+    /// <summary>
+    /// Describes a GeoEntry element with the corresponding value
+    /// GeoEntries are stored in redis as SortedSetEntries
+    /// </summary>
+    public struct GeoEntry : IEquatable<GeoEntry>
```
mgravell: does this need to be `IEquatable<GeoEntry>`? seems unlikely; I'm not saying it doesn't - it just surprised me is all

StackExchange.Redis/StackExchange/Redis/GeoEntry.cs
```diff
@@ -0,0 +1,284 @@
...
+        public override bool Equals(object obj)
+        {
+            return obj is GeoEntry && Equals((GeoEntry)obj);
+        }
+        /// <summary>
+        /// Compares two values for equality
+        /// </summary>
+        public bool Equals(GeoEntry value)
+        {
+            return Equals(geoPos, value.geoPos) && member == value.member;
```
mgravell: boxing?

StackExchange.Redis/StackExchange/Redis/GeoEntry.cs
```diff
@@ -0,0 +1,284 @@
...
+        }
+        /// <summary>
+        /// Compares two values for equality
+        /// </summary>
+        public static bool operator ==(GeoEntry x, GeoEntry y)
+        {
+            return Equals(x.geoPos,y.geoPos) && x.member == y.member;
```
mgravell: boxing?

StackExchange.Redis/StackExchange/Redis/GeoEntry.cs
```diff
@@ -0,0 +1,284 @@
...
+        }
+        /// <summary>
+        /// Compares two values for non-equality
+        /// </summary>
+        public static bool operator !=(GeoEntry x, GeoEntry y)
+        {
+            return !Equals(x.geoPos , y.geoPos) || x.member != y.member;
```
mgravell: boxing?

StackExchange.Redis/StackExchange/Redis/GeoUnit.cs
```diff
@@ -0,0 +1,28 @@
...
+    public enum GeoUnit
+    {
+        /// <summary>
+        /// Meters
+        /// </summary>
+        Meters,
+        /// <summary>
+        /// Kilometers
+        /// </summary>
+        KiloMeters,
```
mgravell: `Kilometers`

StackExchange.Redis/StackExchange/Redis/RedisCommand.cs
```diff
@@ -38,6 +38,13 @@ enum RedisCommand
         FLUSHALL,
         FLUSHDB,
 
+        GEOADD,
```
mgravell: I haven't checked yet, but note that some of these look fine for replica, some look master-only; there's a method somewhere that defines which is which (needs to be done for command routing purposes)

StackExchange.Redis/StackExchange/Redis/RedisDatabase.cs
```diff
@@ -45,6 +47,134 @@ public RedisValue DebugObject(RedisKey key, CommandFlags flags = CommandFlags.No
...
+            var msg = Message.Create(Database,flags, RedisCommand.GEOADD, key, geoEntries);
+            return ExecuteSync(msg, ResultProcessor.Int64);
+        }
+
+        public bool GeoRemove(RedisKey key, RedisValue member, CommandFlags flags = CommandFlags.None)
+        {
+            return SortedSetRemove(key, member, flags);
+        }
+
+        public RedisValue GeoDistance(RedisKey key, RedisValue value0, RedisValue value1, GeoUnit geoUnit = GeoUnit.Meters,CommandFlags flags = CommandFlags.None)
```
mgravell: result here should be a `double`, not a `RedisValue`. It never has to contain the string/number/blob triple that `RedisValue` is intended to represent. It is always just a double, AFAIK

StackExchange.Redis/StackExchange/Redis/RedisDatabase.cs
```diff
@@ -45,6 +47,134 @@ public RedisValue DebugObject(RedisKey key, CommandFlags flags = CommandFlags.No
...
+        {
+            var msg = Message.Create(Database, flags, RedisCommand.GEODIST, key,value0,value1,_redisUnits[(int)geoUnit]);
+            return ExecuteSync(msg, ResultProcessor.RedisValue);
+        }
+
+        public string[] GeoHash(RedisKey key, string[] members, CommandFlags flags = CommandFlags.None)
```
mgravell: `members` should be `RedisValue[]`, not `string[]`

StackExchange.Redis/StackExchange/Redis/RedisDatabase.cs
```diff
@@ -45,6 +47,134 @@ public RedisValue DebugObject(RedisKey key, CommandFlags flags = CommandFlags.No
...
+            var results = ExecuteSync(msg, ResultProcessor.RedisValueArray);
+            var retArray = new string[results.Length];
+            for (int i = 0; i < results.Length; i++)
+            {
+                retArray[i] = (string)results[i];
+            }
+            return retArray;
+        }
+
+        public string[] GeoHash(RedisKey key, string member, CommandFlags flags = CommandFlags.None)
```
mgravell: should be `public string GeoHash(RedisKey key, RedisValue member, ...` surely? The `RedisValue` as above, but since we're only expecting one, we should only return `string`, not `string[]`?

StackExchange.Redis/StackExchange/Redis/RedisDatabase.cs
```diff
@@ -45,6 +47,134 @@ public RedisValue DebugObject(RedisKey key, CommandFlags flags = CommandFlags.No
...
+        {
+            return GeoHash(key, new[] {member}, flags);
+        }
+
+        public GeoPosition?[] GeoPos(RedisKey key, string[] members, CommandFlags flags = CommandFlags.None)
```
mgravell: `RedisValue[] members`

StackExchange.Redis/StackExchange/Redis/RedisDatabase.cs
```diff
@@ -45,6 +47,134 @@ public RedisValue DebugObject(RedisKey key, CommandFlags flags = CommandFlags.No
...
+        {
+            if (members == null) throw new ArgumentNullException(nameof(members));
+            var redisValues = new RedisValue[members.Length];
+            for (var i = 0; i < members.Length; i++) redisValues[i] = members[i];
+            var msg = Message.Create(Database, flags, RedisCommand.GEOPOS, key, redisValues);
+            return ExecuteSync(msg, ResultProcessor.RedisGeoPosition);
+        }
+
+        public GeoPosition? GeoPos(RedisKey key, string member, CommandFlags flags = CommandFlags.None)
```
mgravell: `RedisValue` member

StackExchange.Redis/StackExchange/Redis/RedisDatabase.cs
```diff
@@ -45,6 +47,134 @@ public RedisValue DebugObject(RedisKey key, CommandFlags flags = CommandFlags.No
...
+        {
+            return GeoPos(key, new string[] {member}, flags).FirstOrDefault();
+        }
+
+        public GeoRadiusResult[] GeoRadius(RedisKey key, GeoRadius geoRadius,CommandFlags flags = CommandFlags.None)
```
mgravell: API choice question: it seems unexpected to me that `GeoRadius` *exists*; is there a reason for it? It fees like this should be:

    public GeoRadiusResult[] GeoRadius(RedisKey key, GeoPosition position, double radius, GeoUnit unit = GeoUnit.Meters, GeoRadiusOptions options = {some default}, CommandFlags flags = CommandFlags.None)

I'm not saying it is right or wrong; I just want to understand what we gain by adding `GeoRadius` as a concept...

StackExchange.Redis/StackExchange/Redis/RedisDatabase.cs
```diff
@@ -45,6 +47,134 @@ public RedisValue DebugObject(RedisKey key, CommandFlags flags = CommandFlags.No
...
+        {
+            var redisValues = new List<RedisValue>();
+            redisValues.Add(geoRadius.geoPosition.longitude);
+            redisValues.Add(geoRadius.geoPosition.latitude);
+            redisValues.Add(geoRadius.radius);
+            redisValues.Add(_redisUnits[(int)geoRadius.geoUnit]);
+            if(geoRadius.geoRadiusOptions.HasFlag(GeoRadiusOptions.WITHCOORD))
+                redisValues.Add(GeoRadiusOptions.WITHCOORD.ToString());
```
mgravell: this would be where we apply a map of C# name to redis name on the options

StackExchange.Redis/StackExchange/Redis/RedisDatabase.cs
```diff
@@ -45,6 +47,134 @@ public RedisValue DebugObject(RedisKey key, CommandFlags flags = CommandFlags.No
...
+            if (geoRadius.geoRadiusOptions.HasFlag(GeoRadiusOptions.WITHDIST))
+                redisValues.Add(GeoRadiusOptions.WITHDIST.ToString());
+            if (geoRadius.geoRadiusOptions.HasFlag(GeoRadiusOptions.WITHHASH))
+                redisValues.Add(GeoRadiusOptions.WITHHASH.ToString());
+            if(geoRadius.maxReturnCount >0)
+                redisValues.Add(geoRadius.maxReturnCount);
+            var msg = Message.Create(Database, flags, RedisCommand.GEORADIUS, key, redisValues.ToArray());
+            var items = ExecuteSync(msg, ResultProcessor.RedisGeoRadius);
+
+            var arr = new GeoRadiusResult[items.Length];
```
mgravell: it looks to me like a lot of this should be moved to the processor - in particular, this would avoid having to duplicate it (or refactor it) for the `async` version later

mgravell:
Awesome; I've gone through and added some comments throughout; let me know your thoughts. My biggest query at the moment would be over the `GeoRadius` - "should it exist". Really appreciate the thought here. The mass of comments isn't intended as a negative criticism thing; I'm just trying to make sure we get it right, because it is very hard to change after the fact.

StackExchange.Redis/StackExchange/Redis/RedisDatabase.cs
```diff
@@ -45,6 +47,134 @@ public RedisValue DebugObject(RedisKey key, CommandFlags flags = CommandFlags.No
...
+        }
+
+        public GeoRadiusResult[] GeoRadius(RedisKey key, GeoRadius geoRadius,CommandFlags flags = CommandFlags.None)
+        {
+            var redisValues = new List<RedisValue>();
+            redisValues.Add(geoRadius.geoPosition.longitude);
+            redisValues.Add(geoRadius.geoPosition.latitude);
+            redisValues.Add(geoRadius.radius);
+            redisValues.Add(_redisUnits[(int)geoRadius.geoUnit]);
+            if(geoRadius.geoRadiusOptions.HasFlag(GeoRadiusOptions.WITHCOORD))
```
mgravell: never use `HasFlag`. It is a boxing operation **and** has to do reflection internally (to check the types are compatible). For a simple "has this single bit" test:

    if((val & MyEnum.TheFlagWeWant) != 0) {....}

so:

    if((geoRadius.options & GeoRadiusOptions.WITHCOORD) != 0) {...}

You could of course add something like:

    private static bool HasFlag(GeoRadiusOptions value, GeoRadiusOptions flag) {
        return (value & flag) != 0;
    }

but note that you can't make it an extension method that monkey-patches `HasFlag`, as the resolution order will always prefer the declared instance method over an extension method.

StackExchange.Redis/StackExchange/Redis/RedisDatabase.cs
```diff
@@ -45,6 +47,134 @@ public RedisValue DebugObject(RedisKey key, CommandFlags flags = CommandFlags.No
...
+                RawResult[] item = items[i].GetArrayOfRawResults();
+                if (item == null)
+                {
+                    continue;
+                }
+                else
+                {
+                    var x = 0;
+                    var member = item[x++].AsRedisValue();
+                    var distance = geoRadius.geoRadiusOptions.HasFlag(GeoRadiusOptions.WITHDIST) ? new double?((double)item[x++].AsRedisValue()) : null;
```
mgravell: `HasFlag`

StackExchange.Redis/StackExchange/Redis/GeoEntry.cs
wjdavis5: How do you feel about Point?

StackExchange.Redis.Tests/TestBase.cs
chester89: Maybe an env. variable?

mgravell:
https://www.nuget.org/packages/StackExchange.Redis/1.2.0-alpha1

PR #407, opened 2016-05-19
SE.Redis better error in the exception message by including innerexce…

deepakverma:
…ption details
After this change it would look something like this:
No connection is available to service this operation: SET test ; The remote certificate is invalid according to the validation procedure.


StackExchange.Redis/StackExchange/Redis/ExceptionFactory.cs
```diff
@@ -74,22 +74,46 @@ internal static Exception MultiSlot(bool includeDetail, Message message)
...
                 serverSnapshot = new ServerEndPoint[] { server };
             }
-            string exceptionmessage = "No connection is available to service this operation: " + s ;
-            var ex = new RedisConnectionException(ConnectionFailureType.UnableToResolvePhysicalConnection, exceptionmessage, GetServerSnapshotInnerExceptions(serverSnapshot));
+
+            var innerException = GetServerSnapshotInnerExceptions(serverSnapshot);
+
+            StringBuilder exceptionmessage = new StringBuilder("No connection is available to service this operation: ");
+            exceptionmessage.Append(commandLabel);
+            exceptionmessage.Append("; ");
```
NickCraver: I'd get `GetInnerMostExceptionMessage(innerException)` in a variable, and if it's not `""`, then append. Right now we'd get a trailing `; ` in such cases.

deepakverma: @NickCraver thanks for reviewing the pull request, I have fixed the trailing ;

PR #408, opened 2016-05-20
.Net Core RC2 upgrades to for a 1.604-alpha release

NickCraver:
- RC2 project.json format and references
- NuGet.config added since we're on one feed now
- Unit tests minimally running in core - lots of work left here
- Bumps to netstandard1.5 for BufferedStream, but we have net40, net45,
and net46 builds already so this should be low impact.
- AuthenticateAsClient() (synchronous) is no longer in .Net Core, Async
version must be used there now.
- Version bump to `1.604-*` (alpha suffix is in via dotnet pack now)
- To pack:
```
dotnet pack StackExchange.Redis -o .nupkg -c Release --version-suffix alpha
dotnet pack StackExchange.Redis.StrongName -o .nupkg -c Release --version-suffix alpha
```


BasicTest/Program.cs
```diff
@@ -35,12 +35,12 @@ static void MassiveBulkOpsAsync(int AsyncOpsQty, bool preserveOrder, bool withCo
...
                 muxer.Wait(conn.PingAsync());
 
-#if DNXCORE50
+#if CORE_CLR
                 int number = 0;
 #endif
                 Action<Task> nonTrivial = delegate
                 {
-#if !DNXCORE50
+#if !CORE_CLR
```
agocke: Why not just use the loop and `SpinWait.SpinOnce()`? I think it's netstd1.3, so should be available everywhere.

NickCraver: @agocke I don't have full context here, but the initial port noted it took far more time - has this changed? Link: https://github.com/StackExchange/StackExchange.Redis/commit/e10c20e2351900e3f8d7d88183bd479dd76b3c32

BasicTest_dnxcore50/project.json
```diff
@@ -1,32 +1,31 @@
...
   "commands": {
     "run": "BasicTest_dnxcore50"
   },
-
-  
   "frameworks": {
-    "dnxcore50": {
+    "netcoreapp1.0": {
+      "buildOptions": {
+        "define": [ "CORE_CLR" ]
```
pranavkm: The compiler defines a variable named `NETCOREAPP1_0` in case you were looking for it.

BasicTest/Program.cs
agocke: @NickCraver I see -- yeah, don't worry about it. If what you have works no point in messing around unnecessarily.

BasicTest_dnxcore50/project.json
NickCraver: Yep - thanks! I'm aware of that one, but the define has also changed every time we re-visit it through the pre-releases. For the moment, I think it's better to opt for our self-declared defines until RTM locks them in (to eliminate some git churn).

NickCraver:
No showstoppers found - proceeding with an alpha.

NickCraver:
I just pushed `1.604-alpha` to NuGet:
- https://www.nuget.org/packages/StackExchange.Redis/1.1.604-alpha
- https://www.nuget.org/packages/StackExchange.Redis.StrongName/1.1.604-alpha

Please let me know if you see any issues!

PR #412, opened 2016-05-24
Fixing Documentation for ConfigurationOptions.Proxy

steroberts89:
Spotted this whilst using the library, thought I'd contribute a fix!

The Intellisense documentation for the Proxy Property was the same as AllowAdmin

Probably error through Copy and Paste.


NickCraver:
Good catch, thanks!

PR #414, opened 2016-05-26
Spotted some incorrect details about setting processModel

steroberts89:
I was looking to try to set my minIoThreads and it appears that the processModel is only settable in the Machine.config by default (Read the documentation linked in https://msdn.microsoft.com/en-us/library/vstudio/7w2sway1(v=vs.100).aspx ). 
The documentation indicated otherwise. In revelation of this, It may be an idea to suggest setting the value via code first!.


NickCraver:
Merged in via 208865c43a8d24eb372d375883e442f01a65cdb4

Issue #424, opened 2016-06-17
NullReferenceException on IDatabase calls


Rookian:
```
const string name = "xxxx";
const string key = "xxxx";
var connection = $"{name}.redis.cache.windows.net,abortConnect=false,ssl=true,password={key}";

var redisManager = ConnectionMultiplexer.Connect(connection);

var database = redisManager.GetDatabase();
database.StringSet("34", "343"); // Throws NullRefEx
var keyExists = database.KeyExists("xeer"); // Throws NullRefEx
```

**Stacktrace:**
```
System.NullReferenceException was unhandled by user code
  HResult=-2147467261
  Message=Object reference not set to an instance of an object.
  Source=StackExchange.Redis
  StackTrace:
       at StackExchange.Redis.ServerEndPoint.get_LastException() in C:\TeamCity\buildAgent\work\3ae0647004edff78\StackExchange.Redis\StackExchange\Redis\ServerEndPoint.cs:line 102
       at StackExchange.Redis.ExceptionFactory.GetServerSnapshotInnerExceptions(ServerEndPoint[] serverSnapshot) in C:\TeamCity\buildAgent\work\3ae0647004edff78\StackExchange.Redis\StackExchange\Redis\ExceptionFactory.cs:line 99
       at StackExchange.Redis.ExceptionFactory.NoConnectionAvailable(Boolean includeDetail, RedisCommand command, Message message, ServerEndPoint server, ServerEndPoint[] serverSnapshot) in C:\TeamCity\buildAgent\work\3ae0647004edff78\StackExchange.Redis\StackExchange\Redis\ExceptionFactory.cs:line 88
       at StackExchange.Redis.ConnectionMultiplexer.ExecuteSyncImpl[T](Message message, ResultProcessor`1 processor, ServerEndPoint server) in C:\TeamCity\buildAgent\work\3ae0647004edff78\StackExchange.Redis\StackExchange\Redis\ConnectionMultiplexer.cs:line 1944
       at StackExchange.Redis.RedisBase.ExecuteSync[T](Message message, ResultProcessor`1 processor, ServerEndPoint server) in C:\TeamCity\buildAgent\work\3ae0647004edff78\StackExchange.Redis\StackExchange\Redis\RedisBase.cs:line 81
[...]

PR #426, opened 2016-06-21
fix NullReferenceException when retriving LastException from ServerEn…

u-abramchuk:
…dPoint

This is a fix for issue [#424](https://github.com/StackExchange/StackExchange.Redis/issues/424).


mgravell:
Thanks

PR #428, opened 2016-06-23
Give specific cause of timeouts for known ThreadPool and CPU cases.

JonCole:
This changes the timeout error message to give better guidance around what may have caused the timeout.  Example timeout error message after this change:

> System.TimeoutException: Timeout performing GET q. The number of busy IOCP or WORKER threads in the ThreadPool is greater than the Min setting, which could easily be the cause of this timeout.  See https://github.com/StackExchange/StackExchange.Redis/blob/master/Docs/Timeouts.md#are-you-seeing-high-number-of-busyio-or-busyworker-threads-in-the-timeout-exception for details on how ThreadPool Growth Throttling can affect performance. Diagnostics Info: inst: 2, mgr: Inactive, err: never, queue: 6, qu: 0, qs: 6, qc: 0, wr: 0, wq: 0, in: 51750, ar: 0, clientName: myComputer, IOCP: (Busy=1,Free=999,Min=1,Max=1000), WORKER: (Busy=4,Free=32763,Min=1,Max=32767), Local-CPU: 0%

### DIFF
--- a/StackExchange.Redis/StackExchange/Redis/ConnectionMultiplexer.cs
+++ b/StackExchange.Redis/StackExchange/Redis/ConnectionMultiplexer.cs
@@ -611,8 +611,9 @@ namespace StackExchange.Redis
             {
                 var sb = new StringBuilder();
                 sb.Append(message);
-                string iocp, worker;
-                busyWorkerCount = GetThreadPoolStats(out iocp, out worker);
+                ThreadPoolStats iocp, worker;
+                ThreadPoolStats.Gather(out iocp, out worker);
+                busyWorkerCount = worker.Busy;
                 sb.Append(", IOCP: ").Append(iocp).Append(", WORKER: ").Append(worker);
                 LogLocked(log, sb.ToString());
             }
@@ -1990,14 +1991,25 @@ namespace StackExchange.Redis
                             var lastError = socketManager.LastErrorTimeRelative();
 
 #endif
-                            var sb = new StringBuilder("Timeout performing ").Append(message.CommandAndKey);
+                            var sb = new StringBuilder($"Timeout performing {message.CommandAndKey}.");
                             data = new List<Tuple<string, string>> {Tuple.Create("Message", message.CommandAndKey)};
+                            bool firstAdd = true;
                             Action<string, string, string> add = (lk, sk, v) =>
                             {
                                 data.Add(Tuple.Create(lk, v));
-                                sb.Append(", " + sk + ": " + v);
+                                if (!firstAdd) 
+                                    sb.Append(", ");
+                                sb.Append(sk + ": " + v);
+                                firstAdd = false;
                             };
 
+#if !CORE_CLR
+                            ThreadPoolStats iocp, worker;
+                            string localCpuPercent;
+                            AddTimeoutRootCauseIfPossible(sb, out iocp, out worker, out localCpuPercent);
+#endif
+
+                            sb.Append(" Diagnostics Info: ");
                             int queue = server.GetOutstandingCount(message.Command, out inst, out qu, out qs, out qc, out wr, out wq, out @in, out ar);
                             add("Instantaneous", "inst", inst.ToString());
 #if FEATURE_SOCKET_MODE_POLL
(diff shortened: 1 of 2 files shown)


StackExchange.Redis/StackExchange/Redis/ConnectionMultiplexer.cs
```diff
@@ -1990,14 +1991,29 @@ internal T ExecuteSyncImpl<T>(Message message, ResultProcessor<T> processor, Ser
                             var lastError = socketManager.LastErrorTimeRelative();
 
 #endif
-                            var sb = new StringBuilder("Timeout performing ").Append(message.CommandAndKey);
+                            var sb = new StringBuilder($"Timeout performing {message.CommandAndKey}.");
```
NickCraver: What's the reasoning for these changes? This increases allocations.

StackExchange.Redis/StackExchange/Redis/ConnectionMultiplexer.cs
```diff
@@ -2058,37 +2069,60 @@ internal T ExecuteSyncImpl<T>(Message message, ResultProcessor<T> processor, Ser
...
+            if (server.ConnectionState != PhysicalBridge.State.ConnectedEstablished)
             {
-                return Math.Round(systemCPU, 2) + "%";
+                sb.Append($" The connection to the server was lost and was not restored in the timeout specified.");
+                return; // don't add IOCP and CPU details to the error message because connection management trumps all other causes of timeouts.
+            }
+
+            bool detailsAdded = false;
+            bool timeoutHelpLinkAdded = false;
+            if (iocp.Busy > iocp.Min || worker.Busy > worker.Min)
```
NickCraver: This isn't really an accurate check for any kind of high concurrency environment - I'd question if this is actually a good change. For example on Stack Overflow this would be a constant false positive.

StackExchange.Redis/StackExchange/Redis/ConnectionMultiplexer.cs
```diff
@@ -2058,37 +2069,60 @@ internal T ExecuteSyncImpl<T>(Message message, ResultProcessor<T> processor, Ser
...
+                timeoutHelpLinkAdded = true;
+                sb.Append($" The number of busy IOCP or WORKER threads in the ThreadPool is greater than the 'Min' setting, which could easily be the cause of this timeout.  See https://github.com/StackExchange/StackExchange.Redis/blob/master/Docs/Timeouts.md#are-you-seeing-high-number-of-busyio-or-busyworker-threads-in-the-timeout-exception for details on how ThreadPool Growth Throttling can affect performance.");
+            }
+            
+            if (gotCpuUsageSuccessfully)
+            {
+                localCpuPercent = Math.Round(systemCPU, 2) + "%";
+                if (systemCPU > 85)
+                {
+                    if (detailsAdded)
```
NickCraver: I'd simplify this - the "Additionally, " isn't really needed and the code is much simpler without the differentiation.

StackExchange.Redis/StackExchange/Redis/DebuggingAids.cs
```diff
@@ -299,6 +299,53 @@ public enum CompletionType
     }
 #if !CORE_CLR
 
+    internal class ThreadPoolStats
```
NickCraver: This should be a struct without setters (allocations).

StackExchange.Redis/StackExchange/Redis/ConnectionMultiplexer.cs
JonCole: Fixed

StackExchange.Redis/StackExchange/Redis/DebuggingAids.cs
JonCole: Fixed.

StackExchange.Redis/StackExchange/Redis/ConnectionMultiplexer.cs
JonCole: Will revert

JonCole: Will revert.

StackExchange.Redis/StackExchange/Redis/ConnectionMultiplexer.cs
JonCole: The goal of the check in is to give customers who don't have any experience with debugging these types of errors a better clue as to where to start.  We see many customers come to us to help them diagnose timeouts.  Customers typically fall into two buckets - those who understand how the ThreadPool scales and those that don't.  

For those that do understand this, then a false lead is something they probably have sufficient skill to diagnose themselves and they will be able to look past any false lead.  

However, for those that don't understand the thread pool very well, this is a great hint at a common cause of timeouts.  We help many customers per week adjust their threadpool settings and this fixes the issue for almost all of them.

I would argue that even for a high concurrency system, you should probably adjust your threadpool settings to avoid the growth throttling costs so that you never hit these types of timeouts...

JonCole:
@NickCraver - have I addressed all your concerns?  We see a lot of customer support cases for Azure Redis around timeouts, so getting better error messages would help people get themselves unblocked sooner and would be a win for everyone.  Would be happy to discuss over e-mail or skype if you have further questions.

JonCole:
@NickCraver @mgravell Any further questions on this PR?  As mentioned previously, I would be happy to have a Skype, e-mail or phone conversation about this in detail with you.  I help 70-80 Redis customers a month understand why they are getting timeouts, most of which are caused by incorrect ThreadPool settings or high client side CPU usage.  While there are many causes to timeouts, these are the most common ones I see.

JonCole:
Will re-submit pull request from a different branch in my fork, so that this isn't holding up my master branch.

PR #436, opened 2016-06-29
Fixing comparison of node endpoints when configuring cluster

pbeardshear:
In `ClusterNodesProcessor.Parse`, the server endpoint that `SetClusterConfiguration` is called with won't be the same endpoint object in any of the `ClusterConfiguration.Nodes`, so the reference equality check on line 217 will always return false.

This is because when a new ClusterNode is created, it creates a new endpoint by parsing the raw configuration string

`EndPoint = Format.TryParseEndPoint(parts[1]);`

This causes issues in the cluster when using `CommandFlags.PreferSlave` or `CommandFlags.DemandSlave`, since node.Slaves is always an empty array.


NickCraver:
Looked like this was merged in via #440 with another PR batch, closing out to cleanup but we appreciate the fix and info!

PR #438, opened 2016-07-05
Configuring the minIoThreads

IlincescuMihai:
You can't set the processModel in the web.config  "The processModel section can be set only within the Machine.config" and is confusing to have here to be web.config


NickCraver:
Thanks!

PR #440, opened 2016-07-05
check the endpoints are actually equal to populate the cluster master…

deepakverma:
… and slave nodes correctly
repro:

``` C#
cm.GetDatabase().StringSet("test", "test");
for (int i = 0; i < 10; i++)
{
    try
    {
        Console.WriteLine(cm.GetDatabase().StringGet("test", CommandFlags.DemandSlave));
    }
    catch(Exception e)
    {
        Console.WriteLine("?");
    }
}
```

the get call would fail to get the slave node and end up selecting anynode, which results in moved exception.


deepakverma:
ping

deepakverma:
This pull request fixes a bug for the scenario in the repro, is anyone else not running into it?

mgravell:
Looks good to me, thanks; sorry for the delay

PR #443, opened 2016-07-07
Updating to allow the creation of RedisResults for mocking

nightgoat:
I've spoken with @mgravell a bit about this change via email already. In summary, it is impossible to create mock RedisResult objects due to the internal access modifiers on the abstract class methods. By making the ResultType enum public, I was able to create a static method that allows users to create RedisResult instances of any supported type. This is helpful for me when injecting results of lua script evaluation, but may be useful elsewhere as well.


mgravell:
Well that looks deceptively simple! I'd be more than happy to take that as
a "step zero", but (and I'm not suggesting this is for you to do) we should
perhaps also add some overloads for common types - presumably not even
needing the enum. Or maybe the trick is to take RediaValue instead of
byte[], which would allow a lot of types to be used automatically.

nightgoat:
I think you're right. Simply accepting a RedisValue object works great, doesn't require any changes to the access modifiers, and keeps us from having to deal with byte[]! Here's some example Moq code now:

```C#
var database = new Mock<IDatabase>();
database
	.Setup(db => db.ScriptEvaluateAsync(It.IsAny<LuaScript>(), It.IsAny<object>(), CommandFlags.None))
	.Returns(() => Task.FromResult(RedisResult.Create("1")));
```

I think having the implicit operators in RedisValue may just eliminate the need for separate type overloads.

NickCraver:
Can you please fix the drastic whitespace changes here? While ?w=1 works for PR viewing, it still hoses git history quite a bit with the unnecessary changes.

PR #451, opened 2016-07-22
Surface better error instead of MOVED errors during a network disconnect in redis clusters 

hrishi18pathak:


mgravell:
Looks good, thanks

Issue #321, opened 2015-12-16
Redis Pub/Sub subscriber to slave server


bp74:
Hello

I have an unusual use case where i want to do a subscription to a Redis slave server . A subscription to a slave server works fine but the Redis library does not allow it, and the reason is this line of code:

https://github.com/StackExchange/StackExchange.Redis/blob/master/StackExchange.Redis/StackExchange/Redis/RedisSubscriber.cs#L190

It basically overrules the CommandFlags passed to the method and uses DemandMaster. This makes sense if you want to publish messages, but if you only want to listen as a subscriber this is not necessary (i think). Would it be possible to just use the CommandFlags that are passed to the method? If i do this and by using PreferMaster it works perfectly fine because it also connects to a slave.

mgravell:
I suspect the problem there also ties into reconnect scenarios; it would be
necessary to track how each thing intends to connect - which is...
complicated, especially if you subscribe more than once to the same name.
It can probably be fixed, but it isn't (I don't think) quite as simple as
it appears on the surface.

On 16 December 2015 at 10:21, Bernhard <notifications@github.com> wrote:

> Hello
>
> I have an unusual use case where i want to do a subscription to a Redis
> slave server A subscription to a slave server works fine but the Redis
> library does not allow it, and the reason is this line of code:
>
>
> https://githubcom/StackExchange/StackExchangeRedis/blob/master/StackExchangeRedis/StackExchange/Redis/RedisSubscribercs#L190
>
> It basically overrules the CommandFlags passed to the method and uses
> DemandMaster This makes sense if you want to publish messages, but if you
> only want to listen as a subscriber this is not necessary (i think) Would
> it be possible to just use the CommandFlags that are passed to the method?
> If i do this and by using PreferMaster it works perfectly fine because it
> also connects to a slave
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/StackExchange/StackExchange.Redis/issues/321>.
>



-- 
Regards,

Marc

scottyearsley:
Hi, I'm experiencing a similar issue - simply subscribing in a slave is not working. I'm looking to implement a distributed cache/pub-sub using your library (which works ace for everything else) but this is a show stopper, especially as it is working when testing it with ServiceStack.Redis. I'd much prefer to use your library because ServiceStack doesn't seem to support wildcard subscriptions (at least they don't work for me). I really hope you can help :)

piccit:
[...]

PR #452, opened 2016-07-22
Use the supplied flags when selecting a server to subscribe to

rossipedia:
SUBSCRIBE and PSUBSCRIBE are not write ops and can be
performed on slaved instances, so there's no reason to
explicitly pass DemandMaster when selecting the server.

Instead, use the CommandFlags passed as the parameter to Subscribe()


rossipedia:
Fun fact: this is causing some synchronization failures for Ad Server in CO, as the redis instances there are slaved to NY, and calling `Subscribe()` in a "no master available" scenario is essentially a silent no-op. The call _appears_ to succeed, but the passed callback never gets invoked.

NickCraver:
Won't this cause the same issue on re-subscribe after a disconnect? Since the flags aren't stored with our subscription Dictionary I mean.

mgravell:
yeah, also discussing similar on the team chat; may need to think a bit more

piccit:
Related to issue #321

Issue #446, opened 2016-07-12
How to build Core rtm for netcoreapp 1.0


seriouz:
Is it possible to build a netcoreapp 1.0 nuget package of the rtm branch?

michaellandi:
I am also interested to know what the plans are to introduce core support.  Is it on the timeline?

Thanks.

mgravell:
Paging @NickCraver - did you have a release plan? What more do we need to
do here?

NickCraver:
I have a core-rtm branch which AFAIK is ready to go (I've just been busy selling a house here). I've gotten a few reports of timeouts specific to core, so I'd recommend we do another alpha release, and see if people can reproduce the glitches on the RTM packages underneath - thoughts?

michaellandi:
Congrats on the new house :).  I have some high volume applications I can test on w/ if you have a release.

DamianEdwards:
@NickCraver we'd love to synchronize the shipping of an update to the Redis cache provider for ASP.NET Core with the shipping of your .NET Core support (given we depend on you). If you guys do another preview release to nuget.org, we'll ship a 1.0.1-preview of our package that depends on yours, when we can sync up for a non-preview release after that based on how it lands.

Thoughts?

@muratg

NickCraver:
@DamianEdwards I'm still waiting to hear anything on this: https://github.com/dotnet/corefx/issues/4544#issuecomment-230272419 Is there any update? As it stands I'd have to go move all of the core-rtm solution projects around to get .csproj buildable again.

muratg:
@NickCraver Are you guys still planning to do another alpha release, and if so, do you have a timeline?

NickCraver:
@muratg I'm waiting to hear anything at all from anyone on the .NET Core team on any of the numerous blockers we have currently. https://github.com/dotnet/cli/issues/1342 is another issue that is extremely annoying right now. But it looks like some of these are just closed and never getting fixed. I am admittedly extremely frustrated with .NET Core tooling.

To be blunt: from my view, Microsoft wants us to ship .NET Core packages while simultaneously ignoring the blockers we're hitting with the tooling. I can't spend a large amount of time endlessly fighting tooling here. https://github.com/dotnet/corefx/issues/4544 was last updated 4 months ago. https://github.com/dotnet/cli/issues/1342 was just closed 4 months ago without being fixed.

[...]

PR #467, opened 2016-08-24
Work around build issue for existing csproj's

jeremymeng:
Visual Studio tooling will use `project.json` to restore nuget dependencies
when it finds one beside .csproj. This change uses more specific
`<project name>.project.json` to specify dependencies.

Connect to #446.


mgravell:
Thanks

PR #473, opened 2016-09-02
Workaround issue when connecting via host names

jeremymeng:
Due to https://github.com/dotnet/corefx/issues/8768, on non-Windows platforms
Socket.Connect\* methods throw PaltformNotSupportedException when connecting via
host names. This change resolves the IP endpoints first when the endpoint is a
Dns endpoint, tries to connect to each IP endpoint, then picks the first that
works.

### DIFF
--- a/StackExchange.Redis/StackExchange/Redis/SocketManager.cs
+++ b/StackExchange.Redis/StackExchange/Redis/SocketManager.cs
@@ -174,7 +174,10 @@ namespace StackExchange.Redis
 
         internal SocketToken BeginConnect(EndPoint endpoint, ISocketCallback callback, ConnectionMultiplexer multiplexer, TextWriter log)
         {
-            var socket = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp);
+            var family = AddressFamily.InterNetwork;
+            var socketType = SocketType.Stream;
+            var protocolType = ProtocolType.Tcp;
+            var socket = new Socket(family, socketType, protocolType);
             SetFastLoopbackOption(socket);
             socket.NoDelay = true;
             try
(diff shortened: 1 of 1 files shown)


jeremymeng:
@mgravell @NickCraver This workaround worked on my Linux test VM.  Could you please have a look?

StackExchange.Redis/StackExchange/Redis/SocketManager.cs
```diff
@@ -371,6 +396,28 @@ private void EndConnectImpl(IAsyncResult ar, ConnectionMultiplexer multiplexer,
...
+        private async Task<EndPoint> ResolveWorkingEndpointAsync(DnsEndPoint dnsEndPoint, AddressFamily family, SocketType socketType, ProtocolType protocol)
+        {
+            var hostAddresses = await Dns.GetHostAddressesAsync(dnsEndPoint.Host);
+            foreach (var address in hostAddresses)
+            {
+                try
+                {
+                    var socket = new Socket(AddressFamily.InterNetwork, SocketType.Stream, ProtocolType.Tcp);
+                    var endpoint = new IPEndPoint(address, dnsEndPoint.Port);
+                    await socket.ConnectAsync(endpoint);
```
jeremymeng: This could take long.  Could it be possible that one IP endpoint causes timeout, while in the original windows code, the Dns endpoint still works?

NickCraver:
Looking good, but can you please pass the log down into the resolver so we log as the resolution happens? This needs to be debuggable, so logging along the way is necessary.

jeremymeng:
@NickCraver Thanks for reviewing it!  I've added logging for the resolution.

jeremymeng:
Hi @mgravell or @NickCraver, is this workaround good for now, or we should pursue something else?  Thanks!

PR #484, opened 2016-09-15
Remove password from logging of configuration string.

joroda:
ReconfigureAsync in ConnectionMultiplexer.cs logs the configuration string, including the password.
This change addresses the security concern of logging passwords without changing other behavior.


mgravell:
Makes perfect sense; thanks

PR #497, opened 2016-09-27
Add option AllowMultipleRedundantMasters

cahlbin:
To support a setup with multiple, independent, classic redis server
master nodes, the selection of only one master (via the tie-breaker) can
be disabled and instead allow all server endpoints that are masters to
be added as selectable server endpoints.


NickCraver:
Could you please provide some example cases where this would be desired? We'd also need tests to go along with. I admit I'm just having trouble picturing the usage scenario - some examples would really help.

NickCraver:
Closing out due to inactivity

PR #498, opened 2016-10-01
Check Azure RoleEnvironment.IsAvailable before getting CurrentRoleIns…

wangzq:
…tance to avoid unnecessary exception.


mgravell:
Looks good, thanks; sorry for delay

PR #505, opened 2016-10-10
Add ThreadPool and CPU summary to connection related errors.

JonCole:
The "No connection is available to service this operation" error message needs more information about what is happening on the client side to help diagnose cases like CPU being high, so connectTimeout is always getting hit in the background reconnect attempts.


NickCraver:
+1 From me - @mgravell  I don't want to hose you with a merge, but this looks safe to get in the next batch.

mgravell:
Merged, thanks

PR #520, opened 2016-11-08
Added "When" parameter to SortedSetAdd to support NX and XX options

seniorquico:
This is my first time looking at SE.Redis source code. Please let me know if I'm off base here.

And... before this is really considered ready for merge-- how do you run the unit tests?


NickCraver:
Just noting this would be a breaking change, so has to wait for vNext (2.x) either way. Sorry for the delay, we're trying to get through a large OSS backlog/

mgravell:
Added overload approach to make this not be a breaking change - you might want to look at https://github.com/StackExchange/StackExchange.Redis/commit/ec289038cff70f51964deb1201fc5dceeff1baaa to see how this can be done. Much appreciated.

PR #526, opened 2016-11-19
change from for to foreach

bashrc-real:
Changing for to foreach and replacing loops with linq

### DIFF
--- a/StackExchange.Redis/StackExchange/Redis/ConnectionMultiplexer.cs
+++ b/StackExchange.Redis/StackExchange/Redis/ConnectionMultiplexer.cs
@@ -78,12 +78,11 @@ namespace StackExchange.Redis
         /// </summary>
         public ServerCounters GetCounters()
         {
-            var snapshot = serverSnapshot;
-
+            var snapshots = serverSnapshot;
             var counters = new ServerCounters(null);
-            for (int i = 0; i < snapshot.Length; i++)
+            foreach (var snapshot in snapshots)
             {
-                counters.Add(snapshot[i].GetCounters());
+                counters.Add(snapshot.GetCounters());
             }
             unprocessableCompletionManager.GetCounters(counters.Other);
             return counters;
@@ -585,9 +584,8 @@ namespace StackExchange.Redis
             { }
             // if we get problems, need to give the non-failing ones time to finish
             // to be fair and reasonable
-            for (int i = 0; i < tasks.Length; i++)
+            foreach (var task in tasks)
             {
-                var task = tasks[i];
                 if (!task.IsCanceled && !task.IsCompleted && !task.IsFaulted)
                 {
                     var remaining = timeout - checked((int)watch.ElapsedMilliseconds);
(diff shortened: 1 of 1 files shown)


NickCraver:
What is the purpose of these changes? `foreach` allocates an enumerator (so we explicitly opt not to use it). Linq methods (in general) have inherent overhead as well.

In either case, the changes here somehow are on *every* line (are your line endings configured correctly), making the PR very hard to review. I'd double check your git settings locally.

Here's a diff summary so others can easily see:
```patch
diff -r  
81,82c81
<             var snapshot = serverSnapshot;
< 
---
>             var snapshots = serverSnapshot;
84c83
<             for (int i = 0; i < snapshot.Length; i++)
---
>             foreach (var snapshot in snapshots)
86c85
<                 counters.Add(snapshot[i].GetCounters());
---
>                 counters.Add(snapshot.GetCounters());
588c587
<             for (int i = 0; i < tasks.Length; i++)
---
>             foreach (var task in tasks)
590d588
<                 var task = tasks[i];
682c680
<             for (int i = 0; i < tasks.Length; i++)
---
>             foreach (var task in tasks)
684d681
<                 var task = tasks[i];
975,976c972
<                 for (int i = 0; i < tmp.Length; i++)
<                     tmp[i].OnHeartbeat();
---
>                 foreach (var server in serverSnapshot) server.OnHeartbeat();
1087c1083
<                 for (int i = 0; i < snapshot.Length; i++) total += snapshot[i].OperationCount;
---
>                 long total = snapshot.Sum(server => server.OperationCount);
1668,1672c1664
<             for (int i = 0; i < servers.Length; i++)
<             {
<                 if (string.Equals(Format.ToString(servers[i].EndPoint), endpoint, StringComparison.OrdinalIgnoreCase))
<                     return servers[i];
<             }
---
>             var selectedServer = servers.First(server => string.Equals(Format.ToString(server.EndPoint), endpoint, StringComparison.OrdinalIgnoreCase));
1675,1683c1667,1669
<             for (int i = 0; i < servers.Length; i++)
<             {
<                 if (string.Equals(DeDotifyHost(Format.ToString(servers[i].EndPoint)), deDottedEndpoint, StringComparison.OrdinalIgnoreCase))
<                 {
<                     LogLocked(log, "...but we did find instead: {0}", deDottedEndpoint);
<                     return servers[i];
<                 }
<             }
<             return null;
---
>             selectedServer = servers.First(server => string.Equals(DeDotifyHost(Format.ToString(server.EndPoint)), deDottedEndpoint, StringComparison.OrdinalIgnoreCase));
>             if (selectedServer != null) LogLocked(log, "...but we did find instead: {0}", deDottedEndpoint);
>             return selectedServer;        
1827,1830c1813
<                 var tmp = serverSnapshot;
<                 for (int i = 0; i < tmp.Length; i++)
<                     if (tmp[i].IsConnected) return true;
<                 return false;
---
>                 return serverSnapshot.Any(server => server.IsConnected);
```

mgravell:
I am very much against these changes. If it was `List<T>`, then maybe (custom struct iterator etc), but in the general case and when using arrays: it was fine originally. It is not always good do add LINQ - it might look cleaner, but there is a lot of overhead involved in First, Where, etc.

Issue #410, opened 2016-05-22
[CoreCLR] Can't connect to Redis server by hostname. Only by IP


PabloZaiden:
# Description

When trying to create a ConnectionMultiplexer instance, it throws a PlatformNotSupportedException.
When trying to create the same connection, but using the ip address instead of the hostname, it works just fine.
# Repro steps

Include  `"StackExchange.Redis": "1.1.604-alpha"` in `project.json`
### Code

```
// works fine
var redis1 = StackExchange.Redis.ConnectionMultiplexer.Connect("127.0.0.1");

// thows
var redis2 = StackExchange.Redis.ConnectionMultiplexer.Connect("localhost");
```
# Exception detail

```
at System.Threading.Tasks.TaskFactory`1.FromAsyncImpl[TArg1,TArg2](Func`5 beginMethod, Func`2 endFunction, Action`1 endAction, TArg1 arg1, TArg2 arg2, Object state, TaskCreationOptions creationOptions)
at System.Net.Sockets.SocketTaskExtensions.ConnectAsync(Socket socket, String host, Int32 port)
at StackExchange.Redis.SocketManager.BeginConnect(EndPoint endpoint, ISocketCallback callback, ConnectionMultiplexer multiplexer, TextWriter log)
at StackExchange.Redis.PhysicalConnection.BeginConnect(TextWriter log)
at StackExchange.Redis.PhysicalBridge.GetConnection(TextWriter log)
at StackExchange.Redis.ServerEndPoint..ctor(ConnectionMultiplexer multiplexer, EndPoint endpoint, TextWriter log)
at StackExchange.Redis.ConnectionMultiplexer.<ReconfigureAsync>d__119.MoveNext()
--- End of inner exception stack trace ---
at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
at System.Threading.Tasks.Task.Wait(Int32 millisecondsTimeout, CancellationToken cancellationToken)
at StackExchange.Redis.ConnectionMultiplexer.ConnectImpl(Func`1 multiplexerFactory, TextWriter log)
at ConsoleApplication.Program.Main(String[] args)
---> (Inner Exception #0) System.PlatformNotSupportedException: Sockets on this platform are invalid for use after a failed connection attempt, and as a result do not support attempts to connect to multiple endpoints.
at System.Net.Sockets.Socket.BeginConnect(String host, Int32 port, AsyncCallback requestCallback, Object state)
[...]

PR #544, opened 2016-12-31
Workaround for connecting redis via host name on Linux

cnblogs-dudu:
Address #410

### DIFF
--- a/StackExchange.Redis/StackExchange/Redis/ConfigurationOptions.cs
+++ b/StackExchange.Redis/StackExchange/Redis/ConfigurationOptions.cs
@@ -7,6 +7,9 @@ using System.Net;
 using System.Net.Security;
 using System.Text;
 using System.Threading.Tasks;
+#if CORE_CLR
+using System.Runtime.InteropServices;
+#endif
 
 namespace StackExchange.Redis
 {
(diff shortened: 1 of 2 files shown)


Tratcher:
Taking just the first address is not ideal. e.g. IPv6 addresses are often returned first and the server may or may not be listening on IPv6. A full replacement would loop through the returned addresses.

cnblogs-dudu:
`addresses.FirstOrDefault(ip => ip.AddressFamily == AddressFamily.InterNetwork)` only gets IPv4 addresses.

Tratcher:
Sure, and if there's more than one IPv4 address and the server's only listening on one?

NickCraver:
@Tratcher do you have any thoughts on this? Is something coming in CoreFX? This is something *so many* libraries are going to have to do...it seems silly that they'll all have their own approach, doesn't it? Or am I missing something here?

Tratcher:
@NickCraver I'm as disappointed with https://github.com/dotnet/corefx/issues/8768 as you are. https://github.com/dotnet/corefx/issues/11564 is a mediocre compromise and it doesn't look like it's coming anytime soon.

MichaCo:
@Tratcher @NickCraver Maybe off topic, but, does that "workaround", using the IP instead, even work with Azure Redis? I tested that some time ago and couldn't make it work because you basically only have the domain name via SSL. Connecting via IP just didn't work.

NickCraver:
This is coming back in `netstandard` itself, so I don't think we should have it in our code (the same as *every* other library would have to) - see: https://github.com/dotnet/corefx/pull/16373

Thoughts about the best interim route given all of the issues and corner cases above with IPv4, IPv6, etc.?

PR #549, opened 2017-01-11
Get a random slave

fengzhbo:
获取一个随机的slave

### DIFF
--- a/StackExchange.Redis/StackExchange/Redis/ServerSelectionStrategy.cs
+++ b/StackExchange.Redis/StackExchange/Redis/ServerSelectionStrategy.cs
@@ -240,11 +240,49 @@ namespace StackExchange.Redis
             if (endpoint.IsSlave && endpoint.IsSelectable(command)) return endpoint;
 
             var slaves = endpoint.Slaves;
+
+            return GetRandomSlaveEndPoint(slaves, command) ?? GetFirstSlaveEndPoint(slaves, command);
+        }
+
+        /// <summary>
+        /// get the first available slave
+        /// </summary>
+        /// <param name="slaves"></param>
+        /// <param name="command"></param>
+        /// <returns></returns>
+        private static ServerEndPoint GetFirstSlaveEndPoint(ServerEndPoint[] slaves, RedisCommand command)
+        {
+            ServerEndPoint endpoint = null;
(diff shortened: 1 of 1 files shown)


NickCraver:
This can easily be a breaking behavior change. It changes the behavior of `FindSlave` from first to random, not just adding a random option. Why would this be the case?

The first slave is often a local slave. A slave further downstream (e.g. last resort in the config string) may be off-site in another data center and extremely slow. That cannot be the default behavior. It removes the control via the config string from the user. You also have to deal with replication latency, which in high volume & frequency environments can have problems with. The state in a downstream slave is likely perpetually behind in some regard, issuing commands at it would yield more stale (incorrect) data all of the sudden.

I'm not against adding a *new* `.FindRandomSlave()` for instance (though it should be justified somehow). But changing the default behavior of `.FindSlave()` I'm definitely against.

fengzhbo:
Indeed, this should be a configuration

Thanks for your  suggestion, @NickCraver

NickCraver:
@fengzhbo Can you please explain the use case for this? If this is actually needed, then the only way I'd see it in the core library is via `CommandFlags`. For example, there's currently:
- `CommandFlags. DemandMaster`
- `CommandFlags. PreferMaster`
- `CommandFlags. DemandSlave`
- `CommandFlags. PreferSlave`

If something like a random slave is actually needed, it belongs there IMO. But I'd still like to understand the use case here - I haven't heard on this requested before, only a PR with changes proposed here.

fengzhbo:
@NickCraver 

In my test, i have a cluster include three masters and four slaves, the extra one slave is for [Replicas migration](https://redis.io/topics/cluster-tutorial#replicas-migration).  

No matter which `CommandFlags` is used, there always one slave is idle.  

So i want to add some feature to full use of all the slaves, random is just an easy way.  

Poor english, forgive me, :blush:

NickCraver:
Hey @fengzhbo,
To clarify: my comment was saying such a feature *should* be via `CommandFlags`, e.g. `CommandFlags.PreferRandomSlave`. Rather than the way it's implemented in this PR, which behaves in unintuitive and unexpected ways.

NickCraver:
Closing this out for the reasons above - feel free to submit one for the `CommandFlags` style behavior.

PR #554, opened 2017-01-19
Specify the duration of the timeout when TimeoutException happens

knocte:
Being more verbose on this is important.

While at making this change, I also tried to use TimeSpan
wherever I could, given that it's much more convenient/correct.

### DIFF
--- a/StackExchange.Redis.Tests/PubSub.cs
+++ b/StackExchange.Redis.Tests/PubSub.cs
@@ -206,7 +206,8 @@ namespace StackExchange.Redis.Tests
                 var t2 = pub.PingAsync();
                 Thread.Sleep(100); // especially useful when testing any-order mode
 
-                if (!Task.WaitAll(new[] { t1, t2 }, muxer.TimeoutMilliseconds * 2)) throw new TimeoutException();
+                var timeout = muxer.Timeout + muxer.Timeout;
+                if (!Task.WaitAll(new[] { t1, t2 }, timeout)) throw new TimeoutException(timeout.ToString());
             }
         }
 
--- a/StackExchange.Redis.Tests/SSL.cs
+++ b/StackExchange.Redis.Tests/SSL.cs
@@ -137,7 +137,7 @@ namespace StackExchange.Redis.Tests
                 {
                     db.StringIncrement(key);
                 }
-            }, Threads, timeout: 45000);
+            }, Threads, timeout: TimeSpan.FromSeconds(45));
             value = (long)db.StringGet(key);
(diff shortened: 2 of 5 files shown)


NickCraver:
While I don't disagree with putting timeouts clearly in error messages, just creating a mix of `int` and `TimeSpan` really isn't desirable. And (strictly speaking), it isn't as fast.

I think the proper fix here is using `TimeSpan.FromMilliseconds()` once in exceptions. This can be done cleanly with an overload in `TimeoutException()` that takes an int, auto-appending `", Timeout: <string>"` to the message (and as a `.Data` property) for example.

Converting *all* internal APIs to `TimeSpan` (and it should be all, not some if we do it), should be a separate set of changes.

knocte:
> Converting all internal APIs to TimeSpan (and it should be all, not some if we do it), should be a
> separate set of changes.

That is a sensible review. I mean, generally, I tend to do separation of concerns in all my changes myself.

However, there are reasons why I didn't do this in the first place, which came clearer to me again after I read your review and I tried to separate them:

1. First, in regards to the question: what to implement first? Change to TimeSpan or fix timeoutexceptions to include the timeout? Given that the former is a bigger change than the latter, I was considering to do it first because it's lower hanging fruit, however, it's easier to fix the latter in a codebase which is already using TimeSpan, because for it I just need to call the `ToString()` method of TimeSpan... So then I started looking again at the `int -> TimeSpan` conversion, to make it full instead of partial, and then....

2. I found this line in ConfigurationOptions.cs:

```
        private int? keepAlive, syncTimeout, connectTimeout, responseTimeout, writeBuffer, connectRetry, configCheckSeconds, defaultDatabase;
```

If you ask a contributor to change all those variables to TimeSpan at once, in one commit, you're going to get a giant contribution which is very hard to review. But most important of all: you will not get any contribution at all because any sane contributor will think that there are high chances that after the giant effort of contributing such a change, the PR may never get reviewed/merged.

Given what I explained above, IMHO, the maintainers should be open to accepting an `int -> TimeSpan` progressive conversion. If this is welcomed, then accept my PR. If it is not, I'm afraid it's not worth my time, and you can close this PR.

NickCraver:
@knocte the honest answer is that I wouldn't accept a PR mixing the two, and I think just being very open about that is the correct approach. That being said, I'm not personally against changing *all* of those items to `TimeSpan` in V2 if that's what the community desires. If you want to create an issue on it, I'd do the work myself if that's a widely desired change. 

Closing out (as requested) to cleanup.

PR #584, opened 2017-03-01
Add NextCursor in IScanningCursor in order to limit the result

dobariya:
Add NextCursor varaible  in IScanningCursor in order to limit the result only based on the cursor.


dobariya:
Sorry it was by mistake..please ignore this request

Issue #582, opened 2017-02-27
PlatformNotSupportedException trying to sniff platform in ConnectionMultiplexer.Connect() 


daniel-meismer-zocdoc:
I got an error connecting to redis in an `AWS Lambda` function on `coreclr` (stack trace below). 
The error is triggered by the internal code that tries to optimize the socket for localhost connections on Windows, specifically the code in [SocketManager.cs](https://github.com/StackExchange/StackExchange.Redis/blob/master/StackExchange.Redis/StackExchange/Redis/SocketManager.cs#L272
) that tries to determine if it is running on Windows. 

To verify, I copied the redis client code into my project, commented out the call to `SetFastLoopbackOption` ([on line 178](https://github.com/StackExchange/StackExchange.Redis/blob/master/StackExchange.Redis/StackExchange/Redis/SocketManager.cs#L178)), and did not get the error when I deployed. 

To fix, I'd propose adding a `catch(PlatformNotSupportedException) { /* ignore */ }` to the OS sniffing code to the `try/catch` at [lines 269-280](https://github.com/StackExchange/StackExchange.Redis/blob/master/StackExchange.Redis/StackExchange/Redis/SocketManager.cs#L269-L280) since we don't really care if it fails. 

**Stack trace**
```
One or more errors occurred. (One or more errors occurred. (Operation is not supported on this platform.)): AggregateException
   at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
   at System.Threading.Tasks.Task.Wait(Int32 millisecondsTimeout, CancellationToken cancellationToken)
   at lambda_method(Closure , Stream , Stream , ContextInfo )
   at System.Threading.Tasks.Task.ThrowIfExceptional(Boolean includeTaskCanceledExceptions)
   at System.Threading.Tasks.Task.Wait(Int32 millisecondsTimeout, CancellationToken cancellationToken)
   at StackExchange.Redis.ConnectionMultiplexer.ConnectImpl(Func`1 multiplexerFactory, TextWriter log)
   at Zd.RefDataConsumer.Lambda.EntryPoint.LongLivedConnections.GetRedisConnectionMultiplexer() in /app/src/Zd.RefDataConsumer.Lambda/EntryPoint.cs:line 119
[...]

PR #588, opened 2017-03-06
Suppress platform check failure on connect

daniel-meismer-zocdoc:
Fix for https://github.com/StackExchange/StackExchange.Redis/issues/582
*Note: verified fix in AWS lambda*


NickCraver:
Thanks for the fix!

Issue #316, opened 2015-12-08
Add zrevrangebylex 


mgravell:
See: http://stackoverflow.com/q/34151544/23354

Also: audit all other commands

NickCraver:
Closing this out - tracked in #598 (2.0 change though, or we add an overload)

PR #598, opened 2017-03-16
Support ZREVRANGEBYLEX operation

syedhassaanahmed:
Added support for Redis [ZREVRANGEBYLEX](ZREVRANGEBYLEX) operation which should address #316 
Methods `SortedSetRangeByValue` & `SortedSetRangeByValueAsync` were extended with optional column `Order` which decides whether to use `ZRANGEBYLEX` or `ZREVRANGEBYLEX`

### DIFF
--- a/StackExchange.Redis.Tests/DatabaseWrapperTests.cs
+++ b/StackExchange.Redis.Tests/DatabaseWrapperTests.cs
@@ -725,8 +725,15 @@ namespace StackExchange.Redis.Tests
         [Test]
         public void SortedSetRangeByValue()
         {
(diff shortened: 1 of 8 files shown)


NickCraver:
This is a breaking change, so it'll have to wait for a v2.0 release to be considered.

syedhassaanahmed:
@NickCraver In what way is the change breaking? Maybe we can fix it, since it wasn't the intention.

NickCraver:
@syedhassaanahmed By adding arguments to the method :) I don't have a problem with the change, just need to be bundled in a major release - else we're adding overloads to maintain and strip with the same major version release later.

syedhassaanahmed:
Ah got it .. I thought adding it as optional param at last position would help :)
v2.0 it is then 👍

PR #603, opened 2017-03-21
Add SslProtocols property to ConfigurationOptions

JonCole:
This property allows you to configure the list of acceptable SSL/TLS Protocols for encrypted communication.  Without this setting, users are required to make changes to machine wide settings, which can break other applications that are not compatible.

### DIFF
--- a/Docs/Configuration.md
+++ b/Docs/Configuration.md
@@ -79,11 +79,11 @@ The `ConfigurationOptions` object has a wide range of properties, all of which a
 | serviceName={string}   | `ServiceName`          | `null`                       | Not currently implemented (intended for use with sentinel)                                                |
 | ssl={bool}             | `Ssl`                  | `false`                      | Specifies that SSL encryption should be used                                                              |
 | sslHost={string}       | `SslHost`              | `null`                       | Enforces a particular SSL host identity on the server's certificate                                       |
+| sslProtocols={enum}    | `SslProtocols`         | `null`			             | Ssl/Tls versions supported when using an encrypted connection.  Use '\|' to provide multiple values.      |
 | syncTimeout={int}      | `SyncTimeout`          | `1000`                       | Time (ms) to allow for synchronous operations                                                             |
 | tiebreaker={string}    | `TieBreaker`           | `__Booksleeve_TieBreak`      | Key to use for selecting a server in an ambiguous master scenario                                         |
 | version={string}       | `DefaultVersion`       | (`3.0` in Azure, else `2.0`) | Redis version level (useful when the server does not make this available)                                 |
 | writeBuffer={int}      | `WriteBuffer`          | `4096`                       | Size of the output buffer                                                                                 |
-| sslProtocols={enum}    | `SslProtocols`         | `null`			             | Ssl/Tls versions supported when using an encrypted connection.  Use '\|' to provide multiple values.      |
 
 Additional code-only options:
 - ReconnectRetryPolicy (`IReconnectRetryPolicy`) - Default: `ReconnectRetryPolicy = LinearRetry(ConnectTimeout);`
--- a/Docs/Configuration.md
+++ b/Docs/Configuration.md
@@ -62,27 +62,28 @@ Configuration Options
 
 The `ConfigurationOptions` object has a wide range of properties, all of which are fully documented in intellisense. Some of the more common options to use include:
 
-| Configuration string   | `ConfigurationOptions` | Default                      | Meaning                                                                          |
-| ---------------------- | ---------------------- | ---------------------------- | -------------------------------------------------------------------------------- |
-| abortConnect={bool}    | `AbortOnConnectFail`   | `true` (`false` on Azure)    | If true, `Connect` will not create a connection while no servers are available   |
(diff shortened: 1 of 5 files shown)


Docs/Configuration.md
```diff
@@ -62,27 +62,28 @@ Configuration Options
...
+| sslHost={string}       | `SslHost`              | `null`                       | Enforces a particular SSL host identity on the server's certificate                                       |
+| sslProtocols={enum}    | `SslProtocols`         | `null`			             | Ssl/Tls versions supported when using an encrypted connection.  Use '\|' to provide multiple values.      |
+| syncTimeout={int}      | `SyncTimeout`          | `1000`                       | Time (ms) to allow for synchronous operations                                                             |
+| tiebreaker={string}    | `TieBreaker`           | `__Booksleeve_TieBreak`      | Key to use for selecting a server in an ambiguous master scenario                                         |
+| version={string}       | `DefaultVersion`       | (`3.0` in Azure, else `2.0`) | Redis version level (useful when the server does not make this available)                                 |
+| writeBuffer={int}      | `WriteBuffer`          | `4096`                       | Size of the output buffer                                                                                 |
```
NickCraver: line-ending issue here showing the entire table as changed?

StackExchange.Redis/StackExchange/Redis/ConfigurationOptions.cs
```diff
@@ -65,7 +66,18 @@ internal static Proxy ParseProxy(string key, string value)
...
+            {
+                SslProtocols tmp;
+                //Flags expect commas as separators, but we need to use '|' since commas are already used in the connection string to mean something else
+                value = value?.Replace("|", ","); 
+
+                if (!Enum.TryParse(value, true, out tmp)) throw new ArgumentOutOfRangeException("Keyword '" + key + "' requires an SslProtocol value (multiple values separated by '|').");
 
+                return tmp;                
+            }
+#endif
```
NickCraver: Forgive the ignorance here - but why is this `!CORE_CLR` specific?

StackExchange.Redis/StackExchange/Redis/ConfigurationOptions.cs
```diff
@@ -78,6 +90,10 @@ internal static void Unknown(string key)
                         ConfigChannel = "configChannel", AbortOnConnectFail = "abortConnect", ResolveDns = "resolveDns",
                         ChannelPrefix = "channelPrefix", Proxy = "proxy", ConnectRetry = "connectRetry",
                         ConfigCheckSeconds = "configCheckSeconds", ResponseTimeout = "responseTimeout", DefaultDatabase = "defaultDatabase";
+#if !CORE_CLR
+            internal const string SslProtocols = "sslProtocols";
+#endif
```
NickCraver: Same as comment above - what's specific to non-Core? The documentation fragmenting between the two is not an awesome distinction to have to make.

Docs/Configuration.md
JonCole: When you view this in a text editor, I just added more space at the end of each row so that the last set of |'s would line up since I needed more space for the one row I added.  Nothing significant was changed other than those extra spaces on each row - can revert if it bothers you.

StackExchange.Redis/StackExchange/Redis/ConfigurationOptions.cs
JonCole: The CORE_CLR was purely because I don't have the right tools locally to test my changes for core clr can thus couldn't even compile to see if I got everything correct.  I suspect Core CLR would support the same concept but I don't have a way to validate currently.

StackExchange.Redis/StackExchange/Redis/ConfigurationOptions.cs
JonCole: Same as above - just didn't have a way to validate that I changed the core clr portions of the code correctly, so I left it out.  Shouldn't be too hard to remove this if someone has the right build pieces locally.

NickCraver:
Hmmm, I desperately need to change this over to the VS 2017 project system to make life easier on everyone. That's one reason I'm trying to go through PRs and make sure nothing broken by that move is left behind.

JonCole:
So, I *think* I installed the right version of the .NET Core SDK, but there doesn't seem to be an SslStream.AuthenticateAsClientAsync method/extension that takes the SslProtocols parameter, so I am not sure we can even support this on CORE_CLR.  Suggestions?

NickCraver:
@JonCole it appears to be there for me, as an example:
```c#
ssl.AuthenticateAsClientAsync(host, new X509CertificateCollection(), SslProtocols.Tls11, checkCertificateRevocation: true)
```

JonCole:
@NickCraver - Sounds good.  Would you mind making that change when you merge the PR?

JonCole:
@NickCraver - I went ahead and made the change to use that signature for core clr, but was still unable to build locally, so I am hoping it works now.

mgravell:
I was *going* to say that `ssl` is bad for naming since the one protocol we probably don't want is ssl, but: the code as implemented here is consistent with `UseSsl` etc, so if anything: I can only kick myself for being stupid a few years ago. My bad. Merging.

PR #631, opened 2017-05-17
Update mcs -> csc

rayalex:
Project is using some C# 7 features which can only be compiled using CSC under Mono 5.0


NickCraver:
Sorry I missed this, fix via #691

PR #645, opened 2017-06-13
NRediSearch- Sorting Feature Added

rajeshsubhankar:
Just like in JRediSearch , Sorting functionality for NRediSearch has been implemented.  Sorting functionality can be applied on both TEXT and NUMERIC field.


NRediSearch/Query.cs
```diff
@@ -162,6 +162,16 @@ public Paging(int offset, int count)
         public byte[] Payload { get; set; }
 
         /// <summary>
+        /// Set the query parameter to sort by
+        /// </summary>
+        public string _sortBy { get; set; }
```
mgravell: naming: this is not idiomatic .NET naming conventions for a property - should be `SortBy`

NRediSearch/Query.cs
```diff
@@ -205,6 +215,13 @@ internal void SerializeRedisArgs(List<object> args)
                 args.AddRange(_fields);
             }
 
+            if(_sortBy != null)
+            {
+                args.Add("SORTBY".Literal());
+                args.Add(_sortBy.Literal());
```
mgravell: what is `_sortBy` here? if this is user data related (rather than a fundamental part of the API syntax), it should **not** be `.Literal()`; the point of `.Literal()` is to avoid allocating for *repeated* content by caching and reusing an encoded version - but for user data, this is a bad thing and could lead to a slow memory leak

NRediSearch/Query.cs
```diff
@@ -260,5 +277,18 @@ public Query LimitFields(params string[] fields)
...
             return this;
         }
+
+        /// <summary>
+        /// Set the query to be sorted by a sortable field defined in the schema
+        /// </summary>
+        /// <param name="field">the sorting field's name</param>
+        /// <param name="ascending">if set to true, the sorting order is ascending, else descending</param>
+        /// <returns>the query object itself</returns>
+        public Query SetSortBy(string field, bool ascending)
```
mgravell: recommend: `public Query SetSortBy(string field, bool ascending = true)` - i.e. make it optional with a default of ascending

NRediSearch/Schema.cs
```diff
@@ -95,7 +114,18 @@ public Schema AddGeoField(string name)
...
+            Fields.Add(new Field(name, FieldType.Numeric, false));
+            return this;
+        }
+
+        /// <summary>
+        /// Add a numeric field that can be sorted on
+        /// </summary>
+        /// <param name="name">the field's name</param>
+        /// <returns>the schema object</returns>
+        public Schema AddSortableNumericField(string name)
```
mgravell: this *could* also be done by tweaking `AddNumericField(string name)` to be `AddNumericField(string name, bool sortable = false)` - i.e. so that it defaults to the old non-sortable behavior. Technically that's a breaking change (the signature is no longer compatible), but I *suspect* that's OK at this point in time. It wouldn't be OK in fully established lib without a major (semver) incr.

NRediSearch/Query.cs
```diff
@@ -162,6 +162,16 @@ public Paging(int offset, int count)
...
 
         /// <summary>
+        /// Set the query parameter to sort by
+        /// </summary>
+        public string _sortBy { get; set; }
+
+        /// <summary>
+        /// Set the query parameter to sort by ASC by default
+        /// </summary>
+        public bool _sortAsc = true;
```
mgravell: public bool SortAscending {get; set;} = true;

docs/ReleaseNotes.md
```diff
@@ -1,6 +1,13 @@
 # Release Notes
 
-## Unreleased changes:
+# Important: .NET 4.0 support is currently **disabled**; if you need .NET 4.0, please stick with 1.2.1 while we keep investigating
```
mgravell: something seems off with the baselining here; what is the baseline for this PR? it seems to be showing relative to some very old version; is there any chance of rebasing against current?

mgravell:
Awesome! I've added some feedback - mostly just idiomatic .NET things like property naming and public fields. It does seem to have an odd baseline, though, which is making it very hard to view the changes relative to the current codebase; it looks like it is baselined against before the first code was laid down?

NRediSearch/Query.cs
rajeshsubhankar: Thanks. Appropriate changes has been done.

NRediSearch/Query.cs
rajeshsubhankar: Thanks. Appropriate changes has been done.

NRediSearch/Query.cs
rajeshsubhankar: Thanks. Appropriate changes has been done.

NRediSearch/Query.cs
rajeshsubhankar: Thanks. Appropriate changes has been done.

NRediSearch/Schema.cs
rajeshsubhankar: Little confusion over here. Are you recommending to change `AddNumericField(string name)` to  `AddNumericField(string name, bool sortable = false)` or no change required as of now?

docs/ReleaseNotes.md
rajeshsubhankar: Yes. It's baseline against `StackExchange:NRediSearch` . I will close this PR and reopen with updated code base against Master branch. Thanks.

dvirsky:
Thanks @rajeshsubhankar for adding this!

PR #647, opened 2017-06-22
Fixes CommandFlags duplicate values.

farukterzioglu:
CommandFlags had duplicated values for "None" and "PreferMaster".
PreferMaster updated to "3"


NickCraver:
This was absolutely intentional - the default behavior is to prefer the master. This minor looking change would break most of the library :)

PR #658, opened 2017-07-05
Update URL for the processModel reference

peorth80:
Fixed the URL to the properties of <processModel in machine.web, added a comment regarding using the value in web.config (it's not possible)


NickCraver:
Thank you!

PR #660, opened 2017-07-06
Support for netstandard1.4 and netappcore1.0.

uatec:
AWS Lambda only supports netcoreapp1.0, and will never support 1.1. 

This means that to run StackExchange.Redis in AWS Lambda, it must run on netstandard1.4 or lower.

This PR adjusts the targeted version to 1.4, and excludes BufferedStream, which is only supported in 1.5.

I welcome feedback on how to handle this shortfall.

### DIFF
--- a/Directory.build.props
+++ b/Directory.build.props
@@ -22,7 +22,7 @@
     <DefaultLanguage>en-US</DefaultLanguage>
     <IncludeSymbols>false</IncludeSymbols>
 
-    <LibraryTargetFrameworks>net45;net46;netstandard1.5</LibraryTargetFrameworks><!--net40;-->
+    <LibraryTargetFrameworks>net45;net46;netstandard1.5;netstandard1.4</LibraryTargetFrameworks><!--net40;-->
     <CoreFxVersion>4.3.0</CoreFxVersion>
   </PropertyGroup>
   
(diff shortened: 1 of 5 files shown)


mgravell:
I'm fine with the *aim* of this, however I'm concerned about the impact of simply removing BufferedStream; the code above is *expecting* that layer, and without it we could see severe packet fragmentation. What I'd rather do is take the intent of this, but combine it with a custom BufferedStream implementation (which I could provide if you don't fancy that challenge). I think it would be a mistake to merge without that part.

uatec:
You are quite right.

We could attempt to import the .net BufferedSteam implementation (https://github.com/dotnet/corefx/blob/03d74ce754abacc65bbe288ac34f630cd9327ff1/src/System.Runtime.Extensions/src/System/IO/BufferedStream.cs).

Are you happy with importing code from corefx?

mgravell:
If the license permits, then: sure - although I would recommend changing the namespaces etc so that there is no possible ambiguity.

uatec:
Both projects are MIT.

I can exclude the imported implementation when compiling against anything other than 1.4. That would avoid collisions. 

I'll have a crack at it tomorrow.

alexvaluyskiy:
@uatec you could run any version of `netstandard` up to 1.6 on `netcoreapp1.0`

NickCraver:
As @alexvaluyskiy says, `netstandard1.6` is supported by `netcoreapp1.0`, so I don't there's any reason to introduce the complexity and risk here.

uatec:
My use case for this was flakey at best.

To do with runtime versions and implicit upgrades to versions that are not available in AWS Lambda.

However my use case has now evaporated for other reasons.  I will close this issue now, however it could be reopened it anybody else has a similar requirement.

PR #686, opened 2017-08-13
feat: use endpoint address family instead of hardcoded InterNetwork

### LINKED ISSUE #252: IPv6 support
Hi!

I am using  Redis client (and in the past, BookSleeve) in production.  The connector works and useful in IPV4 but could not connect to IPV6 Redis.

Redis itself responding to V6 (tested via `telnet`).

Will be change?

thanks

laskoviymishka:
This should allow us to use ipv6 based hosts and endpoints. Closes #252


laskoviymishka:
just tested it on ipv6 only network in docker - works perfect.

NickCraver:
FYI I'm getting unit tests into shape over in #700, after merge of that I plan on regression testing this for merge :)

laskoviymishka:
Great, right now this fix works in our production environment, but to make it working I had to use some really dirty hacks. Can't wait when it will be available in nuget.

PR #691, opened 2017-08-18
use roslyn csc compiler on mono for C#7 support

### LINKED ISSUE #674: Compilation error: Feature 'declaration expression' cannot be used
I downloaded the latest source (1.2.5) zip, and tried compiling it using `monobuild.cmd` under Windows Server with latest mono stable (5.0.1.1), but it throws an error during compilation:

> Building StackExchange.Redis.dll ...
> StackExchange.Redis\StackExchange\Redis\CommandMap.cs(179,49): error CS1644: Feature `declaration expression' cannot be used because it is not part of the C# 6.0 language specification
> Compilation failed: 1 error(s), 0 warnings

Any clues why this would occur, and how to resolve it?

**Edit:** Same error when compiling it using `monobuild.bash` under Ubuntu Server 16.04 with latest mono stable (5.0.1.1).

rufusl:
Fixes #674 for Mono builds.

### DIFF
--- a/monobuild.bash
+++ b/monobuild.bash
@@ -3,9 +3,9 @@ rm -rf BasicTest/bin/mono
 mkdir -p StackExchange.Redis/bin/mono
 mkdir -p BasicTest/bin/mono
 echo -e "Building StackExchange.Redis.dll ..."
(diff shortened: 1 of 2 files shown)


NickCraver:
Pulling in, thanks!

Issue #678, opened 2017-08-07
I found the problem code that caused the timeout under CoreCLR.


majian159:
test code:

            var connection = ConnectionMultiplexer.Connect("localhost:6379");
            var database = connection.GetDatabase();

            Parallel.For(0, 100000, i =>
            {
                try
                {
                    database.StringGet("test");
                }
                catch (Exception e)
                {
                    Console.WriteLine(e);
                }
            });

            //wait...
            Console.ReadLine();

exception:
StackExchange.Redis.RedisTimeoutException: Timeout performing GET test, inst: 0, queue: 10, qu: 0, qs: 10, qc: 0, wr: 0, wq: 0, in: 45, ar: 0

**I changed it 'PhysicalConnection.cs',and set up 'NETSTANDARD2_0'.
The problem seems to come from TAP,.net framework APM is not problem.**

@mgravell #601 
**I hope you fix it as soon as possible.**

This is the modified code:

    internal sealed partial class PhysicalConnection : IDisposable, ISocketCallback
    {

        internal readonly byte[] ChannelPrefix;

        private const int DefaultRedisDatabaseCount = 16;

        private static readonly byte[] Crlf = Encoding.ASCII.GetBytes("\r\n");

#if CORE_CLR && !NETSTANDARD2_0
        readonly Action<Task<int>> endRead;
        private static Action<Task<int>> EndReadFactory(PhysicalConnection physical)
        {
            return result =>
            {
                // can't capture AsyncState on SocketRead, so we'll do it once per physical instead
                if (result.IsFaulted)
                {
                    GC.KeepAlive(result.Exception);
                }
                try
                {
                    physical.Multiplexer.Trace("Completed asynchronously: processing in callback", physical.physicalName);
                    if (physical.EndReading(result)) physical.BeginReading();
                }
                catch (Exception ex)
                {
                    physical.RecordConnectionFailed(ConnectionFailureType.InternalFailure, ex);
                }
            };
        }
#else
        static readonly AsyncCallback endRead = result =>
        {
            PhysicalConnection physical;
            if (result.CompletedSynchronously || (physical = result.AsyncState as PhysicalConnection) == null) return;
            try
            {
                physical.Multiplexer.Trace("Completed asynchronously: processing in callback", physical.physicalName);
                if (physical.EndReading(result)) physical.BeginReading();
            }
            catch (Exception ex)
            {
                physical.RecordConnectionFailed(ConnectionFailureType.InternalFailure, ex);
[...]

PR #699, opened 2017-08-23
.netstandard2.0 use APM reduce timeout exception.

majian159:
#678

### DIFF
--- a/Directory.build.props
+++ b/Directory.build.props
@@ -22,7 +22,7 @@
     <DefaultLanguage>en-US</DefaultLanguage>
     <IncludeSymbols>false</IncludeSymbols>
 
-    <LibraryTargetFrameworks>net45;net46;netstandard1.5</LibraryTargetFrameworks><!--net40;-->
+    <LibraryTargetFrameworks>net45;net46;netstandard1.5;netstandard2.0</LibraryTargetFrameworks><!--net40;-->
     <CoreFxVersion>4.3.0</CoreFxVersion>
   </PropertyGroup>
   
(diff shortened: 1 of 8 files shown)


majian159:
This is not valid in linux

NickCraver:
We need a little more information here, I'm still not sure what we're looking at. What is the goal we're trying to fix? The `netstandard2.0` setup in general will need a different approach, I recommend we look at whatever the intended functionality here is after that merge (I'm working on getting Sentinel in first).

zhouqi6876:
@majian159  Do you known how to solve the timeout exception with .net core in linux？

majian159:
@zhouqi6876 
https://github.com/StackExchange/StackExchange.Redis/blob/master/docs/Timeouts.md
In .Net Core, add Environment Variable COMPlus_ThreadPool_ForceMinWorkerThreads to overwrite default MinThreads setting, according to Environment/Registry Configuration Knobs

zhouqi6876:
@majian159 I try to export ComPlus_ThreadPool_ForceMinWorkerThreads=50, it doesn't work ,have you ever tried？

majian159:
@zhouqi6876 can refer to https://github.com/dotnet/cli/blob/rel/1.0.0/Documentation/specs/runtime-configuration-file.md.

wjdavis5:
@majian159 
You can also use the SetMinWorkers() method as documented here:

https://docs.microsoft.com/en-us/dotnet/api/system.threading.threadpool.setminthreads?view=netstandard-2.0

NickCraver:
The next build will have `netstandard2.0` in it - I'm closing this one out as there's no update or information on how it relates really. The commit here introduces another package entirely which doesn't include support for TwemProxy (which isn't enabled by default anyway)...I'm not sure what condition this would make sense. Happy to discuss further, but we need more info.

PR #700, opened 2017-08-26
WIP: Unit Test Fixes and Upgrades

NickCraver:
Before converting most of the project to C# 7, `netstandard2.0`, etc. I want to get unit tests in shape. There's a fair bit of complication to doing so in that we need to test against:
- Different versions
- Clusters
- Master/Slave setups
- Multiple-test unfriendly tests (like SHUTDOWN commands)
- Simulate remote (read: latent) connections

This work does a few things:
- Moves all unit tests and assertions to xUnit
- Enables parallelization for tests
- Utilizes the xUnit output semantics, which means things like connection logs are on the test in VS, console output, build servers, etc. (it was a giant pool of `Console.WriteLine` mixed before)
- Cleans up the `\packages` folder completely (it wasn't used for anything but the `redis-*.exe` binaries
- Cleans up all test code formatting in general
- Re-enables `net462` in testing (was only `netcoreapp1.0` before)

#### Still TODO
- Better handling of Skips (e.g. if you aren't running clusters or 4.0 locally)
  - `[Theory]` skips
  - Feature-based skips
  - SSL-based skips
- Analyze all current failures and fix up if possible

#### Long-term TODOs
- Look at Redis 4.0 on Windows via WSL
- Configure endpoints in a `.json` config override that's in `.gitignore`
- Find a way to handle master/slave breaks on underpowered machines better

Posting this for review as I go, hopefully if people have time to eyeball they can spot existing errors with unit tests as well. Certainly far more cleanup can be done, and comments are appreciated.


NickCraver:
Merging into master to continue work on other PRs and getting decent CI validation up. Remaining tests are: "I don't think the test is wrong" failures I need to get with Marc on. The landscape has changed a bit.

PR #701, opened 2017-08-31
Rider friendly solution.

azinoviev:
GeoTests.cs converted to UTF-8, .idea dir added to .gitignore.


NickCraver:
Totally reasonable, thanks!

PR #702, opened 2017-08-31
Set SocketClosed on EndOfStreamException.

azinoviev:
It was unreachable because EndOfStreamException is inherited from IOException.


NickCraver:
Good catch, verified fix locally, thanks!

PR #707, opened 2017-09-05
Added a note on .NET Core and thread counts

JoeStead:
It threw us a bit when we saw this exception but not the same information :)


NickCraver:
I'd amend this slightly (as it won't be true soon), to:

> Also note that the IOCP and WORKER threads will _not_ be shown on .NET Core if using `netstandard` < 2.0.

Alongside Sentinel, I'll do a lot of cleanup and `netstandard2.0` work here shortly.

NickCraver:
Merged in, thank you!

PR #714, opened 2017-09-14
add solution doc for resolve timeout issue

RavenZZ:


NickCraver:
Good tip for users, thanks!

PR #735, opened 2017-10-30
http -> https for links in codebase.

nphmuller:
Except for redisearch.io and www.ideawu.com, because they don't have an https endpoint atm.


NickCraver:
Thank you!

PR #736, opened 2017-10-31
async timeout

deepakverma:
Hi Marc & Nick,

I have a proposal for introducing async timeout to the library. Can you please review it and let me know your inputs.

I have pending performance testing to be done for TimeoutAfter method to see if there's any scope for improvement.

Thanks,
Deepak

### DIFF
--- a/StackExchange.Redis.Tests/AsyncTests.cs
+++ b/StackExchange.Redis.Tests/AsyncTests.cs
@@ -1,4 +1,6 @@
-﻿using System.Linq;
+﻿using System;
+using System.Linq;
+using System.Threading.Tasks;
 using Xunit;
 using Xunit.Abstractions;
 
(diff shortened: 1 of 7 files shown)


deepakverma:
Hey Nick/Mark,

Any idea why the AppVeyor build would be failing? It looks like it's failing to restore nuget package for .netcore. It builds fine on my local machine with a clean build. I feel it shouldn't be caused by my changes to tests, but if you have seen this before?

Thanks,
Deepak

mgravell:
My main concern on this is that it adds a *lot* of `Task.Delay`, which is a lot of callbacks and thread-pool interactions. I wonder whether the "right" thing to do here is to keep an ordered queue of items that should timeout and have the *reader* kill them if they are overdue. That would work for the active reader implementation, but not really for the passive async reader code. And in either event it would be a massive change to the code. But I'm interested in your thoughts. Is `Task.Delay` scaleable for this usage?

NickCraver:
@deepakverma thoughts here? I'm in agreement with Marc on the overhead for everything here - it's just a very expensive approach even for the in the nominal case. Note: there's a mismatch on "(defaults to 1 second)" here, I'm not sure what the intent it, but the overhead when set here done really add up either way.

NickCraver:
Going to close this out due to inactivity, @deepakverma please ping if you want to continue discussions here, hopefully we could find some lower-overhead way of this happening if it's still desired.

PR #749, opened 2017-12-07
Feature/NRediSearch-TAGS

rajeshsubhankar:
Redisearch **TAG** feature has been added. Now user can create schema with TAG fields and separators. This PR also includes **FT.TAGVALS** to get all distinct values in a TAG field.

[TAGS](http://redisearch.io./Tags/)
[tagvals](http://redisearch.io/Commands/#fttagvals)

### DIFF
--- a/NRediSearch/Client.cs
+++ b/NRediSearch/Client.cs
@@ -139,6 +139,39 @@ namespace NRediSearch
         }
 
         /// <summary>
+        /// Return Distinct Values in a TAG field
+        /// </summary>
+        /// <param name="fieldName">TAG field name</param>
+        /// <returns>List of TAG field values</returns>
+        public RedisValue[] TagVals(string fieldName)
+        {
+            var args = new List<object>();
+            args.Add(_boxedIndexName);
+            args.Add(fieldName);
+
+            var resp = (RedisValue[])DbSync.Execute("FT.TAGVALS", args);
+            
+            return resp;
+        }
+
+        /// <summary>
+        /// Return Distinct Values in a TAG field
+        /// </summary>
+        /// <param name="fieldName">TAG field name</param>
+        /// <returns>List of TAG field values</returns>
+        public async Task<RedisValue[]> TagValsAsync(string fieldName)
+        {
+            var args = new List<object>();
+            args.Add(_boxedIndexName);
+            args.Add(fieldName);
+
+            var resp = (RedisValue[])await _db.ExecuteAsync("FT.TAGVALS", args).ConfigureAwait(false);
+            
+            return resp;
+        }
+
+
+        /// <summary>
         /// Add a single document to the query
         /// </summary>
         /// <param name="docId">the id of the document. It cannot belong to a document already in the index unless replace is set</param>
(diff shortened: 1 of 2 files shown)


NRediSearch/Client.cs
```diff
@@ -138,6 +138,39 @@ public SearchResult Search(Query q)
...
+        /// </summary>
+        /// <param name="fieldName">TAG field name</param>
+        /// <returns>List of TAG field values</returns>
+        public RedisValue[] TagVals(string fieldName)
+        {
+            var args = new List<object>();
+            args.Add(_boxedIndexName);
+            args.Add(fieldName);
+
+            var resp = (RedisValue[])DbSync.Execute("FT.TAGVALS", args);
```
mgravell: if the number of args is known and fixed: you don't need to use a list - there should be a `params object[]` overload, which means this should work?

    var resp = (RedisValue[])DbSync.Execute("FT.TAGVALS", _boxedIndexName, fieldName);

NRediSearch/Schema.cs
```diff
@@ -15,7 +15,8 @@ public enum FieldType
         {
             FullText,
             Geo,
-            Numeric
+            Numeric,
+            TAG
```
mgravell: is tag an abbreviation? or is this literally a "tag"? If the latter: `Tag` (capitalization) would be preferred

NRediSearch/Schema.cs
```diff
@@ -129,5 +131,36 @@ public Schema AddSortableNumericField(string name)
...
+        {
+            public string Separator { get; }
+            internal TagField(string name, string separator = ",") : base(name, FieldType.TAG, false)
+            {
+                Separator = separator;
+            }
+            internal override void SerializeRedisArgs(List<object> args)
+            {
+                base.SerializeRedisArgs(args);
+                if (!Separator.Equals(","))
```
mgravell: would prefer `if(Separator != ",")` here - avoids a possible NRE; note that C# has simpler operator usage than Java, so `==` does the obvious / correct thing

mgravell:
looks good; I've added some feedback

mgravell:
oof; `@@ -1,453 +1,445 @@` - looks like you did a ctrl+k,ctrl+d ? whitespace / document reformats are *fine*, but it is usually a good idea to keep those in a separate commit with *just* a reformat it in - it is virtually impossible to see the actual code changes in there now :(

mgravell:
hmmm; on second glance I'm not quite sure *what* happened; git isn't very happy with it, but it might not have been you; I'll try to wade through the red...

mgravell:
LGTM; merged; I can try to build/deploy later - thanks

rajeshsubhankar:
Thanks a lot.

dvirsky:
@mgravell can you please release a version of the client with this PR merged? Thanks.

rajeshsubhankar:
@dvirsky We have one more pending PR @ https://github.com/StackExchange/StackExchange.Redis/pull/754

It will be great if we can include that as well. Thanks.

mgravell:
Can you please confirm that 7 is the right flag? This is UseTermOffsets  | KeepFieldFlags  | UseScoreIndexes.

Are you sure this shouldn't be 8?

mgravell:
@rajeshsubhankar  k; I've sought clarification on that #754, but yes: let's get that merged (when clarified) and deployed

PR #753, opened 2017-12-13
Fix ResolveDns Documentation

nitinag:


NickCraver:
Ah, good fix - thanks!

PR #754, opened 2017-12-13
Disable Stop-Words added

rajeshsubhankar:
An option to disable Stop-Words completely while creating schema.

### DIFF
--- a/NRediSearch/Client.cs
+++ b/NRediSearch/Client.cs
@@ -34,7 +34,11 @@ namespace NRediSearch
             /// <summary>
             /// The default indexing options - use term offsets and keep fields flags
             /// </summary>
-            Default = UseTermOffsets | KeepFieldFlags
+            Default = UseTermOffsets | KeepFieldFlags ,
+            /// <summary>
+            /// If set, we will disable the Stop-Words completely
+            /// </summary>
+            DisableStopWords = 7
         }
         private static void SerializeRedisArgs(IndexOptions flags, List<object> args)
         {
@@ -49,6 +53,11 @@ namespace NRediSearch
             if ((flags & IndexOptions.UseScoreIndexes) == 0)
             {
                 args.Add("NOSCOREIDX".Literal());
+            }
+            if ((flags & IndexOptions.DisableStopWords) == IndexOptions.DisableStopWords)
+            {
+                args.Add("STOPWORDS".Literal());
+                args.Add(0);
             }
         }
         private readonly IDatabaseAsync _db;
--- a/NRediSearch/Client.cs
+++ b/NRediSearch/Client.cs
@@ -38,7 +38,7 @@ namespace NRediSearch
             /// <summary>
             /// If set, we will disable the Stop-Words completely
             /// </summary>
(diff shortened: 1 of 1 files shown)


mgravell:
Can you please confirm that 7 is the right flag? This is UseTermOffsets | KeepFieldFlags | UseScoreIndexes.

Are you sure this shouldn't be 8?

rajeshsubhankar:
Making it 8 will enable `NOOFFSETS`,`NOFIELDS `and `NOSCOREIDX `while someone has opted for **DisableStopWords** only.

What I assumed was, disabling stop words should not affect any other flags (here, it will not pass any other flags and let the RediSearch engine to create the schema with default configuration) .

Similarly if someone has opted for `UseTermOffsets  `or `KeepFieldFlags `, it will not enable **DisableStopWords** by default.

Please let me know if my assumption is wrong.

mgravell:
I don't know the interplay of how these flags are meant to interact. But if you use 7, then it *absolutely* affects other flags - you cannot specify `DisableStopWords` **without** also enabling `UseTermOffsets`, `KeepFieldFlags` and `UseScoreIndexes`. It would be impossible to specify `DisableStopWords` by itself, or `DisableStopWords | UseScoreIndexes`; anything that involves `DisableStopWords` automatically includes `UseTermOffsets`, `KeepFieldFlags` and `UseScoreIndexes` .

Likewise and for the same reason: if someone has intentionally enabled `UseTermOffsets`, `KeepFieldFlags` and `UseScoreIndexes` - then with the choice of 7: they will suddenly find that `DisableStopWords` is enabled. 

If that isn't what is intended, 7 is an inappropriate choice.

mgravell:
> Making it 8 will enable NOOFFSETS,NOFIELDSand NOSCOREIDXwhile someone has opted for DisableStopWords only.

That's surely expected? If they didn't specify `UseTermOffsets`, and *not specifying* `UseTermOffsets` means `NOOFFSETS` is added, then... that's exactly right, no?

rajeshsubhankar:
Well, that's correct.  But `UseTermOffsets` is by default enabled in `FT.CREATE` . So, from a developer point of view making it mandatory to provide this flag so as to enable it seems little odd to me.

rajeshsubhankar:
We can go with 8 so as to satisfy our previous flag configurations.  I will make the changes right now. Thanks.

mgravell:
> Well, that's correct. But `UseTermOffsets` is by default enabled in `FT.CREATE` . So, from a developer point of view making it mandatory to provide this flag so as to enable it seems little odd to me.

Ultimately that was a design choice you made when you switched from opt-out flags (`NOOFFSETS`) to opt-in flags (`UseTermOffsets`).

But: if they want "defaults plus no stopwords", then: `IndexOptions.Default | IndexOptions.DisableStopWords` (and changing to `8`, obviously)

mgravell:
k; merging; let me know if that all looks good to you - about to head out for food, but will try to build/deploy when I get back later

rajeshsubhankar:
Looks good. Thanks.

PR #761, opened 2017-12-27
Give defaultClientName a value when it is null

jeremymeng:
This prevents calling TryGetAzureRoleInstanceIdNoThrow() repeatedly in the null
case.

### DIFF
--- a/StackExchange.Redis.Tests/StackExchange.Redis.Tests.csproj
+++ b/StackExchange.Redis.Tests/StackExchange.Redis.Tests.csproj
@@ -25,8 +25,6 @@
     <PackageReference Include="Jil" Version="2.15.4" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.0.0" />
     <PackageReference Include="Moq" Version="4.7.99" />
-    <!-- Temp workaround for analyzer issue -->
-    <PackageReference Include="xunit.analyzers" Version="0.6.1" />
     <PackageReference Include="xunit" Version="$(xUnitVersion)" />
     <PackageReference Include="xunit.runner.visualstudio" Version="$(xUnitVersion)" />
     <DotNetCliToolReference Include="dotnet-xunit" Version="$(xUnitVersion)" />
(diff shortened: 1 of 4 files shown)


mgravell:
In principal: ok. But that final default value...
less OK. Something like "StackExchange.Redis" would be Ok, I guess.

jeremymeng:
Thanks @mgravell for reviewing.  Default value updated as suggested.

mgravell:
@NickCraver do you know anything about that temp reference that is being
removed as a side artefact here? I have zero context

jeremymeng:
Not sure about which xunit issue either. The xunit changes fix the build (although netcore tests are not running on appveyor). In xunit 2.3.1 xunit.analyzer is pulled in implicitly.

If needed I can split xunit changes into a separate PR.

jeremymeng:
netcoreapp tests are now running after upgrading to xunit 2.4 beta1 (and changes responding to obsoleted xunit API).

StackExchange.Redis.Tests/StackExchange.Redis.Tests.csproj
```diff
@@ -25,8 +25,6 @@
     <PackageReference Include="Jil" Version="2.15.4" />
     <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.0.0" />
     <PackageReference Include="Moq" Version="4.7.99" />
-    <!-- Temp workaround for analyzer issue -->
-    <PackageReference Include="xunit.analyzers" Version="0.6.1" />
```
jeremymeng: @NickCraver do you remember which issue this workaround refers to?

NickCraver:
Fixed conflicts here (including ref removal), that was a temp workaround for a null ref exception in a pre-release analyzer. After build here I'll get this merged in so a rebase of the netstandard2.0 bits is good to go.

NickCraver:
Test failures are existing (which we need to work on) - this doesn't regress any. Merging in, and thanks!

PR #767, opened 2018-01-16
Add netstandard2.0 support

jeremymeng:


jeremymeng:
This PR is re-based on the other open PR to allow netcoreapp1.0 tests in appveyor builds. I will rebase again after that PR is merged.

BasicTest/BasicTest.csproj
```diff
@@ -15,7 +15,11 @@
   </ItemGroup>
 
   <PropertyGroup Condition=" '$(TargetFramework)' == 'netcoreapp1.1' ">
-    <DefineConstants>$(DefineConstants);CORE_CLR</DefineConstants>
+    <DefineConstants>$(DefineConstants);NETSTANDARD1</DefineConstants>
+  </PropertyGroup>
+
+  <PropertyGroup Condition=" '$(TargetFramework)' == 'netcoreapp2.0' ">
+    <DefineConstants>$(DefineConstants)</DefineConstants>
   </PropertyGroup>
```
jeremymeng: Not used

StackExchange.Redis.Tests/StackExchange.Redis.Tests.csproj
```diff
@@ -17,16 +17,17 @@
     <DefineConstants>$(DefineConstants);FEATURE_MOQ</DefineConstants>
   </PropertyGroup>
   <PropertyGroup Condition=" '$(TargetFramework)' == 'netcoreapp1.0' ">
-    <DefineConstants>$(DefineConstants);FEATURE_MOQ;CORE_CLR</DefineConstants>
+    <DefineConstants>$(DefineConstants);FEATURE_MOQ;NETSTANDARD1</DefineConstants>
+  </PropertyGroup>
+  <PropertyGroup Condition=" '$(TargetFramework)' == 'netcoreapp2.0' ">
+    <DefineConstants>$(DefineConstants);FEATURE_MOQ</DefineConstants>
```
jeremymeng: `FEATURE_MOQ` can be removed now as all platforms have it

jeremymeng:
@NickCraver @mgravell please take a look.

wjdavis5:
This library already has .NETStandard 1.5 support - Whats the point in adding explicit support for 2.0? By supporting 1.5, they automatically support 2.0.

image

If you change the level of support to 2.0 you loose support for everything below it.:

https://github.com/dotnet/standard/blob/master/docs/versions.md

MichaCo:
@wjdavis5 the .NETStandard1.5 target depends on NETStandardLibrary 1.6.1 for example, and many other coreclr packages (a NETstandard2.0 library doesn't).
When you use .NETStandard2.0, those (most) dependencies are implicit.

That being said, you are right that the library will just work anyways in your projects which might target .NETCoreApp2.0 or something like that. I'm just not sure how well the nuget hell is handled... You could run into version conflicts when mixing the new packages with things which rely on older packages...

Regarding 
> If you change the level of support to 2.0 you loose support for everything below it

You can multitarget netstandard1.5 and netstandard2.0.

wjdavis5:
> the .NETStandard1.5 target depends on NETStandardLibrary 1.6.1 for example
Can you provide the docs for where you are getting this info? 

According to the package:
.NETStandard1.5 depends on
- NETStandard1.3
- NETStandard1.2
- NETStandard1.1

You can review the dependencies here:
https://www.nuget.org/packages/NETStandard.Library/
Like wise, .NETStandard 2.0 relies on everything below it.


Really I guess the point is, there is already .NETStandard 2.0 support

NickCraver:
I didn't see this come in - @jeremymeng I agree with these changes (and was about to do this same set today = so thanks!). I'll give these a go but have 2 asks please:
- Let's go with the built-in `NETSTANDARD1_5` now (and remove our def), since it'll all be minor-version dependent for surfaces after that anyway
- Can you please rebase?

I'll happily grab, test, and merge these ASAP.

NickCraver:
@wjdavis5 That's not really how it works - trust us that this is needed! You're looking at the NETStandard.Library package, which was *a dependency* in 1.5. Now we don't actually *need* this dependency, it was implicitly added at some point by tooling (which results in a really poor user experience). I never agreed with NETStandard.Library because of the litany of packages it presents a user at install time. Most of them which aren't needed. This is still causing me pain in other libs because dependencies beneath somewhere have it when they really only need 1 or 2 packages.

If we have this, users will get a drastically smaller install area, new features lit up, and with no downsides.

When libs have netstandard support, that doesn't mean all future versions are useless. New versions of the standard have increased API surface area, and if that can be useful then a lib can add a build...which is what's happening here too.

jeremymeng:
Thank you @NickCraver for reviewing it!

I've updated the PR.

> Let's go with the built-in NETSTANDARD1_5 now (and remove our def), since it'll all be minor-version dependent for surfaces after that anyway

I also was swinging between NETSTANDARD1 and the built-ins. The only slight benefit of NETSTANDARD1 is that I could use it in both library projects and the test project. I am going with the built-ins now.

NickCraver:
We're looking good here - I'm going to merge in and fix the already-broken tests separately. I'll also do some package investigations before we push anything up to NuGet. Thanks @jeremymeng!

wjdavis5:
@NickCraver  / @MichaCo  - I know you're both pretty busy but was hoping you could help me understand a little bit more. There is a similar conversation happening on the log4net mailing list and Im afraid there may be some misunderstanding. Just want to help encourage the right decision . Could you help us understand a little more how you benefit from this?

Issue #280, opened 2015-09-16
Enabling profiling causes random UnableToResolvePhysicalConnection exceptions


TheCloudlessSky:
I'm using StackExchange.Redis (1.0.481) successfully for a long time. I'm looking into adding profiling to report in our production environment. I have an MVC 5.2.3 application that's wired up via Ninject (DI). I've created an HttpModule that begins/ends profiling during the `BeginRequest` and `EndRequest` events (pretty much copied straight from the Docs):

``` csharp
public class HttpContextRedisProfiler : IProfiler
{
    private const string key = "HttpContextRedisProfiler.Context";

    public object GetContext()
    {
        var httpContext = HttpContext.Current;
        if (httpContext == null) return null;

        return httpContext.Items[key];
    }

    public object CreateContextForCurrentRequest()
    {
        var httpContext = HttpContext.Current;
        if (httpContext == null) return null;

        var result = new object();
        httpContext.Items[key] = result;
        return result;
    }
}

public class MeasureRedisModule : IHttpModule
{
    private readonly HttpContextRedisProfiler profiler;
    private readonly ConnectionMultiplexer connectionMultiplexer;

    public MeasureRedisModule(HttpContextRedisProfiler profiler, ConnectionMultiplexer connectionMultiplexer)
    {
        this.profiler = profiler;
        this.connectionMultiplexer = connectionMultiplexer;
    }

    public void Init(HttpApplication context)
    {
        context.BeginRequest += OnBeginRequest;
        context.EndRequest += OnEndRequest;
    }

    private void OnBeginRequest(object sender, EventArgs e)
    {
        var context = profiler.CreateContextForCurrentRequest();
        if (context != null)
        {
            connectionMultiplexer.BeginProfiling(context);
        }
    }

    private void OnEndRequest(object sender, EventArgs e)
    {
        var context = profiler.GetContext();
        if (context == null) return;

        var profiledCommands = connectionMultiplexer.FinishProfiling(context);

        foreach (var result in profiledCommands)
        {
            // Do something with the commands (e.g. log them to our metrics API).
        }
    }

    public void Dispose()
    {
    }
}

// Setup

var config = new ConfigurationOptions();
config.EndPoints.Add("localhost:6379");
// Increase the timeout so that AOF writes do not cause exceptions:
// https://github.com/StackExchange/StackExchange.Redis/issues/83
config.SyncTimeout = 2000;
[...]

PR #769, opened 2018-01-19
Removed invalid call to SetRequestSent for IMultiMessage

gliljas:
Hopefully solves #280. 

For `IMultiMessage`:s, such as a `ScriptEvalMessage`, `SetRequestSent` was called twice, which triggered an exception in `ProfileStorage`. The exception is weirdly rethrown as `UnableToResolvePhysicalConnection`.

Possible solutions:

1. Don't throw the exception in `ProfileStorage`
2. Detect in `Message.SetRequestSent` if the message was already sent, and don't forward the call to  `ProfileStorage` if it was.
3. Detect in `PhysicalBridge` whether the `IMultiMessage` instance was returned from `GetMessages` and ignore the extra call to `SetRequestSent` if it was.

I chose solution 3, since it scoped the change to `IMultiMessage`:s, but the other solutions should work equally fine. If it's really regarded as an error to call `SetRequestSent` twice, I think the exception should be thrown in `Message` instead.

### DIFF
--- a/Directory.build.props
+++ b/Directory.build.props
@@ -24,6 +24,6 @@
 
     <LibraryTargetFrameworks>net45;net46;netstandard1.5</LibraryTargetFrameworks>
     <CoreFxVersion>4.3.0</CoreFxVersion>
-    <xUnitVersion>2.3.0-beta5-build3750</xUnitVersion>
+    <xUnitVersion>2.3.1</xUnitVersion>
   </PropertyGroup>
 </Project>
--- a/StackExchange.Redis.Tests/Profiling.cs
+++ b/StackExchange.Redis.Tests/Profiling.cs
@@ -35,39 +35,42 @@ namespace StackExchange.Redis.Tests
                 db.StringSet("hello", "world");
                 var val = db.StringGet("hello");
                 Assert.Equal("world", (string)val);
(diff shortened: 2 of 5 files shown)


mgravell:
Will have a look right a way; intrigued.

gliljas:
I'm leaning towards option 2 instead. Have you had time to look at this @mgravell? It's preventing profiling to be used together with EVAL.

mgravell:
LGTM;  going to merge and dogfood  (if I push to public NuGet, I'll mark it as preview)

mgravell:
https://www.nuget.org/packages/StackExchange.Redis/1.2.7-alpha-00001

gliljas:
@mgravell Brilliant! I will test it right away!

gliljas:
@mgravell It indeed seems to have solved the problem.

footcha:
> https://www.nuget.org/packages/StackExchange.Redis/1.2.7-alpha-00001

Hello @mgravell,
will this fix be released as 1.2.x package? Or is the package 1.x unsupported since version 2.0 was released?
Thank you!
Petr

mgravell:
It is not my current plan to release any further 1.* packages; there would
need to be a pretty significant justification, I suspect.

PR #778, opened 2018-02-07
Add Transaction Conditions for Set Contains and Sorted Set Contains

AlphaGremlin:
This adds Transaction Conditions to check if a Set or Sorted Set member exists.

Extends the ExistsCondition type to support Set and SortedSet.
For Set, SISMEMBER works well with the same pattern as HEXISTS and EXISTS.
For SortedSet, there's no exact equivalent, so I picked ZSCORE (which is O(1)) and check if the result is null or not. ZRANK and ZREVRANK are O(log(N)).

All transaction tests passed against my local Redis instance.


NickCraver:
@mgravell any chance you can give this eyes? Have explicitly avoided this one as to not cause a conflict in the overall cleanup happening. It looks good to me.

NickCraver:
This looks good, thank you!

AlphaGremlin:
Awesome :)

PR #787, opened 2018-02-26
Provide a way to inject where to do background work.

aquartulli:
This changeset provides an injectable way of specifying where to execute background work in case developers want an alternative to the managed thread pool. It also provides a default implementation that uses the existing behavior.


NickCraver:
@aquartulli Can you please describe the use case here? Currently this is a breaking change (adding optional parameters still changes the method signature and breaks binary compatibility), so to properly support this we have to add additional overloads to most of the methods here. That adds a bunch of code hence the question: what's the use case in mind?

aquartulli:
We're looking to avoid contention between thread pool consumers by injecting a dedicated thread pool for Redis to use. This changeset is just the first step to put the existing behavior behind an interface.

NickCraver:
@aquartulli as-is, this is a breaking change - to do such an addition we would need to have multiple constructors. But, I don't think this is the proper place anyway. Look at SocketManager for instance, see how that's pluggable on the ConfigurationOptions? That's where such an option would belong, as a new property. Since `ConnectionMultiplexer` already has `.Connect*()` overloads taking this, there's also no breaking change in play.

aquartulli:
@NickCraver Planning on revisiting this. Is this something you'd be interested in accepting given a non-breaking change?

NickCraver:
@aquartulli Had to catch up on planning here - I would say at this time, this isn't a change we'd want to make. We're about to replace the networking stack with `async` machinery based on pipelines and how that acts with the threadpool is something we want to be simple at first (or forever, honestly) but at least as a step for. In short: we'd probably break any interface constructed here immediately, so it's not good time to tackle such a thing.

NickCraver:
Talked this over - confirmed above. Let's see what things look like after pipelines (currently in progress), but an interface isn't likely something we'd want to introduce here.

PR #796, opened 2018-03-10
Update Timeouts.md

wjdavis5:
The .NET Core documentations states that hte SetMinThreads Method is available in .NET Core 1.0+. Updated instructions to identify that. Also updated the link to the new documentation.


NickCraver:
Good changes, thanks!

PR #797, opened 2018-03-10
Remove TaskSource.CreateDenyExecSync

NickCraver:
After 50547eba453ff42b7b1b965f3f710dfcb177ca98, TaskSource.Create and TaskSource.CreateDenyExecSync methods are the same. There was some crud left everywhere that appeared like differences but it's really determined once in the static initializer for TaskSource based on the platform anyway.

I've fixed the tests as well - they were a bit backwards on logic/assumptions so hard to read, and have been failing for some time. Now we'd expect ContinueWithExecSync to *not* be hijacked, and everything else to be...that's how the tests are as well.

Note: the #if DEBUG change is because internals are already visible to tests...just simpler.

cc @mgravell for eyes - I've tested and running cross-platform. This is part of an overall effort to get tests green and cleanup in prep for network rewrite.

### DIFF
--- a/NRediSearch.Test/ExampleUsage.cs
+++ b/NRediSearch.Test/ExampleUsage.cs
@@ -36,7 +36,23 @@ namespace NRediSearch.Test
                 .AddTextField("body", 1.0)
                 .AddNumericField("price");
 
-            Assert.True(client.CreateIndex(sc, Client.IndexOptions.Default));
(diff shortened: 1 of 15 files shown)


NickCraver:
SSH hosed locally so can't force push a rebase - just opened a PR on another branch for now.

PR #799, opened 2018-03-11
AsyncSockets configuration option

paveliak:
We use single-box test environment with multiple services deployed. Each of those may create one or more ```ConnectionMultiplexer``` objects (for connection pooling) resulting overall in dozens of ```ConnectionMultiplexer``` objects across all processes talking to the local ```Redis``` instance. Each of ```ConnectionMultiplexer``` objects create two polling socket threads and consume CPU even if services are idle and slowly but surely draining the battery.

With 100 ```ConnectionMultiplexer``` objects I observe ~30% CPU consumption on my laptop. Settings ```HighPrioritySocketThreads``` to false does not really help. The proposal is to allow using asynchronous sockets for non-SSL connections which are enabled by default (and actually required) for SSL connections.

I am introducing new ```AsyncSockets``` configuration option which when set to true configures socket mode to ```SocketMode.Async``` even for non-SSL connections.

### DIFF
--- a/StackExchange.Redis.Tests/Config.cs
+++ b/StackExchange.Redis.Tests/Config.cs
@@ -335,5 +335,32 @@ namespace StackExchange.Redis.Tests
                 }
             }
         }
+        
+        [Fact]
+        public void AsyncSocketMode()
+        {
+            var options = ConfigurationOptions.Parse($"socketMode={SocketMode.Async}");
+            Assert.True(options.SocketMode.Equals(SocketMode.Async));
+        }
+
+        [Fact]
+        public void PollSocketMode()
+        {
+            var options = ConfigurationOptions.Parse($"socketMode={SocketMode.Poll}");
+            Assert.True(options.SocketMode.Equals(SocketMode.Poll));
+        }
+
+        [Fact]
+        public void InvalidSocketMode()
+        {
+            Assert.Throws<ArgumentOutOfRangeException>(() => ConfigurationOptions.Parse($"socketMode={SocketMode.Abort}"));
+        }
+
+        [Fact]
+        public void DefaultSocketMode()
+        {
+            var options = new ConfigurationOptions();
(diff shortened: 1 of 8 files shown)


NickCraver:
Question: why are you not using the same socket manager for many multiplexers instead? This is intended and supported behavior to resolve the situation you are in.

paveliak:
I was not aware of such configuration (thanks!) but pooling is not a problem in the single box test environment since we can configure pool size to 1. We usually have multiple services deployed and each service has at least 2 processes (worker and web role). With 5 services deployed (which is typical) there are at minimum 10 ConnectionMutiplexer objects.

paveliak:
@mgravell A colleague of mine brought another point in our conversation today towards this change. For the services that use Azure Redis there is a discrepancy in the communication mechanisms between the test and production deployments. This PR allows alleviating such discrepancy.

NickCraver:
@mgravell Does this make sense with pipelines approach coming up? I'm not sure which controls you have in mind.

As-is, I'd suggest using the existing enum for this option (e.g. "Poll" or "Async", defaulting to "Poll" as we do today) *if* it makes sense. We'd need to expose that enum for programmatic setting though, or create another that maps to it (and doesn't include Abort) - could definitely see that both ways.

paveliak:
@NickCraver I updated pull request per your proposal. Socket mode is now explicitly specified in the configuration options.

paveliak:
@NickCraver What is your opinion on the second iteration? Please let me know if you have any concerns in accepting this change.

NickCraver:
@paveliak I talked with Marc here - we want to hold off on this since it's something getting wholesale replaces with a new network layer soon. Since this option will disappear, having it as an example or in connection strings that get upgraded would be unnecessarily breaking.

NickCraver:
Closing this out for the reasons above - going to focus on sockets layer replacement instead.

PR #786, opened 2018-02-26
Add IsConnecting method to ConnectionMultiplexor, ServerEndpoint and …

tapmantwo:
…PhysicalBridge.

This is to assist with issue reported: https://github.com/StackExchange/StackExchange.Redis/issues/783

I'm not sure on best approach to write unit tests for this particular issue - if you have pointers, let me know and I'll happily add them in.

Issue #783, opened 2018-02-19
Safely reinitialising the multiplexer after network failure


tapmantwo:
We identified that if we receive `RedisConnectionException` with failure types of SocketFailure or SocketClosed, the multiplexer no longer works and we need to reinstantiate it.

This is tricky, because we could be in the middle of servicing many requests, and we want a singleton instance.  As such, we've wrapped the calling code in a critical section, and tried to determine whether the connection has happened by another caller by looking at the IsConnected property.

A bit like this (we've taken some Lazy type optimisations out for clarity);
```
            if (_connection != null &&_connection.IsConnected) return;

            lock (_lock)
            {

                if (_connection == null || _connection.IsConnected == false)
                {
                    _connection?.Dispose();
                    _connection = return ConnectionMultiplexer.Connect(_options);
                }                
            }
```

However, we are still getting race conditions - which is if the lock is released and the next caller enters and makes the IsConnected check whilst the multiplexer is 'Connecting'.  

There doesn't seem to be a way to directly determine that the servers are actually not yet connected.  I note in `PhysicalBridge.State`  has the following values `Connecting`, `ConnectedEstablishing` and `          ConnectedEstablished`.

Is it feasible for us to find out these states, so we can change our logic to somthing like;

```
            if (_connection != null &&_connection.IsConnected) return;

            lock (_lock)
            {

                if (_connection == null || (_connection.IsConnected == false || _connection.IsConnecting == false))
                {
                    _connection?.Dispose();
                    _connection = return ConnectionMultiplexer.Connect(_options);
                }                
            }
```

I'm happy to submit a PR for a `IsConnecting` method or similar on `ConnectionMultiplexer`, or take advice on better options :)

Thanks

Richard

PR #803, opened 2018-03-13
Isconnecting

tapmantwo:
…ing to establish a connection.

This is to assist with issue reported: #783

I'm not sure on best approach to write unit tests for this particular issue - if you have pointers, let me know and I'll happily add them in.

(This was previously submitted #786, but now updated to fix line ending probs).

### DIFF
--- a/StackExchange.Redis/StackExchange/Redis/PhysicalBridge.cs
+++ b/StackExchange.Redis/StackExchange/Redis/PhysicalBridge.cs
@@ -70,6 +70,8 @@ namespace StackExchange.Redis
 
         public bool IsConnected => state == (int)State.ConnectedEstablished;
 
+        public bool IsConnecting => state == (int)State.ConnectedEstablished || state == (int)State.Connecting;
+
         public ConnectionMultiplexer Multiplexer { get; }
 
         public ServerEndPoint ServerEndPoint { get; }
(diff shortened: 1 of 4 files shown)


StackExchange.Redis/StackExchange/Redis/PhysicalBridge.cs
```diff
@@ -70,6 +70,8 @@ public enum State : byte
 
         public bool IsConnected => state == (int)State.ConnectedEstablished;
 
+        public bool IsConnecting => state == (int)State.ConnectedEstablished || state == (int)State.Connecting;
```
NickCraver: This one doesn't make sense to me, why is `state == (int)State.ConnectedEstablished` considered "connecting"? It's already the case for IsConnected above, I'd expect this to only return `state == (int)State.Connecting`

tapmantwo: The rationale was that if it's not yet `Connected`, it is still `Connecting`.  However if it's easier to understand I'd be happy to adjust, as long as there is a mechanism to determine that the multiplexor is currently in the process of connecting.

NickCraver: @tapmantwo Not sure I follow, are you saying `state == (int)State.ConnectedEstablished` is "not yet connected"? That's the part I'm having a disconnect on, as we read that as "IsConnected" above

tapmantwo: Oh my, that is a good spot!  It was supposed to be State.ConnectedEstablishing.  I will update now :)

Issue #807, opened 2018-03-28
IPv6 is not worked by hostname


theanurin:
1.  Symptom
    Unhandled Exception: StackExchange.Redis.RedisConnectionException: It was not possible to connect to the redis server(s); to create a disconnected multiplexer, disable AbortOnConnectFail. SocketFailure on PING
       at StackExchange.Redis.ConnectionMultiplexer.ConnectImpl(Func1 multiplexerFactory, TextWriter log) in 
    C:\code\StackExchange.Redis\StackExchange.Redis\StackExchange\Redis\ConnectionMultiplexer.cs:line 876
       at ipv6issue.EntryPoint.Main(String[] args) in 
    C:\Users\me\Work\StackExchange.Redis.IPv6Issue\TestApp.cs:line 21

1. Version: 1.2.7-alpha-00002

1. Condition to reproduce the issue:
    Hostname "redis-server" that has only one DNS record AAAA = 2xxx:xxxx:xxxx::1

1.  Source code to reproduce the issue
   var options = new ConfigurationOptions();
   options.EndPoints.Add("redis-server", 6379);
   //options.EndPoints.Add("2xxx:xxxx:xxxx::1", 6379); // with this row works fine
   var redis = ConnectionMultiplexer.Connect(options);
   var sub = redis.GetSubscriber();
   sub.Publish("messages", "hello");

NickCraver:
Closing out to cleanup.

alexvaut:
We are facing the exact same issue with StackExchange.Redis 2.2.4

PR #808, opened 2018-03-29
Pull request for an issue #807. Add AddressFamily resolver to use correct AddressFamily with IPv6

theanurin:
I wrote separate method ResolveAddressFamily to resolve AddressFamily.
Previous implementation 
 `var addressFamily = endpoint.AddressFamily == AddressFamily.Unspecified ? AddressFamily.InterNetwork : endpoint.AddressFamily;`
is not worked when endpoint's hostname resolves as IPv6. In this case, we have IPv6 address but pass to Socket constructor AddressFamily.InterNetwork (IPv4).

### DIFF
--- a/StackExchange.Redis/StackExchange/Redis/SocketManager.cs
+++ b/StackExchange.Redis/StackExchange/Redis/SocketManager.cs
@@ -179,7 +179,7 @@ namespace StackExchange.Redis
 
         internal SocketToken BeginConnect(EndPoint endpoint, ISocketCallback callback, ConnectionMultiplexer multiplexer, TextWriter log)
         {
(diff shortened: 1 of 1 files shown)


NickCraver:
If you see in the build here: this introduces new failures and exception types. If a DNS entry is not found, a `SocketException` is thrown instead. See the DNS tests failing in the link above...this would be a behavior change and elude anyone currently catching and handling `RedisConnectionException` in their current code.

NickCraver:
I'm going to close this out because it's now handled in #893, which removes the need for us to resolve it. We can do this simpler now, and also support IPv6 addresses in connection strings directly :)

PR #812, opened 2018-04-04
Clarify the documentation for IDatabase(Async).HashSet(Async)

jasonmalinowski:
The documentation for the overload of IDatabaseAsync.HashSetAsync that takes multiple fields/values was somewhat ambiguous. It could have been read to mean that fields not set as a part of the HashSetAsync are deleted rather than preserved. The documentation for HMSET was clearer, so this borrows some of the verbiage from the Redis documentation into the documentation for HashSetAsync.


mgravell:
This change probably applies to IDatabase as well. Are you up for changing it there too? If not, I can merge and re-apply, so : fine either way.

jasonmalinowski:
@mgravell: Fixed!

Issue #789, opened 2018-03-01
Cluster pub/sub: keyspace notification issues


klords:
Hoping this is just something I'm doing wrong, but I have a cluster with two master-slaves. I've found I'm unable to use a pattern subscription because it just subscribes to one master's notifications (and redis doesn't broadcast notifications across cluster by design). I tried subscribing with a few individual keys that I know are on different masters and only one subset of that group is coming back. I'm not sure if this is because of a slot miscalculation or what.

A way to remedy this would be to allow manual endpoint selection at subscription, but endpoint selection seems to be done automatically based on channel during subscription.

klords:
I can't imagine I'm the first to want to use a pattern subscription to keyspace notifications in a cluster. Is there a workaround for this using this client?

mgravell:
For *regular* pub/sub (done manually), the events propagate around all nodes in the cluster, so it is only necessary to subscribe to a single node. Becuase of this, SE.Redis essentially uses a hash of the channel name to choose a node arbitrarily, and it will get the message.

It sounds like keyspace notifications *do not* propagate, which is causing this confusion. I can see two possible fixes for this:

1. detect and special-case the keyspace notification keys, and subscribe to all primary nodes
2. implement a new API specifically to make keyspace notifications a first-class citizen in the API

(or, of course: both)

This doesn't exist today, but i certainly something that we could look at.

NickCraver:
I'm going to close this out since a PR is open (and welcome!). Going through issues first, but #813 with some efficiency tweaks and tests added seems very appealing to me. Thanks for kicking this off @klords :)

PR #813, opened 2018-04-04
Cluster Keyspace Notification Support

klords:
Attempt at adding support for cluster keyspace notifications to subscriber by allowing subscription to multiple nodes.

The basic goal with this was to alter the internal logic of the `Subscription` class to accommodate connection to multiple endpoints for a single channel when a pattern-based keyspace/keyevent subscription is detected (rather than having `Subscription` simply represent a 1:1 relationship between channel/endpoint). The hope was that a subscriber could handle subscriptions uniformly, regardless of the special cases.

However, this approach has been problematic for me so far. This could be my setup, but it seems that there's a significant delay for the subscription to connect, which is causing the initial `PSUBSCRIBE` command messages to fail. Other times, the application will intermittently work, receiving notifications from the various masters in the cluster, and then will just start throwing repeated "No connection found to handle X operation" exceptions, occasionally correcting itself. So, it seems like something somewhere doesn't like me forcing these connections. I'm not sure if there's a flaw with my understanding of how the Subscriber fits into the application, or just the implementation (or both). 

I appreciate any feedback/insights. I'm happy to put in the time to make this work, with any guidance. Also, wasn't sure how to proceed with the tests as (last I checked) there were a few failing already. Happy to implement those too.

implements: #789

### DIFF
--- a/StackExchange.Redis/StackExchange/Redis/RedisChannel.cs
+++ b/StackExchange.Redis/StackExchange/Redis/RedisChannel.cs
@@ -12,6 +12,7 @@ namespace StackExchange.Redis
 
         internal readonly byte[] Value;
         internal readonly bool IsPatternBased;
+        internal readonly bool IsKeyspaceChannel;
 
         /// <summary>
         /// Indicates whether the channel-name is either null or a zero-length value
@@ -38,6 +39,7 @@ namespace StackExchange.Redis
         {
             Value = value;
             IsPatternBased = isPatternBased;
+            IsKeyspaceChannel = value != null && Encoding.UTF8.GetString(value).ToLower().StartsWith("__key");
         }
 
         private static bool DeterminePatternBased(byte[] value, PatternMode mode)
--- a/StackExchange.Redis/StackExchange/Redis/RedisSubscriber.cs
+++ b/StackExchange.Redis/StackExchange/Redis/RedisSubscriber.cs
@@ -1,5 +1,6 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Linq;
 using System.Net;
 using System.Threading;
 using System.Threading.Tasks;
@@ -144,7 +145,7 @@ namespace StackExchange.Redis
         private sealed class Subscription
         {
             private Action<RedisChannel, RedisValue> handler;
-            private ServerEndPoint owner;
+            private List<ServerEndPoint> owners = new List<ServerEndPoint>();
 
             public Subscription(Action<RedisChannel, RedisValue> value) => handler = value;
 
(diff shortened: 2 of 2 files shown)


StackExchange.Redis/StackExchange/Redis/RedisChannel.cs
```diff
@@ -38,6 +39,7 @@ private RedisChannel(byte[] value, bool isPatternBased)
         {
             Value = value;
             IsPatternBased = isPatternBased;
+            IsKeyspaceChannel = value != null && Encoding.UTF8.GetString(value).ToLower().StartsWith("__key");
```
NickCraver: Instead of creating a `byte[]` for each comparison - it'd be far cheaper to pre-compute the byte array of `"__key"` once in a static and compare them instead :)

StackExchange.Redis/StackExchange/Redis/RedisSubscriber.cs
```diff
@@ -170,33 +171,81 @@ public bool Remove(Action<RedisChannel, RedisValue> value)
...
-                if (oldOwner == null) return null;
+                if (owners.Count == 0) return null;
 
+                List<Task> queuedTasks = new List<Task>();
                 var cmd = channel.IsPatternBased ? RedisCommand.PUNSUBSCRIBE : RedisCommand.UNSUBSCRIBE;
                 var msg = Message.Create(-1, flags, cmd, channel);
                 if (internalCall) msg.SetInternalCall();
-                return oldOwner.QueueDirectAsync(msg, ResultProcessor.TrackSubscriptions, asyncState);
+                foreach (var owner in owners)
+                    queuedTasks.Add(owner.QueueDirectAsync(msg, ResultProcessor.TrackSubscriptions, asyncState));
```
NickCraver: `owners` isn't locked here, so this can go boom under load. The previous interlock above prevented this scenario.

StackExchange.Redis/StackExchange/Redis/RedisSubscriber.cs
```diff
@@ -170,33 +171,81 @@ public bool Remove(Action<RedisChannel, RedisValue> value)
...
+                owners.Clear();
+                return Task.WhenAll(queuedTasks.ToArray());
             }
 
-            internal ServerEndPoint GetOwner() => Interlocked.CompareExchange(ref owner, null, null);
+            internal ServerEndPoint GetOwner()
+            {
+                var owner = owners?[0]; // we subscribe to arbitrary server, so why not return one
+                return Interlocked.CompareExchange(ref owner, null, null);
```
NickCraver: This doesn't make sense anymore...would need to lock `owners` here instead.

Davidnovarro:
Somebody, please merge this PR.

klords:
Wow this completely fell off my radar. I'll have a look at the feedback and
current state of things and update as appropriate.

StackExchange.Redis/StackExchange/Redis/RedisChannel.cs
Davidnovarro: Here is the fixed version.
https://pastebin.com/Q9T0Q99K

G31st:
Will this feature by available in 2.1 ? its not on the release notes.

DanoThom:
Bumping this as it's been a while and would be a fantastic feature to be able to utilize.

ercangorgulu:
> Bumping this as it's been a while and would be a fantastic feature to be able to utilize.

I adapted the feature with new code and created pull request. You can look at [here](https://github.com/StackExchange/StackExchange.Redis/pull/1536)

NickCraver:
Closing in favor of #1536 to get this going, though we still need to add testing there.

PR #839, opened 2018-05-28
Makes Exception constructors public

### LINKED ISSUE #825: Mocking exceptions in tests
We've created a wrapper service around an IDatabase and we want to test how our code behaves when the IDatabase methods throw exceptions.
We cannot create the exceptions, since their constructors are internal. Is there a way to make this possible?

The unittest code is as follows (we use Moq as mocking library):
```
mockDbService.Setup(
	m => m.GetAsync<string>(KEY_NOT_EXIST, null)
)
.Throws(new RedisConnectionException("mock exception"));

string result = await dbService.GetAsync<string>(KEY_NOT_EXIST);

Assert.Null(result);
```

Is there a way around the `internal`ness of the exception classes?

One way to do this is to have a separate assembly `[assembly:InternalsVisibleTo("Stackoverflow.Redis.TestHelper")]`. The helper should expose some methods to create the exceptions...

q42jaap:
In unit tests mocking out an IDatabase with implementations that throw
an error was not possible. Constructing the Exception instances wasn't
accessible in user code. It is now.

This fixes #825.

### DIFF
--- a/StackExchange.Redis/StackExchange/Redis/Exceptions.cs
+++ b/StackExchange.Redis/StackExchange/Redis/Exceptions.cs
@@ -76,8 +76,8 @@ namespace StackExchange.Redis
     /// </summary>
     public sealed partial class RedisCommandException : Exception
     {
-        internal RedisCommandException(string message) : base(message) { }
-        internal RedisCommandException(string message, Exception innerException) : base(message, innerException) { }
+        public RedisCommandException(string message) : base(message) { }
+        public RedisCommandException(string message, Exception innerException) : base(message, innerException) { }
     }
 
     /// <summary>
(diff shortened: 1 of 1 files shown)


NickCraver:
Thanks @q42jaap! This seems perfectly reasonable, it'll be on MyGet shortly and in the next release :)

NickCraver:
Hmm build error though I'm not spotting it here - grabbing local for a look before this goes in.

StackExchange.Redis/StackExchange/Redis/Exceptions.cs
```diff
@@ -76,16 +76,16 @@ public sealed partial class RedisServerException : RedisException
...
+        public RedisCommandException(string message, Exception innerException) : base(message, innerException) { }
     }
 
     /// <summary>
     /// Indicates the time allotted for a command or operation has expired.
     /// </summary>
     public sealed partial class RedisTimeoutException : TimeoutException
     {
-        internal RedisTimeoutException(string message, CommandStatus commandStatus) : base(message)
+        pulic RedisTimeoutException(string message, CommandStatus commandStatus) : base(message)
```
NickCraver: `public` - always be sure to try to build locally ;)

NickCraver:
Pushed some fixes up - just waiting on the build.

PR #850, opened 2018-06-08
Added support for UnixDomainSocketEndPoint introduced in .NET Core 2.1

VasiliyNovikov:
.NET Core 2.1 introduced `UnixDomainSocketEndPoint` and it doesn't work with redis client because of 2 things:  

1. `ProtocolType` should be `Unspecified` instead of `Tcp` otherwise Socket constructor throws that it is not supported. [See these tests in corefx repo](https://github.com/dotnet/corefx/blob/b21589f01d699617fb2fba6589af4e43c02d4f31/src/System.Net.Sockets/tests/FunctionalTests/UnixDomainSocketTest.netcoreapp.cs) - they use `ProtocolType.Unspecified`  
2.  `Socket.NoDelay` is not supported by Unix domain sockets - it throws when trying to set it

### DIFF
--- a/StackExchange.Redis/StackExchange/Redis/SocketManager.cs
+++ b/StackExchange.Redis/StackExchange/Redis/SocketManager.cs
@@ -190,9 +190,13 @@ namespace StackExchange.Redis
             }
 
             var addressFamily = endpoint.AddressFamily == AddressFamily.Unspecified ? AddressFamily.InterNetwork : endpoint.AddressFamily;
-            var socket = new Socket(addressFamily, SocketType.Stream, ProtocolType.Tcp);
+            var protocolType = addressFamily == AddressFamily.Unix ? ProtocolType.Unspecified : ProtocolType.Tcp;
+            var socket = new Socket(addressFamily, SocketType.Stream, protocolType);
             SetFastLoopbackOption(socket);
-            socket.NoDelay = true;
+            if (addressFamily != AddressFamily.Unix)
+            {
+                socket.NoDelay = true;
+            }
             try
             {
                 var formattedEndpoint = Format.ToString(endpoint);


mgravell:
lgtm; did you have an example of how to initialize this usage?

VasiliyNovikov:
To test this you'd need to enable unix sockets in redis configuration. E.g.:
```
unixsocket /var/run/redis/redis.sock
unixsocketperm 770
```
Add current user to redis group (if redis runs under own user and group)
In the code you can do the following:
```csharp
var redisConfig = new ConfigurationOptions();
redisConfig.EndPoints.Add(new UnixDomainSocketEndPoint("/var/run/redis/redis.sock"));
var redis = await ConnectionMultiplexer.ConnectAsync(redisConfig);
```

mgravell:
It pleases me greatly that it works so cleanly. I wonder if we should also
support this in the stringified parse/format. Trying to think if there's a
good unified way to do that. Happy to merge as-is, though.

VasiliyNovikov:
Not sure if it is good example but here is url format supported by redis-py:
https://github.com/andymccurdy/redis-py/blob/5109cb4f6b610e8d5949716a16435afbbf35075a/redis/connection.py#L789

mgravell:
that's useful context, thanks

NickCraver:
@mgravell I think, we'd need to parse the endpoint for type in the config generically (replacing the `addressFamily` ternary above this change). IMO, let's do that in a separate pass. It overlaps with #808, as we currently don't recognize IPv6 correctly either. We probably want a parse for these to recognize the big ones (IPv6/Unix may be it) then default down to IPv4 as we do today.

Thoughts? I can take a look at that if agreed.

VasiliyNovikov:
Hi, what the status of the PR?

mgravell:
good to merge; just haven't done it yet

VasiliyNovikov:
Thanks

NickCraver:
Merging in - will need to merge over to pipelines.

VasiliyNovikov:
Thanks for merging it

PR #858, opened 2018-06-16
Pipelines/2.0: Update to GitVersioning, fix build, and .StrongName cleanup

NickCraver:
This updates to GitVersioning, `version.json`, and fixes the CI build.

Note: the downside is breaks @mgravell locally (by removing the package source), but if that `Nuget.config` is checked in, anyone lacking that directory will break, including our CI builds. I recommend we just not checkin that change, or remove the `<clear/>` while working on this branch (or forever?) and allow local configs to work. I'm thinking the `<clear />` removal is safest overall given the constraints here...we can always restore it later if we find a better option.

Thoughts?


NickCraver:
Checked with Marc on screwing-his-local...all good. Merging in to fix CI.

Issue #354, opened 2016-02-04
Can't write to slave


tallesl:
I have a slave with `slave-read-only` set to `no` but when I attempt to do a SADD to it I got a `No connection is available to service this operation`.
When I issue SMEMBERS it works fine, I'm assuming it's a write problem.

```
at StackExchange.Redis.ConnectionMultiplexer.ExecuteSyncImpl[T](Message message, ResultProcessor`1 processor, ServerEndPoint server) in c:\TeamCity\buildAgent\work\3ae0647004edff78\StackExchange.Redis\StackExchange\Redis\ConnectionMultiplexer.cs:line 1857
at StackExchange.Redis.RedisBase.ExecuteSync[T](Message message, ResultProcessor`1 processor, ServerEndPoint server) in c:\TeamCity\buildAgent\work\3ae0647004edff78\StackExchange.Redis\StackExchange\Redis\RedisBase.cs:line 80
at StackExchange.Redis.RedisDatabase.SetAdd(RedisKey key, RedisValue[] values, CommandFlags flags) in c:\TeamCity\buildAgent\work\3ae0647004edff78\StackExchange.Redis\StackExchange\Redis\RedisDatabase.cs:line 921
```

Inspecting the variables while debugging, I noticed that the Server object has `SlaveReadOnly` set to `false` as expected, but there's `AllowSlaveWrites` also set to `false`. I don't know how those are set (or what exactly they affect), but I forcefully changed `AllowSlaveWrites` to `true` with no luck.

Setting `abortConnect=false` on configuration string didn't help either.

If I set the instance to master everything works fine.
It's reproducible in my development machine where the redis configuration is pretty much vanilla.
This is not an intermittent error, I get it as soon I start the application.

tallesl:
Looks like that is [IsMasterOnly](https://github.com/StackExchange/StackExchange.Redis/blob/567cdf7/StackExchange.Redis/StackExchange/Redis/Message.cs#L382) fault.

Is writing to a slave even supported? If not, why not?

mgravell:
I will need to revisit the code for determining when slave writes are
legal. At the moment IIRC the code assumes the "safe" default of
master-only writes. But : this is indeed configurable in Redis (and before
it was configurable: slave writes were allowed).

So: yes, I agree we should look at this. But: a part of me will always
wonder why people would want out-of-step slaves :)

tallesl:
[...]

PR #860, opened 2018-06-17
Initial implementation of Redis Streams.

ttingen:
The Streams data type is available in release 5.0 RC1 and above.

- Implemented Sync & Async methods for all Stream related commands (minus the blocking options) as of 5.0 RC1.
- Added tests for the synchronous versions of the Streams API but the testing is a work in progress. Need to refactor for reuse within the streams tests and write a more thorough suite of tests.
- Added a NameValueEntry struct which mimicks HashEntry. Using HashEntry for the name/value pairs of stream entries seemed wrong. Perhaps refactor the usage of HashEntry to the more generic NameValueEntry and deprecate HashEntry?


mgravell:
Oh, awesome. I can see I have quite some reading ahead, but I guess that's
better than reading + writing :)

Can I confirm that you are able and willing to freely contribute this PR
under the project's license? Essentially the "no I won't sue you and claim
I own that part of the code later" legals check?

NickCraver:
Lots of reading to do indeed! I just went through a quick cursory scan of this and I wanted to say: this is **fantastic**.

It's the largest first time contribution I've seen to any project and also appears to be very well polished. **Thank you**. Marc and I were already talking about adding this and the other APIs we haven't caught up with (including overloads) in v2 that we're working on now. This is perfect timing too. Marc's busy on the pipelines branch at the moment but we'll give this a shakedown and get it in ASAP. The test suite helps immensely here. Again: very well done. I'm impressed.

ttingen:
@mgravell Yes, I'm willling to contribute this PR under the project's license.

Thanks for the positive feedback, streams is a feature that I look forward to using. 

I already see some things that my hasty review last night at midnight (EDT) missed. Copypasta got me on some of the unit test comments and the XML documentation on the NameValueEntry struct still references HashEntry, those things need to be cleaned up.

I also didn't like the naming of some of the API methods and parameters (e.g., StreamClaimMessagesReturningIds, and some others).  The names are sufficiently descriptive but I think they could be improved. Naming is hard...

NickCraver:
Need to go through code next, but have 5.0-rc3 up and running and test suite against it here via WSL:
image
image

I'll start on code review right now, but will take a bit since you've done so much work here :)

StackExchange.Redis/StackExchange/Redis/RedisStream.cs
```diff
@@ -0,0 +1,32 @@
...
+
+        /// <summary>
+        /// The key for the stream.
+        /// </summary>
+        public RedisKey Key => key;
+
+        /// <summary>
+        /// An arry of entries contained within the stream.
+        /// </summary>
+        public RedisStreamEntry[] Entries => entries;
```
NickCraver: Can simplify this and other structs down to `{ get; }`-only properties, e.g. `public RedisStreamEntry[] Entries { get; }` here.

StackExchange.Redis/StackExchange/Redis/ResultProcessor.cs
```diff
@@ -1301,6 +1325,461 @@ protected override bool SetResultCore(PhysicalConnection connection, Message mes
...
+                    // 1) 1) "mystream"                     <== Skip the stream name
+                    //    2) 1) 1) 1519073278252 - 0        <== Index 1 contains the array of messages
+                    //          2) 1) "foo"
+                    //             2) "value_1"
+                    //       2) 1) 1519073279157 - 0
+                    //          2) 1) "foo"
+                    //             2) "value_2"
+
+                    result = result.GetItems()[0]  // Initial result array, then the first stream.
+                                   .GetItems()[1]; // Skip the stream name and return the array of stream entries.
```
NickCraver: Should use a local for `result.GetItems()` here and access it twice.

StackExchange.Redis/StackExchange/Redis/ResultProcessor.cs
```diff
@@ -1301,6 +1325,461 @@ protected override bool SetResultCore(PhysicalConnection connection, Message mes
...
+            {
+                this.skipStreamName = skipStreamName;
+            }
+
+            protected override bool SetResultCore(PhysicalConnection connection, Message message, RawResult result)
+            {
+                if (result.IsNull)
+                {
+                    // Server returns 'nil' if no entries are returned for the given stream.
+                    SetResult(message, new RedisStreamEntry[0]);
```
NickCraver: Use `Array.Empty<RedisStreamEntry>()` instead here

StackExchange.Redis/StackExchange/Redis/ResultProcessor.cs
```diff
@@ -1301,6 +1325,461 @@ protected override bool SetResultCore(PhysicalConnection connection, Message mes
...
+                            3) "surname"
+                            4) "Austen"
+            */
+
+            protected override bool SetResultCore(PhysicalConnection connection, Message message, RawResult result)
+            {
+                if (result.IsNull)
+                {
+                    // Nothing returned for any of the requested streams. The server returns 'nil'.
+                    SetResult(message, new RedisStream[0]);
```
NickCraver: `Array.Empty<RedisStream>()`

StackExchange.Redis/StackExchange/Redis/ResultProcessor.cs
```diff
@@ -1301,6 +1325,461 @@ protected override bool SetResultCore(PhysicalConnection connection, Message mes
...
+                    return false;
+                }
+
+                var arr = result.GetItems();
+                var parsedItems = new T[arr.Length];
+
+                for (var i = 0; i < arr.Length; i++)
+                {
+                    parsedItems[i] = ParseItem(arr[i]);
+                }
```
NickCraver: Can simplify this and other similar loops using `Array.ConverAll`, e.g.
```c#
var parsedItems = Array.ConvertAll(arr, ParseItem);
```

StackExchange.Redis/StackExchange/Redis/ResultProcessor.cs
```diff
@@ -1301,6 +1325,461 @@ protected override bool SetResultCore(PhysicalConnection connection, Message mes
...
+                var groupCount = (int)arr[7].AsRedisValue();
+
+                // Note: Even if there is only 1 message in the stream, this command returns
+                //       the single entry as the first-entry and last-entry in the response.
+
+                var entries = ParseRedisStreamEntries(new RawResult(new RawResult[]
+                {
+                    arr[9],
+                    arr[11]
+                }));
```
NickCraver: This one is a bit round-about to get the parsed result - we should probably break up the parsing so it's accessible/this is more direct without the intermediate creations.

StackExchange.Redis/StackExchange/Redis/ResultProcessor.cs
```diff
@@ -1301,6 +1325,461 @@ protected override bool SetResultCore(PhysicalConnection connection, Message mes
...
+                    return null;
+                }
+
+                NameValueEntry[] pairs;
+
+                int count = arr.Length / 2;
+
+                if (count == 0)
+                {
+                    pairs = new NameValueEntry[0];
```
NickCraver: `Array.Empty<NameValueEntry>()`

Also: would just `return` directly here and in the else, keep it simple!

Note: it's okay to drop `net45` from the `.csproj` to get this to compile...we're already on `netstandard2.0`-only in the `pipelines` branch already, so no loss :)

StackExchange.Redis/StackExchange/Redis/StreamConstants.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        /// <summary>
+        /// The "-" value used in the XRANGE, XREAD, and XREADGROUP commands. Indicates the minimum message ID from the stream.
+        /// </summary>
+        public static readonly RedisValue StreamMinValue = "-";
+
+        /// <summary>
+        /// The "+" value used in the XRANGE, XREAD, and XREADGROUP commands. Indicates the maximum message ID from the stream.
+        /// </summary>
+        public static readonly RedisValue StreamMaxValue = "+";
```
NickCraver: Naming: suggest `ReadMinValue`/`ReadMaxValue` (rather than repeating `Stream`), thoughts?

StackExchange.Redis/StackExchange/Redis/StreamConsumer.cs
```diff
@@ -0,0 +1,19 @@
+﻿
+namespace StackExchange.Redis
+{
+    /// <summary>
+    /// Describes a consumer off a Redis Stream.
+    /// </summary>
+    public class StreamConsumer
```
NickCraver: Would make this a `struct` (with constructor)

StackExchange.Redis/StackExchange/Redis/RedisStreamEntry.cs
```diff
@@ -0,0 +1,42 @@
+﻿namespace StackExchange.Redis
+{
+    /// <summary>
+    /// Describes an entry contained in a Redis Stream.
+    /// </summary>
+    public struct RedisStreamEntry
+    {
+        internal readonly RedisValue id;
+        internal readonly NameValueEntry[] values;
```
NickCraver: Can use `{ get; }`-only properties here

StackExchange.Redis/StackExchange/Redis/RedisDatabase.cs
```diff
@@ -2164,6 +2805,171 @@ private Message GetSortedSetRemoveRangeByScoreMessage(RedisKey key, double start
...
+                                    RedisCommand.XPENDING,
+                                    values);
+        }
+
+        private Message GetStreamReadGroupMessage(RedisKey key, RedisValue groupName, RedisValue consumerName, RedisValue readFromId, int? count, CommandFlags flags)
+        {
+            // Example: > XREADGROUP GROUP mygroup Alice COUNT 1 STREAMS mystream >
+            if (count.HasValue && count <= 0)
+            {
+                throw new ArgumentOutOfRangeException("count must be greater than 0.");
```
NickCraver: `throw new ArgumentOutOfRangeException(nameof(count), "count must be greater than 0.");`
(current code is calling the `string paramName` overload)

StackExchange.Redis/StackExchange/Redis/RedisDatabase.cs
```diff
@@ -2164,6 +2805,171 @@ private Message GetSortedSetRemoveRangeByScoreMessage(RedisKey key, double start
...
+            // 2) 1) 1526569506935 - 0
+            //    2) "Bob"
+            //    3) (integer)74170458
+            //    4) (integer)1
+
+            // See https://redis.io/topics/streams-intro.
+
+            if (count <= 0)
+            {
+                throw new ArgumentOutOfRangeException("count must be greater than 0.");
```
NickCraver: `throw new ArgumentOutOfRangeException(nameof(count), "count must be greater than 0.");`

StackExchange.Redis/StackExchange/Redis/RedisDatabase.cs
```diff
@@ -2164,6 +2805,171 @@ private Message GetSortedSetRemoveRangeByScoreMessage(RedisKey key, double start
...
+
+            return Message.Create(Database, flags, RedisCommand.XACK, arr);
+        }
+
+        private Message GetStreamAddMessage(RedisKey key, RedisValue entryId, int? maxLength, bool useApproximateMaxLength, CommandFlags flags, params NameValueEntry[] streamFields)
+        {
+            // See https://redis.io/commands/xadd.
+
+            if (streamFields == null) throw new ArgumentNullException(nameof(streamFields));
+            if (streamFields.Length == 0) throw new ArgumentOutOfRangeException("streamFields must contain at least one field.");
```
NickCraver: `throw new ArgumentOutOfRangeException(nameof(streamFields), "streamFields must contain at least one field.");`

StackExchange.Redis/StackExchange/Redis/RedisDatabase.cs
```diff
@@ -2164,6 +2805,171 @@ private Message GetSortedSetRemoveRangeByScoreMessage(RedisKey key, double start
...
+
+            if (maxLength.HasValue && maxLength <= 0)
+            {
+                throw new ArgumentOutOfRangeException("maxLength must be greater than 0.");
```
NickCraver: `throw new ArgumentOutOfRangeException(nameof(maxLength), "maxLength must be greater than 0.");`

StackExchange.Redis/StackExchange/Redis/RedisDatabase.cs
```diff
@@ -1579,6 +1579,605 @@ public IEnumerable<SortedSetEntry> SortedSetScan(RedisKey key, RedisValue patter
             return ExecuteAsync(msg, ResultProcessor.NullableDouble);
         }
 
+        public long StreamAcknowledge(RedisKey key, RedisValue groupName, string messageId, CommandFlags flags = CommandFlags.None)
```
NickCraver: I think `messageId` (here and elsewhere - returns included) should be a `RedisValue` going in (for consistency as well), since we're converting underneath on every path used. Thoughts?

StackExchange.Redis/StackExchange/Redis/RedisDatabase.cs
```diff
@@ -1579,6 +1579,605 @@ public IEnumerable<SortedSetEntry> SortedSetScan(RedisKey key, RedisValue patter
...
+        {
+            var msg = GetStreamPendingMessagesMessage(key, groupName, minId, maxId, count, RedisValue.Null, flags);
+            return ExecuteAsync(msg, ResultProcessor.StreamPendingMessages);
+        }
+
+        public Task<StreamPendingMessageInfo[]> StreamPendingMessageInfoGetAsync(RedisKey key, RedisValue groupName, RedisValue minId, RedisValue maxId, int count, RedisValue consumerName, CommandFlags flags = CommandFlags.None)
+        {
+            var msg = GetStreamPendingMessagesMessage(key, groupName, minId, maxId, count, consumerName, flags);
+            return ExecuteAsync(msg, ResultProcessor.StreamPendingMessages);
+        }
```
NickCraver: nit: can we keep sync/async pattern on methods here? last 4 aren't paired up

StackExchange.Redis/StackExchange/Redis/RedisDatabase.cs
```diff
@@ -1579,6 +1579,605 @@ public IEnumerable<SortedSetEntry> SortedSetScan(RedisKey key, RedisValue patter
...
+
+        public RedisStreamEntry[] StreamRange(RedisKey key, RedisValue minId, RedisValue maxId, CommandFlags flags = CommandFlags.None)
```
NickCraver: Same for below - instead of 2 methods - how about we make `count` an `int?` on these? That'd take us from 4 (sync+async) methods down to 2.

StackExchange.Redis/StackExchange/Redis/RedisDatabase.cs
```diff
@@ -1579,6 +1579,605 @@ public IEnumerable<SortedSetEntry> SortedSetScan(RedisKey key, RedisValue patter
...
+                    maxId,
+                    minId,
+                    StreamConstants.Count,
+                    count
+                });
+
+            return ExecuteSync(msg, ResultProcessor.SingleStream);
+        }
+
+        public RedisStreamEntry[] StreamRangeReverse(RedisKey key, RedisValue maxId, RedisValue minId, CommandFlags flags = CommandFlags.None)
```
NickCraver: Same as above - by naming `count` nullable, we can collapse overloads here and simplify

Overall: for API consistency instead of `StreamRangeReverse`, look at how `SortedSetRangeByRank` and others work - add a `Order order = Order.Ascending` argument in the same way - this again halves our overloads needed :)

StackExchange.Redis/StackExchange/Redis/RedisDatabase.cs
```diff
@@ -1579,6 +1579,605 @@ public IEnumerable<SortedSetEntry> SortedSetScan(RedisKey key, RedisValue patter
...
+                {
+                    StreamConstants.Streams,
+                    key.AsRedisValue(),
+                    afterId
+                });
+
+            return ExecuteAsync(msg, ResultProcessor.SingleStreamWithNameSkip);
+        }
+
+        public Task<RedisStreamEntry[]> StreamReadAsync(RedisKey key, RedisValue afterId, int count, CommandFlags flags = CommandFlags.None)
```
NickCraver: Same as elsewhere: `int? count` and collapse overloads

StackExchange.Redis/StackExchange/Redis/RedisDatabase.cs
```diff
@@ -1579,6 +1579,605 @@ public IEnumerable<SortedSetEntry> SortedSetScan(RedisKey key, RedisValue patter
...
+            return ExecuteAsync(msg, ResultProcessor.SingleStreamWithNameSkip);
+        }
+
+        public Task<RedisStream[]> StreamReadAsync(IList<KeyValuePair<RedisKey, RedisValue>> streamWithAfterIdList, CommandFlags flags = CommandFlags.None)
+        {
+            var msg = GetMultiStreamReadMessage(streamWithAfterIdList, flags);
+            return ExecuteAsync(msg, ResultProcessor.MultiStream);
+        }
+
+        public Task<RedisStream[]> StreamReadAsync(IList<KeyValuePair<RedisKey, RedisValue>> streamWithAfterIdList, int countPerStream, CommandFlags flags = CommandFlags.None)
```
NickCraver: Collapse: `int? count` with above.

ttingen:
Lots to update. I will start on these changes tomorrow evening.

StackExchange.Redis/StackExchange/Redis/RedisDatabase.cs
mgravell: I thought there used to be a test that checked this with reflection... If not, I can add

StackExchange.Redis/StackExchange/Redis/RedisDatabase.cs
```diff
@@ -2164,6 +2805,171 @@ private Message GetSortedSetRemoveRangeByScoreMessage(RedisKey key, double start
                     GetRange(start, exclude, true), GetRange(stop, exclude, false));
         }
 
+        private Message GetStreamAcknowledgeMessage(RedisKey key, RedisValue groupName, CommandFlags flags, params string[] messageIds)
```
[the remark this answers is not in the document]
mgravell: Passing query: is messageId a strict string? Or should it be RedisValue?

StackExchange.Redis/StackExchange/Redis/ResultProcessor.cs
mgravell: Nit for @NickCraver: prefer:

```
var parsedItems = Array.ConvertAll(arr, item => ParseItem(item));
```

Unless I'm mistaken, they're still subtly different. Happy to find out they fixed the compiler, of course.

StackExchange.Redis/StackExchange/Redis/ResultProcessor.cs
ttingen: Something like this?

image

ttingen: Something like this? 

image

ttingen: This looks cleaner than the previous version. Another option would be to specify the constructor parameters by name instead of with the comment. I like to have an explicit identifier (named param or comment) of what it is I'm passing if it's not readily apparent from the value being passed (just a personal preference).

StackExchange.Redis/StackExchange/Redis/RedisDatabase.cs
ttingen: I'm not sure I understand this one, there are sync/async methods for the StreamPendingMessageInfoGet operation. Or is the naming too similar to StreamPendingInfoGet?

StreamPendingInfoGet returns a summary of pending information for the stream, whereas StreamPendingMessageInfoGet returns the information for each pending message.

image

StackExchange.Redis/StackExchange/Redis/Interfaces/IDatabase.cs
```diff
@@ -1389,6 +1389,318 @@ public interface IDatabase : IRedis, IDatabaseAsync
...
+        /// Adds an entry using the specified values to the given stream key. If key does not exist, a new key holding a stream is created. The command returns the ID of the newly created stream entry.
+        /// </summary>
+        /// <param name="key">The key of the stream.</param>
+        /// <param name="streamFields">The fields and their associated values to set in the stream entry.</param>
+        /// <param name="flags">The flags to use for this operation.</param>
+        /// <param name="maxLength">The maximum length of the stream.</param>
+        /// <param name="useApproximateMaxLength">If true, use the "~" argument to keep the stream length at or slightly above the maxLength.</param>
+        /// <returns>The ID of the newly created message.</returns>
+        /// <remarks>https://redis.io/commands/xadd</remarks>
+        string StreamAdd(RedisKey key, NameValueEntry[] streamFields, int? maxLength = null, bool useApproximateMaxLength = false, CommandFlags flags = CommandFlags.None);
```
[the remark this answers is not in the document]
ttingen: I like streamPairs, will go with that.

StackExchange.Redis/StackExchange/Redis/RedisDatabase.cs
NickCraver: In this case there are 2 issues:
- Ordering (should be method, async equivalent, method, async equivalent) for consistency. 
- I don't think we need 2 sets of methods here - if consumer name is null, we don't return details (just default it to `RedisValue.Null`)
...that gets the same functionality while simplifying the API. We should document this in the usage though.

ttingen:
Just commited many of the smaller items from the change request. I will try to knock out the remainder of them tomorrow.

I still need to do the following:
- Reduce the number of overloads.
- Remove net45 target to allow Array.Empty usage and then make those changes.
- Convert messageId from string to RedisValue.

NickCraver:
Note: I pushed a launch-all-servers bash script in https://github.com/StackExchange/StackExchange.Redis/commit/18c607d463a259271f34dfc3b3d21a334943c742 that I'm using to fire up all servers in WSL for testing. This should be useful for anyone testing this PR. @ttingen If you can yank in master on the next pass, that'll make life easier :)

Appreciate all the work here - looking good.

ttingen:
I believe I completed all of the requested changes (minus the Array.Empty changes). I added a few more tests and I will continue that effort.

I didn't collapse any of the StreamAdd overloads but I would like to remove half of them. If the streamEntryId parameter were a string, we could default it to "*". But I understand the reasoning to want to standardize on using the RedisValue type for those values.

StackExchange.Redis/StackExchange/Redis/Interfaces/IDatabase.cs
```diff
@@ -1389,6 +1389,250 @@ public interface IDatabase : IRedis, IDatabaseAsync
...
+        /// <param name="key">The key of the stream.</param>
+        /// <param name="streamEntryId">The ID to assign to the stream entry. This must be a unique value per entry.</param>
+        /// <param name="streamField">The field name for the stream entry.</param>
+        /// <param name="streamValue">The value to set in the stream entry.</param>
+        /// <param name="flags">The flags to use for this operation.</param>
+        /// <param name="maxLength">The maximum length of the stream.</param>
+        /// <param name="useApproximateMaxLength">If true, the "~" argument is used to allow the stream to exceed max length by a small number. This improves performance when removing messages.</param>
+        /// <returns>The ID of the newly created message.</returns>
+        /// <remarks>https://redis.io/commands/xadd</remarks>
+        RedisValue StreamAdd(RedisKey key, RedisValue streamEntryId, RedisValue streamField, RedisValue streamValue, int? maxLength = null, bool useApproximateMaxLength = false, CommandFlags flags = CommandFlags.None);
```
NickCraver: How about we move `streamEntryId` as the first optional arg (on both sync and async)? That'll eliminate down 4 overloads in a hurry (2 on single fields, 2 on pairs overloads).

StackExchange.Redis/StackExchange/Redis/Interfaces/IDatabase.cs
```diff
@@ -1389,6 +1389,250 @@ public interface IDatabase : IRedis, IDatabaseAsync
...
+        bool StreamCreateConsumerGroup(RedisKey key, RedisValue groupName, RedisValue readFrom, CommandFlags flags = CommandFlags.None);
+
+        /// <summary>
+        /// Retrieve information about the given stream. This is the equivalent of calling "XINFO STREAM key".
+        /// </summary>
+        /// <param name="key">The key of the stream.</param>
+        /// <param name="flags">The flags to use for this operation.</param>
+        /// <returns>A <see cref="StreamInfo"/> instance with information about the stream.</returns>
+        /// <remarks>https://redis.io/topics/streams-intro</remarks>
+        StreamInfo StreamInfoGet(RedisKey key, CommandFlags flags = CommandFlags.None);
```
NickCraver: IMO, this should be just `StreamInfo` (we don't have `Get` as a convention or anything - just in the few places it makes sense).

A few options to consider:
1. `StreamGetInfo`, `StreamGetGroupInfo`, `StreamGetConsumerInfo`
2. `StreamInfo`, `StreamGroupInfo`, `StreamConsumerInfo`
3. `StreamGetInfo`, `StreamGetGroups`, `StreamGetConsumers`

They're all `XINFO` under the covers, and IMO it'd be nice to indicate that in naming if we can - I'm personally a fan of option 2. Thoughts?

StackExchange.Redis/StackExchange/Redis/Interfaces/IDatabase.cs
```diff
@@ -1389,6 +1389,250 @@ public interface IDatabase : IRedis, IDatabaseAsync
...
+        /// <summary>
+        /// View information about pending messages for a stream. A pending message is a message read using StreamReadGroup (XREADGROUP) but not yet acknowledged.
+        /// </summary>
+        /// <param name="key">The key of the stream.</param>
+        /// <param name="groupName">The name of the consumer group</param>
+        /// <param name="flags">The flags to use for this operation.</param>
+        /// <returns>An instance of <see cref="StreamPendingInfo"/>. <see cref="StreamPendingInfo"/> contains the number of pending messages, the highest and lowest ID of the pending messages, and the consumers with their pending message count.</returns>
+        /// <remarks>The equivalent of calling XPENDING key group.</remarks>
+        /// <remarks>https://redis.io/commands/xpending</remarks>
+        StreamPendingInfo StreamPendingInfoGet(RedisKey key, RedisValue groupName, CommandFlags flags = CommandFlags.None);
```
NickCraver: `StreamPending`?

StackExchange.Redis/StackExchange/Redis/Interfaces/IDatabase.cs
```diff
@@ -1389,6 +1389,250 @@ public interface IDatabase : IRedis, IDatabaseAsync
...
+        /// <param name="maxId">The maximum ID to read to within the stream of pending messages. Pass "+" or <see cref="StreamConstants.ReadMaxValue"/> to read to the end of the stream.</param>
+        /// <param name="count">The maximum number of pending messages to return.</param>
+        /// <param name="consumerName">The consumer name for the pending messages. Pass RedisValue.Null to include pending messages for all consumers.</param>
+        /// <param name="flags">The flags to use for this operation.</param>
+        /// <returns>An instance of <see cref="StreamPendingMessageInfo"/> for each pending message.</returns>
+        /// <remarks>Equivalent of calling XPENDING key group start-id end-id count consumer-name.</remarks>
+        /// <remarks>https://redis.io/commands/xpending</remarks>
+        StreamPendingMessageInfo[] StreamPendingMessageInfoGet(RedisKey key, RedisValue groupName, RedisValue minId, RedisValue maxId, int count, RedisValue consumerName, CommandFlags flags = CommandFlags.None);
```
NickCraver: `StreamPendingMessages`?

StackExchange.Redis/StackExchange/Redis/Interfaces/IDatabase.cs
```diff
@@ -1389,6 +1389,250 @@ public interface IDatabase : IRedis, IDatabaseAsync
...
+        /// <summary>
+        /// Read from multiple streams.
+        /// </summary>
+        /// <param name="streamAndIdPairs">The list of streams and the ID from which to begin reading for each stream.</param>
+        /// <param name="countPerStream">The maximum number of messages to return from each stream.</param>
+        /// <param name="flags">The flags to use for this operation.</param>
+        /// <returns>An instance of <see cref="RedisStream"/> for each stream.</returns>
+        /// <remarks>Equivalent of calling XREAD COUNT num STREAMS key1 key2 id1 id2.</remarks>
+        /// <remarks>https://redis.io/commands/xread</remarks>
+        RedisStream[] StreamRead(IList<KeyValuePair<RedisKey, RedisValue>> streamAndIdPairs, int? countPerStream = null, CommandFlags flags = CommandFlags.None);
```
NickCraver: @mgravell thoughts here? We don't have this type of input type (AFAIK) anywhere else in the API, the closest equivalent is `SortedSetEntry[]` for adds, which incurs the array overhead you'd likely want to avoid in streams in general...suggestions?

ttingen: I’ve always thought this one looked clunky. An input type tailored for this method would look a lot cleaner.

Something like this?
StreamRead(StreamIdPair[] streamIdPairs, int? countPerStream, ...)

StackExchange.Redis/StackExchange/Redis/Interfaces/IDatabase.cs
ttingen: +1

StackExchange.Redis/StackExchange/Redis/Interfaces/IDatabase.cs
ttingen: +1

StackExchange.Redis/StackExchange/Redis/Interfaces/IDatabase.cs
ttingen: I like #2 as well.

StackExchange.Redis/StackExchange/Redis/Interfaces/IDatabase.cs
ttingen: Change the type to a string since we can’t default to RedisValue.Null?

magmasystems:
Thanks for this. I did some initial work in my own framework (sitting on top of StackExchange.Redis) to put in support for Streams, and all I had to change in the Redis 1.26 source was to add the Stream enum to RedisType.cs. Everything else was handled by my own framework, and a healthy dose of calling Execute() and ExecuteAsync() and hand-coding the stream parsing. I am looking forward to the official support for streams that you guys are implementing.

ttingen:
@magmasystems, thanks for mentioning RedisType. That needed to be added.

magmasystems:
No probs, Todd. I had to do it because I wrote a custom WPF-based GUI and Test Harness for Redis that needed to query the data type of what a key pointed to in order to have a custom WPF data template.

mgravell:
Presumably the "mutate" commands are "master only". It may be worthwhile doing a quick review and adding those ones to `IsMasterOnly` in `Message.cs` - that will avoid them being sent incorrectly to replica (slave) nodes.

NickCraver:
@mgravell It shouldn't stop this PR because it's really a more global/separate issue, but note #354 is related here - I added some notes/a branch there.

ttingen:
@NickCraver, 4 more overloads removed. The streamEntryId parameter was renamed to messageId for the StreamAdd overloads. 

Also added a StreamIdPair struct and updated one of the StreamRead overloads to use that instead of IList<KeyValuePair<RedisKey, RedisValue>>. Looks a lot cleaner now.

magmasystems:
@ttingen - Just wondering if you have ReSharper and are doing any C# 7 refactoring as you go along, or eliminating any of the ReSharper "yellows"?

StackExchange.Redis/StackExchange/Redis/Interfaces/IDatabase.cs
```diff
@@ -1389,6 +1389,225 @@ public interface IDatabase : IRedis, IDatabaseAsync
...
+        /// </summary>
+        /// <param name="key">The key of the stream.</param>
+        /// <param name="consumerGroup">The consumer group.</param>
+        /// <param name="claimingConsumer">The consumer claiming the given message(s).</param>
+        /// <param name="minIdleTimeInMs">The minimum message idle time to allow the reassignment of the message(s).</param>
+        /// <param name="messageIds">The IDs of the messages to claim for the given consumer.</param>
+        /// <param name="flags">The flags to use for this operation.</param>
+        /// <returns>The message IDs for the messages successfully claimed by the given consumer.</returns>
+        /// <remarks>https://redis.io/topics/streams-intro</remarks>
+        RedisValue[] StreamClaimMessagesReturningIds(RedisKey key, RedisValue consumerGroup, RedisValue claimingConsumer, long minIdleTimeInMs, RedisValue[] messageIds, CommandFlags flags = CommandFlags.None);
```
NickCraver: I was thinking about these, and it finally occurs to me: everything is `Messages`. We need not say it. For example, we don't call it `StreamAddMessage()`, `StreamAcknowledgeMessages()`, etc.

What if we shortened these `StreamClaim()` and `StreamClaimIdsOnly()`? Thoughts?

StackExchange.Redis/StackExchange/Redis/RedisDatabase.cs
```diff
@@ -1579,6 +1579,330 @@ public IEnumerable<SortedSetEntry> SortedSetScan(RedisKey key, RedisValue patter
...
+        }
+
+        public Task<bool> StreamCreateConsumerGroupAsync(RedisKey key, RedisValue groupName, RedisValue readFrom, CommandFlags flags = CommandFlags.None)
+        {
+            var msg = Message.Create(Database,
+                flags,
+                RedisCommand.XGROUP,
+                key,
+                groupName,
+                readFrom);
```
NickCraver: Shouldn't this match non-async?

StackExchange.Redis/StackExchange/Redis/Interfaces/IDatabase.cs
```diff
@@ -1389,6 +1389,225 @@ public interface IDatabase : IRedis, IDatabaseAsync
...
+
+        /// <summary>
+        /// Delete messages in the stream.
+        /// </summary>
+        /// <param name="key">The key of the stream.</param>
+        /// <param name="messageIds">The IDs of the messages to delete.</param>
+        /// <param name="flags">The flags to use for this operation.</param>
+        /// <returns>Returns the number of messages successfully deleted from the stream.</returns>
+        /// <remarks>https://redis.io/topics/streams-intro</remarks>
+        long StreamMessagesDelete(RedisKey key, RedisValue[] messageIds, CommandFlags flags = CommandFlags.None);
```
NickCraver: I think we can call this `StreamDelete`, as the `messageIds` are required and make it clear it's not deleting a stream.

StackExchange.Redis/StackExchange/Redis/Interfaces/IDatabase.cs
ttingen: I like it, making those changes now.

StackExchange.Redis/StackExchange/Redis/RedisDatabase.cs
ttingen: Indeed it should. That's what I get for not testing the Async versions yet...

ttingen:
@NickCraver, thanks and it's been a pleasure.

mgravell:
Heya; there's a *lot* of work here, so once again: thanks hugely, this PR is great. I've been deliberately and consciously avoiding jumping in so that we had a second set of "clean eyes" for a final review.

Firstly: the code is looking in great shape, thanks. It is also without doubt the single biggest PR we've received for SE.Redis, so: (awards sticky gold star)

There's a couple of things I want to quickly ask about before we click that final merge button, if you don't mind (but: this is definitely very close!)

Firstly, `XREAD`; my understanding is that there is a multi-key version of this that allows the consumer to fetch from multiple streams; I couldn't see an API that surfaces this - is that intentional?

---

Secondly, I'm still a little ... "uncomfortable" about the ambiguity of message id; in some places (mostly: add) they are a string, and in other (most) places they are `RedisValue`. I have seen your comment above about wanting to default to `"*"`, but ... I think this is actually an implementation detail that we don't need to surface in the API. I suspect the most appropriate API here is probably:

```c#
RedisValue messageId = default
```

if that compiles, or

```c#
RedisValue? messageId = null
```

if it doesn't; then just check in the *implementation* code whether the `messageId` is nil; if it is, then the *implementation* code can substitute the wildcard.

---

This made me think about `StreamConstants` in general; it looks like almost all of these are internal implementation details, which is good. I'm a little suspicious of `StreamConstants.ReadMinValue` etc for similar reasons to `messageId` above - I wonder if that should *also* just default to a nil `RedisValue` using the same approach used by `messageId` above. Essentially: "if you don't supply this, it'll use the min/max", which is *I think* the intent? If not, please let me know!

With that done, I wonder whether `StreamConstants` can be made `internal` - as far as I can see, there's nothing left that isn't already handled by API options rather than magic values on the API.

Thoughts?

magmasystems:
Apologies for butting in here with something that may be slightly off-topic to the PR.

I have spent a good part of my working life architecting and developing low-latency trading systems and other systems that use messaging. Usually that messaging is done through a JMS or AMQP interface over a bus like Tibco EMS, RabbitMQ, or ActiveMQ.

I think that Marc and Nick's philosophy for StackExchange.Redis is to maintain a fairly thin wrapper over the Redis API, and not to introduce any opinionated higher-level framework over StackExchange.Redis. But I am wondering if there may be an appetite for some higher-level framework that sits over the streams. For example, a RedisStreamProducer that can publish a stream to a "topic" (where the "topic" is the key of a stream) and a RedisStreamConsumer which fires a C# event when data is read from a "topic".

When I first tested out the new Redis streams, I wrote something like that in my own code, and I did not consider altering the StackExchange.Redis code in any way, given the fact that I view StackExchange.Redis as a wrapper. If Marc/Nick thinks that a higher-level framework is interesting, I can try to put something together. Otherwise, I will just keep the JMS stuff in my own code,

mgravell:
@magmasystems  my main concern there is: is there an existing that would allow it to be consumed from multiple places? or would it have to be provider specific. If it is provider specific, I'd rather keep it out of the main lib purely for dependency reasons; there's scope for sibling-libraries, though, that re-expose the "raw" library in the context of what that means for "tool X"

NickCraver:
> I think this is actually an implementation detail that we don't need to surface in the API. I suspect the most appropriate API here is probably

Agreed

> I wonder whether StreamConstants can be made `internal`

Also agree, same with making them null by default and min/max in the implementation.

@magmasystems I think this would be where another library sits atop the APIs here, could be my view there. We could always add more concepts later if doing so was very appealing.

ttingen:
> Firstly, XREAD; my understanding is that there is a multi-key version of this that allows the consumer to fetch from multiple streams; I couldn't see an API that surfaces this...

The StreamRead overload that takes an array of StreamIdPair values is the API that reads from multiple streams. It's the one that previously used an IList<KeyValuePair<RedisKey, RedisValue>> as an input parameter.

`RedisStream[] StreamRead(StreamIdPair[] streamIdPairs, int? countPerStream = null, ...)`

The current version looks a lot cleaner than the original but perhaps the naming of the StreamIdPair struct could be improved. I started with `StreamAndIdPair` but didn't like that name too much. Thoughts?

The StreamReadGroup API does however only expose the ability to read from a single stream at a time. Do you want me to add an overload that allows reading multiple streams there as well?

> ...about the ambiguity of message id...

Totally agree, I will make the change to `RedisValue? messageId = null`.  The `RedisValue messageId = default` wouldn't compile with the current language setting (7.0) for the library.

> This made me think about StreamConstants [being changed to] `internal`...

The only other two "constants" I was concerned about were "$" (new messages) and ">" (undelivered messages), they would be used with the `XGROUP CREATE` and `XREADGROUP` commands. But those values would work as sensible defaults for those APIs as well. 

I will update the documentation to include details on the default values.

@mgravell/@NickCraver, let me know what you think about `StreamIdPair` and the additional `StreamReadGroup` overload. I will start making these changes later this evening.

ttingen:
Additionally, I only created the single-stream-reading version of StreamReadGroup to keep the API simple and straightforward. 

I felt it would minimize confusion to only expose the single stream version. For example, to read multiple streams at once using `XREADGROUP`, each stream would need to have created a Consumer Group with the same name. Not that setting up the consumer groups is difficult, I just like the "don't make me think" mentality and decided to keep it simple.

But I'll certainly add the overload if you want it there.

mgravell:
>  I started with StreamAndIdPair but didn't like that name too much. Thoughts?

Naming is hard; it does what it says, I guess; I think it is fine - thanks for the clarification

> EDIT: I gave this more consideration and I will add the overload. Since this is a more generalized library, I shouldn’t limit the scope to just my use cases or sensibilities.

Fair enough; I wasn't going to ask you to add it, but if it is there : great

Will glance over code again :)

mgravell:
In a bunch of the tests, things there's a lot of changes like:

```
- db.StreamCreateConsumerGroup(key, groupName, StreamConstants.ReadMinValue);
+ db.StreamCreateConsumerGroup(key, groupName, "-");
```

Just so I understand: is this because you're being explicit? i.e. would this be semantically equivalent? or is it different?

```
db.StreamCreateConsumerGroup(key, groupName);
```

(just so I understand)

mgravell:
API looks a lot cleaner, thanks!

ttingen:
I defaulted the readFrom parameter to `"$"` (only read new messages) for  `StreamCreateConsumerGroup` and `">"` (only read undelivered messages) for `StreamReadGroup`. 

Those are the options used in the first mentions of the `XGROUP CREATE` and `XREADGROUP` commands in the [intro to streams](https://redis.io/topics/streams-intro) docs and IIRC those options are only valid for those commands. I believe they are reasonable defaults but they can be overridden by passing "-" (start from the lowest ID in the stream) or by passing an actual ID from the stream. Many of the tests for Consumer Groups needed to read from the beginning of the stream so I was passing "-" when creating the consumer group. 

Another interesting caveat when using `"-"` with `XREADGROUP` ... when used in conjunction with a consumer group that was created to only read new messages, passing `"-"` to `XREADGROUP` will read from the beginning of the stream of messages created AFTER the consumer group was created, not the actual beginning of the stream.

Consumer Groups is definitely one of the more complex features of Redis Streams. I suppose the API could be modeled to expose each of the available options as a different method, at least that would be very explicit. Leaving it as-is certainly enables a lot of flexibility in how each method can be utilized but proper use requires a sound understanding of the underlying commands and their options.

Or maybe, instead of explicit methods, the API takes and options struct that would detail how the command should be constructed. Hmm, I'll put some thought into how this functionality could be made more intuitive in the API.

I updated the method documentation to highlight the default value that would be passed to the command if a value is not supplied for the parameter.

...Sorry for lengthy response...I will work on the remaining overload tomorrow morning.

mgravell:
It is also entirely possible that what you have is perfectly fine and usable. Trying to make an API completely idiot-proof can also mean bending it out of shape; I'm not sure that is necessarily an improvement :) I'd welcome any more thoughts you have on this, but "I've considered it, and I'm happy with it like this" might also be fine and reasonable.

ttingen:
> Trying to make an API completely idiot-proof can also mean bending it out of shape.

Great point. :)

ttingen:
@NickCraver I just [added a commit](https://github.com/ttingen/StackExchange.Redis/commit/7d283c1862fdc3b7692a772633d98c7413762774) to my fork that adds the multi-stream reading overload for StreamReadGroup. I also updated some documentation and parameter names to be "more correct". Check it out and let me know if you want me to do another PR to add those changes.

mgravell:
merged into 2.0 - the only awkward bit  was the temp/fake multi-bulk around `ParseRedisStreamEntries` - I *think* I've converted that code correctly. Thanks again.

(/cc @ttingen  @NickCraver )

ttingen:
@mgravell  @NickCraver I tinkered with passing an options struct to the StreamCreateConsumerGroup method, I think I like it. Now there's no need for the user to know any of the magic strings. What do you think?

`db.StreamCreateConsumerGroupAsync("events", "consumerGroup", GroupCreateOptions.BeginningOfStream);`

image

image

mgravell:
I quite like that, yes. Seems clean. I'm in favour of a new PR to apply
that. I can't speak for Nick, of course.

ttingen:
@NickCraver @mgravell And thanks again for taking so much time to work with me on this PR, I know you two are very busy individuals.

mgravell:
Meh, everyone is busy :) But *because* we've been busy doing 47 other things is why we live so much that someone else found the time to look at this and do the thinking for us :)

mgravell:
blocking commands are incompatible with the connection-sharing approach used by SE.Redis; my instinct would be "no"; see the discussion around "blocking pops" [here](https://stackexchange.github.io/StackExchange.Redis/PipelinesMultiplexers#multiplexing) for the same problem.

PR #874, opened 2018-07-03
API: Add SetPop/SetPopAsync count overloads

### LINKED ISSUE #203: support SPOP count parameter
SPOP key [count] http://redis.io/commands/spop
Redis command SPOP can use with default parameter 'count' .

But currently there is no parameter for count option.
RedisValue SetPop(RedisKey key, CommandFlags flags = CommandFlags.None);

I think need to implement below method. 
RedisValue SetPop(RedisKey key, long count = 0, CommandFlags flags = CommandFlags.None);

Thanks

NickCraver:
This API was missing the newer overload added in Redis 3.2 - resolves #203 for v2.


mgravell:
We've discussed the issue of the implicit zero and overload resolution separately (off github); conclusion: we think it is OK for 2.0 as long as documented.

With that proviso: 👍

NickCraver:
Documented in #868 as a breaking change - merging in!

PR #877, opened 2018-07-05
Discussion required: remove PreserveAsyncOrder

mgravell:
I see no scenarios in which we can continue to support this option *and* not doom users to pain; it is fundamentally problematic, in that it *enforces* a queue, *and* that queue is impossible to correctly pump due to "thread stealing" when completing tasks

Note: incomplete (seems glitched right now)

Note: review request is about the idea, not the code.

This would *need* to be a breaking change release note.


mgravell:
Possible idea re pub/sub to respect order: `Channel<T>`, which lets the consumer deal with it - we just drop stuff in the top

NickCraver:
`Channel<T>` is appealing if we have the default that's basically the implementation here - could make that pluggable to replace, is that the thinking?

Overall thoughts:
- I agree, this needs to go.
- I like the completion pool to handle the functionality, I think this sidesteps the deadlock issues we were trying to resolve with it in the first place.
- On *overall* code - I'd remove most references internally. I think for compatibility we can leave that it parses and that you can (fruitlessly) set it, and remove all other traces beyond that in tests, etc.
- I don't think we need to wait on `Channel<T>`, if the plan is our implementation and then extensible, we can just have that in mind when switching things up internally. I'm 100% okay with removing that without a concrete plan for `Channel<T>` beforehand.

Happy to take a look at what's glitching on this if you're working on other bits, just lemme know I'll take the bits you're not on.

mgravell:
"I don't think we need to wait on `Channel<T>`" - it is RTM: https://www.nuget.org/packages/System.Threading.Channels/

NickCraver:
Sorry what I meant there was: an implementation *based on it*, in case there are any gotchas and it's just generally more work...I'm +1 for removing and doing another pass at Channel<T>-based if we think that's valuable. Since it's completely internal otherwise, we could add it at any time really.

mgravell:
Here's the proposal replacement API for pub/sub:

```
var channel = sub.Subscribe(channelName);
...
while (!channel.IsCompleted)
{
    var next = await channel.ReadAsync();
    YourWhateverHere(next.Channel, next.Value);
}
```

mgravell:
@NickCraver  OK, I had an idea. Let me rubber-duck this against you a sec.

consider: we add two methods onto `ChannelMessageQueue`:

```c#
void OnMessage(Action<RedisChannel, RedisValue> handler)
void OnMessage(Func<RedisChannel, RedisValue, Task> handler)
```

These would only work once - if you try to `OnMessage` a second time it laughs at you. Implementation: *exactly as shown above*:

```c#
while (!queue.IsCompleted)
{
    var next = await queue.ReadAsync();
    try { handler(next.Channel, next.Value); }
    catch (ChannelClosedException) { break; } // an inbuilt
    catch (Exception anythingElse) { ...log?...}
}
```

and for the `async` case:

```c#
    try { await handler(next.Channel, next.Value) ?? Task.CompletedTask; }
```

--- 

Another more ambitious option:

I think that now that the machinery is in place, I could actually use this to *reinstate* `PreserveAsyncOrder`, but with a slightly different meaning:

- it would only apply on pub/sub
- it would only apply *inside each individual subscription*

We could also consider whether it makes sense to *request* this mode on a per-subscription basis (presumably via `CommandFlags`), but that's separate.

Basically, the idea is: if `PreserveAsyncOrder` is enabled (whether globally or per-subscription, implementation detail), we spin up a `ChannelMessageQueue` *without showing the user*, and call `OnMessage` with their handler.

This would, IMO, allow a much easier transition for many users, but I wonder if it propagates bad habits.

---

Thoughts?

mgravell:
Hmmm, actually unsubscribe becomes a PITA in the second case. Maybe just doing the first is already more than sufficient.

NickCraver:
I like it, though still have concerns about very long running user code on the path...it's an existing problem and this is still a net win over today (and I see no way that maintains order + allows user code on the path without indefinite constraints anyway) 👍

mgravell:
`OnMessage` implemented; I kinda like it - it is very cute!

StackExchange.Redis.Tests/Booksleeve/PubSub.cs
```diff
@@ -117,10 +118,197 @@ private void TestMassivePublish(ISubscriber conn, string channel, string caption
...
+                var subChannel = await sub.SubscribeAsync(channel);
+                subChannel.OnMessage((key, val) =>
+                {
+                    int i = int.Parse(Encoding.UTF8.GetString(val));
+                    bool pulse = false;
+                    lock (data)
+                    {
+                        data.Add(i);
+                        if (data.Count == count) pulse = true;
+                        if ((data.Count % 10) == 99) Output.WriteLine(data.Count.ToString());
```
NickCraver: Wouldn't ever be hit - mean `% 100` or `== 9` maybe?

StackExchange.Redis/StackExchange/Redis/ChannelMessageQueue.cs
```diff
@@ -0,0 +1,198 @@
...
+                _parent = null;
+                _channel.Writer.TryComplete(error);
+            }
+        }
+        internal async Task UnsubscribeAsyncImpl(Exception error = null, CommandFlags flags = CommandFlags.None)
+        {
+            var parent = _parent;
+            if (parent != null)
+            {
+                await _parent.UnsubscribeAsync(_redisChannel, HandleMessage, flags);
```
NickCraver: There's a race in these - if someone called it concurrently, `_parent` could be a null ref - could avoid by using `parent.UnsubscribeAsync` here (similar in the other equivalents).

mgravell: was meant to be that, thanks; good eyes

NickCraver:
Gave it one last pass, +1 on merging over - nice work!

PR #878, opened 2018-07-07
Eliminate the need for Stream magic strings & added additional XGROUP…

ttingen:
… subcommands.

- Replaced the need for magic strings with Range, ReadOffset, and GroupReadOffset structs.
- Condensed some of the method signatures due to the new structs.
- Added methods & tests for two of the XGROUP subcommands (DESTROY & DELCONSUMER).


ttingen:
Maybe I went a bit overboard but I took what I did for GroupCreateOptions (mentioned in #860) and updated the methods that read from the streams as well.

The need for the magic strings has been eliminated through the new structs added. No more having to pass "-", "+", ">", or "$". Here's what it looks like:

`db.StreamCreateConsumerGroup(streamKey, groupName, GroupCreateOptions.ReadNew);`
`db.StreamCreateConsumerGroup(streamKey, groupName, GroupCreateOptions.ReadBeginning);`
`db.StreamCreateConsumerGroup(streamKey, groupName, GroupCreateOptions.ReadAfterId(afterId));`


`db.StreamReadGroup(streamKey, groupName, consumerName, GroupReadOffset.All);`
`db.StreamReadGroup(streamKey, groupName, consumerName, GroupReadOffset.New);`
`db.StreamReadGroup(streamKey, groupName, consumerName, GroupReadOffset.AfterId(afterId));`

`db.StreamRead(streamKey, ReadOffset.FromBeginning);`
`db.StreamRead(streamKey, ReadOffset.AfterId(afterId));`

`db.StreamRange(streamKey, Range.All);`
`db.StreamRange(streamKey, Range.AllDescending);`
`db.StreamRange(streamKey, Range.Ascending(minId, maxId));`
`db.StreamRange(streamKey, Range.Descending(maxId, minId));`

Not a big fan of `ReadOffset.FromBeginning` but `ReadOffsetset.All` didn't seem to fit either. Same with `Range.AllDescending`, perhaps only have `Range.All` with a defaulted Order to Ascending?

ttingen:
> I wonder if just `StreamOffset` 
I'm good with that.

> I wonder if this should be just StreamEntry?
Also good with StreamEntry, tighter and a more concise.

> Inconsistent naming...
I felt the divide in these two as well but I saw a slight difference in that when a group is created, the command just sets the point in the stream from which to read. And the other struct (GroupReadOffset) is the one that is used when reading the stream. But like you said, that's probably a distinction without a difference since they both indicate a position in the stream.

I'm good with combining the types (and even combine ReadOffset as well). How would you suggest restricting the use of certain tokens (e.g., "$" & ">") in commands where those values aren't valid? Were you thinking of something other than validating that, for example, ">" is an invalid token for StreamRead?

> Range - I'm not sure about this type.
Yep, this one is where I felt I was overstepping. It's probably fine to leave the StreamRange methods as they were but you can pass the min ("-") and max ("+") magic strings to those command as well, but it's probably not a big deal that users know/use those magic strings.

That being the case, maybe trying to eliminate the magic strings to make it idiot-proof will end up bending it out of shape. What do you think?

mgravell:
For the "-"/"+" - I was thinking more: nulls/defaults. If no min is
specified: infer "-", and if no max is specified: infer "+".

On the position thing, I was thinking something like:

internal enum PositionKind {
    Explicit, New, Beginning
}

(Or whatever the meanings are), and:

internal PositionKind Kind {get;}

the the combined type. The existing constructor specifies Explicit, the two
static members specify the other two, along with a nil value.

The message-building code would check the .Kind *first*, adding the correct
symbol. Only for Explicit would it use the RedisValue that is stored.

Does that work?

ttingen:
> For the "-"/"+" - I was thinking more: nulls/defaults. If no min is
specified: infer "-", and if no max is specified: infer "+".

This is how it was working prior to the Range struct. I agree, it's better the way it was. Will revert those changes.

I like what you're suggesting, will start on those changes tonight or tomorrow.

ttingen:
Since `RedisStreamEntry` will be changed to `StreamEntry`, should I also change `RedisStream` to `Stream`?

I already had that change queued up, just wanted to run that by you before committing.

ttingen:
Here's what I have so far. I moved the resolution of the enum to the value used in the command to the StreamPosition struct itself. The Read overloads build two different types of command messages (one of single-stream, the other for multi-stream) so I didn't want the resolution code duplicated in the message building methods.

Just want to run this by you to make sure I'm on the right path.

The read overload would look like this:
image

And here is what the `StreamPosition` struct looks like:
image

mgravell:
Don't use the name Stream by itself - huge conflict with a key BCL type.

ttingen:
> Don't use the name Stream by itself - huge conflict with a key BCL type.

Will revert that change. I figured that would be the answer. :)

ttingen:
You mentioned `StreamOffset` as a potential name for the `StreamIdPair` replacement, and `StreamPosition` as the struct to represent a position for the read commands.

So there wouldn't be confusion between "Position" and "Offset", I named the struct that represents a stream position just `Position` and the pair struct was created as `StreamPosition`.

Not sure if those names are quite what you want but the API does "feel" a lot better now.

mgravell:
@ttingen  because this has been iterating - can you let me know when you're at a state when you're happy for this to be re-reviewed/merged? that could be "now" for all I know, but I don't want to jump back in prematurely

ttingen:
@mgravell I made a final pass with a couple small tweaks, I think it's good now. I like the API a lot more with the single struct to represent a position within a stream. 

The only questions I had were about the naming, particularly for the new `Position` and `StreamPosition` (which replaces StreamIdPair) structs but also naming in general. I want things to line up with your expectations and style.

mgravell:
Awesome, I'll give this some eyeballs this evening. Thanks again for your
continued input.

StackExchange.Redis/StackExchange/Redis/Position.cs
```diff
@@ -0,0 +1,60 @@
...
+            ExplicitValue = readAfter;
+        }
+
+        private Position(PositionKind kind)
+        {
+            Kind = kind;
+            ExplicitValue = RedisValue.Null;
+        }
+
+        private PositionKind? Kind { get; }
```
mgravell: In what scenario can this actually be null? The only way I can see is `default(Position)` - in that case, we should make the zero-value of `Kind` be whatever we want the default to be

StackExchange.Redis/StackExchange/Redis/Position.cs
```diff
@@ -0,0 +1,60 @@
...
+
+        /// <summary>
+        /// Read from the beginning of a stream.
+        /// </summary>
+        public static Position Beginning = new Position(PositionKind.Beginning);
+
+        internal RedisValue ResolveForCommand(RedisCommand command)
+        {
+            // Handle the default struct value.
+            var actualKind = Kind ?? PositionKind.Beginning;
```
mgravell: see above

StackExchange.Redis/StackExchange/Redis/Position.cs
```diff
@@ -0,0 +1,60 @@
+﻿using System;
+
+namespace StackExchange.Redis
+{
+    /// <summary>
+    /// A position within a stream. Defaults to <see cref="Position.New"/>.
+    /// </summary>
+    public struct Position
```
mgravell: I'm having thoughts about this name, in particular how it relates alongside `StreamPosition`. I've been thinking about how this works in some other APIs, and I wonder if the correct name here is `StreamOrigin`? I'm ... unsure, but I think that's closer and less confusing

NickCraver: FWIW, I'm unsure about `StreamOrigin` (either positively or negatively), but I do think `Position` is going to conflict with enough code outside our library we should avoid a name that generic in the root namespace.

StackExchange.Redis/StackExchange/Redis/Position.cs
ttingen: Yes, it would only be null with `default(Position)`, I suppose I was going out of my way to make it super-explicit with the nullable field. I will update the enum with the appropriate zero value, I had already rearranged the order of the enum values so the implicit zero value would be `PositionKind.Beginning` but I will make it explicit.

StackExchange.Redis/StackExchange/Redis/Position.cs
ttingen: Naming _is_ hard. If you've seen Narcos, [this is relevant/funny](https://i.redd.it/zi8clnpxxr811.jpg).

If you consider the meaning of the value, it's the point or position to begin reading a stream. So I'll just throw some names out. I'm not a huge fan of any of them but they're close. 

ReadPosition
ReadOffset 
ReadStart
InitialReadPosition (too verbose?)
StreamOffset
StreamReadOffset
StreamReadPosition

I'm not sure about including the word `Stream` in the name either, that seems to clutter the name of the pair struct that's passed to the multi-stream reading APIs (e.g., `StreamStreamOffset`, `RedisStreamStreamOffset`), those names would seem off somehow.

Hoping for a variable naming epiphany while I sleep tonight...

ttingen:
@mgravell I updated `PositionKind` with an explicit zero value. I haven't had any different names come to me as of yet, do you have thoughts on any of the options I mentioned in the previous comment?

I'll keep thinking about other potential names.

ttingen:
@mgravell @NickCraver I'm still stuck on the names I gave a few days ago. Any new ideas on your end? I thought it might help to see the names I gave in use so as to get a "feel" for how they flow or fit. I'm still not attached to any of them, but some feel better than others.

`db.StreamRead(streamName, ReadPosition.Beginning);`
`db.StreamRead(streamName, StreamReadPosition.Beginning);`

`db.StreamRead(streamName, ReadOffset.Beginning);`
`db.StreamRead(streamName, StreamReadOffset.Beginning);`

`db.StreamRead(streamName, ReadStart.Beginning);`
`db.StreamRead(streamName, StreamReadStart.Beginning);`

`db.StreamRead(streamName, ReadStartPosition.Beginning);`
`db.StreamRead(streamName, StreamReadStartPosition.Beginning);`

`db.StreamRead(streamName, InitialReadPosition.Beginning);`
`db.StreamRead(streamName, StreamInitialReadPosition.Beginning);`

`db.StreamRead(streamName, StreamOffset.Beginning);`

ttingen:
@NickCraver sounds great and I understand; you want a simple, straightforward, easy to use API.

NickCraver:
Here's current state of API diffs of Streams overall for discussion:
```diff
@@ -1,19 +1,146 @@
 namespace StackExchange.Redis
 {
     public class static ExtensionMethods
     {
+        public static RedisValue[] ToRedisValueArray(this string[] values) { }
     }
     public interface IDatabase : IDatabaseAsync, IRedis, IRedisAsync
     {
+        long StreamAcknowledge(RedisKey key, RedisValue groupName, RedisValue messageId, CommandFlags flags = 0);
+        long StreamAcknowledge(RedisKey key, RedisValue groupName, RedisValue[] messageIds, CommandFlags flags = 0);
+        RedisValue StreamAdd(RedisKey key, RedisValue streamField, RedisValue streamValue, System.Nullable<RedisValue> messageId = null, System.Nullable<int> maxLength = null, bool useApproximateMaxLength = False, CommandFlags flags = 0);
+        RedisValue StreamAdd(RedisKey key, NameValueEntry[] streamPairs, System.Nullable<RedisValue> messageId = null, System.Nullable<int> maxLength = null, bool useApproximateMaxLength = False, CommandFlags flags = 0);
+        StreamEntry[] StreamClaim(RedisKey key, RedisValue consumerGroup, RedisValue claimingConsumer, long minIdleTimeInMs, RedisValue[] messageIds, CommandFlags flags = 0);
+        RedisValue[] StreamClaimIdsOnly(RedisKey key, RedisValue consumerGroup, RedisValue claimingConsumer, long minIdleTimeInMs, RedisValue[] messageIds, CommandFlags flags = 0);
+        bool StreamConsumerGroupSetPosition(RedisKey key, RedisValue groupName, Position position, CommandFlags flags = 0);
+        StreamConsumerInfo[] StreamConsumerInfo(RedisKey key, RedisValue groupName, CommandFlags flags = 0);
+        bool StreamCreateConsumerGroup(RedisKey key, RedisValue groupName, System.Nullable<Position> position = null, CommandFlags flags = 0);
+        long StreamDelete(RedisKey key, RedisValue[] messageIds, CommandFlags flags = 0);
+        long StreamDeleteConsumer(RedisKey key, RedisValue groupName, RedisValue consumerName, CommandFlags flags = 0);
+        bool StreamDeleteConsumerGroup(RedisKey key, RedisValue groupName, CommandFlags flags = 0);
+        StreamGroupInfo[] StreamGroupInfo(RedisKey key, CommandFlags flags = 0);
+        StreamInfo StreamInfo(RedisKey key, CommandFlags flags = 0);
+        long StreamLength(RedisKey key, CommandFlags flags = 0);
+        StreamPendingInfo StreamPending(RedisKey key, RedisValue groupName, CommandFlags flags = 0);
+        StreamPendingMessageInfo[] StreamPendingMessages(RedisKey key, RedisValue groupName, int count, RedisValue consumerName, System.Nullable<RedisValue> minId = null, System.Nullable<RedisValue> maxId = null, CommandFlags flags = 0);
+        StreamEntry[] StreamRange(RedisKey key, System.Nullable<RedisValue> minId = null, System.Nullable<RedisValue> maxId = null, System.Nullable<int> count = null, Order messageOrder = 0, CommandFlags flags = 0);
+        StreamEntry[] StreamRead(RedisKey key, Position position, System.Nullable<int> count = null, CommandFlags flags = 0);
+        RedisStream[] StreamRead(StreamPosition[] streamPositions, System.Nullable<int> countPerStream = null, CommandFlags flags = 0);
+        StreamEntry[] StreamReadGroup(RedisKey key, RedisValue groupName, RedisValue consumerName, System.Nullable<Position> position = null, System.Nullable<int> count = null, CommandFlags flags = 0);
+        RedisStream[] StreamReadGroup(StreamPosition[] streamPositions, RedisValue groupName, RedisValue consumerName, System.Nullable<int> countPerStream = null, CommandFlags flags = 0);
+        long StreamTrim(RedisKey key, int maxLength, bool useApproximateMaxLength = False, CommandFlags flags = 0);
     }
     public interface IDatabaseAsync : IRedisAsync
     {
+        Task<long> StreamAcknowledgeAsync(RedisKey key, RedisValue groupName, RedisValue messageId, CommandFlags flags = 0);
+        Task<long> StreamAcknowledgeAsync(RedisKey key, RedisValue groupName, RedisValue[] messageIds, CommandFlags flags = 0);
+        Task<RedisValue> StreamAddAsync(RedisKey key, RedisValue streamField, RedisValue streamValue, System.Nullable<RedisValue> messageId = null, System.Nullable<int> maxLength = null, bool useApproximateMaxLength = False, CommandFlags flags = 0);
+        Task<RedisValue> StreamAddAsync(RedisKey key, NameValueEntry[] streamPairs, System.Nullable<RedisValue> messageId = null, System.Nullable<int> maxLength = null, bool useApproximateMaxLength = False, CommandFlags flags = 0);
+        Task<StreamEntry[]> StreamClaimAsync(RedisKey key, RedisValue consumerGroup, RedisValue claimingConsumer, long minIdleTimeInMs, RedisValue[] messageIds, CommandFlags flags = 0);
+        Task<RedisValue[]> StreamClaimIdsOnlyAsync(RedisKey key, RedisValue consumerGroup, RedisValue claimingConsumer, long minIdleTimeInMs, RedisValue[] messageIds, CommandFlags flags = 0);
+        Task<bool> StreamConsumerGroupSetPositionAsync(RedisKey key, RedisValue groupName, Position position, CommandFlags flags = 0);
+        Task<StreamConsumerInfo[]> StreamConsumerInfoAsync(RedisKey key, RedisValue groupName, CommandFlags flags = 0);
+        Task<bool> StreamCreateConsumerGroupAsync(RedisKey key, RedisValue groupName, System.Nullable<Position> position = null, CommandFlags flags = 0);
+        Task<long> StreamDeleteAsync(RedisKey key, RedisValue[] messageIds, CommandFlags flags = 0);
+        Task<long> StreamDeleteConsumerAsync(RedisKey key, RedisValue groupName, RedisValue consumerName, CommandFlags flags = 0);
+        Task<bool> StreamDeleteConsumerGroupAsync(RedisKey key, RedisValue groupName, CommandFlags flags = 0);
+        Task<StreamGroupInfo[]> StreamGroupInfoAsync(RedisKey key, CommandFlags flags = 0);
+        Task<StreamInfo> StreamInfoAsync(RedisKey key, CommandFlags flags = 0);
+        Task<long> StreamLengthAsync(RedisKey key, CommandFlags flags = 0);
+        Task<StreamPendingInfo> StreamPendingAsync(RedisKey key, RedisValue groupName, CommandFlags flags = 0);
+        Task<StreamPendingMessageInfo[]> StreamPendingMessagesAsync(RedisKey key, RedisValue groupName, int count, RedisValue consumerName, System.Nullable<RedisValue> minId = null, System.Nullable<RedisValue> maxId = null, CommandFlags flags = 0);
+        Task<StreamEntry[]> StreamRangeAsync(RedisKey key, System.Nullable<RedisValue> minId = null, System.Nullable<RedisValue> maxId = null, System.Nullable<int> count = null, Order messageOrder = 0, CommandFlags flags = 0);
+        Task<StreamEntry[]> StreamReadAsync(RedisKey key, Position position, System.Nullable<int> count = null, CommandFlags flags = 0);
+        Task<RedisStream[]> StreamReadAsync(StreamPosition[] streamPositions, System.Nullable<int> countPerStream = null, CommandFlags flags = 0);
+        Task<StreamEntry[]> StreamReadGroupAsync(RedisKey key, RedisValue groupName, RedisValue consumerName, System.Nullable<Position> position = null, System.Nullable<int> count = null, CommandFlags flags = 0);
+        Task<RedisStream[]> StreamReadGroupAsync(StreamPosition[] streamPositions, RedisValue groupName, RedisValue consumerName, System.Nullable<int> countPerStream = null, CommandFlags flags = 0);
+        Task<long> StreamTrimAsync(RedisKey key, int maxLength, bool useApproximateMaxLength = False, CommandFlags flags = 0);
     }
+    public struct NameValueEntry : System.IEquatable<NameValueEntry>
+    {
+        public NameValueEntry(RedisValue name, RedisValue value) { }
+        public RedisValue Name { get; }
+        public RedisValue Value { get; }
+        public override bool Equals(object obj) { }
+        public bool Equals(NameValueEntry other) { }
+        public override int GetHashCode() { }
+        public static bool ==(NameValueEntry x, NameValueEntry y) { }
+        public static System.Collections.Generic.KeyValuePair<RedisValue, RedisValue> op_Implicit(NameValueEntry value) { }
+        public static NameValueEntry op_Implicit(System.Collections.Generic.KeyValuePair<RedisValue, RedisValue> value) { }
+        public static bool !=(NameValueEntry x, NameValueEntry y) { }
+        public override string ToString() { }
+    }
+    public struct Position
+    {
+        public static Position Beginning;
+        public static Position New;
+        public Position(RedisValue readAfter) { }
+    }
     public struct RedisFeatures
     {
+        public bool Streams { get; }
     }
+    public struct RedisStream
+    {
+        public StreamEntry[] Entries { get; }
+        public RedisKey Key { get; }
+    }
     public enum RedisType
     {
-        Unknown = 6,
+        Stream = 6,
+        Unknown = 7,
     }
+    public struct StreamConsumer
+    {
+        public RedisValue Name { get; }
+        public int PendingMessageCount { get; }
+    }
+    public struct StreamConsumerInfo
+    {
+        public long IdleTimeInMilliseconds { get; }
+        public string Name { get; }
+        public int PendingMessageCount { get; }
+    }
+    public struct StreamEntry
+    {
+        public RedisValue Id { get; }
+        public bool IsNull { get; }
+        public static StreamEntry Null { get; }
+        public NameValueEntry[] Values { get; }
+    }
+    public struct StreamGroupInfo
+    {
+        public int ConsumerCount { get; }
+        public string Name { get; }
+        public int PendingMessageCount { get; }
+    }
+    public struct StreamInfo
+    {
+        public int ConsumerGroupCount { get; }
+        public StreamEntry FirstEntry { get; }
+        public StreamEntry LastEntry { get; }
+        public int Length { get; }
+        public int RadixTreeKeys { get; }
+        public int RadixTreeNodes { get; }
+    }
+    public struct StreamPendingInfo
+    {
+        public StreamConsumer[] Consumers { get; }
+        public RedisValue HighestPendingMessageId { get; }
+        public RedisValue LowestPendingMessageId { get; }
+        public int PendingMessageCount { get; }
+    }
+    public struct StreamPendingMessageInfo
+    {
+        public RedisValue ConsumerName { get; }
+        public int DeliveryCount { get; }
+        public long IdleTimeInMilliseconds { get; }
+        public RedisValue MessageId { get; }
+    }
+    public struct StreamPosition
+    {
+        public StreamPosition(RedisKey key, Position position) { }
+        public RedisKey Key { get; }
+        public Position Position { get; }
+    }
 }
```

NickCraver:
@ttingen We finally got time to meet this morning - going to change up how `Positon` works slightly but after merge. @mgravell is going to pull this in and tweak. Thanks again for all the work here!

PR #884, opened 2018-07-10
v2: Socket Changes

NickCraver:
I think we're ready to merge in - this simplifies and fixes the Appveyor hanging issue as well. It also seems to resolve the cluster errors on appveyor with duplicate endpoints.

I'd say let's merge this into pipelines ASAP and I'll keep poking tests. Note that `BasicOps.TestQuit` is throwing a behavioral difference vs. before (I think this is a v2 issue that predates these changes...just writing it down):
image

The issue is that an exception is now thrown:
```
typeof(Pipelines.Sockets.Unofficial.ConnectionResetException): An existing connection was forcibly closed by the remote host
```
But, this only happens with latency (e.g. Appveyor). Locally, it won't throw at all because we'll issue `QUIT` and `PING` to the pipe before the `QUIT` severs the connection and the next `.FlushAsync()` fails. So the question is: how do we want to handle this? I think we need to wrap:
```c#
var flush = _ioPipe.Output.FlushAsync();
if (!flush.IsCompletedSuccessfully) flush.AsTask().Wait();
```
... in ` try`/`catch` and `RecordConnectionFailed`, but there's still a race there on the pipelining fundamentals. Should we just catch/wrap/re-throw?


NickCraver:
Okay merging this in to simplify, another PR to follow...will move notes to #868

PR #891, opened 2018-07-17
Proposed profiling API refactor 

mgravell:
The existing implementation was... confusing; we're trying to check whether it was necessary or overkill. This PR is an experimental refactor which moves all the hard bits *out* of SE.Redis, essentially just exposing (as a breaking change):

    public void RegisterProfiler(Func<ProfilingSession> profilingSessionProvider) 

and *that's it*; everything else is the calling-code's problem (i.e. if necessary, tracking multiple `ProfilingSession` based on the ambient context, and returning the same one, or `null` if not needed, etc). The code invokes this when needed. None of the "context" / `BeginProfiling` / `EndProfiling` exists any more. To consume commands against a `ProfilingSession`, there is a:

    public ProfiledCommandEnumerable GetCommands()

method which **detaches** the buffered commands from the session and makes them available (any commands that enlist after this will be retrievable via an additional call to `GetCommands`)

Reasons:

- much much simpler
- completely removes a thread-race / safety issue we were seeing where commands get attributes to the wrong session
- can't leak (at least, not within the lib) - since session management is now at the caller (and whatever context makes sense to them)


mgravell:
additional thought: should `GetCommands()` perhaps be called `Flush()` instead? that makes it clearer that it resets the state

PR #893, opened 2018-07-21
Initial IPv6 connection support

NickCraver:
TL;DR: We previously did IP:Port parsing based on splitting at the semicolon...that's pretty hostile for IPv6 and breaks everything. This adopts better parsing from Microsoft.AspNetCore since it doesn't exist in the BCL. The BCL/API conversation is started here: https://github.com/dotnet/corefx/issues/23463

A quick way to test this across all tests is via TestConfig, for example:
```json
{
  "MasterServer": "[::1]",
  "SlaveServer": "[::1]",
  "SecureServer": "[::1]"
}
```

Note: there's still an `Unspecified` == `AddressFamily.InterNetwork` in `SocketManager` which may be affecting DNS endpoints that are v6...need to look at that next. This change allows actually connecting to an IPv6 endpoint via IP though, that's step 1.

cc @mgravell as a heads up this is in-progress

### DIFF
--- a/StackExchange.Redis.Tests/Performance.cs
+++ b/StackExchange.Redis.Tests/Performance.cs
@@ -6,6 +6,7 @@ using Xunit.Abstractions;
 
 namespace StackExchange.Redis.Tests
 {
+    [Collection(NonParallelCollection.Name)]
     public class Performance : TestBase
     {
         public Performance(ITestOutputHelper output) : base(output) { }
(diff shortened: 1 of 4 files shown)


NickCraver:
@mgravell alrighty this is ready for eyes/merging. It should handle connecting to IPv6 endpoints whether directly or via DNS across the board.

StackExchange.Redis.Tests/Performance.cs
```diff
@@ -6,6 +6,7 @@
 
 namespace StackExchange.Redis.Tests
 {
+    [Collection(NonParallelCollection.Name)]
```
NickCraver: This isn't related to the PR, just fixing the flakiness of this test in general.

PR #899, opened 2018-07-27
add a command sentinels 

liuqiyun520:
1.Node about sentinels 
Sometimes wo need get a redisMaster‘s seninels info,so i add a command sentinels
2.Node about scan
SCAN is a cursor based iterator.Sometime because some reasons we don't need all keys,we just need 
one pageSize keys,and the value of the cursor argument in the next call. So I add a method .
RedisKey[] Keys(ref long refCursor, int database = 0, RedisValue pattern = default(RedisValue), int pageSize = RedisBase.CursorUtils.DefaultPageSize, long cursor = RedisBase.CursorUtils.Origin, int pageOffset = 0, CommandFlags flags = CommandFlags.None);
refCursor is the cursor argument in the next call.

### DIFF
--- a/StackExchange.Redis/StackExchange/Redis/Interfaces/IServer.cs
+++ b/StackExchange.Redis/StackExchange/Redis/Interfaces/IServer.cs
@@ -646,6 +646,25 @@ namespace StackExchange.Redis
         /// <remarks>https://redis.io/topics/sentinel</remarks>
         Task SentinelFailoverAsync(string serviceName, CommandFlags flags = CommandFlags.None);
 
+        /// <summary>  
+        /// show a list of sentinels info about a redis master  
+        /// </summary>  
+        /// <param name="serviceName">The sentinel service name.</param>  
+        /// <param name="flags">The command flags to use.</param>  
+        /// <returns>an array of sentinel info KeyValuePair arrays</returns>  
+        /// <remarks>https://redis.io/topics/sentinel</remarks>  
+        KeyValuePair<string, string>[][] SentinelSentinels(string serviceName, CommandFlags flags = CommandFlags.None);  
+   
+        /// <summary>  
+        /// show a list of sentinels info about a redis master  
+        /// </summary>  
+        /// <param name="serviceName">The sentinel service name.</param>  
+        /// <param name="flags">The command flags to use.</param>  
+        /// <returns>an array of sentinel info KeyValuePair arrays</returns>  
+        /// <remarks>https://redis.io/topics/sentinel</remarks>  
+        Task<KeyValuePair<string, string>[][]> SentinelSentinelsAsync(string serviceName, CommandFlags flags = CommandFlags.None);
+
+
         #endregion
     }
 }
--- a/StackExchange.Redis/StackExchange/Redis/RedisLiterals.cs
+++ b/StackExchange.Redis/StackExchange/Redis/RedisLiterals.cs
@@ -1,4 +1,4 @@
-using System;
+﻿using System;
 using System.Text;
 
 namespace StackExchange.Redis
@@ -66,6 +66,7 @@ namespace StackExchange.Redis
             GETMASTERADDRBYNAME = "GET-MASTER-ADDR-BY-NAME",
 //            RESET = "RESET",
             FAILOVER = "FAILOVER",
+            Sentinels = "Sentinels",
 
             // Sentinel Literals as of 2.8.4
             MONITOR = "MONITOR",
--- a/StackExchange.Redis/StackExchange/Redis/RedisServer.cs
+++ b/StackExchange.Redis/StackExchange/Redis/RedisServer.cs
@@ -812,6 +812,18 @@ namespace StackExchange.Redis
             return ExecuteAsync(msg, ResultProcessor.SentinelArrayOfArrays);
         }
 
+        public KeyValuePair<string, string>[][] SentinelSentinels(string serviceName, CommandFlags flags = CommandFlags.None)
+        {  
+            var msg = Message.Create(-1, flags, RedisCommand.SENTINEL, RedisLiterals.Sentinels, (RedisValue)serviceName);  
+            return ExecuteSync(msg, ResultProcessor.SentinelArrayOfArrays);  
+        }  
+            
+        public Task<KeyValuePair<string, string>[][]> SentinelSentinelsAsync(string serviceName, CommandFlags flags = CommandFlags.None)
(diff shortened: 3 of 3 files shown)


mgravell:
I appreciate the input, but I'm going to reject 2; that's not a good way to implement that, and it doesn't actually give you what you want. We already have a fully implemented version of `Keys` (based on `SCAN` when possible):

```c#
IEnumerable<RedisKey> Keys(int database = 0, RedisValue pattern = default(RedisValue), int pageSize = RedisBase.CursorUtils.DefaultPageSize, long cursor = RedisBase.CursorUtils.Origin, int pageOffset = 0, CommandFlags flags = CommandFlags.None);
```

This also full support for resuming cursors etc, although you need to know to cast the enumerable to `IScanningCursor`. It sounds like what you're actually looking for is:

```C#
var keys = server.Keys(pattern: pattern).Take(count).ToArray();
```

In particular, the `count` (`pageSize`) parameter to redis *is only a hint* - it doesn't actually say anything about how much data will come back per page *even if there are no pattern-based misses. It can return *anything it wants*, so making it sound too much like "the amount you'll get" is just setting people up for confusion.

But ultimately, my main issue here is that it doesn't add anything that isn't already available through the existing `Keys` API. As a minor issue: the proposed API isn't `async`-compatible; that's something that could be fixed, but: I really don't think there is a need to add this method.

If you *really, really* want to issue a "first page from `SCAN`" query, you should be able to use `Execute`:

```c#
var arr = (RedisResult[])whatever.Execute("SCAN", 0, "MATCH", pattern, "COUNT", 20);
var nextCursor = (long)arr[0];
var keys = (RedisKey[])arr[1];
```

But again; my primary recommendation is to use the existing `Keys(...)` API, presumably with `.Take(lim)`

mgravell:
I think I'm OK with the `SentinelSentinels[Async]` part - that seems to expose new functionality in a reasonable way. I'm not familiar with what the shape of the `sentinel sentinels` result is, but the documentation seems to strongly suggest it is directly comparable to `SENTINEL slaves`.

@NickCraver  - you know more about sentinel than me - thoughts? but: that bit seems fine and reasonable to me.

liuqiyun520:
Thanks for you replay. 
Keys(...) API, presumably with .Take(lim).This way will scan all keys in the server.I don't want this.But I think it's a good idea by using 
`var arr = (RedisResult[])whatever.Execute("SCAN", 0, "MATCH", pattern, "COUNT", 20);`
Smilarly, I have resolve this problem by using luascript.
```
string Script = "return redis.call('scan', @cursorStart,'match',@match,'count',@count)";
var prepared = LuaScript.Prepare(Script);
var loaded = prepared.Load(server);
var result = loaded.Evaluate(dataBase, new { cursorStart = cursorStartFrom, match = pattern, count = count });
var realResult = (RedisResult[])result;
start = cursorEndTo = (long)realResult[0];
return ((RedisResult[])realResult[1]).Select(i => (string)i).ToArray(); 
```

liuqiyun520:
Abount sentinels,
 we have the command about sentinel like Masters,master,slaves in the file RedisLiterals.cs
 These commands are link to 
 ```
SENTINEL masters 
 SENTINEL master mastername
 SENTINEL slaves mastername

``` The commands is far from complete, so I add a command sentinels which link to
 SENTINEL sentinels mastername(list all sentinels about a redisMaster named mastername)
 Like method name SentinelMaster,SentinelMasters,SentinelSlaves I named method as SentinelSentinels.
 it's a fact that my English is really, really poor.Actually, to improve my English level, I recite words every day. I hope you can understand all what I say.

mgravell:
> Keys(...) API, presumably with .Take(lim).This way will scan all keys in the server.I don't want this.But I think it's a good idea by using

No, it won't. Context: I wrote it. The custom enumerable it yields is a lazy paged source - it will only perform `SCAN` queries when it runs out of data on the current page while being enumerated.

mgravell:
On the sentinels - I understand that; I'm comfortable with adding this - my slight concern was simply that I don't know the shape of the data that comes back; unusually, it isn't stated clearly in the documentation.

I'm very comfortable with the addition of `SentinelSentinels[Async]` - that's great. Perhaps what is missing is a test for it. Could you please add a test that exercises this method (or methods, but either would probably be fine) in to `StackExchange.Redis.Tests\Sentinel.cs` ? There is code to spin up test servers including sentinels in `RedisConfigs`; if you're on windows, `start-sentinel-cmd` or `start-all.cmd`; if you're on linux or WSL, `start-all.sh`. If we can see it being tested (including seeing what we expect the results to look like), I'm happy to merge the  `SentinelSentinels[Async]` piece.

mgravell:
Can I also provide some side feedback; this is **not** intended as a "dig" (an attack), etc; your contribution here is very much appreciated, and I thank you for your input - I very much look forward to merging the sentinel work, in particular.

This is simply meant very genuinely as positive advice for future PRs, to keep things moving smoothly: if you're making two unrelated sets of changes, it is usually preferable to make them as two separate PRs. This avoids the awkward situation here where we need to rework the PR to selectively take pieces of it. It also means that the two PRs can progress independently (for example, if one can merge "now" but the other needs more work).

Again; simply meant as a tip for future PRs - and I'm happy to help get this one into the right shape so we can merge it.

liuqiyun520:
OK, in the future,I will  give a PRs for every independent change. This time is my first time to make a PRs.
I will never make a mistake like this.Thank you.

mgravell:
No problem! That's just part of the learning curve. My own first PR attempts were unmitigated disasters, so you're already doing better than me!

NickCraver:
@liuqiyun520 Can you please remove the `Keys` overload as discussed above so we can pull in the `SentinelSentinels` bits ASAP? :)

NickCraver:
@liuqiyun520 Any news? I'd love to merge the sentinel bits merged in if you're willing.

liuqiyun520:
@NickCraver I am happy you will merge it. You please!

liuqiyun520:
@NickCraver I have not remove the keys. Now I will do it. Then I will pull request again. Please Have a wait.

liuqiyun520:
@NickCraver I have remove the keys. Now you can merge.

liuqiyun520:
Leave a comment

NickCraver:
@liuqiyun520 Thanks! I'll pull this in now and add a few tests since that's quick and straightforward to test.

mgravell:
merged into pipelies / "v2"; thanks

PR #902, opened 2018-07-31
fix disconnected handling when redis server down

kenji-andoh:
FIX. All Operation will timeout on SSL connection
When  NetworkStream.BeginRead () throws an IOException, the ConnectionMultiplexer.IsConnected will keep true.


mgravell:
it is very unclear what this fixes - how it used to manifest, vs how it manifests now, in what exact scenario; can you clarify a little? it *looks* like it might be fine, but... trying to think how to verify it. Also, it is worth noting that this code literally doesn't exist in 2.0, but it might be a very shortlived fix :/

kenji-andoh:
The `PhysicalConnection.BeginReading ()` method is called by the `PhysicalConnection.endRead` delegate and the
`SocketManager.EndConnectImpl ()` method (via the `PhysicalConnection.StartReading()`).

`PhysicalConnection.RecordConnectionFailed()` is being handled by catching an exception in these functions.

If the IOException is not thrown again, the caller's disconnection detection is hindered.
And `endRead` is never called again.

kenji-andoh:
I use Windows Azure Redis Cache.  ( `ConfigurationOptions.AbortOnConnectFail = false ) 
In this case, all operation of IDatabase occurs timeout. And never Reconnection.

mgravell:
ok; I'm pretty comfortable with [the original change](https://github.com/StackExchange/StackExchange.Redis/pull/902/commits/a985ac885204535d91bc72092ed019ceb593d444), but I'm very confused by the [merge commit](https://github.com/StackExchange/StackExchange.Redis/pull/902/commits/bd0f54b0de98a8fab7dbc25d78ba26427b32075a) - looking at the [combined effects](https://github.com/StackExchange/StackExchange.Redis/pull/902/files), it *looks to me* like it might be a bad merge, perhaps reapplying old code? or maybe it is just showing incorrectly in github?

I pulled the code locally - it reflects the same combined change as shown above, and it does not compile. Please can you try to fix? If it was me, in my local branch I'd revert that second commit completely, and then rebase. I can't merge it "as is", as it is a bad merge.

kenji-andoh:
Oh, I have failed to merge.
I colse this commit and will submit change from master head.

PR #907, opened 2018-08-02
fix `PhysicalConnection.BeginReading()` exception handling

kenji-andoh:
FIX. All Operation will timeout on SSL connection.
When NetworkStream.BeginRead () throws an IOException, the ConnectionMultiplexer.IsConnected will keep true.

The PhysicalConnection.BeginReading () method is called by the PhysicalConnection.endRead delegate and the SocketManager.EndConnectImpl () method (via the PhysicalConnection.StartReading()).

PhysicalConnection.RecordConnectionFailed() is being handled by catching an exception in these functions.

If the IOException is not thrown again, the caller's disconnection detection is hindered.
And endRead is never called again.


NickCraver:
Bringing in for 1.x, thanks!

Issue #909, opened 2018-08-06
Support UNLINK command


craigfo:
https://redis.io/commands/unlink

NickCraver:
This is added and in 2.0, it'll be in a stable release soon but is already in an alpha on NuGet :)

PR #910, opened 2018-08-06
Add UNLINK as .KeyDelete() automatic usage underneath

NickCraver:
Submitting as a PR so test are reusable, but maybe we want this as an overload via some enum on `.KeyDelete()` instead? (breaking change) I'm good either way, but getting the ball rolling on #909 here.

@mgravell thoughts on separate vs. changing `.KeyDelete()`? That binary change in v2 would break pretty much *everyone*, so leaned this direction. Then again, strong name means we're breaking everyone so...enum? I could go either way.

cc @craigfo


mgravell:
I wonder whether we should make it automatic on `KeyDelete` if the server version is high enough and the command isn't explicitly disallowed. General consensus seems to be "prefer unlink". https://stackoverflow.com/q/45818371/23354 . I *did* think we had precedent for one where there was a global mode toggle, but I can't find that now...

NickCraver:
Thinking of the KEYS replacement?

mgravell:
No, I thought there was another

NickCraver:
It looks like KeyExpiry switches between `TTL` (seconds) and `PTTL` (milliseconds) depending on if it's available (`MillisecondExpiry`, version >= 2.6.0), might be thinking of that?

NickCraver:
Bah submitted early, here's the usage on that:
```c#
var features = GetFeatures(Database, key, flags, out server);
processor = StringGetWithExpiryProcessor.Default;
if (server != null && features.MillisecondExpiry && multiplexer.CommandMap.IsAvailable(RedisCommand.PTTL))
{
    return new StringGetWithExpiryMessage(Database, flags, RedisCommand.PTTL, key);
}
// if we use TTL, it doesn't matter which server
server = null;
return new StringGetWithExpiryMessage(Database, flags, RedisCommand.TTL, key);
```

NickCraver:
@mgravell made the changes discussed, can you give eyes when available? Should be good to merge unless changes wanted here, lemme know!

PR #912, opened 2018-08-09
Streams: simplify Position down into RedisValue

NickCraver:
Not 100% sure this is what we figured out, but good first pass at getting there if not.

cc @mgravell @ttingen for eyes


mgravell:
LGTM

ttingen:
Looks great.

Issue #913, opened 2018-08-11
Protocol Error: Unexpected response (Tracking issue)


NickCraver:
This is an issue Marc and I are tracking down in v2 transactions that's really, really hard to repro. Listing builds here so we can find them quickly.

Example error:
> RedisConnectionException : Unexpected response to HLEN: SimpleString: PONG

Links to known failure builds:
- https://ci.appveyor.com/project/StackExchange/stackexchange-redis/build/1.0.1421/tests
- https://ci.appveyor.com/project/StackExchange/stackexchange-redis/build/2.0.0-alpha.346+gb6aa4348be/tests

NickCraver:
Fixed! This was due to a bug in the pipelines code and a very rare race basically discarding buggers and not handling the synchronous follow-up action correctly. Fixed in the .Unofficial v88 release.

PR #914, opened 2018-08-12
Build: add Ubuntu

NickCraver:
This won't pass yet (some tests are failing on Ubuntu) and it ~1.5x double build times. I'll inquire about paying for concurrent builds Monday to resolve the latter.

But! Those tests are going against the pre-installed Redis 4.0.8 official build instead of redis-64. \o/

What's interesting is right out of the gate it reproduced #913 (protocol error): https://ci.appveyor.com/project/StackExchange/stackexchange-redis/build/2.0.0-alpha.357+g67ded86dd3/tests


NickCraver:
Merging into pipelines build

PR #917, opened 2018-08-16
Exceptions: add timeout info everywhere

NickCraver:
Note: this doesn't include command and key, which is still behind the IncludeDetailInExceptions flag


mgravell:
LGTM - does it work?

PR #926, opened 2018-08-26
Cancel Task.Delays

benaadams:
Cancel `Task.Delay` timers in `Task.WhenAll` when pair task completes prior to timeout; so the timer doesn't hang around until the `Task.Delay` itself times out.

Pattern from [corefx/.../Common/tests/System/Threading/Tasks/TaskTimeoutExtensions.cs](https://github.com/dotnet/corefx/blob/851a53bee21639cd65e109b72fd5f9ef90cc0f39/src/Common/tests/System/Threading/Tasks/TaskTimeoutExtensions.cs)

### DIFF
--- a/StackExchange.Redis/StackExchange/Redis/ConnectionMultiplexer.cs
+++ b/StackExchange.Redis/StackExchange/Redis/ConnectionMultiplexer.cs
@@ -623,8 +623,7 @@ namespace StackExchange.Redis
                 }
 
                 var allTasks = Task.WhenAll(tasks).ObserveErrors();
-                var any = Task.WhenAny(allTasks, Task.Delay(remaining)).ObserveErrors();
-                bool all = await any.ForAwait() == allTasks;
+                bool all = await allTasks.TryComplete(timeoutMs: remaining).ObserveErrors().ForAwait();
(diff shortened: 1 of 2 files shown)


NickCraver:
We really need to get merged into master (we're just at that point now) - current code for v2.0 is the pipelines branch (which still may need some love here, refactoring was still in full swing last time I looked) - see https://github.com/StackExchange/StackExchange.Redis/blob/pipelines/src/StackExchange.Redis/TaskExtensions.cs#L34

Any chance you want to peek at `pipelines`? ;) We're close enough to 2.0 that I think we decided there won't be another 1.2.x release (as we wouldn't likely support it), and little has changed aside from Streams additions. `master` not being "current" if causing too much confusion...I'll talk to @mgravell and see if we're cool with merging into master tomorrow.

NickCraver:
Just checked - a different form, but this would be a cleaner fix for the first and a 1:1 fix for the second in pipelines. If you have time to re-point PR I'd appreciate it, if not I can apply changes manually - lemme know!

benaadams:
Looks like you have code, but not using it :)

Will leave this open for reference

NickCraver:
Tidied up to the extensions in 53427697e8745dd8de1200c7f11a4d66f2260033, thanks!

PR #927, opened 2018-08-26
Clear ResultBox state prior to caching

benaadams:
Else it keeps a bunch of stuff around (e.g. the AsyncLocal+Execution context for TaskCompletionSources, which can end up being a big tree fanning out from HttpContext)

As seen in https://github.com/aspnet/KestrelHttpServer/issues/2840#issuecomment-416017410 (comment now deleted)


benaadams:
Will push to pipelines branch

benaadams:
Superseded by https://github.com/StackExchange/StackExchange.Redis/pull/928

PR #928, opened 2018-08-26
Clear ResultBox state prior to caching

benaadams:
Else it keeps a bunch of stuff around (e.g. the AsyncLocal+Execution context for TaskCompletionSources, which can end up being a big tree fanning out from HttpContext)


NickCraver:
Good catch! Merging in

PR #932, opened 2018-08-29
remove indirect vector usage (on < net472, at least); it breaks the world

mgravell:
giving up on making "vectors" happy; instead, just give up and use black-box knowledge of which code paths touch vectors, and just manually remove those path dependencies via manual fallbacks

### DIFF
--- a/src/StackExchange.Redis/BufferReader.cs
+++ b/src/StackExchange.Redis/BufferReader.cs
@@ -155,7 +155,7 @@ namespace StackExchange.Redis
                 if (reader.RemainingThisSpan == 0) continue;
 
                 var span = reader.SlicedSpan;
-                int found = span.IndexOf(value);
+                int found = span.VectorSafeIndexOf(value);
                 if (found >= 0) return totalSkipped + found;
 
                 totalSkipped += span.Length;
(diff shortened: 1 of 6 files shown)


footcha:
Last two days I spent debugging our global application (.NET471) just to find in the end that assembly redirect for `System.Numerics.Vectors` was missing in `Web.config`:
```xml
<dependentAssembly>
    <assemblyIdentity name="System.Numerics.Vectors" publicKeyToken="b03f5f7f11d50a3a" />
    <bindingRedirect oldVersion="0.0.0.0-4.1.4.0" newVersion="4.1.4.0"/>
</dependentAssembly>
```
 We use `StackExchange.Redis(2.0.601)` and our application requires `System.Numerics.Vectors` Nuget package v4.5.0 (assembly version 4.1.4.0).
See the transitive dependency `StackExchangeRedis(v2.0.601)->System.IO.Pipelines(4.5.1)->System.Memory(4.5.1)->System.Numerics.Vectors(4.4.0)`

Is there a way to include some message details to `RedisConnectionException`? Debugging with FusionLog is not something that comes to my mind always in case of problems.

mgravell:
If I recall: we've tried and it isn't reliable. I've also observed that the
currently-preview MS packages behave a lot better, so there's a good chance
that when they get released and we update: a lot of this pain goes away.

footcha:
OK, thank you very much for a quick reply. I believe the future library upgrade will help.
I keep fingers crossed!

Issue #943, opened 2018-09-06
Commited transaction with task not completed in Redis 2.x


BenedicteLeo:
I have updated a project from Redis 1.2.6 to Redis 2.0.495 under target framework net462. Since the update, some tasks that are part of a transaction are not completed after executing the transaction. I suspect it is a 2.x bug since it was working consistently before, and I haven't seen a related release note (do tell if I missed it).
Here is a simplified code sample:

```ITransaction transaction = this.Database.CreateTransaction();
ConditionResult keyNotExists = transaction.AddCondition(Condition.KeyNotExists(someKey));
Task hashSetTask = transaction.HashSetAsync(someKey, someValues);
Task<bool> expireTask = transaction.KeyExpireAsync(someKey, ttl);

bool committed = await transaction.ExecuteAsync();
Debug.Assert(hashSetTask.IsCompleted, "This fails sometimes");
Debug.Assert(expireTask.IsCompleted, "This also fails sometimes");
```
Both Debug.Asserts fail very frequently (as I run tests), but they don't fail all of the time. In addition, the asserts also fail when I put them inside an "if(commited)" conditional. Thank you.
[Edit] I forgot to mention I'm using  Redis 3.2.100 (00000000/0) 64 bit

NickCraver:
Are you deleting (or setting) `someKey` before the test? Or is it inconsistent? e.g. does the transaction fail or succeed?

BenedicteLeo:
The transaction succeeds. The key is generated at random.

Here is the script I just ran in LinqPad, and the issue repros.
`ConnectionMultiplexer conn = ConnectionMultiplexer.Connect("localhost,abortConnect=false");
var db = conn.GetDatabase();

for (int i = 0; i < 40000; i++)
{
	string key = "key~" + Guid.NewGuid();
	HashEntry[] hashEntries = new HashEntry[]
	{
	new HashEntry("blah", DateTime.UtcNow.ToString("R"))
	};

	ITransaction transaction = db.CreateTransaction();
	ConditionResult keyNotExists = transaction.AddCondition(Condition.KeyNotExists(key));
	Task hashSetTask = transaction.HashSetAsync(key, hashEntries);
	Task<bool> expireTask = transaction.KeyExpireAsync(key, TimeSpan.FromSeconds(30));
	bool committed = await transaction.ExecuteAsync();
	
	if (committed)
	{
		if (!hashSetTask.IsCompleted) "Oh no!".Dump();
[...]

PR #950, opened 2018-09-12
make transaction inner tasks use RunContinuationsAsynchronously

mgravell:
For discussion, see issue #943 

- make transaction inner tasks use `RunContinuationsAsynchronously`
- detect `RunContinuationsAsynchronously` in `ResultBox` completion, and set directly rather than deferred

### DIFF
--- a/src/StackExchange.Redis/TaskSource.cs
+++ b/src/StackExchange.Redis/TaskSource.cs
@@ -9,7 +9,8 @@ namespace StackExchange.Redis
         /// </summary>
         /// <typeparam name="T">The type for the created <see cref="TaskCompletionSource{TResult}"/>.</typeparam>
         /// <param name="asyncState">The state for the created <see cref="TaskCompletionSource{TResult}"/>.</param>
-        public static TaskCompletionSource<T> Create<T>(object asyncState)
-            => new TaskCompletionSource<T>(asyncState, TaskCreationOptions.None);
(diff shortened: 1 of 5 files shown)


src/StackExchange.Redis/TaskSource.cs
```diff
@@ -9,7 +9,8 @@ internal static class TaskSource
         /// </summary>
         /// <typeparam name="T">The type for the created <see cref="TaskCompletionSource{TResult}"/>.</typeparam>
         /// <param name="asyncState">The state for the created <see cref="TaskCompletionSource{TResult}"/>.</param>
-        public static TaskCompletionSource<T> Create<T>(object asyncState)
-            => new TaskCompletionSource<T>(asyncState, TaskCreationOptions.None);
+        /// <param name="options">The options to apply to the task</param>
+        public static TaskCompletionSource<T> Create<T>(object asyncState, TaskCreationOptions options = TaskCreationOptions.None)
+            => new TaskCompletionSource<T>(asyncState, options);
```
NickCraver: FWIW, we could delete this helper now that it’s a wrap with no logic over built-in.

mgravell: @NickCraver  I'd rather keep it all in one place even if it is just inlined by the JIT; I might add a JIT hint, but frankly I'd expect it to inline like a beast

PR #980, opened 2018-10-18
Improve error messaging in unreachable-after-moved situation

hpk:
`TryResend` sets `CommandFlags.NoRedirect`, so the check on line 211 was always `true`, even if the original message didn't have `NoRedirect` specified. This was reporting some really confusing error messages until I tracked it down to here.

Also improved the MOVED error message to be more clear about which endpoint was involved.


NicolaAtorino:
image

PR #1000, opened 2018-11-12
WIP: #969 testable ChannelMessageQueue

### LINKED ISSUE #969: ISubscriber no longer mockable
With the introduction of `ChannelMessageQueue Subscribe(...)` and the async version, `ISubscriber` is no longer mockable (because of `sealed class ChannelMessageQueue`) so it's difficult/impossible to unit test code that uses Redis.

anurse:
Fixes #969 

This PR sketches out what it would look like to make `ChannelMessageQueue` constructable by a test, so that mock implementations of `ISubscriber` could use it. 

It could also be useful outside of a test, since `ISubscriber` is a public interface and (in 2.0) has methods that an external assembly *can't implement* because it is unable to construct `ChannelMessageQueue`

I believe this to be non-breaking, but there are some quirks about it. Comments below.

cc @BrennanConroy @davidfowl

### DIFF
--- a/src/StackExchange.Redis/ChannelMessageQueue.cs
+++ b/src/StackExchange.Redis/ChannelMessageQueue.cs
@@ -15,7 +15,7 @@ namespace StackExchange.Redis
         /// <summary>
         /// See Object.ToString
         /// </summary>
-        public override string ToString() => ((string)Channel) + ":" + ((string)Message);
+        public override string ToString() => Channel + ":" + Message;
 
         /// <summary>
         /// See Object.GetHashCode
@@ -28,7 +28,14 @@ namespace StackExchange.Redis
         /// <param name="obj">The <see cref="object"/> to compare.</param>
         public override bool Equals(object obj) => obj is ChannelMessage cm
             && cm.Channel == Channel && cm.Message == Message;
-        internal ChannelMessage(ChannelMessageQueue queue, RedisChannel channel, RedisValue value)
+
+        /// <summary>
+        /// Create a new <see cref="ChannelMessage"/> representing a message written to a <see cref="ChannelMessageQueue"/>
+        /// </summary>
+        /// <param name="queue">The <see cref="ChannelMessageQueue"/> associated with this message.</param>
+        /// <param name="channel">A <see cref="RedisChannel"/> identifying the channel from which the message was received.</param>
+        /// <param name="value">A <see cref="RedisValue"/> representing the value of the message.</param>
+        public ChannelMessage(ChannelMessageQueue queue, RedisChannel channel, RedisValue value)
         {
             _queue = queue;
             Channel = channel;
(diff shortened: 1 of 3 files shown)


src/StackExchange.Redis/ChannelMessageQueue.cs
```diff
@@ -56,64 +63,59 @@ internal ChannelMessage(ChannelMessageQueue queue, RedisChannel channel, RedisVa
...
-        internal ChannelMessageQueue(RedisChannel redisChannel, RedisSubscriber parent)
+        /// <summary>
+        /// Constructs a <see cref="ChannelMessageQueue" /> from a <see cref="System.Threading.Channels.ChannelReader{RedisValue}"/> representing
+        /// incoming Redis values on the channel.
+        /// </summary>
+        /// <param name="channel">The name of the channel this subscription is listening on.</param>
+        /// <param name="incomingValues">A channel reader representing the incoming values.</param>
+        /// <param name="onUnsubscribe">A delegate to call when <see cref="Unsubscribe(CommandFlags)"/> is called.</param>
+        /// <param name="onUnsubscribeAsync">A delegate to call when <see cref="UnsubscribeAsync(CommandFlags)"/> is called.</param>
+        /// <param name="onInternalError">REVIEW: Need more context here</param>
```
anurse: I'm not super clear on the purpose of this `onInternalError` callback (represents the previous call directly to `ConnectionMultiplexer.OnInternalError`). It was only called when a non-`ChannelClosedException` was thrown by `ReadAsync` in the message loop (which is highly unlikely).

src/StackExchange.Redis/ChannelMessageQueue.cs
```diff
@@ -162,17 +175,29 @@ private async Task OnMessageSyncImpl()
             while (!Completion.IsCompleted)
             {
                 ChannelMessage next;
-                try { if (!TryRead(out next)) next = await ReadAsync().ConfigureAwait(false); }
-                catch (ChannelClosedException) { break; } // expected
-                catch (Exception ex)
+
+                // Keep trying to read values
```
anurse: This is the pattern we generally use to read from a channel and gracefully stop without an exception. If there's a reason for the older behavior, I'm happy to reset it.

src/StackExchange.Redis/ChannelMessageQueue.cs
```diff
@@ -162,17 +175,29 @@ private async Task OnMessageSyncImpl()
...
+                    // so we keep TryReading in the loop.
                 }
 
                 try { handler(next); }
                 catch { } // matches MessageCompletable
             }
+
+            if (Completion.IsFaulted)
+            {
+                _onInternalError(Completion.Exception.InnerException);
```
anurse: This is technically new behavior, it will trigger this callback if the channel is closed with an `Exception`, but we don't ever do that in the normal flow. A custom implementor that uses their own Channel in `ChannelMessageQueue` might though...

src/StackExchange.Redis/ChannelMessageQueue.cs
```diff
@@ -193,68 +218,63 @@ private async Task OnMessageAsyncImpl()
...
-            try
-            {
-                return handler?.Target is ChannelMessageQueue
-                    && handler.Method.Name == nameof(HandleMessage);
-            }
-            catch
-            {
-                return false;
-            }
+            // REVIEW: Need more context here to properly replace this.
```
anurse: I think this can be easily restored by moving it to RedisSubscriber, since that's where the handler lives now.

src/StackExchange.Redis/ChannelMessageQueue.cs
mgravell: it was simply to do *something* in the case that an unexpected exception happened - agree "highly unlikely", but: at least in the "real" code, should do *something*; probably doesn't need to be mockable, if it presents a problem

anurse: We could have two constructors. One that takes a `RedisSubscriber` (i.e. the internal one, that can access `OnInternalError` directly and one that just takes a `ChannelReader` and an action to fire when unsubscribe is called. Since this is primarily (exclusively?) for testing, it's less important that the testable entry point be able to perform all the necessary functions.

PR #1026, opened 2018-12-28
Add pop functionality to sorted set (ZPOPMIN, ZPOPMAX)

martinekvili:
Redis 5.0.0 added functionality for popping from sorted sets, see [ZPOPMIN](https://redis.io/commands/zpopmin) and [ZPOPMAX](https://redis.io/commands/zpopmax).

### DIFF
--- a/src/StackExchange.Redis/Enums/RedisCommand.cs
+++ b/src/StackExchange.Redis/Enums/RedisCommand.cs
@@ -187,6 +187,8 @@
         ZINCRBY,
         ZINTERSTORE,
         ZLEXCOUNT,
+        ZPOPMAX,
+        ZPOPMIN,
         ZRANGE,
         ZRANGEBYLEX,
         ZRANGEBYSCORE,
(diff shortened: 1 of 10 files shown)


NickCraver:
On a text read, this looks good! 2 pieces of feedback on current:
1. It needs some testing love, any chance you could add a few to the suite? The pattern there from normal sets should be a quick thing to mirror :)
2. We should probably add a feature check for this as well (quick few-line addition in `RedisFeatures.cs`) - is it 4.9.1 or added after 5.x somewhere? Docs just say 5.0.0 so not sure where it landed...but since we're on 5.0 RTM now I'm fine just calling it 5.0. It'd look like this:
```c#
public bool SortedSetPop => Version >= v5_0_0;
```
(and needs a version add for 5.0.0 up top)

martinekvili:
Thank you very much for the quick response! 

1. I added tests similar to the normal sets.
2. Regarding the feature check, I looked into the release notes (avaliable [here](https://github.com/antirez/redis/blob/5.0/00-RELEASENOTES)), and it seems the new sorted set commands landed in 5.0 RC1 (line 2558: _"5. New sorted set commands: ZPOPMIN/MAX and blocking variants."_) – the same time as the `Stream` data type. Therefore, for the sake of consistency I used the same `v4_9_1` version that was already there for checking whether `Streams` are available.

mgravell:
Looks good to me; merging

mgravell:
and: thanks!

Issue #1027, opened 2018-12-29
Redis Streams Consumer groups busted in Redis 5.0.3


NickCraver:
I'm not sure what happened here but 5.0.0-rc3 worked, 5.0.3 does not. There's a break somewhere I'm looking into now. I thought the build was flaky, but it turns out the Ubuntu AppVeyor image just updated Redis to latest.

Upon building from source locally for 5.0.3 on WSL, I see the same test failures that didn't exist on a 5.0.0-rc3 run just before. Logging here to track it, but going through release notes first ([link](https://raw.githubusercontent.com/antirez/redis/5.0/00-RELEASENOTES)) to see if there are any noted protocol changes.

Sample test output:
```
[xUnit.net 00:00:12.90]     PubSub.TestMassivePublishWithWithoutFlush_Local [SKIP]
Skipped  PubSub.TestMassivePublishWithWithoutFlush_Local
[xUnit.net 00:00:19.98]     Scripting.PurgeLuaScriptOnFinalize [SKIP]
Skipped  Scripting.PurgeLuaScriptOnFinalize
[xUnit.net 00:00:20.50]     Streams.StreamConsumerGroupReadFromStreamBeginning [FAIL]
[xUnit.net 00:00:20.52]     Streams.StreamDeleteConsumer [FAIL]
[xUnit.net 00:00:20.52]     Streams.StreamConsumerGroupAcknowledgeMessage [FAIL]
Failed   Streams.StreamConsumerGroupReadFromStreamBeginning
Error Message:
 Assert.True() Failure
Expected: True
Actual:   False
Stack Trace:
   at StackExchange.Redis.Tests.Streams.StreamConsumerGroupReadFromStreamBeginning() in C:\git\StackExchange\StackExchange.Redis\tests\StackExchange.Redis.Tests\Streams.cs:line 249
Standard Output Messages:
 Compiled under .NETCoreApp2.1
Running on: Microsoft Windows 10.0.17763
 01:47:27.773: 01:47:27.773: Service Counts: (Scheduler) Queue: 43436, Pool: 47, (Completion) Queue: 11906, Pool: 1157
 01:47:27.773: 01:47:27.773: Service Counts: (Scheduler) Queue: 43436, Pool: 47, (Completion) Queue: 11906, Pool: 1157


Failed   Streams.StreamDeleteConsumer
Error Message:
 Assert.Equal() Failure
Expected: 2
Actual:   0
Stack Trace:
[...]

PR #1028, opened 2018-12-31
Fix streams consumer group tests.

ttingen:
Fix for #1027 

From the XREADGROUP docs on [redis.io](https://redis.io/commands/xreadgroup):

> The ID to specify in the STREAMS option when using XREADGROUP can be one of the following two:
> The special > ID, which means that the consumer want to receive only messages that were never delivered to any other consumer. It just means, give me new messages.
Any other ID, that is, 0 or any other valid ID or incomplete ID (just the millisecond time part), will have the effect of returning entries that are pending for the consumer sending the command. **So basically if the ID is not >, then the command will just let the client access its pending entries: delivered to it, but not yet acknowledged.**

Do you think there's a need to improve the `StreamReadGroup` method signatures to make this change more intuitive to users? Or maybe a possible addition to the StreamPosition class?

### DIFF
--- a/tests/StackExchange.Redis.Tests/Streams.cs
+++ b/tests/StackExchange.Redis.Tests/Streams.cs
@@ -243,7 +243,7 @@ namespace StackExchange.Redis.Tests
 
                 db.StreamCreateConsumerGroup(key, groupName, StreamPosition.Beginning);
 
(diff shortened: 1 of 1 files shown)


NickCraver:
Thanks for the fixes here - going to yank this in as-is but I'm not sure what to do about making it more intuitive...I agree it's problematic today since the docs are in paragraph form only and not that quick to find information in.

PR #1030, opened 2019-01-03
Update lastReadTickCount in ReadFromPipe

hpk:
lastReadTickCount is never updated after the connection is established, which leads to some confusing error messages.


mgravell:
Must have been an oversight when we rewrote the network code for v2 - looks good, thanks - merging

PR #1032, opened 2019-01-07
Rename index to readme & fix broken documentation links

dc-p8:
Documentation links in index are broken (at least on github). And I think it would be better to rename the main documentation file as readme so you don't have to search it.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,7 +1,7 @@
 StackExchange.Redis
 ===================
 
-[Release Notes](ReleaseNotes)
+[Release Notes](ReleaseNotes.md)
 
 ## Overview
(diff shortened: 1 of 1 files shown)


mgravell:
This PR may be based on an incorrect premise; these files are the basis for the github pages site, https://stackexchange.github.io/StackExchange.Redis/ - the links aren't broken **when viewed in that context**, and index is correct as the root file. I genuinely appreciate the time taken here, but I don't think this is something that needs "fixing".

dc-p8:
You right, I didn't think there were dedicated github pages for documentation. I didn't see that there was this link in the project home page.

Issue #1029, opened 2019-01-03
`ResponseTimeout` is unused


hpk:
I've noticed this while trying to track down some flaky behavior with connections to ElastiCache. We'll periodically see total Redis connectivity failure on a single container for a number of minutes (manifested as every single operation timing out), followed by a connection reset message:

```
StackExchange.Redis.ConnectionFailureType - SocketFailure (ReadSocketError/ConnectionReset,
last-recv: 4777) on <host>/Interactive, Idle, last: GET, origin: ReadFromPipe, outstanding: 9161,
last-read: 1996s ago, last-write: 0s ago, unanswered-write: 1996s ago, keep-alive: 10s, state:
ConnectedEstablished, mgr: 9 of 10 available, in: 0, last-heartbeat: 0s ago, last-mbeat: 0s ago,
global: 0s ago, v: 2.0.513.63329 [StackExchange.Redis.RedisConnectionException]
```

In version 1.*, there was a `ResponseTimeout` configuration option that would reset the connection when no reply was received after a certain number of milliseconds. That config option is still present in version 2, but I don't see it being used anywhere.

Leaving aside the issue of what's causing these connection failures in the first place, how can we get SE.Redis to reset the connection in a reasonable time frame if it's throwing `RedisTimeoutException`s consistently?

hpk:
Any update on this? Is the functionality previously provided by `ResponseTimeout` no longer supported? If so, it should be removed from the config options, or at least deprecated.

NickCraver:
Removed by #1042, thanks for the cleanup! Closing this out to tidy up :)

PR #1042, opened 2019-01-16
Remove obsolete responseTimeout setting

hpk:
Addresses #1029 

Alternatively, is there a recommended way of accomplishing what this used to do in v1?


mgravell:
Seems fine; if we need to revisit this later re the v1 bits, we can do that then. Much obliged.

PR #1090, opened 2019-03-11
try pipelines v2 - now with "ref T Current"

mgravell:


src/StackExchange.Redis/RedisTransaction.cs
```diff
@@ -505,12 +505,11 @@ protected override bool SetResultCore(PhysicalConnection connection, Message mes
                                     connection?.BridgeCouldBeNull?.Multiplexer?.OnTransactionLog($"processing {arr.Length} wrapped messages");
 
                                     int i = 0;
-                                    var iter = arr.GetEnumerator();
-                                    while(iter.MoveNext())
+                                    foreach(ref RawResult item in arr)
```
Rodrigo-Andrade: this is in the realm of micro optimizations, but wont this introduce a try catch block to dispose of the enumerable?

mgravell:
It is a custom iterator and isn't disposable. So: it shouldn't, but I'll
check the IL

src/StackExchange.Redis/RedisTransaction.cs
mgravell: it is a custom enumerator that isn't disposable; no `try`/`finally` block is emitted: https://sharplab.io/#v2:EYLgtghgzgLgpgJwD4AEAMACFBGA3AWACgUBmLAJgwGEMBvIjRrMlAFgwFkAKASjoaaCAZgHsEcCAGMAFlwBuEBBgCW8MCoB2GAOJwYvHgMGN6hY0wC+RjFbNMAYiJEAeZRpgA+HXt4YAvF4AJnBCEACuADYwBIS2RLAIYZIwGI4uACoeRKYAkKQYAEKKzpneMACiGmFgiBAwYr4BGMGhkdFEtglJKUUIJVmEuekA2gC6KjGC+eJCGOnUYQji7v5eMyrDaKOTTPnAThGcInJwAHJwAB76fE0wiXCTFkA

PR #1102, opened 2019-03-24
Delete Profiling_v1.md and Profiling_v1.md

Evahtr:
Deleted Profiling_v1.md and Profiling_v1.md which were contained in Profiling.md to make doc looks concise.


mgravell:
This just leaves 2 broken links and removes the actual useful content, no? I'm ... confused. Can you clarify the intent here? Because "as-is", this seems harmful.

NickCraver:
Since there's no response here and as Marc notes, these are used - going to close this out to tidy up.

PR #1104, opened 2019-03-29
Fix #1103 - add explicit operator support for ulong on RedisValue

### LINKED ISSUE #1103: StringSet last digit of uint64 value is wrong
I have an issue where StringSet messes up the last digit of my uint64 value.
I want to store 14220525521023800**5** but redis-cli > get key1 returns me 14220525521023800**0**.

See here an example :
```
using System;
using StackExchange.Redis;

namespace redisUint64
{
    class Program
    {
        static void Main(string[] args)
        {
            ulong value = 142205255210238005;
            var conn = ConnectionMultiplexer.Connect("127.0.0.1");
            var db = conn.GetDatabase();
            db.StringSet("key1", value);
            var retVal = db.StringGet("key1");
            Console.WriteLine(retVal);
        }
    }
}
```

Stackexchange.Redis version 2.0.571
Redis server version 5.0.4
.Net Core 2.2
Arch linux 64bit

mgravell:
context: #1103

This necessitated a new sentinel value for handling very large (64-bit) `ulong`, as we need to encode/decode them differently (obviously, they're not large negative numbers) - but we treat everything up to 63-bit as the previous `long`, for simplicity.

Adding these operators made some of the *other* operators ambiguous (from the compiler's perspective), so additional operators were also added to make this work without complaint.

Additionally:

- some tweaks to remove unnecessary string allocations (by using the `Utf8Parser`) were exploited
- all existing *direct* access to the overloaded field were removed; that was getting dangerous - there are now properties for all of these uses

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -1,5 +1,10 @@
 # Release Notes
 
+## (unreleased)
+
+- add `ulong` support to `RedisValue` and `RedisResult`
+- fix: remove odd equality: `"-" != 0` (we do, however, still allow `"-0"`, as that is at least semantically valid, and is logically `== 0`)
+
 ## 2.0.571
 
 - performance: use new [arena allocation API](https://mgravell.github.io/Pipelines.Sockets.Unofficial/docs/arenas) to avoid `RawResult[]` overhead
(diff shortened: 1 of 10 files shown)


mgravell:
PR appears to impact `Streams.StreamPositionValidateBeginning` and `Streams.StreamPositionDefaultValueIsBeginning` - I'll investigate why when I can

PR #1122, opened 2019-04-14
add optional timeout parameter to allow for XREAD/XREADGROUP BLOCK msec

Jens-G:
The [XREAD](https://redis.io/commands/xread) and [XREADGROUP](https://redis.io/commands/xreadgroup) commands have an optional Parameter BLOCK which comes handy in cases where a server application just waits for incoming messages to be processed. Without the BLOCK Parameter, the only option is to run a polling loop and waste precious CPU cycles. The current implementation does not support that parameter at all, so here's the patch to add it.

### DIFF
--- a/src/StackExchange.Redis/Interfaces/IDatabase.cs
+++ b/src/StackExchange.Redis/Interfaces/IDatabase.cs
@@ -1681,22 +1681,24 @@ namespace StackExchange.Redis
         /// <param name="key">The key of the stream.</param>
         /// <param name="position">The position from which to read the stream.</param>
         /// <param name="count">The maximum number of messages to return.</param>
+        /// <param name="timeout">Optionally wait for the specified number of msec for data to be read.</param>
         /// <param name="flags">The flags to use for this operation.</param>
         /// <returns>Returns a value of <see cref="StreamEntry"/> for each message returned.</returns>
         /// <remarks>Equivalent of calling XREAD COUNT num STREAMS key id.</remarks>
         /// <remarks>https://redis.io/commands/xread</remarks>
-        StreamEntry[] StreamRead(RedisKey key, RedisValue position, int? count = null, CommandFlags flags = CommandFlags.None);
+        StreamEntry[] StreamRead(RedisKey key, RedisValue position, int? count = null, int? timeout = null, CommandFlags flags = CommandFlags.None);
 
         /// <summary>
         /// Read from multiple streams.
         /// </summary>
         /// <param name="streamPositions">Array of streams and the positions from which to begin reading for each stream.</param>
         /// <param name="countPerStream">The maximum number of messages to return from each stream.</param>
+        /// <param name="timeout">Optionally wait for the specified number of msec for data to be read.</param>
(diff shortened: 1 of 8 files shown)


Jens-G:
That looks unrelated, I'd say …

```
dotnet: [xUnit.net 00:00:16.78]     Failover.ConfigVerifyReceiveConfigChangeBroadcast [FAIL]
At C:\projects\stackexchange-redis\build.ps1:49 char:9
+         dotnet test -c Release
+         ~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : NotSpecified: ([xUnit.net 00:0...roadcast [FAIL]:String) [], RemoteException
    + FullyQualifiedErrorId : NativeCommandError
[xUnit.net 00:00:16.80]     Issues.Issue1101.ExecuteWithUnsubscribeViaChannel 

```

mgravell:
This is a very deliberate omission. Blocking operations are anathema to the current implementation. It would interfere massively with both regular operation of the pipe from other threads, and background activity such as heartbeats. This is explained more here: https://stackexchange.github.io/StackExchange.Redis/PipelinesMultiplexers#multiplexing

If we want to add this (and I don't know whether we do, I haven't thought about it recently), we'd need some kind of dedicated leased connection that your code would "own" in isolation for the duration of this operation.

Jens-G:
Anything I can do?

Jens-G:
I see, I was not aware of that restriction. I guess that essentially means I can close the PR and forget about it?

mgravell:
I don't see a way that we can merge this currently without it causing
breakages. It does, however, add to the bucket of "people want X", which
can influence architectural decisions re possible options in the future.

Sorry if that sounds dismissive. It isn't meant to, and I really appreciate
the contributions - just : for reasons as described: it isn't quite that
simple today.

Jens-G:
Understood. Keep up the good work!

chris-ray:
If I may add to the bucket @mgravell, I find the explicit omission of the blocking features of redis to be.... blocking me from using this library.

A seemingly common use case for 5.x streams is a consoleapp, microservice if you will, that simply acts as a consumer waiting for messages to process, single-thread. XREADGROUP BLOCK is exactly what is desired in this case.

I think future devs coming to this library will find it distressing that this ability doesn't exist, like I did, especially since the alternative, more minimalist C# redis libraries are so woefully out of date.

mgravell:
@chris-ray noted; that's good feedback, as it allows me to add a +1 to the notion of dedicated additional sockets, which is what we'd need in order to support this; it also ties in nicely to the wider "pool" work. So: indeed I don't have a good solution for that *today*, but maybe that's a thing we can get in the not-too-distant future.

afederici75:
Hi :) Any news or updates on this? I am in the same situation of @chris-ray

mgravell:
Closing this down; when we come back to this with blocking support, we'll have a wave of things to do at the same time, including this.

mgravell:
really appreciate the effort, to be clear; but it needs to be part of a bigger piece of work

PR #1141, opened 2019-05-11
Updated StreamCreateConsumerGroup methods to use the MKSTREAM option.

ttingen:
Added the `createStream` parameter to the `StreamCreateConsumerGroup` methods, the parameter defaults to `true`. Also added appropriate tests for the new parameter and its potential values.

PR #1154, opened 2019-05-29
Added support for NOACK in the StreamReadGroup methods.

ttingen:
The StreamReadGroup methods were updated to include a `noAck` parameter that defaults to `false`. When `true`, the NOACK option is sent with the XREADGROUP command. According to the [documentation](https://redis.io/commands/xreadgroup), this is the equivalent of acknowledging the message when it is read and the message will not be added to the pending message list.

Tests were also added to verify that messages are not added to the pending message list when this option is used.


mgravell:
Adding an optional parameter (`bool noAck = false`) is a breaking change that will result in `MissingMethodException` errors. Given that we've recently done a "major", I'm not in a hurry to force *another* major - so: we'll need to do it another way. This is usually done by adding an overload.

The question then becomes:

1. add a `bool`
2. add a new `[Flags]` enum, enabling scope for additional options later
3. (ab)use `CommandFlags` by adding one there

I think 2 would probably be the safest option, but I'm open to other options.

ttingen:
Thanks Marc, I will update the change to be an overload. It looks like I need to make the same change for #1141 as well.

mgravell:
Yes,same thing there. Comment added.

mgravell:
Note on making this change: the usual pattern when there are existing optional parameters is to make the **old version** non-optional throughout, i.e.

from:

``` c#
Whatever SomeMethod(int a, string b, float? c = null, SomeEnum d = SomeEnum.None) {...}
```

to

``` c#
Whatever SomeMethod(int a, string b, float? c, SomeEnum d)
    => SomeMethod(a, b, c, blah, d);

Whatever SomeMethod(int a, string b, float? c = null, NewArgHere yay = blah, SomeEnum d = SomeEnum.None) {...}
```

This minimizes problems due to ambiguous method resolution, while ensuring that old code a: compiles, and b: continues to execute without compiling. Additionally, newly compiled code will almost always pick the **new** method, avoiding a stack-frame (which may or may not be inlined by the JIT).

ttingen:
Apologies for the delay in getting this done...summer vacation, work, etc...

I will take care of the overload for the MKSTREAM option as well. Thanks Marc!

mgravell:
Merging, thanks

PR #1159, opened 2019-06-06
Updated documentation as PreserveAsyncOrder is deprecated

odyth:
Updated documentation to reflect that PreserveAsyncOrder is deprecated and ChannelMessageQueue should be used instead


NickCraver:
This may need some manual merge love after #1162

PR #1175, opened 2019-07-06
Make GeoRadiusResult constructor public to make it mockable

firenero:
It's impossible to mock `GeoRadius` and `GeoRadiusAsync` methods now because `GeoRadiusResult` constructor is internal. Given that constructors for all other GEO calls are public I think it's possible to make `GeoRadiusResult` constructor public as well.


firenero:
@NickCraver Can I help somehow to get it merged? Lack of public constructor here makes unit testing really hard for a while already.

NickCraver:
I'm unsure on timeline for the next release, but it'll be on MyGet shortly.

PR #1181, opened 2019-07-15
Dotnet core 2.2 upgrade

kishoretvk:
updated references to .net core 2.2 upgrade and changed language test case


mgravell:
Question: what is the thing you're trying to do here? Changing from ns20 to netcoreapp2.2 isn't an upgrade - it is a breaking change. The old code would have already worked just fine on netcoreapp2.2 unless there's a very specific issue you're working around. And it is only an *upgrade" if you're exploiting some new feature iof a higher TFM. Removing TFMs is something that needs serious consideration, especially removing ns2.0

So: what is he thing you're trying to achieve here, so that we can understand the context.

Adding a new TFM for the lib may be warranted if there's a new feature to exploit

Adding a new test TFM : probably fine, especially if it adds rather than replaces

Removing TFMs... needs discussion. As does adding a TFM without an explicit stated reason.

mgravell:
This is not an upgrade, and no explanation / communication offered. Closing.

PR #1190, opened 2019-07-21
Include details of abbreviations seen in exceptions

marafiq:
I have been getting Redis Timeout under high load, but it was not making sense at all. So i opened the source to understand what these abbreviations meant. I hope it will help people understand error messages a lot better.
My CPU Use case:
If Redis dedicated thread pool state looks good, and global thread pool has more free threads than minimum threads, but your CPU is usage is 90%, more likely you will face time outs, depending upon the size of bytes client is trying to read or write.


NickCraver:
First, good idea! I recommend using a Markdown table here instead though (we could add a description column as well), an example of this can be found at: https://github.com/StackExchange/StackExchange.Redis/blob/master/docs/Configuration.md#configuration-options

marafiq:
Updated PR to include Markdown table, i have written descriptions as per my understanding, kindly provide feedback if descriptions can be better.

mgravell:
nice job, thanks; merging

Issue #1198, opened 2019-07-29
NRediSearch set return parameter of search


oruchreis:
Hi,
We can use `infields` parameter to filter the used fields in the query with `LimitFields` method. Could you implement also `RETURN {num} {field}` parameter to limit returned fields in the result? It is very useful if you have documents that have a lot of fields and you don't need all fields in every query result.

NickCraver:
Fixed in #1199, thank you for the contribution!

PR #1199, opened 2019-07-29
Added "Return" parameter to Query.

oruchreis:
PR for #1198

### DIFF
--- a/src/NRediSearch/Query.cs
+++ b/src/NRediSearch/Query.cs
@@ -142,7 +142,8 @@ namespace NRediSearch
         /// Set the query language, for stemming purposes; see http://redisearch.io for documentation on languages and stemming
         /// </summary>
         public string Language { get; set; }
-        internal string[] _queryFields = null;
+        internal string[] _fields = null;
+        internal string[] _keys = null;
         internal string[] _returnFields = null;
         /// <summary>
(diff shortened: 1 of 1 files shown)


src/NRediSearch/Query.cs
```diff
@@ -317,9 +324,20 @@ public Query AddFilter(Filter f)
         /// </summary>
         /// <param name="fields">a list of TEXT fields in the schemas</param>
         /// <returns>the query object itself</returns>
-        public Query LimitFields(params string[] fields)
```
mgravell: this is a breaking change; any changes should preserve the existing API (it is OK to forward if needed)

mgravell:
most of the API here is a snapshot of what JRediSearch was at some point in time; do you know (I don't) what JRediSearch exposes here?

oruchreis:
Ok, I didn't know that this is a snapshot of JRediSearch API. It seems JRediSearch have return parameter also inkeys parameter with different method name. https://github.com/RediSearch/JRediSearch/blob/5d8ea3c8d2f71eefde59d88c17bb92c65dfb95ed/src/main/java/io/redisearch/Query.java#L413
 I'll update than.

mgravell:
Seems great; merging, thanks

PR #1230, opened 2019-09-19
Accurate timespan if the stopwatch is high resolution

thomhurst:
We were using profiling, and our durations were crazy high.

We traced it down to this, the stopwatch can be a high or low resolution.

When constructing the timespan, we should check if the stopwatch was high resolution and do some calculations for a more accurate time.

### DIFF
--- a/src/StackExchange.Redis/Profiling/TimeSpanHelper.cs
+++ b/src/StackExchange.Redis/Profiling/TimeSpanHelper.cs
@@ -6,14 +6,14 @@ namespace StackExchange.Redis.Profiling
     /// <summary>
     /// A helper class for dealing with Low/High Resolution Stopwatches and their ticks
     /// </summary>
-    public class TimeSpanHelper
+    internal static class TimeSpanHelper
     {
         /// <summary>
-        ///
+        /// Used to construct a timespan from ticks obtained using Stopwatch.GetTimestamp()
         /// </summary>
         /// <param name="ticks"></param>
-        /// <returns>A TimeSpan represented from the ticks</returns>
-        public static TimeSpan FromStopwatchTicks(long ticks)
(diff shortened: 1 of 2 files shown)


src/StackExchange.Redis/Profiling/TimeSpanHelper.cs
```diff
@@ -0,0 +1,21 @@
+using System;
+using System.Diagnostics;
+
+namespace StackExchange.Redis.Profiling
+{
+    /// <summary>
+    /// A helper class for dealing with Low/High Resolution Stopwatches and their ticks
+    /// </summary>
+    public class TimeSpanHelper
```
mgravell: should probably be `internal static class` ?

mgravell:
in principle: fine; I've added a comment just on the public API exposure, but: fix that and it should be good

src/StackExchange.Redis/Profiling/TimeSpanHelper.cs
thomhurst: Good shout. Have pushed an update :)

thomhurst:
Changed!

mgravell:
Awesome; all merged, thanks!

NickCraver:
Soooo, turns out this is what broke the profiling tests, because it's reducing accuracy. For example, in `Profiling.Simple`, we end up with:
```c#
TimeSpanHelper.FromStopwatchTicks(4085501507216-4085501507028)
```
...which yields `TimeSpan.Zero`...but it's very not-zero and that's breaking profiling.

I think we need to change whatever this was trying to solve. Can you please explain a bit more about the problem this was trying to address?

NickCraver:
I am reverting this for now in #1250 - once we better understand the problem here, we can find another solution. Admittedly, I don't know what the issue is since ultimately both formulas should work being tick-based.

PR #1232, opened 2019-09-22
add support to sentinel version v1.2.x

shadim:


NickCraver:
Just seeing this - there will be no further releases on SE.Redis 1.x, so we will not be accepting any more commits either.

However, we still welcome Sentinel support (_with tests_, which so far no one has taken up) against current master :)

PR #1234, opened 2019-09-24
add CheckCertificateRevocation of SslStream.AuthenticateAsClient to Configuration

BLun78:
Hello,

in an enterprise infrastructure there is not always the possibility to check certificates for revocation. The target server from a certificate is not always accessible, e.g. AWS Cloud on Premise. I need to be able to disable the certificate revocation check.

I customized the documentation in the configuration and added the CheckCertificateRevocation attribute. The SslStream is now extended by a variation in the AuthenticateAsClient method. 
The risk that a defect is caused by this change is low. Also there is no breaking change, because the default: CheckCertificateRevocation=true remains the same. 
Please record this changerequest. Thank you.


mgravell:
I'm fine with this; makes perfect sense, thanks; merging

PR #1250, opened 2019-10-14
Revert "Accurate timespan if the stopwatch is high resolution"

NickCraver:
Reverts StackExchange/StackExchange.Redis#1230 - this change was not correct and broke timing by slimming down the the nearest millisecond. Since most often less than 1ms passes in the library before most things happen, this only worked on a very slow machine.

We know this breaks thing because less than 1ms comes out as `TimeSpan.Zero` and is the cause of our broken `Simple` tests (`Profiling.Simple` and `Cluster.SimpleProfiling`). I'm not sure exactly what the underlying cause of #1230 is (I've asked the question there), but we need to solve it in a different way.


NickCraver:
Merging in - this fixes the tests - we'll track down what was being hit in #1230 a bit later

PR #1265, opened 2019-10-31
Fix TestConfig.cs link

inkel:
The previous link was relative to the document, however, when displayed in GitHub Pages it ends in a 404. I've changed to point to the source file in GitHub UI.


gkorland:
LGTM!

PR #1267, opened 2019-11-04
Bring NRediSearch "even" with JRediSearch

tombatron:
Ported AggregationBuilder with Tests from JRediSearch.

Added Cursor support to the Client and AggregationResult.

Added RETURN support to the Query class.

Added INKEYS support to the Query class.

Added support for the FT.ALTER command.

Introduced new result type for the FT.INFO command. Removed TODO.

Updated “TestHighlightSummarize” test to match that of JRediSearch.

Added support for the FT.MGET command.

Expanded support for suggestions. Ported Suggestion class (with builder) and SuggestionOptions class (with builder).

Added ability to add multiple documents to an index with a single call.

Added ability to specify that an underlying indexed document be deleted when calling `DeleteDocument`.

Added ability to specify multiple documents for deletion. 

Refactored IndexOptions to allow it to behave more like JRediSearch. Removed NOSCOREIDX index option as that has been deprecated.


tombatron:
@mgravell As mentioned [here](https://twitter.com/tombatron/status/1189683468550184962) I'm taking a stab at bringing NRediSearch "even" with JRediSearch.

I look forward to your feedback!

gkorland:
@tombatron thanks for the great contribution!

mgravell:
This is awesome, thanks; I'm going to try and steal an hour this morning to read through all of this

src/NRediSearch/AggregationResult.cs
```diff
@@ -9,6 +9,12 @@ namespace NRediSearch
     public sealed class AggregationResult
     {
         private readonly Dictionary<string, RedisValue>[] _results;
+
+        internal AggregationResult(RedisResult result, long cursorId) : this(result)
+        {
+            CursorId = cursorId;
+        }
+
         internal AggregationResult(RedisResult result)
```
mgravell: since this is `internal`, we can add an optional parameter without breaking the public API; suggest just one ctor: `internal AggregationResult(RedisResult result, long cursorId = -1) { ...}`

src/NRediSearch/Aggregation/AggregationBuilder.cs
```diff
@@ -0,0 +1,157 @@
+﻿// .NET port of https://github.com/RedisLabs/JRediSearch/
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Linq;
+using NRediSearch.Aggregation.Reducers;
+
+namespace NRediSearch.Aggregation
+{
+    public class AggregationBuilder
```
mgravell: if this isn't using inheritance, we should probably mark this `sealed` until there's a good reason to un-seal it

src/NRediSearch/Aggregation/AggregationBuilder.cs
```diff
@@ -0,0 +1,157 @@
...
+        }
+
+        public AggregationBuilder SortBy(int max, params SortedField[] fields)
+        {
+            SortBy(fields);
+
+            if (max > 0)
+            {
+                _args.Add("MAX");
+                _args.Add(max.ToString());
```
mgravell: invariant culture

src/NRediSearch/Aggregation/AggregationBuilder.cs
```diff
@@ -0,0 +1,157 @@
...
+{
+    public class AggregationBuilder
+    {
+        private readonly List<object> _args = new List<object>();
+
+        public bool IsWithCursor { get; private set; }
+
+        public ReadOnlyCollection<object> Args => new ReadOnlyCollection<object>(_args);
+
+        public string ArgsString => string.Join(" ", _args.ToArray());
```
mgravell: is this `ToArray` needed? Most TFMs should be fine with just `string.Join(" ", _args);` - also; this isn't "free" - so having it as a property is misleading; suggest `GetArgsString()` unless there's a reason not to?

src/NRediSearch/Aggregation/AggregationBuilder.cs
```diff
@@ -0,0 +1,157 @@
...
+
+namespace NRediSearch.Aggregation
+{
+    public class AggregationBuilder
+    {
+        private readonly List<object> _args = new List<object>();
+
+        public bool IsWithCursor { get; private set; }
+
+        public ReadOnlyCollection<object> Args => new ReadOnlyCollection<object>(_args);
```
mgravell: ditto re "not free"; what is the use-case of this API? I wonder if it can be simply omitted, or exposed in a different way

src/NRediSearch/Aggregation/AggregationBuilder.cs
```diff
@@ -0,0 +1,157 @@
...
+
+        public string ArgsString => string.Join(" ", _args.ToArray());
+
+        public AggregationBuilder() : this("*")
+        {
+        }
+
+        public AggregationBuilder(string query) => _args.Add(query);
```
mgravell: combine constructor? i.e. `public AggregationBuilder(string query = "*")` ?

src/NRediSearch/Aggregation/AggregationBuilder.cs
```diff
@@ -0,0 +1,157 @@
...
+
+            return this;
+        }
+
+        public AggregationBuilder Limit(int count) => Limit(0, count);
+
+        public AggregationBuilder SortBy(params SortedField[] fields)
+        {
+            _args.Add("SORTBY");
+            _args.Add((fields.Length * 2).ToString());
```
mgravell: invariant culture (note to self: see if this is exposed anywere in a single place)

src/NRediSearch/Aggregation/AggregationBuilder.cs
```diff
@@ -0,0 +1,157 @@
...
+
+        public AggregationBuilder Cursor(int count, long maxIdle)
+        {
+            IsWithCursor = true;
+
+            if (count > 0)
+            {
+                _args.Add("WITHCURSOR");
+                _args.Add("COUNT");
+                _args.Add(count.ToString());
```
mgravell: invariant culture

src/NRediSearch/Aggregation/AggregationBuilder.cs
```diff
@@ -0,0 +1,157 @@
...
+
+                if (maxIdle < long.MaxValue && maxIdle >= 0)
+                {
+                    _args.Add("MAXIDLE");
+                    _args.Add(maxIdle.ToString());
```
mgravell: invariant culture

src/NRediSearch/Aggregation/AggregationBuilder.cs
```diff
@@ -0,0 +1,157 @@
...
+            foreach(var arg in _args)
+            {
+                args.Add(arg);
+            }
+        }
+
+        private static void AddCommandLength(List<object> list, string command, int length)
+        {
+            list.Add(command);
+            list.Add(length.ToString());
```
mgravell: invariant culture

mgravell:
Note on the integers: I started adding notes to use invariant culture, but:
actually the library should handle this intrernally. Try just adding them
as integers?

src/NRediSearch/Aggregation/AggregationBuilder.cs
mgravell: actually, what is the use-case of this API?

src/NRediSearch/Aggregation/AggregationBuilder.cs
```diff
@@ -0,0 +1,157 @@
+﻿// .NET port of https://github.com/RedisLabs/JRediSearch/
```
mgravell: wait; is this class the same thing as `AggregationRequest` ? they *look* pretty similar..?

src/NRediSearch/Document.cs
```diff
@@ -18,6 +18,8 @@ public class Document
         public Document(string id, double score, byte[] payload) : this(id, null, score, payload) { }
         public Document(string id) : this(id, null, 1.0, null) { }
 
+        public Document(string id, Dictionary<string, RedisValue> fields) : this(id, fields, 1.0) { }
+
         public Document(string id, Dictionary<string, RedisValue> fields, double score) : this(id, fields, score, null) { }
```
mgravell: suggest make this optional instead, with `, double score = 1.0` - this is not breaking, and keeps the API count down

src/NRediSearch/Extensions.cs
```diff
@@ -3,20 +3,11 @@
...
 namespace NRediSearch
 {
     public static class Extensions
     {
-        /// <summary>
-        /// Set a custom stopword list
-        /// </summary>
-        /// <param name="options">The <see cref="IndexOptions"/> to set stopwords on.</param>
-        /// <param name="stopwords">The stopwords to set.</param>
-        public static ConfiguredIndexOptions SetStopwords(this IndexOptions options, params string[] stopwords)
```
mgravell: removing an API is a hard break; what is the reasoning for this? there may be better options; for example, to avoid a hard break, if an old API gains a method that was *previously* being provided as an extension, then you should retain the old API as a forward, but just remove the `this` modifier; this means it still works if old IL requests it, but newly compiled code will not use it; it might also be worth adding `[Obsolete]` in that scenario

src/NRediSearch/Suggestion.cs
```diff
@@ -0,0 +1,108 @@
...
+        }
+
+        public override bool Equals(object obj)
+        {
+            if (this == obj)
+            {
+                return true;
+            }
+
+            if (obj == null || obj.GetType() != GetType())
```
mgravell: prefer something like `if (!(obj is Suggestion that)) return false;`, and remove the `that` from 2 lines below

src/NRediSearch/Suggestion.cs
```diff
@@ -0,0 +1,108 @@
...
+            {
+                return false;
+            }
+
+            var that = obj as Suggestion;
+
+            return Score.Equals(that.Score) && String.Equals(that.String) && Payload.Equals(that.Payload);
```
mgravell: `.Equals` here is dangerous re nullability; are they known to be not null?

src/NRediSearch/Suggestion.cs
```diff
@@ -0,0 +1,108 @@
+﻿// .NET port of https://github.com/RedisLabs/JRediSearch/
+using System;
+using StackExchange.Redis;
+
+namespace NRediSearch
+{
+    public class Suggestion
```
mgravell: suggest `sealed` if not known to need inheritance

mgravell: I actually wonder whether this should be `readonly struct`, but... meh, that might limit us if `JRediSearch` gets exotic in the future

src/NRediSearch/SuggestionOptions.cs
```diff
@@ -0,0 +1,92 @@
...
+        private With _with;
+        public With GetWith() => _with;
+
+        public bool Fuzzy { get; }
+
+        public int Max { get; } = 5;
+
+        public SuggestionOptionsBuilder ToBuilder() => new SuggestionOptionsBuilder(this);
+
+        public class With
```
mgravell: we know that java enums are always going to be slightly different to .NET; I wonder if this is just:

``` c#
[Flags]
public enum With {
    None = 0,
    Payloads = 1,
    Scores = 2,
    PayloadsAndScores = Payloads | Scores,
}
```

(and worry about the flags in a private method on `SuggestionOptions`)

src/NRediSearch/Client.cs
```diff
@@ -1,17 +1,18 @@
...
+using StackExchange.Redis;
+using static NRediSearch.Schema;
 
 namespace NRediSearch
 {
     public sealed class Client
     {
         [Flags]
-        public enum IndexOptions
+        public enum IndexOption
```
mgravell: renaming a type is a breaking change; are there reasons? note: `[Flags]` enums are traditionally pluralized

src/NRediSearch/Client.cs
```diff
@@ -30,23 +31,21 @@ public enum IndexOptions
...
+            Default = UseTermOffsets | KeepFieldFlags | KeepTermFrequencies,
             /// <summary>
             /// If set, we keep an index of the top entries per term, allowing extremely fast single word queries
             /// regardless of index size, at the cost of more memory
             /// </summary>
-            UseScoreIndexes = 4,
-            /// <summary>
-            /// If set, we will disable the Stop-Words completely
-            /// </summary>
-            DisableStopWords = 8
```
mgravell: I don't know the back-story here; is this an API that got removed? or...?

The key point for me here is that this is a breaking change; that isn't necessarily a blocker, but there are usually better ways to manage a breaking change, for example adding `[Obsolete(..., true)]` for a while, and raising suitable exceptions - so that it isn't *as* surprising

src/NRediSearch/Client.cs
```diff
@@ -55,45 +54,36 @@ public ConfiguredIndexOptions(IndexOptions options)
...
-
-            internal static void SerializeRedisArgs(IndexOptions flags, List<object> args)
-            {
-                if ((flags & IndexOptions.UseTermOffsets) == 0)
+                if (!_options.HasFlag(IndexOption.UseTermOffsets))
                 {
                     args.Add("NOOFFSETS".Literal());
                 }
-                if ((flags & IndexOptions.KeepFieldFlags) == 0)
+                if (!_options.HasFlag(IndexOption.KeepFieldFlags))
```
mgravell: note that `Enum.HasFlag<T>` is a boxing operation; perhaps add a helper method:

``` c#
private bool HasOption(IndexOption option) => (_options & option) != 0;
```

then

``` c#
if (HasOption(IndexOption.KeepFieldFlags)) {...}
```

etc

src/NRediSearch/Client.cs
```diff
@@ -30,23 +31,21 @@ public enum IndexOptions
...
             /// </summary>
-            UseScoreIndexes = 4,
-            /// <summary>
-            /// If set, we will disable the Stop-Words completely
-            /// </summary>
-            DisableStopWords = 8
+            KeepTermFrequencies = 4,
         }
 
-        public sealed class ConfiguredIndexOptions
```
mgravell: this break is ... problematic; I acknowledge that JRediSearch did this; I'm not sure we want to mirror their casual approach to compatibility :) needs thought, I think

src/NRediSearch/Client.cs
```diff
@@ -421,52 +479,95 @@ public async Task<bool> AddHashAsync(RedisKey docId, double score, bool replace)
         /// <summary>
         /// Get the index info, including memory consumption and other statistics.
         /// </summary>
-        /// <remarks>TODO: Make a class for easier access to the index properties</remarks>
         /// <returns>a map of key/value pairs</returns>
-        public Dictionary<string, RedisValue> GetInfo() =>
-            ParseGetInfo(DbSync.Execute("FT.INFO", _boxedIndexName));
+        public InfoResult GetInfo() =>
```
mgravell: I think we should leave the old API alone here (breaking changes, etc), and add a new API for the parsed version; perhaps `GetInfo()` and `GetInfoParsed()` ?

src/NRediSearch/Client.cs
```diff
@@ -486,22 +587,6 @@ public async Task<bool> DropIndexAsync()
             return (string)await _db.ExecuteAsync("FT.DROP", _boxedIndexName).ConfigureAwait(false) == "OK";
         }
 
-        /// <summary>
-        /// Optimize memory consumption of the index by removing extra saved capacity. This does not affect speed
-        /// </summary>
-        public long OptimizeIndex()
```
mgravell: what is the nature of the change here? does this no longer exist in the API? I would suggest:

``` c#
[Obsolete]
public long OptimizeIndex() => -1;
```

src/NRediSearch/Client.cs
```diff
@@ -486,22 +587,6 @@ public async Task<bool> DropIndexAsync()
...
-        {
-            return (long)DbSync.Execute("FT.OPTIMIZE", _boxedIndexName);
-        }
-
-        /// <summary>
-        /// Optimize memory consumption of the index by removing extra saved capacity. This does not affect speed
-        /// </summary>
-        public async Task<long> OptimizeIndexAsync()
```
mgravell: ditto

src/NRediSearch/Client.cs
```diff
@@ -562,43 +675,109 @@ public async Task<bool> DeleteSuggestionAsync(string value)
         /// Get completion suggestions for a prefix
         /// </summary>
         /// <param name="prefix">the prefix to complete on</param>
-        /// <param name="fuzzy"> if set,we do a fuzzy prefix search, including prefixes at levenshtein distance of 1 from the prefix sent</param>
-        /// <param name="max">If set, we limit the results to a maximum of num. (Note: The default is 5, and the number cannot be greater than 10).</param>
+        /// <param name="suggestionOptions"> the options on what you need returned and other usage</param>
         /// <returns>a list of the top suggestions matching the prefix</returns>
-        public string[] GetSuggestions(string prefix, bool fuzzy = false, int max = 5)
+        public Suggestion[] GetSuggestions(string prefix, SuggestionOptions options)
```
mgravell: no problem with having a structured API, but the old API should be retained and *proxy* to the new one, if that is at all achievable

src/NRediSearch/Client.cs
```diff
@@ -611,10 +790,12 @@ public AggregationResult Aggregate(AggregationRequest query)
 
             return new AggregationResult(resp);
         }
+
         /// <summary>
         /// Perform an aggregate query
         /// </summary>
         /// <param name="query">The query to watch</param>
+        [Obsolete("Use `AggregateAsync` method that takes an `AggregationBuilder`.")]
```
mgravell: why? I'm fuzzy on what the difference is between these APIs, since they look *very* similar; could do with some discussion here, I think

src/NRediSearch/Client.cs
```diff
@@ -674,6 +997,92 @@ public Document GetDocument(string docId)
...
+            };
+
+            foreach (var docId in docIds)
+            {
+                args.Add(docId);
+            }
+
+            var queryResults = (RedisResult[])DbSync.Execute("FT.MGET", args);
+
+            var result = new Document[docIds.Length];
```
mgravell: perhaps special-case zero length result with `Array.Empty<Document>()`

src/NRediSearch/InfoResult.cs
```diff
@@ -0,0 +1,153 @@
...
+
+        public Dictionary<string, RedisResult[]> Fields { get; } = new Dictionary<string, RedisResult[]>();
+
+        public long NumDocs { get; }
+
+        public long NumTerms { get; }
+
+        public long NumRecords { get; }
+
+        public double InvertedSzMb { get; }
```
mgravell: nomenclature; is the megabit (Mb)? megabyte (MB)? or mebibyte? (MiB)? I suspect the last, but I'm pretty sure it isn't the first, so `Mb` is misleading

src/NRediSearch/InfoResult.cs
```diff
@@ -0,0 +1,153 @@
...
+
+        public Dictionary<string, RedisResult> GcStats { get; } = new Dictionary<string, RedisResult>();
+
+        public Dictionary<string, RedisResult> CursorStats { get; } = new Dictionary<string, RedisResult>();
+
+        public InfoResult(RedisResult result)
+        {
+            var results = (RedisResult[])result;
+
+            for (var i = 0; i < results.Length; i += 2)
```
mgravell: do we actually need to parse these all out? most of these feel like things that we could lookup on-the-fly cheaply, i.e.

``` c#
public double DocTableSizeMb => GetDouble("doc_table_size_mb");
private double GetDouble(string key)
    => all.TryGetValue(key, out var value) ? (double)value : default;
```

mgravell:
First: this is awesome, thanks.

I've gone through and added a first pass at feedback; my biggest concern is re back-compat. However! We're also at a point where we haven't released version that targets v2 of the underlying client, so: in theory we *could* break compat at the same time. But even then I'd be keen to *minimize* any unexpected / unnecessary breaks.

there's a few discussion items in the feedback to chew through, but: looking good

src/NRediSearch/Aggregation/AggregationBuilder.cs
tombatron: The only use case that I can see is internal for the tests present in the QueryBuilder\BuilderTest.cs file.

How about I convert to a method and make it internal?

src/NRediSearch/Aggregation/AggregationBuilder.cs
tombatron: Yeah, the `ArgsString` was for testing, I believe I ported this because it was there. 

I'll go ahead and remove it. If it becomes useful to someone later I suppose we could just expose a property that returns the private `_args` List.

src/NRediSearch/Aggregation/AggregationBuilder.cs
tombatron: I thought the same exact thing. 

While I was reviewing the Java implementation I noticed that the `aggregate` method that accepted an instance of `AggregationRequest` had been deprecated in favor of the the `AggregationBuilder`.

src/NRediSearch/Client.cs
tombatron: Yes, apparently all the optimization is done behind the curtains without user intervention required.

mgravell: no problem with that; my concern is the library break - I'd rather deprecate it properly so people are directed to stop using it, rather than getting a `MissingMethodException`.

src/NRediSearch/Client.cs
tombatron: Yeah I think I might have gotten a little carried away with making NRediSearch "look" like JRediSearch. You think I should just revert my changes while taking into account the changes that need to be applied (missing and new index options)?

src/NRediSearch/Client.cs
tombatron: Sounds good.

I added OptimizeIndex and OptimizeIndexAsync back in, they now return `default` and are marked as obsolete.

src/NRediSearch/Client.cs
tombatron: I hate to "be that guy", but I have no idea. Maybe @gkorland could shed some light on this?

tombatron: To be clear, my intent was to bring parity between the two libraries. 😬

src/NRediSearch/Client.cs
tombatron: At the top of this method I'm checking to see if the length of the `docIds` array is 0, if so then I'm returning an `Array.Empty<Document>()` (I just changed that it was `new Document[]{}`. 

The intent here is to return `null` for any document not retrieved.

tombatron:
I must have overlooked the `TestStopWords` test. That appears to be failing locally for me on the last assertion, though I have no idea why. 

This will be fun...

src/NRediSearch/Aggregation/AggregationBuilder.cs
tombatron: @gkorland? Any insight from the Java side of things?

gkorland: Yes AggregationRequest was replaced by AggregationBuilder

src/NRediSearch/Client.cs
gkorland: no need to add AggregationRequest

src/NRediSearch/Suggestion.cs
tombatron: Nice!

src/NRediSearch/Suggestion.cs
tombatron: 😬 Point taken. 

To answer your question, the builder would ensure that the value of `String` was set, but not that the value of `Payload`. 

Updating these.

src/NRediSearch/SuggestionOptions.cs
tombatron: Fair enough.

src/NRediSearch/InfoResult.cs
tombatron: You suspect correct!

[source](https://github.com/RediSearch/RediSearch/blob/fdbb3b21bb875cdcd798248b6f1346242048cb4c/src/module.c#L188)

Looks like the number of bytes divided by `0x100000`

src/NRediSearch/InfoResult.cs
tombatron: That seems like it's be a lot less gross.

src/NRediSearch/Client.cs
tombatron: I reverted this logic back to how it was before.

src/NRediSearch/Client.cs
tombatron: I readded the `DisableStopWords` enum. It was removed initially when I was going a bit mustang trying to match the JRediSearch api.

src/NRediSearch/Client.cs
tombatron: Reverted.

src/NRediSearch/Client.cs
tombatron: Done!

src/NRediSearch/Client.cs
tombatron: Fixed.

tombatron:
It looks like the AggregationBuilder added the ability to FILTER and CURSOR. It shouldn't be an issue to port that to the AggregationRequest if that's the direction you wanna take.

As for why the AggregationRequest was replaced by the builder... Perhaps some kind of idiomatic Java practice?

gkorland:
@tombatron it was replaced as part of a code "refactoring"

gkorland:
p.s. @tombatron thanks for the great work, it looks great

mgravell:
Appreciate the input here - the code and the eyeballs. I'm AFK for a bit
over the holiday season, but let's try and get this merged ASAP in January.

tombatron:
Have a good holiday my man!

NickCraver:
Note: I've added NRediSearch to the Linux test suite in #1316 - once that's merged, we should pull it in here to get PR testing :)

tombatron:
Awesome! Happy New Year!

NickCraver:
Merging master in so we light up testing!

tombatron:
Sweet!

tombatron:
Hey @NickCraver is there anything outstanding that I need to do, or are we just waiting on some extra bandwidth?

mgravell:
Looking great, thanks; merging

tombatron:
Thanks @mgravell ! It was a little intimidating opening the PR but you and @NickCraver I really appreciate y'all for making it as painless as possible.

mgravell:
no, thank *you*; there's a lot of work here that **I didn't have to do**, which is great.

Issue #1282, opened 2019-11-19
ERR wrong number of arguments for 'sadd' command


omeaart:
Not entirely sure if this is expected behaviour, but it surprised me. When the SetAdd(Async) methods are used with an empty array of values, there is a broken call made to redis.

To reproduce:
```
var redisConnection = ConnectionMultiplexer.Connect("localhost");
redisConnection.GetDatabase().SetAdd("set1", new RedisValue[0]);
```

This also fails for SetAddAsync.

Version: 2.0.601
Tested with dotnetcore 2.2 and 3.0

mgravell:
Agreed - we should detect that and no-op it (avoiding the redis call
completely, whenever possible)

omeaart:
This was resolved in https://github.com/StackExchange/StackExchange.Redis/pull/1283

PR #1283, opened 2019-11-20
fix for #1282: do no-op when adding 0 values to a set

omeaart:
This provides a fix for #1282. I hope the tests are sufficient.


mgravell:
That's great, thanks! All looks good, including tests (I reviewed the CI fail: not your fault). Merging, with thanks.

PR #1291, opened 2019-11-28
Add support for TOUCH command 

### LINKED ISSUE #1278: Expose a friendly way to run the "TOUCH" command
We currently run the [`TOUCH`](https://redis.io/commands/touch) command like this:
```csharp
byte[] keyToTouch = /* ... */;
object[] args = { keyToTouch };
db.Execute("TOUCH", args, CommandFlags.FireAndForget);
```

It would be nice to have a friendlier version of this.

gkorland:
Fix #1278

### DIFF
--- a/src/StackExchange.Redis/Enums/RedisCommand.cs
+++ b/src/StackExchange.Redis/Enums/RedisCommand.cs
@@ -161,6 +161,7 @@
         SYNC,
 
         TIME,
+        TOUCH,
         TTL,
         TYPE,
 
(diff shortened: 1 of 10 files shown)


gkorland:
@mgravell any idea what is going on here? the tests errors line make no sense.

gkorland:
@mgravell I managed to clean the Linux tests but on windows I'm getting this error
`StackExchange.Redis.RedisServerException : ERR unknown command 'TOUCH'`.

Is there a plan to upgrade the Redis version on windows behind 3.0.503?

mgravell:
The problem there is that redis-64 was abandoned sometime around V3.

The simple fix here is to just to use server version detection in the code, and if TOUCH isn't expected : either don't test it, or assert that it fails in this way. Lots of tests do similar.

For the other fix on Windows, though is for us to do one of 3 things:

- use an external Linux server
- use WSL/WSL2
- use the more recent code that fell out of redis-64 (can't recall the name off-hand, but it exists, but isn't totally F/OSS)

Probably best to just use server version detection for now :)

gkorland:
>Probably best to just use server version detection for now :) 
Ok I'll do it, BTW there are much newer versions for Redis on Windows 4.x and I saw there is already a work in progress to support 5.0.6

mgravell:
are you referrring to Memurai? or...?

gkorland:
https://github.com/tporadowski/redis

gkorland:
@mgravell added the server feature test

gkorland:
@mgravell thanks,

Issue #1058, opened 2019-02-12
IServer.Time();


Tzott:
The return time of the time command to the current server is 8 hours different from the server time.

Tzott:
It's like a time zone problem.

mgravell:
Dammit, that's annoying and problematic. There are times when it is hugely
inconvenient to develop in a GMT area - timezone problems can whizz past
you! I'll see what I can find out.

mgravell:
Can you check your server settings? We've spoken to antirez, who has confirmed that the time reported is UTC; our code treats it as UTC. Is it possible that either your server or client has their time incorrect? i.e. instead of being in the correct timezone, they've been set in the wrong time with a UTC shift?

mgravell:
Perhaps also: can you provide an example that shows the time being reported incorrectly? It *is* possible that you just need a `.ToLocalTime()` here!

Tzott:
My sercer time zone configuration is UTC+8.

The `IServer.Time().Kind` is `DateTimeKind.Utc`.  

But we used to use local time, as `DateTime.Now.Kind` is `Local`.

That's why I always thought there was an eight-hour gap in time at first.

As you said, I just need to use `.ToLocalTime()`, I'll get the right time.

mgravell:
so: what I'll do is convert this into a documentation issue, to make it clear in the intellisense that the time reported is in UTC, and to call `ToLocalTime()` if the caller wants local time. I think that's the best thing to do here.

Tzott:
I also think so.

MZar2019:
Using StackExchange.Redis 2.0.601, IServer.Time() and IServer.TimeAsync both return the Utc time from the server, however, the DateTime.Kind for TimeAsync() is **_Local, which is incorrect_**, while DateTime.Kind for Time() is Utc, which is correct.

mgravell:
They both use the exact same processor; the code here is:

``` c#

        public DateTime Time(CommandFlags flags = CommandFlags.None)
        {
            var msg = Message.Create(-1, flags, RedisCommand.TIME);
            return ExecuteSync(msg, ResultProcessor.DateTime);
        }

        public Task<DateTime> TimeAsync(CommandFlags flags = CommandFlags.None)
        {
            var msg = Message.Create(-1, flags, RedisCommand.TIME);
            return ExecuteAsync(msg, ResultProcessor.DateTime);
        }
```

do you have a repro that demonstrates this difference?

[...]

PR #1311, opened 2019-12-23
Extending description of the IServer.Time and IServer.TimeAsync

DillonAd:
This PR addresses #1058 by changing the XML documentation in the `IServer` interface to specify the timezone returned by the `Time` and `TimeAsync` methods (UTC) and what to do to get the local time from the value returned from those methods.


mgravell:
Fair enough, merged thanks

PR #1313, opened 2019-12-25
add support for netcoreapp3.1 and build on none windows

gkorland:


mgravell:
A couple of questions:

1. Is there any reason to add the extra TFM? It should already work just fine on netcoreapp3.1 - the reasons for adding extra TFMs (upwards) are usually to exploit additional features - we don't seem to do that here, so I'm wondering if this is actually necessary
2. Is there a reason to limit the TFM to windows only? I'm unclear what the OS restriction is for in this case

gkorland:
I couldn't find another way to make it compile on Ubuntu with netcoreapp3.1.
Without these changes I'm getting these errors:
```bash 
$ dotnet build
...
/snap/dotnet-sdk/57/sdk/3.1.100/Microsoft.Common.CurrentVersion.targets(1175,5): error MSB3644: The reference assemblies for .NETFramework,Version=v4.7.2 were not found. To resolve this, install the Developer Pack (SDK/Targeting Pack) for this framework version or retarget your application. You can download .NET Framework Developer Packs at https://aka.ms/msbuild/developerpacks [/home/guy/dotnetworkspace/StackExchange.Redis/tests/BasicTest/BasicTest.csproj]
```

mgravell:
Right, that was the context I was missing (on the platform build bit) - it seems obvious in hindsight, but it didn't click at the time - probably would have been worth clarifying in the original description :)

So; sounds OK to me, and we probably should indeed make sure we test on the current runtime. @NickCraver - you know a lot more about the CI implications than me - any views here?

NickCraver:
Will take a peek at this tonight - it'd be better to use https://www.nuget.org/packages/Microsoft.NETFramework.ReferenceAssemblies/ probably and be buildable on all but only test on the appropriate for the platform.

Cleaning up after holidays here, but should be able to poke some afterwards.

NickCraver:
Hey all, I threw up a simpler PR on how to achieve this in https://github.com/StackExchange/StackExchange.Redis/pull/1316 :)

NickCraver:
I'm going to close this out since #1316 is the correct path there - let's discuss any issues you have building there, but it appears to be missing runtimes (which we require) only after the fixes!

PR #1316, opened 2019-12-31
Fix builds for all packages on Linux

NickCraver:
Note that we're only building `StackExchange.Redis` and `NRediSearch` in the CI build so this wasn't apparent as an issue, but it's also easily remedied. Cleaning up the xUnit MyGet feed as well since it was an intermittent issue earlier in testing this...and turns out we don't use any betas or need them anymore, so why not simplify anyhow.

Overall, this is a simpler alternative to #1313 and more build simplification/cleanup:

Total changes:
- Moves to 3.x build SDK (just to update)
  - Moves to `Visual Studio 2019` AppVeyor build image, for 3.x support (removes need to install SDK)
  - Rolls forward to latest version
- Moves to `Build.csproj` for building
  - Simplifies `build.ps1`
- Adds `NRediSearch` to the test suite (on Linux in CI, since Windows is against 3.x)
  - Adds `RedisFeatures.Module` (against v 4.0.0) to guard tests against
  - Makes `NRediSearch.Tests` reference `StackExchange.Redis.Tests` and support test skipping
- Makes copyright year range dynamic for the library
- Moves `LangVersion` to latest and in one place (root `Directory.Build.props`)
- Uses `Microsoft.NETFramework.ReferenceAssemblies` to build `net4x` on Linux
- Uses `IsPackable` to default to excluded then reverses to `true` only for the `src/` directory
- Removes `xUnit` MyGet feed, since we're on stable builds now
- Adds new files (and corrects old) in `.sln`
- Removes `LibTargetFrameworks` to simplify since it's less complex given the above changes

cc @mgravell for eyes - this should be all set now!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,4 @@ t8.shakespeare.txt
 launchSettings.json
 *.vsp
 *.diagsession
+TestResults/
--- a/Build.csproj
+++ b/Build.csproj
@@ -1,8 +1,6 @@
 <Project Sdk="Microsoft.Build.Traversal/2.0.24">
   <ItemGroup>
     <ProjectReference Include="src\**\*.csproj" />
-  </ItemGroup>
-  <ItemGroup Condition="$(Packing) != 'true'">
     <ProjectReference Include="tests\**\*.csproj" />
     <ProjectReference Include="toys\**\*.csproj" />
   </ItemGroup>
--- a/Build.csproj
+++ b/Build.csproj
@@ -1,0 +1,9 @@
+<Project Sdk="Microsoft.Build.Traversal/2.0.24">
+  <ItemGroup>
+    <ProjectReference Include="src\**\*.csproj" />
+  </ItemGroup>
+  <ItemGroup Condition="$(Packing) != 'true'">
+    <ProjectReference Include="tests\**\*.csproj" />
+    <ProjectReference Include="toys\**\*.csproj" />
+  </ItemGroup>
+</Project>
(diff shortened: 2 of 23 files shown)


gkorland:
@NickCraver thanks! it seems like it did solve the build issue.
But, now I'm getting this error `dotnet test` 
```
Testhost process exited with error: It was not possible to find any compatible framework version
The framework 'Microsoft.NETCore.App', version '2.1.0' was not found.
  - The following frameworks were found:
      3.1.0 at [/snap/dotnet-sdk/57/shared/Microsoft.NETCore.App]
You can resolve the problem by installing the specified framework and/or SDK.
The specified framework can be found at:
  - https://aka.ms/dotnet-core-applaunch?framework=Microsoft.NETCore.App&framework_version=2.1.0&arch=x64&rid=ubuntu.19.10-x64
. Please check the diagnostic logs for more information.
```

NickCraver:
@gkorland where are you running `dotnet test`? Regardless of directory, you'll still need to have *each* of the runtimes installed to test against, e.g. 2.1.x here. While the SDK *builds* previous versions, it cannot test them. This isn't something we can or should change - to test you'll need to have each test target version installed.

This happens on the CI build agents as well, the versions are just already in the images (since testing against various versions is common).

NickCraver:
@mgravell when you're back in action - this is good for a look.

All changes list above, but the big ones that matter mot:
- Simplifies the build
- Builds all projects not just the lib to ensure no breaks for any PR
- Tests NRediSearch in the build whenever possible

This doesn't fix any of the current test stability, just moving cheese and improving how we build.

guykorlandredis:
@NickCraver after this PR trying to build locally using `dotnet build`, I started to the get the following error.
```
Microsoft (R) Build Engine version 16.4.0+e901037fe for .NET Core
Copyright (C) Microsoft Corporation. All rights reserved.

MSBUILD : error MSB1011: Specify which project or solution file to use because this folder contains more than one project or solution file.
```

PR #1318, opened 2020-01-06
Tests/CI: Explore a Redis-on-Windows 4.0 changeover

NickCraver:
From https://github.com/tporadowski/redis - let's give it a test run spin just to check things out :)


NickCraver:
Fixed on the build side by beefing up Appveyor - nuking this!

PR #1332, opened 2020-01-23
Add Condition.SortedSetLengthEqual with min and max score

### LINKED ISSUE #1331: Transaction Condition.SortedSetLengthEqual with min and max score
Looking at source, `Condition.SortedSetLengthEqual` transaction condition is currently using `ZCARD key` to determine sorted set's length.

It would be nice to have an overload of this method that accepts minimum and maximum score range, using `ZCOUNT key min max`. This would look at the number of members with score in the given range, not the cardinality of the entire set.

phosphene47:
Resolves #1331

### DIFF
--- a/src/StackExchange.Redis/Condition.cs
+++ b/src/StackExchange.Redis/Condition.cs
@@ -232,6 +232,15 @@ namespace StackExchange.Redis
         public static Condition SortedSetLengthEqual(RedisKey key, long length) => new LengthCondition(key, RedisType.SortedSet, 0, length);
 
         /// <summary>
+        /// Enforces that the given sorted set contains a certain number of members with scores in the given range
+        /// </summary>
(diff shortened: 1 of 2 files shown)


mgravell:
Perfectly fine with the concept - looks fine. Code looks OK at first
eyeball. Will need to check it in more detail, but: passes the sniff test.
Thanks, good job.

Can I confirm that you're able and willing (and thus doing so) to freely
contribute this code in line with the projects licence, assigning any and
all IP/copyright in the code here to the project?

(Note to self: we need to get a CLA-bot!)

phosphene47:
> Can I confirm that you're able and willing (and thus doing so) to freely contribute this code in line with the projects licence, assigning any and all IP/copyright in the code here to the project

Yes.

mgravell:
Merged, thanks

PR #1351, opened 2020-02-14
Improve pub/sub order documentation

vcsjones:
First, the link from Basics doesn't work. Second, the docs code samples appear backward, in addition to the casing of the `Subscribe` method being off.

### DIFF
--- a/docs/Basics.md
+++ b/docs/Basics.md
@@ -107,7 +107,7 @@ sub.Publish("messages", "hello");
 
 This will (virtually instantaneously) write `"hello"` to the console of the subscribed process. As before, both channel-names and messages can be binary.
 
-Please also see [Pub / Sub Message Order](PubSubOrder) for guidance on sequential versus concurrent message processing.
+Please also see [Pub / Sub Message Order](PubSubOrder.md) for guidance on sequential versus concurrent message processing.
 
 Accessing individual servers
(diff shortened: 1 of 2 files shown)


NickCraver:
Good one the fix - need to revert the link bit though, unfortunately there's an either/or with GitHub or the site working, e.g. https://stackexchange.github.io/StackExchange.Redis/PubSubOrder vs. https://stackexchange.github.io/StackExchange.Redis/PubSubOrder.md which we'd link to here :( It's possible I'm an idiot but it seems like GitHub's publishing is anti-doing-one-thing here

docs/Basics.md
```diff
@@ -107,7 +107,7 @@ sub.Publish("messages", "hello");
 
 This will (virtually instantaneously) write `"hello"` to the console of the subscribed process. As before, both channel-names and messages can be binary.
 
-Please also see [Pub / Sub Message Order](PubSubOrder) for guidance on sequential versus concurrent message processing.
+Please also see [Pub / Sub Message Order](PubSubOrder.md) for guidance on sequential versus concurrent message processing.
```
NickCraver:
```suggestion
Please also see [Pub / Sub Message Order](PubSubOrder) for guidance on sequential versus concurrent message processing.
```

PR #1362, opened 2020-02-24
Fix: CommandFlags PreferSlave and PreferMaster are handled incorrectly for clusters. 

aravindyeduvaka:
Fixes a bug where the behavior of PreferSlave and PreferMaster are identical to DemandSlave and DemandMaster respectively.

### DIFF
--- a/src/StackExchange.Redis/ServerSelectionStrategy.cs
+++ b/src/StackExchange.Redis/ServerSelectionStrategy.cs
@@ -293,12 +293,14 @@ namespace StackExchange.Redis
                         return FindSlave(endpoint, command) ?? Any(command, flags);
                     case CommandFlags.PreferSlave:
                         testing = FindSlave(endpoint, command);
+                        if (testing == null) testing = FindMaster(endpoint, command);
                         if (testing != null) return testing;
                         break;
                     case CommandFlags.DemandMaster:
                         return FindMaster(endpoint, command) ?? Any(command, flags);
                     case CommandFlags.PreferMaster:
                         testing = FindMaster(endpoint, command);
+                        if (testing == null) testing = FindSlave(endpoint, command);
                         if (testing != null) return testing;
                         break;
                 }


mgravell:
I'd have to check, but isn't this already covered by the line beneath:

``` c#
if (endpoint.IsSelectable(command)) return endpoint;
```

?

Not at a PC right now to check; is this making a difference in behaviour for yo

aravindyeduvaka:
@mgravell, Yeah it seems like IsSelectable does handle it. 

I was trying to trace a bug where people are getting multiple MOVED errors after a disconnect with slaves on clustered configurations. 

I think I found the problem - it seems that SE.Redis is not sending READONLY command to Server (on Clustered slave) when reconnecting after Server goes down (or rather this command is being sent before AUTH, so its not reaching Redis Server.) I was able to verify this successfully. 

This line which does the READONLY is run before the line below whichwrites the message (containing AUTH during reconnects). 

https://github.com/StackExchange/StackExchange.Redis/blob/bb981525a99299d676a99d85725def8c7e1a7a30/src/StackExchange.Redis/PhysicalBridge.cs#L1203 

https://github.com/StackExchange/StackExchange.Redis/blob/bb981525a99299d676a99d85725def8c7e1a7a30/src/StackExchange.Redis/PhysicalBridge.cs#L1236

Also since the command is Fire and Forget, we never get NoAuth error. 
https://github.com/StackExchange/StackExchange.Redis/blob/ef5af00d5febd0dd3e4daca397fae84737d44277/src/StackExchange.Redis/PhysicalConnection.cs#L37

Do you have any suggestions on how this should be fixed.

NickCraver:
@aravindyeduvaka ooh that is a very nice find! This may explain a few unrecoverable failures in the queue - we'll chat tomorrow for sure.

aravindyeduvaka:
Going to close this and open a new PR with the correct fix.

PR #1376, opened 2020-03-08
[WIP] Add cancellation support to `IDatabaseAsync()`

longfin:
I'm adding to cancellation support to `IDatabaseAsync` interface by creating an overloaded method that received `CancellationToken`.

- 🚧  I'm working on drafts until I can get the review, but please leave comments from now on if you have any concerns. (I'll following upstream by rebasing during drafting.)
- 👶  Because I'm new to contributing to this project, there may be things I don't know. (e.g. design rationale, conventions) References are also welcome.

### DIFF
--- a/src/StackExchange.Redis/PhysicalBridge.cs
+++ b/src/StackExchange.Redis/PhysicalBridge.cs
@@ -181,6 +181,13 @@ namespace StackExchange.Redis
         }
 
         public ValueTask<WriteResult> TryWriteAsync(Message message, bool isSlave)
+            => TryWriteAsync(message, isSlave, default);
+
+        public ValueTask<WriteResult> TryWriteAsync(
+            Message message,
+            bool isSlave,
+            CancellationToken cancellationToken
+        )
         {
             if (isDisposed) throw new ObjectDisposedException(Name);
             if (!IsConnected) return new ValueTask<WriteResult>(QueueOrFailMessage(message));
@@ -188,7 +195,7 @@ namespace StackExchange.Redis
             var physical = this.physical;
             if (physical == null) return new ValueTask<WriteResult>(FailDueToNoConnection(message));
 
-            var result = WriteMessageTakingWriteLockAsync(physical, message);
+            var result = WriteMessageTakingWriteLockAsync(physical, message, cancellationToken);
             LogNonPreferred(message.Flags, isSlave);
             return result;
         }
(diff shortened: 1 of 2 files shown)


mgravell:
There's lots of things to consider in such a thing, but the biggest are:

- how is it exposed to the consumer?
- what actually happens if something is cancelled?
- before it is written and after it is written
- and how does this impact what is on the connection / other work?

I would suggest it is a good idea to figure out the answers to these before getting too deep.

longfin:
> how is it exposed to the consumer?

I thought of adding an interface like this at the end of this work:...

```csharp
        // Previous API
        Task<RedisValue[]> StringGetAsync(RedisKey[] keys, CommandFlags flags = CommandFlags.None) => StringGetAsync(keys, flags, default);

        // Added API
        Task<RedisValue[]> StringGetAsync(RedisKey[] keys, CancellationToken cancellationToken, CommandFlags flags = CommandFlags.None);
```

To do this, I needed to modify `RedisBase` and `ServerEndPoint`, but before proceeding, I would like to add the innermost interfaces they depend on.


> what actually happens if something is cancelled?

Currently, it will bypass [`OperationCanceledException`](https://docs.microsoft.com/en-us/dotnet/api/system.operationcanceledexception?view=netstandard-2.0) from [`PipeWriter.FlushAsync()`](https://docs.microsoft.com/en-us/dotnet/api/system.io.pipelines.pipewriter.flushasync?view=dotnet-plat-ext-3.1&viewFallbackFrom=netstandard-2.0). And, affected interfaces also throw `OperationCanceledException` when something is canceled. 


> before it is written and after it is written 

- If a task canceled before written, `PipeWriter.FlushAsync()` will fail and `OperationCanceldException` will be raised.
- If a task canceled after written, Task completes without any exception.

https://github.com/StackExchange/StackExchange.Redis/pull/1376/commits/7510c5494fa5872ebaddf24f40c5dcba0f232271 is a very small piece of the total work. so the remaining tasks may depend on how we define the behavior. (My main purpose in this PR is just applying the cancellation pattern in .NET to this library. 😅 )


> and how does this impact what is on the connection / other work?

Basically, I'll feed `CancellationToken` to .NET API(e.g. `PipeWriter`) that can receive it for the cancellation mechanism. it means that the cancellation scope is divided by the .NET APIs.

For example, I guess https://github.com/StackExchange/StackExchange.Redis/pull/1376/commits/7510c5494fa5872ebaddf24f40c5dcba0f232271 will only impact on its function invocation scope due to it only relies on `PipeWrite.Flush()`'s cancellation.


> I would suggest it is a good idea to figure out the answers to these before getting too deep.

I totally agree 😄 . it's a reason that I've opened this PR early.

NickCraver:
We've been looking at cancellation a bit, but I think we'd want to go a different direction here. The proposal of effectively doubling the async API surface area is a hard sell for maintenance so unlike to go that route. The main questions aren't really around implementing the API (which this PR does), but _what we do with a cancelled request_, which we should probably discuss in an issue to figure out. Do we not send it? Do we toss any result? Do we not retry it? What happens to items after it? Do we need a cancellation token ref on every message? etc. Figuring that stuff out is probably best done not in a flowing PR :)

NickCraver:
Going to close this out because of the notes above - if we do this, it'll be far more global and breaking.

PR #1379, opened 2020-03-10
fix forgotten clone config attribute CheckCertificateRevocation

BLun78:
Hi,

i have make a pull request: @https://github.com/StackExchange/StackExchange.Redis/pull/1234
but it has a Bug. When the Config is Clone i have forgot the copy for the attribute CheckCertificateRevocation.
Now this request do that. Sorry for my fail.


NickCraver:
hmm this breaks tests because we null coalesce to true in the getter so when comparing fields they don't match. Taking a peek.

PR #1374, opened 2020-03-06
PhysicalBridge always returns NoConnection after exception during message write / connection flush

hamish-omny:
We recently encountered an issue where we repeatedly (for several hours) got the exception "No connection is available to service this operation". Using version 2.0.601 (with `AbortOnConnectFail` set to false)  running on asp.net core, in Azure and connecting to an Azure Redis instance. This was only resolved by a restart of the instance, other application instance could connect to the redis instance during this time

The full exception was as follows
```
No connection is available to service this operation: SMEMBERS XXXX; An existing connection was forcibly closed by the remote host; IOCP: (Busy=1,Free=999,Min=2,Max=1000), WORKER: (Busy=0,Free=32767,Min=2,Max=32767), Local-CPU: n/a
SocketFailure on xxx.redis.cache.windows.net:6380/Interactive, Idle/Faulted, last: GET, origin: ReadFromPipe, outstanding: 3, last-read: 0s ago, last-write: 0s ago, unanswered-write: 0s ago, keep-alive: 10s, state: ConnectedEstablished, mgr: 8 of 10 available, in: 0, last-heartbeat: 0s ago, last-mbeat: 0s ago, global: 0s ago, v: 2.0.601.3402
```
I was able to reproduce this situation with a small console app that repeatably writes and reads keys and while this is running i kill the connection from the redis cli `client kill type normal`. This sometimes causes an exception from within one of the `WriteMessage*` methods to be thrown. This in turns causes the `_activeMessage` field in `PhysicalBridge` to not be set back to null and from that point forward the compare exchange logic in `WriteMessageInsideLock` bridge will always return "NoConnection". At this point any request for the `ConnectionMultiplexer` fails and it won't recover (at least in my testing)

I note there have been a number of issues opened that have have a similar error message and symptoms. Looking at the commit history this [commit](https://github.com/StackExchange/StackExchange.Redis/commit/687394109f9b7faa6a25851b3744417bfae5cd51#diff-e9ff80307b03460becea3fe93c80bdf7) changed from clearing `_activeMessage` in `ReleaseSingleWriterLock` which was called from the finally block to the current approach, although obviously along with a significant number of other changes. Version 2.0.519 doesn't include this change and that might explain why people have had some success rolling back to it.

This PR moves back to always clearing `_activeMessage` within the finally block of the three relevant write methods, ensuring it's cleared even in the case of an exception occuring. Happy to try write some tests for this, just wanted to validate the change is acceptable first

[...]

PR #1381, opened 2020-03-12
Let uses explicty configure ServerSelectionStrategy.ServerType in cases they are using a Redis Cluster.

dceravigupta:
We are using 1.2.6 version of StackExchange.Redis with Azure Redis cache and occasionally we get bunch of "No connection is available to service this operation" and sometime it goes into a state where it doesn't ever recover from this issue and we are forces to reboot those VMs to fix the problem.

I was looking at [1374](https://github.com/StackExchange/StackExchange.Redis/pull/1374) and using the same setup but using code of 1.2.6 SDK version with actual Azure Redis cache. I was able to repo the problem and noticed that whenever it happens, ConnectionMultiplexer -> ReconfigureAsyc function fails while PINGing Redis Server and as a result the clusterCount remain 0.

image

Now since we are not sure what type of Redis we are talking to, we default to ServerSelectionStrategy.ServerType i.e. ServerType.Standalone. I didn't went that deep into code, but somehow the endpoint is able to recover but we are still stuck with ServerSelectionStrategy.ServerType as Standalone due to which all subsequent calls fails with "No connection is available to service this operation". Given the part of ReconfigureAsyc I'm talking about is same in latest version as well, I'm pretty sure the problem will exists in there as well. Happy to try it out with latest version as well, just wanted to validate the change is acceptable first.

image

By giving option to explicitly specify, I'm thinking that way in case of failure, we don't have to assume about the type of Redis we are talking to and hence more deterministic behavior.

### DIFF
--- a/src/StackExchange.Redis/ConnectionMultiplexer.cs
+++ b/src/StackExchange.Redis/ConnectionMultiplexer.cs
@@ -1675,6 +1675,10 @@ namespace StackExchange.Redis
                         {
                             ServerSelectionStrategy.ServerType = ServerType.Twemproxy;
                         }
+                        if (RawConfig.Proxy == Proxy.Cluster)
+                        {
+                            ServerSelectionStrategy.ServerType = ServerType.Cluster;
+                        }
                         else if (standaloneCount == 0 && sentinelCount > 0)
                         {
                             ServerSelectionStrategy.ServerType = ServerType.Sentinel;
--- a/src/StackExchange.Redis/Enums/Proxy.cs
+++ b/src/StackExchange.Redis/Enums/Proxy.cs
@@ -12,6 +12,10 @@
         /// <summary>
         /// Communication via <a href="https://github.com/twitter/twemproxy">twemproxy</a>
         /// </summary>
-        Twemproxy
+        Twemproxy,
+        /// <summary>
+        /// Direct communication to the redis cluster
+        /// </summary>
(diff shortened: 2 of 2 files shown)


NickCraver:
I get the issue, but I'm not sure this is the right place to plugin a change - the main thing is _it's not a proxy_. `None` is correct here. It confuses the abstraction/type to have `Cluster` there, perhaps we can find a better place to force the fact we're talking to a cluster even though it may be failing when we first connect, etc.

Alternatively, if we're always re-evaluating this on reconnect (we should be) and the issue only happens _during failure_, then perhaps what we need is better "hey...assume it's still a cluster" logic in our retry mechanism. I'd say it's a safe assumption to assume someone didn't change the server _types_ while we disconnected, in 99.99% of cases. And even if they did, we'd just be making bad assumptions until reconnect. And honestly, in that low-percentage case they'd have to shut down the works to change to/from cluster, etc.

mgravell:
What concerns me here is that regular cluster is *not* a proxy; however there *is* a cluster proxy that *is* a proxy, and I think this would cause huge confusion. I think we should reserve this change for https://github.com/RedisLabs/redis-cluster-proxy

NickCraver:
Closing this out since it's not the right approach - the #1374 change should help here though. We'll take a look at assuming last server state as well, just not in this manner - hopefully the above explains why!

PR #1390, opened 2020-03-16
Fix documentation URLs

Badkoubehei:
Added .md extension to links

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -30,19 +30,19 @@ PM> Install-Package StackExchange.Redis
 Documentation
 ---
 
-- [Basic Usage](Basics) - getting started and basic usage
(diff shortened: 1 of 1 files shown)


NickCraver:
I get the confusion, but this _breaks_ the documentation URLs. They are meant to be accessed from https://stackexchange.github.io/StackExchange.Redis/. If you try adding a `.md` there, you'll notice it'll only result in 404s :)

Badkoubehei:
Oh! Sorry. You are right. My mistake was accessing the documentation from https://github.com/StackExchange/StackExchange.Redis/blob/master/docs/index.md instead of github.io. Apologies :)

NickCraver:
No worries, thanks for the effort! It's still appreciated :)

PR #1400, opened 2020-03-21
Fix up boadcasts, tests, and logging across the board

NickCraver:
This changes does the following:
- Moves the broadcast to be only before the master reconfiguration to both before *and* after - a fix following https://github.com/StackExchange/StackExchange.Redis/commit/88dcf0c989b25623fb8ffb1f4d48c24593246cfe to fix the gap.
- Tweaks tests by overall lessening runtime (and thus build server time). Overall, fixes a few static timeouts to be conditional (so they short circuit faster if met), and brings some tests that were some variant of the above into RELEASE since they're safe now.
- Changes `UntilCondition` to take a `TimeSpan`, just because clearer. Even though I IntelliSense completed `.FromMinutes()` earlier and watched it like an idiot for a while...I stand by this decision!
- Locks the `ITestOutputHelper` writer because...that was jacked up in races:

off to the races

------

Note: a lot of the test changes are just optimizations to delays which allow longer but short-circuit sooner. The important changes are in broadcast and locking around the test runner. I can think of downsides to neither, but want some @mgravell eyes. This should resolve a lot of flaky-ness with local (and build agent) tests. Not all of it, but a lot of it!

### DIFF
--- a/tests/StackExchange.Redis.Tests/Failover.cs
+++ b/tests/StackExchange.Redis.Tests/Failover.cs
@@ -182,6 +182,8 @@ namespace StackExchange.Redis.Tests
                     Assert.Equal(secondary2.EndPoint, db2.IdentifyEndpoint(key, CommandFlags.DemandSlave));
                 }
 
+                await UntilCondition(TimeSpan.FromSeconds(10), () => !primary.IsSlave && secondary.IsSlave, waitPerLoop: TimeSpan.FromMilliseconds(50));
+
                 Assert.False(primary.IsSlave, $"{primary.EndPoint} should be a master.");
                 Assert.True(secondary.IsSlave, $"{secondary.EndPoint} should be a slave.");
 
--- a/tests/StackExchange.Redis.Tests/Failover.cs
+++ b/tests/StackExchange.Redis.Tests/Failover.cs
@@ -104,7 +104,7 @@ namespace StackExchange.Redis.Tests
             }
         }
 
-#if DEBUG
+
         [Fact]
         public async Task DeslaveGoesToPrimary()
         {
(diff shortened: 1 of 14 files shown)


ejsmith:
Looking good!

NickCraver:
@mgravell Should be good for review - this isn't perfect but does fix more of the flaky. I propose we merge this so I can get the other PRs using fewer builds to get green - we're eating build minutes on flaky nonsense at the moment. This also just generally reduces the arbitrary pausing (instead spin checking the condition we want with a timeout) to eat fewer build minutes as well. Our builds basically doubled in expense when Sentinel landed due to these pauses. I should have paid more attention then.

tests/StackExchange.Redis.Tests/Failover.cs
```diff
@@ -281,10 +287,19 @@ public async Task SubscriptionsSurviveMasterSwitchAsync()
 
                     Assert.True(a.GetServer(TestConfig.Current.FailoverMasterServerAndPort).IsSlave, $"A Connection: {TestConfig.Current.FailoverMasterServerAndPort} should be a slave");
                     Assert.False(a.GetServer(TestConfig.Current.FailoverSlaveServerAndPort).IsSlave, $"A Connection: {TestConfig.Current.FailoverSlaveServerAndPort} should be a master");
+                    await UntilCondition(TimeSpan.FromSeconds(10), () => b.GetServer(TestConfig.Current.FailoverMasterServerAndPort).IsSlave).ForAwait();
                     var sanityCheck = b.GetServer(TestConfig.Current.FailoverMasterServerAndPort).IsSlave;
                     if (!sanityCheck)
                     {
-                        Skip.Inconclusive("Not enough latency.");
+                        Log("FAILURE: B has not detected the topology change.");
+                        foreach (var server in b.GetServerSnapshot().ToArray())
```
mgravell: why `ToArray()`? if this is because of spans+await, a: it'll *probably* work anyway, but if it doesn't (sometimes the compiler can, sometimes it can't), then suggest you create a static local method just for the log, that takes `b` and whatever `Log` needs (the `this` instance?)

mgravell:
oh, ignore my change; didn't realize that was a test; in that case: don't care about a `ToArray`; approved!

tests/StackExchange.Redis.Tests/Failover.cs
NickCraver: Yes indeed, await/span won't compile just went simple for test case!

PR #1401, opened 2020-03-22
Actions: Colors Test

NickCraver:
Let's get more info on this issue!

This attempts to add colorization to the build segment - researching the `dotnet test` issue.


NickCraver:
Whelp, that doesn't work - was a test for https://github.com/microsoft/vstest/issues/2370. Leaving in draft here while we try and chase down where colors aren't happening.

ejsmith:
Did you see https://github.com/dotnet/runtime/issues/33980?  Upvotes would be good.

PR #1406, opened 2020-03-25
ThreadTheft.md add 1.2.6

pengweiqhca:


NickCraver:
There are 2 issues here:
1. The 1.2.6 issue is completely different - the only course there is to actually upgrade (so the don't don't help users here - we don't want to point them at this version at all really)
2. The `.md` links actually break the https://stackexchange.github.io/StackExchange.Redis/ site - them not being there is very intentional :)

PR #1410, opened 2020-03-28
allow auth with user (redis 6)

mgravell:
- adds new config options for cn+pw authentication against redis 6
- currently no inbuilt test, because we need to figure out a redis 6 testing setup
- no support for extended `ACL` command set yet
  1. because they aren't entirely nailed down yet, and
  2. because they aren't critical path to being able to use it, because
  3. everything still works fine via all-encompassing `Execute` fallback

### DIFF
--- a/src/StackExchange.Redis/ConfigurationOptions.cs
+++ b/src/StackExchange.Redis/ConfigurationOptions.cs
@@ -76,6 +76,7 @@ namespace StackExchange.Redis
                 HighPrioritySocketThreads = "highPriorityThreads",
                 KeepAlive = "keepAlive",
                 ClientName = "name",
+                User = "user",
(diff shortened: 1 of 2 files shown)


NickCraver:
For local, you can change the 5 to a 6 in the `tests\RedisConfigs\Dockerfile` up top to quickly get a 6.x (5.9 really) running:
```diff
- FROM redis:5
+ FROM redis:6.0-rc
```
To rebuild:
```bash
docker-compose build --no-cache
docker-compose up
```

NickCraver:
FWIW Did a quick run against local docker 6 as above: `dotnet test .\Build.csproj -c Release` and only had 1 flaky failure...6 is looking good. Probably viable to kick it up.

PR #1417, opened 2020-04-01
Add the ability to remap cluster nodes responses, etc to different hosts/ports

N6UDP:
This is still a bit of WIP but it works and allows you to have TLS with client certificate auth for clients to a cluster behind a load balancer with the cluster talking to itself in the clear on a private network.  I'm open to suggestions as to a better way to do this / if I've missed something (and to name changes/description + doc update requests/tests/etc).

The idea is if you have a cluster with a layer-4 load balancer configured something like this: 
diagram
e.g. one port (6380 in this case) for discovery (green) talks to any node through stunnel (or a local haproxy/traefik/whatnot proxying to redis) and then a static per-host port mapping for each cluster node (black) through stunnel as well.  The cluster replicates (blue) and does cluster communications in the clear (not shown) via the standard ports.

You could then take a `cluster nodes` output like so 
```
df7927b560da64a5cecc9e95b066640733d0ac91 10.0.0.10:6379@16379 slave 1d9bb9764d233ceda62801b015de78bb04b99a6f 0 1585698945482 18 connected
1d9bb9764d233ceda62801b015de78bb04b99a6f 10.0.0.11:6379@16379 myself,master - 0 1585698945000 18 connected 0-5460
b855ca587f6ff7f223271c1a25f4fdf9e9840074 10.0.0.12:6379@16379 master - 0 1585698946485 2 connected 5461-10922
f778573815de8fb3921ff420f80d288a08eeccef 10.0.0.13:6379@16379 slave b855ca587f6ff7f223271c1a25f4fdf9e9840074 0 1585698944981 2 connected
a303387b3e0129ab405b1d0bd92aef99d8b94159 10.0.0.14:6379@16379 master - 0 1585698946000 3 connected 10923-16383
c174a32d94862a0ec40b8a8fe4722eb688794db6 10.0.0.15:6379@16379 slave a303387b3e0129ab405b1d0bd92aef99d8b94159 0 1585698946084 3 connected
```
and transform it client side to:
```
df7927b560da64a5cecc9e95b066640733d0ac91 myexternal.host.name:6310@16379 slave 1d9bb9764d233ceda62801b015de78bb04b99a6f 0 1585698945482 18 connected
1d9bb9764d233ceda62801b015de78bb04b99a6f myexternal.host.name:6311@16379 myself,master - 0 1585698945000 18 connected 0-5460
b855ca587f6ff7f223271c1a25f4fdf9e9840074 myexternal.host.name:6312@16379 master - 0 1585698946485 2 connected 5461-10922
f778573815de8fb3921ff420f80d288a08eeccef myexternal.host.name:6313@16379 slave b855ca587f6ff7f223271c1a25f4fdf9e9840074 0 1585698944981 2 connected
a303387b3e0129ab405b1d0bd92aef99d8b94159 myexternal.host.name:6314@16379 master - 0 1585698946000 3 connected 10923-16383
c174a32d94862a0ec40b8a8fe4722eb688794db6 myexternal.host.name:6315@16379 slave a303387b3e0129ab405b1d0bd92aef99d8b94159 0 1585698946084 3 connected
```

I have handled remapping for cluster responses and `moved`.  I also added it to the sentinel related responses but haven't confirmed the behavior there.  I wrote a quite test application with a cluster w/load balancer configured as per above and validated it works as expected including in the event of master failures etc.

```csharp
var remapping = new Dictionary<string, string>();
for(int i = 10; i <= 15; i++)
{
    remapping.Add("10.0.0." + i + ":6379", "my.host.name:63" + i);
}
{
    ClientName = "testclient-" + Environment.MachineName,
    EndPoints = { { "my.host.name:6380" } },
    Ssl = true,
    AbortOnConnectFail = true, //false 
    AllowAdmin = true,
    AddressRemapping = remapping,
    SslProtocols = System.Security.Authentication.SslProtocols.Tls12, 
    SslHost = "my.host.name"
};

options.CertificateSelection += delegate
{
...return a client cert here
};
ConnectionMultiplexer muxer = ConnectionMultiplexer.Connect(options);
```

Thoughts? Total insanity?

### DIFF
--- a/src/StackExchange.Redis/ClusterConfiguration.cs
+++ b/src/StackExchange.Redis/ClusterConfiguration.cs
@@ -157,10 +157,12 @@ namespace StackExchange.Redis
         private readonly Dictionary<EndPoint, ClusterNode> nodeLookup = new Dictionary<EndPoint, ClusterNode>();
 
         private readonly ServerSelectionStrategy serverSelectionStrategy;
-        internal ClusterConfiguration(ServerSelectionStrategy serverSelectionStrategy, string nodes, EndPoint origin)
+        
+        internal ClusterConfiguration(ServerSelectionStrategy serverSelectionStrategy, string nodes, EndPoint origin, Dictionary<string, string> addressRemapping)
         {
             // Beware: Any exception thrown here will wreak silent havoc like inability to connect to cluster nodes or non returning calls
             this.serverSelectionStrategy = serverSelectionStrategy;
+            this.AddressRemapping = addressRemapping;
             Origin = origin;
             using (var reader = new StringReader(nodes))
             {
@@ -220,6 +222,8 @@ namespace StackExchange.Redis
         /// The node that was asked for the configuration
         /// </summary>
         public EndPoint Origin { get; }
+
+        internal readonly Dictionary<string, string> AddressRemapping;
 
         /// <summary>
         /// Obtain the node relating to a specified endpoint
@@ -291,6 +295,10 @@ namespace StackExchange.Redis
             var ep = parts[1];
             int at = ep.IndexOf('@');
             if (at >= 0) ep = ep.Substring(0, at);
+            if ((configuration?.AddressRemapping?.ContainsKey(ep)).GetValueOrDefault(false))
+            {
+                ep = configuration.AddressRemapping[ep];
+            }
 
             EndPoint = Format.TryParseEndPoint(ep);
             if (flags.Contains("myself"))
--- a/src/StackExchange.Redis/ConfigurationOptions.cs
+++ b/src/StackExchange.Redis/ConfigurationOptions.cs
@@ -357,7 +357,9 @@ namespace StackExchange.Redis
         public string ServiceName { get; set; }
 
         /// <summary>
-        /// A key value pair of address remappings
+        /// A key value pair of server address remappings
+        /// Any key found will be replaced with the value when parsing cluster / sentinel responses
+        /// This allows you to use different addressing, ports and tls settings outside the cluster vs inside
         /// </summary>
         public Dictionary<string, string> AddressRemapping { get; set; }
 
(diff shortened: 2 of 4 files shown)


mgravell:
Can I just make sure I understand the scenario here? Is the point that the addresses reported by the cluster are unreachable, and should be substituted for alternative addresses?

Also: have you looked at the new cluster proxy in redis 6? Maybe it would be simpler to use that as the sole gateway, with the library not doing any of the shard work?

N6UDP:
Yep that's exactly it; yeah there are certainly other options like the new cluster proxy (we're running redis 5 currently) or doing DNAT shenanigans w/cluster announce ip/port using the reachable addresses.  Or of course modifying the server side to support advertising differing addresses for clustering vs for clients.

NickCraver:
Overall, I would argue this isn't something that belongs in the client. From a systems design standpoint, all clients knowing significant details about what's _behind_ a proxy or load balancer defeats a large part of the purpose of the proxy or load balancer. This seems like adapting the client to a very specific scenario and for anyone who'd need this: you'd get a nasty surprise when any topology changes (and there's no mechanism to update it).

I'd rather we not put this in the client. If we need to expose enhancement to sever selection strategy then _maybe_ there where you could plugin code needed...but I'd really revisit the proxy design and consider what it's doing if the client needs to know about it. For example, all DNS entries pointed at the LB with equivalent ports and have it forward to what's needed underneath (even if the nodes think _they're_ that DNS endpoint, etc. are valid options. Generally you'd want a load balancer's results to point _at itself_ and not depend on a downstream replacement for clients to determine that relationship. Imagine for a moment if cloud providers did this, we'd be in quite a pickle. Instead, they surface a few ports and the rest of the internals of how it's load balanced/failed over, etc. are hidden away...that should be the goal, and what out client supports, IMO.

N6UDP:
Yeah, we moved to some DNAT iptables rules to avoid this.  It allows us to advertise TLS'd ports with public IPs to the clients but redirect internal/cluster traffic to avoid going out to the LB.  It would be better if the server supported announcing different IPs for cluster communication vs client communication.

PR #1421, opened 2020-04-05
Centralized package versioning with Directory.Packages.props

NickCraver:
Explicitly didn't do `/toys` here because the versions aren't shared but if it makes sense: yeah let's add em. This structure is made for easier migration to https://github.com/NuGet/Home/wiki/Centrally-managing-NuGet-package-versions later (we should just need to remove the manual import in each project file (necessary for ordering):
```xml
<Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Packages.props', '$(MSBuildThisFileDirectory)../'))" />
```

If you're wondering why `Directory.Build.targets` doesn't work - I'm not sure. It's fine for `netcoreapp*` builds, but .NET Full framework throws a fit:
```
C:\git\StackExchange\StackExchange.Redis\tests\NRediSearch.Test\NRediSearch.Test.csproj : error NU1701: Package 'xunit 1.7.0.1540' was restored using '.NETFramework,Version=v4.6.1, .NETFramework,Version=v4.6.2, .NETFramework,Version=v4.7, .NETFramework,Version=v4.7.1, .NETFramework,Version=v4.7.2, .NETFramework,Version=v4.8' instead of the project target framework '.NETCoreApp,Version=v2.1'. This package may not be fully compatible with your project. [C:\git\StackExchange\StackExchange.Redis\Build.csproj]
C:\git\StackExchange\StackExchange.Redis\tests\BasicTestBaseline\BasicTestBaseline.csproj : error NU1604: Project dependency BenchmarkDotNet does not contain an inclusive lower bound. Include a lower bound in the dependency version to ensure consistent restore results. [C:\git\StackExchange\StackExchange.Redis\Build.csproj]
C:\git\StackExchange\StackExchange.Redis\tests\BasicTest\BasicTest.csproj : error NU1604: Project dependency BenchmarkDotNet does not contain an inclusive lower bound. Include a lower bound in the dependency version to ensure consistent restore results. [C:\git\StackExchange\StackExchange.Redis\Build.csproj]
C:\git\StackExchange\StackExchange.Redis\tests\BasicTest\BasicTest.csproj : error NU1701: Package 'BenchmarkDotNet 0.5.0' was restored using '.NETFramework,Version=v4.6.1, .NETFramework,Version=v4.6.2, .NETFramework,Version=v4.7, .NETFramework,Version=v4.7.1, .NETFramework,Version=v4.7.2, .NETFramework,Version=v4.8' instead of the project target
framework '.NETCoreApp,Version=v2.0'. This package may not be fully compatible with your project. [C:\git\StackExchange\StackExchange.Redis\Build.csproj]
C:\git\StackExchange\StackExchange.Redis\tests\BasicTestBaseline\BasicTestBaseline.csproj : error NU1701: Package 'BenchmarkDotNet 0.5.0' was restored using '.NETFramework,Version=v4.6.1, .NETFramework,Version=v4.6.2, .NETFramework,Version=v4.7, .NETFramework,Version=v4.7.1, .NETFramework,Version=v4.7.2, .NETFramework,Version=v4.8' instead of the project target framework '.NETCoreApp,Version=v2.0'. This package may not be fully compatible with your project. [C:\git\StackExchange\StackExchange.Redis\Build.csproj]
```
...but the approach in this PR is both works and is closer to future plans.


schemburkar:
In your tests folder, there is already a file `Directory.Build.targets`. The build system will pickup the first .targets file when building every project file, traversing up.

If I add the `<PackageReference Update ... />` in both the targets files (one in root and other in tests), it seems to build fine.
OR
Add just this to your .targets in tests folder

  `<Import Project="$([MSBuild]::GetPathOfFileAbove('Directory.Build.targets', '$(MSBuildThisFileDirectory)../'))" />`

This also build fine and single place too!

(I already removed the individual imports for Directory.Packages.props in every .csproj first)

NickCraver:
@schemburkar Thanks for the eyes here, obvious fail after seeing it - much appreciated! Simplifying this further.

PR #1423, opened 2020-04-06
IPv6 parsing for addresses without brackets which can be seen from

### LINKED ISSUE #1422: IPv6 Cluster connection errors
I've got an IPv6 only Redis cluster. The client connects fine to any hash slots on the existing node, but if it needs to go to another node it fails. Have proven the network is fine as another client will connect fine not in .Net code. Odd how the port is in the endpoint. 

Connection string was "[abcd:1234:3:a::a]:7000,[abcd:1234:3:a::b]:7000"

The GetEndpoints has 8 endpoints when there is only 6 in the cluster (3 master, 3 slave).

'Endpoint [abcd:1234:3:a::c:7000]:0 serving hashslot 6438 is not reachable at this point of time. Please check connectTimeout value. If it is low, try increasing it to give the ConnectionMultiplexer a chance to recover from the network disconnect. IOCP:

lewisvive:
Adjusted the TryParseEndPoint to understand IPv6 address without brackets, but with ports. This is how they come from Redis on the cluster nodes command. 

Fixes #1422

### DIFF
--- a/src/StackExchange.Redis/Format.cs
+++ b/src/StackExchange.Redis/Format.cs
@@ -236,7 +236,7 @@ namespace StackExchange.Redis
             var lastColonIndex = addressWithPort.LastIndexOf(':');
             if (lastColonIndex > 0)
             {
-                // IPv4 with port or IPv6
+                // IPv4 with port or IPv6 with brackets
                 var closingIndex = addressWithPort.LastIndexOf(']');
                 if (closingIndex > 0)
                 {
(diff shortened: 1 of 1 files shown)


mgravell:
Whatever the new format is here should be added to `FormatTests.EndpointData`. Can you confirm: is it that the IPs are coming from the cluster as IPv6 without the correct brackets?

NickCraver:
@lewisvive any more context on what we're dealing with here?

mgravell:
Closing down; requested additional data that wasn't forthcoming. You're probably right about the changes, but without that context: we can't evaluate it, so I'm going to bump it to when we actually have that data to play with. Happy to reopen if you can clarify.

PR #1425, opened 2020-04-08
add a new cluster kind for redis-cluster-proxy

mgravell:
in most ways it is similar to twemproxy, and you can make it just about work by manually configuring things in the config, but this makes everything better integrated, so all you need to do is use a config like `"127.0.0.1,proxy=RedisClusterProxy"`, and everything else clicks into place automatically.

In particular:
- configure the command map
- default the port to 7777
- disable the subscriber/server APIs
- set a minimum redis version of 3 (because: cluster)
- only work with DB zero (again, because: cluster)

work todo:

- [x] consider `PROXY MULTIPLEXING on`
  - [x] find out whether `MULTI` *resets* `MULTIPLEXING` *afterwards* (it disables it during)
- [ ] look into `*SCAN` cursor change

(it turns out that `PROXY MULTIPLEXING on` isn't a thing; it is on by default, and you can turn it *off*, or it gets turned off when you do the first `MULTI`; conclusion: just let the server worry about it)

### DIFF
--- a/src/StackExchange.Redis/CommandMap.cs
+++ b/src/StackExchange.Redis/CommandMap.cs
@@ -21,7 +21,7 @@ namespace StackExchange.Redis
         public static CommandMap Default { get; } = CreateImpl(null, null);
 
         /// <summary>
-        /// The commands available to <a href="twemproxy">https://github.com/twitter/twemproxy</a>
+        /// The commands available to <a href="https://github.com/twitter/twemproxy/">twemproxy</a>
         /// </summary>
         /// <remarks>https://github.com/twitter/twemproxy/blob/master/notes/redis.md</remarks>
         public static CommandMap Twemproxy { get; } = CreateImpl(null, exclusions: new HashSet<RedisCommand>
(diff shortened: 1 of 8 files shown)


mgravell:
Do not merge yet; need to think about whether or not to enable multiplexing on the connection, I'm guessing "yes" - if so, need to issue `PROXY MULTIPLEXING on`

artix75:
@mgravell Hello, I'm the developer of Redis Cluster Proxy, thank you for trying it.

As for `PROXY MULTIPLEXING ON`,  just some tip:

- By default, every client is connected to a shared (multiplex) connection, as you said. The multiplexing mode is automatically disabled by some commands (ie. `MULTI` or blocking commands) that could not work properly in a shared connection, so that the client gets its own private connection to the cluster (multiplexing off). Alternatively, the command `PROXY MULTIPLEXING OFF` can be used to give the client a private connection, anyway, it has been designed for debugging purpose.
- Disabling multiplexing for all the clients is an option, but it's dangerous since it could open too many file descriptors depending on the number of clients, so it's there for testing and debugging purposes too.
- Every thread has a "connections pool" used to provide ready-to-use (already connected) sockets to clients with private connections, in order to speed-up query execution. The pool is automatically re-populated and its size and re-population (re-spawn) criteria can be set in the configuration.

NickCraver:
@mgravell ping, check out #1982 to base on for simplicity

mgravell:
Consencus is that this proxy failed to launch; community guidance (there are literally no official responses) from here https://github.com/RedisLabs/redis-cluster-proxy/issues/81#issue-comment-box is to look to envoy instead, which is already covered above. Closing this.

PR #1428, opened 2020-04-12
Add auth to commandmap for sentinel servers

ejsmith:
Sentinel servers can be set to require auth like regular servers.


mgravell:
Yep, it should have Auth - thanks

PR #1429, opened 2020-04-12
Fix RedisConfig sln file paths

ejsmith:


mgravell:
Thanks - an oversight!

Issue #1430, opened 2020-04-12
UpdateSentinelAddressList is failing with error: EndPoints must be unique


ejsmith:
I am getting the following error attempting to use sentinel support. I'm guessing this is some sort of concurrency issue, but I'm not sure. I will try and get a repro together.

```
EndPoints must be unique (Parameter 'item')
System.ArgumentException: EndPoints must be unique (Parameter 'item')
   at StackExchange.Redis.EndPointCollection.InsertItem(Int32 index, EndPoint item) in /_/src/StackExchange.Redis/EndPointCollection.cs:line 71
   at StackExchange.Redis.ConnectionMultiplexer.UpdateSentinelAddressList(String serviceName) in /_/src/StackExchange.Redis/ConnectionMultiplexer.cs:line 2400
   at StackExchange.Redis.ConnectionMultiplexer.SwitchMaster(EndPoint switchBlame, ConnectionMultiplexer connection, TextWriter log) in /_/src/StackExchange.Redis/ConnectionMultiplexer.cs:line 2376
   at StackExchange.Redis.ConnectionMultiplexer.GetSentinelMasterConnection(ConfigurationOptions config, TextWriter log) in /_/src/StackExchange.Redis/ConnectionMultiplexer.cs:line 2255
   at Exceptionless.Insulation.Bootstrapper.GetRedisConnection(Dictionary`2 options) in /app/src/Exceptionless.Insulation/Bootstrapper.cs:line 166
```

ejsmith:
Is there any reason why `EndPointCollection.InsertItem` wouldn't just return if the item is already in the collection? I don't really understand why you would want to blow up. Seems like there is no harm in basically assuring that an item is in the collection.

ejsmith:
I guess maybe just because you are expecting to insert it at a specific index and if it already exists in another index position then it can not complete the requested operation.

ejsmith:
I added a `TryAdd` method to the `EndPointCollection` class that just adds the endpoint if it doesn't already exist.

NickCraver:
This was fixed in #1431, closing out to cleanup!

Issue #1427, opened 2020-04-10
Issues with Sentinel


ctlajoie:
I implemented support for Redis Sentinel in an application yesterday and I had a rough time getting it to work. I had to actually dig into the library code to figure out what I was doing wrong. If you make a few minor tweaks to the API you could entirely avoid the problems I ran into. 
Allow me to explain. This is not the exact code I had; I modified it to illustrate the point.

```csharp
IConnectionMultiplexer sentinel = ConnectionMultiplexer.Connect(new ConfigurationOptions {
    ServiceName = _redisSentinelServiceName,
    EndPoints = { { _redisSentinelHost, 26379 } },
    CommandMap = CommandMap.Sentinel
});

IConnectionMultiplexer master = sentinel.GetSentinelMasterConnection(new ConfigurationOptions {
    ServiceName = _redisSentinelServiceName
});
```
Looks like it should work... but there are 2 problems with this code.
1. `IConnectionMultiplexer` does not have a `GetSentinelMasterConnection` function.
2. The sentinel `ConfigurationOptions` **MUST** have `TieBreaker = ""` to work.

I have a few suggestions:
1. Add `GetSentinelMasterConnection` to `IConnectionMultiplexer`
2. Have a separate `ConnectionMultiplexer.ConnectSentinel(...)` function, which sets `CommandMap` and `TieBreaker` to the required values, and make endpoints default to port 26379.
3. Just use the `ServiceName` provided in the Sentinel configuration, so I don't have to provide it twice

I realize there may be any number of reasons why these suggestions might not make sense (this isn't my code), but please consider if there's some improvements to be made with regard to the library's Sentinel functionality.

marcelbeeker:
I have a rough time too getting this to work for the Redis Sentinel running on my local machine. I'm glad to hear I'm not the only one.

mgravell:
Ultimately, this is one of the key concerns I had about merging the sentinel code. Fundamentally, we (the core maintainers) aren't routinely sentinel users, so it is hard to say "yes, this is the right API and does what it needs to do". So we held off on the sentinel code for ages. Eventually we merged it, and: it looks like there's already some gaps and problem areas. Now I'm in that awkward "what option is best now" position again :/

ctlajoie:
@mgravell That makes sense. Now I understand why this functionality was only recently merged in. I still believe the API can be made much better with a few "minor" changes.

A refinement of my original suggestions would be to have an interface like the following.
```csharp
[...]

PR #1403, opened 2020-03-24
Sentinel: remove Thread.Sleep and throws

NickCraver:
We were throwing in event handlers...that's not good. Also post-thread sleep in timer callbacks.

PR #1431, opened 2020-04-12
Sentinel improvements

ejsmith:
Add SentinelConnect and SentinelMasterConnect to ConnectionMultiplexer for working with sentinel setups (#1427)
Fix issue with duplicate endpoints being added in the UpdateSentinelAddressList method (#1430).

This change makes it a lot easier and more discoverable how to connect to a sentinel server while also allowing connecting with just a connection string change which allows existing libs that are using SE.Redis to be used in sentinel mode.

Adding a `serviceName` parameter to the connection string triggers sentinel mode. It will connect to the sentinel and discover the current master and return a managed connection that follows the master. 

```csharp
var conn = ConnectionMultiplexer.Connect("localhost,serviceName=mymaster");
var db = conn.GetDatabase();
db.StringSet("key", "value");
```


src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -994,14 +994,74 @@ public static ConnectionMultiplexer Connect(string configuration, TextWriter log
...
+        /// <summary>
+        /// Create a new ConnectionMultiplexer instance that connects to a sentinel server, discovers the current master server
+        /// for the specified ServiceName in the config and returns a managed connection to the current master server
+        /// </summary>
+        /// <param name="configuration">The string configuration to use for this multiplexer.</param>
+        /// <param name="log">The <see cref="TextWriter"/> to log to.</param>
+        public static ConnectionMultiplexer SentinelMasterConnect(string configuration, TextWriter log = null)
+        {
+            var options = ConfigurationOptions.Parse(configuration);
+            var sentinelConnection = SentinelConnect(options, log);
```
mgravell, line 1047: feels like this could jump to `SentinelMasterConnect(options, log)` here to avoid code dup

src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -994,14 +994,74 @@ public static ConnectionMultiplexer Connect(string configuration, TextWriter log
...
+
+        /// <summary>
+        /// Create a new ConnectionMultiplexer instance that connects to a sentinel server, discovers the current master server
+        /// for the specified ServiceName in the config and returns a managed connection to the current master server
+        /// </summary>
+        /// <param name="configuration">The configuration options to use for this multiplexer.</param>
+        /// <param name="log">The <see cref="TextWriter"/> to log to.</param>
+        public static ConnectionMultiplexer SentinelMasterConnect(ConfigurationOptions configuration, TextWriter log = null)
+        {
+            var sentinelConnection = SentinelConnect(configuration, log);
```
mgravell, line 1060: what is the lifetime of `sentinelConnection` here? it feels like the initial discovery connection could be `using` ?

src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -994,14 +994,74 @@ public static ConnectionMultiplexer Connect(string configuration, TextWriter log
...
+            return Connect(sentinelConfigurationOptions, log);
+        }
+
+        /// <summary>
+        /// Create a new ConnectionMultiplexer instance that connects to a sentinel server, discovers the current master server
+        /// for the specified ServiceName in the config and returns a managed connection to the current master server
+        /// </summary>
+        /// <param name="configuration">The string configuration to use for this multiplexer.</param>
+        /// <param name="log">The <see cref="TextWriter"/> to log to.</param>
+        public static ConnectionMultiplexer SentinelMasterConnect(string configuration, TextWriter log = null)
```
mgravell, line 1044: note that `Connect` is twinned with `ConnectAsync` - it feels like this should be maintained here too, i.e. matching `SentinelMasterConnectAsync` methods

mgravell:
some crossover / thoughts here; views? https://github.com/StackExchange/StackExchange.Redis/issues/1427#issuecomment-613450272

ejsmith:
I actually attempted to go down that path with it just being part of the connection string, but it was causing some issues as well as not being intuitive as to what it was doing. I do think it’s super important for existing things that make use of SE.Redis to work without code changes so I can take another stab at this and see how it goes.

src/StackExchange.Redis/ConnectionMultiplexer.cs
ejsmith: I believe this connection is kept open to be able to query the sentinel server for master changes.

src/StackExchange.Redis/ConnectionMultiplexer.cs
ejsmith: Good call.

NickCraver:
FWIW, I totally agree with connection string via existing `.Connect()`/`.ConnectAsync()`- so many things call us and will never be adapted to Sentinel overloads (not should they be, IMO). For this to be viable for so many use cases, it needs to be invoked via that route.

ejsmith:
What are you guys thinking for the configuration name? I thought about triggering it just off of ServiceName, but it doesn't seem very intuitive.

src/StackExchange.Redis/ConnectionMultiplexer.cs
ejsmith: I am working on adding the Async implementations.

ctlajoie:
Any updates?

src/StackExchange.Redis/ConnectionMultiplexer.cs
ejsmith: All of the Async overloads should be there now.

ejsmith:
@mgravell @NickCraver sorry for the delay. I finally got around to updating this PR to allow connecting to sentinel server by just setting the serviceName in the connection string. Also added the missing async overloads and added some more tests.

ejsmith:
@ctlajoie sorry for the delay. Just got around to updating this PR.

docs/Configuration.md
```diff
@@ -29,6 +29,14 @@ This will connect to a single server on the local machine using the default redi
 var conn = ConnectionMultiplexer.Connect("redis0:6380,redis1:6380,allowAdmin=true");
 ```
 
+If you specify a serviceName in the connection string, it will trigger sentinel mode. This will connect to a sentinel server on the local machine
```
NickCraver, line 32:
```suggestion
If you specify a serviceName in the connection string, it will trigger sentinel mode. This example will connect to a sentinel server on the local machine
```
IMO adding "example" here would make it a lot clearer - I read this as it _always_ connects local on first pass.

tests/RedisConfigs/Sentinel/redis-7011.conf
```diff
@@ -1,9 +1,11 @@
...
+#masterauth changeme
 repl-diskless-sync yes
 repl-diskless-sync-delay 0
 maxmemory 100mb
 appendonly no
 dir "../Temp"
 dbfilename "sentinel-target-7011.rdb"
-save ""
\ No newline at end of file
+save ""
```
NickCraver, line 11: Can we revert changes to these files please?

tests/RedisConfigs/Sentinel/redis-7010.conf
```diff
@@ -1,4 +1,6 @@
 port 7010
+#requirepass changeme
+#masterauth changeme
```
NickCraver, line 3: Can we revert changes to these files please?

tests/RedisConfigs/Sentinel/sentinel-26379.conf
```diff
@@ -1,5 +1,7 @@
 port 26379
+#requirepass "changeme"
```
NickCraver, line 2: This seems to be a weird addition from testing perhaps? Let's revert all these commented config changes please - PR doesn't need to touch those files :)

tests/StackExchange.Redis.Tests/Sentinel.cs
```diff
@@ -12,7 +12,7 @@ namespace StackExchange.Redis.Tests
     public class Sentinel : TestBase
     {
         private string ServiceName => TestConfig.Current.SentinelSeviceName;
-        private ConfigurationOptions ServiceOptions => new ConfigurationOptions { ServiceName = ServiceName, AllowAdmin = true };
+        private ConfigurationOptions ServiceOptions => new ConfigurationOptions { ServiceName = ServiceName, AllowAdmin = true, Password = "changeme" };
```
NickCraver, line 15: This is commented in config and present here - should be removed here I think, guessing it was during testing?

src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -876,13 +882,20 @@ private static async Task<ConnectionMultiplexer> ConnectImplAsync(object configu
         /// </summary>
         /// <param name="configuration">The configuration options to use for this multiplexer.</param>
         /// <param name="log">The <see cref="TextWriter"/> to log to.</param>
-        public static Task<ConnectionMultiplexer> ConnectAsync(ConfigurationOptions configuration, TextWriter log = null)
+        public static async Task<ConnectionMultiplexer> ConnectAsync(ConfigurationOptions configuration, TextWriter log = null)
         {
             SocketConnection.AssertDependencies();
-            return ConnectImplAsync(configuration, log);
+
+            bool sentinel = !String.IsNullOrEmpty(configuration.ServiceName);
```
NickCraver, line 889:
```suggestion
            bool sentinel = !string.IsNullOrEmpty(configuration.ServiceName);
```

NickCraver: Previously, a user would hit `PrepareConfig` here ultimately, which does a null check. This would cause a null ref instead - we'd need a null arg check above this access to compensate.

src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -876,13 +882,20 @@ private static async Task<ConnectionMultiplexer> ConnectImplAsync(object configu
...
+
+            if (!sentinel)
```
NickCraver, line 891: Let's restructure this as `if (sentinel) { ...handling... } so we encapsulate the case, rather than default to it inversely.

src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -899,7 +912,21 @@ internal static ConfigurationOptions PrepareConfig(object configuration)
                 throw new ArgumentException("Invalid configuration object", nameof(configuration));
             }
             if (config.EndPoints.Count == 0) throw new ArgumentException("No endpoints specified", nameof(configuration));
-            config.SetDefaultPorts();
+
+            if (!sentinel)
```
NickCraver, line 916: Same as above, let's invert this please

src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -876,13 +882,20 @@ private static async Task<ConnectionMultiplexer> ConnectImplAsync(object configu
...
         {
             SocketConnection.AssertDependencies();
-            return ConnectImplAsync(configuration, log);
+
+            bool sentinel = !String.IsNullOrEmpty(configuration.ServiceName);
+
+            if (!sentinel)
+                return await ConnectImplAsync(PrepareConfig(configuration), log).ForAwait();
+
+            var conn = await ConnectImplAsync(PrepareConfig(configuration, true), log).ForAwait();
```
NickCraver, line 894: With the move above, let's be explicit so this is more readable please:
```suggestion
            var conn = await ConnectImplAsync(PrepareConfig(configuration, sentinel: true), log).ForAwait();
```

src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -987,22 +1014,121 @@ private static ConnectionMultiplexer CreateMultiplexer(object configuration, Log
...
-        private static ConnectionMultiplexer ConnectImpl(object configuration, TextWriter log)
+        /// <summary>
+        /// Create a new ConnectionMultiplexer instance that connects to a sentinel server
+        /// </summary>
+        /// <param name="configuration">The string configuration to use for this multiplexer.</param>
+        /// <param name="log">The <see cref="TextWriter"/> to log to.</param>
+        public static ConnectionMultiplexer SentinelConnect(string configuration, TextWriter log = null)
+        {
+            SocketConnection.AssertDependencies();
+            return ConnectImpl(PrepareConfig(configuration, true), log);
```
NickCraver, line 1047:
```suggestion
            return ConnectImpl(PrepareConfig(configuration, sentinel: true), log);
```

src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -987,22 +1014,121 @@ private static ConnectionMultiplexer CreateMultiplexer(object configuration, Log
...
+
+        /// <summary>
+        /// Create a new ConnectionMultiplexer instance that connects to a sentinel server
+        /// </summary>
+        /// <param name="configuration">The string configuration to use for this multiplexer.</param>
+        /// <param name="log">The <see cref="TextWriter"/> to log to.</param>
+        public static Task<ConnectionMultiplexer> SentinelConnectAsync(string configuration, TextWriter log = null)
+        {
+            SocketConnection.AssertDependencies();
+            return ConnectImplAsync(PrepareConfig(configuration, true), log);
```
NickCraver, line 1058:
```suggestion
            return ConnectImplAsync(PrepareConfig(configuration, sentinel: true), log);
```

src/StackExchange.Redis/ConnectionMultiplexer.cs
NickCraver: It can be GCed right after this method, so that doesn't seem right?

src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -987,22 +1014,121 @@ private static ConnectionMultiplexer CreateMultiplexer(object configuration, Log
...
+
+        /// <summary>
+        /// Create a new ConnectionMultiplexer instance that connects to a sentinel server, discovers the current master server
+        /// for the specified ServiceName in the config and returns a managed connection to the current master server
+        /// </summary>
+        /// <param name="configuration">The string configuration to use for this multiplexer.</param>
+        /// <param name="log">The <see cref="TextWriter"/> to log to.</param>
+        public static Task<ConnectionMultiplexer> SentinelMasterConnectAsync(string configuration, TextWriter log = null)
+        {
+            return SentinelMasterConnectAsync(PrepareConfig(configuration, true), log);
```
NickCraver, line 1115:
```suggestion
            return SentinelMasterConnectAsync(PrepareConfig(configuration, sentinel: true), log);
```

src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -987,22 +1014,121 @@ private static ConnectionMultiplexer CreateMultiplexer(object configuration, Log
...
+
+        /// <summary>
+        /// Create a new ConnectionMultiplexer instance that connects to a sentinel server, discovers the current master server
+        /// for the specified ServiceName in the config and returns a managed connection to the current master server
+        /// </summary>
+        /// <param name="configuration">The string configuration to use for this multiplexer.</param>
+        /// <param name="log">The <see cref="TextWriter"/> to log to.</param>
+        public static ConnectionMultiplexer SentinelMasterConnect(string configuration, TextWriter log = null)
+        {
+            return SentinelMasterConnect(PrepareConfig(configuration, true), log);
```
NickCraver, line 1091:
```suggestion
            return SentinelMasterConnect(PrepareConfig(configuration, sentinel: true), log);
```

src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -987,22 +1014,121 @@ private static ConnectionMultiplexer CreateMultiplexer(object configuration, Log
...
+
+        /// <summary>
+        /// Create a new ConnectionMultiplexer instance that connects to a sentinel server
+        /// </summary>
+        /// <param name="configuration">The configuration options to use for this multiplexer.</param>
+        /// <param name="log">The <see cref="TextWriter"/> to log to.</param>
+        public static Task<ConnectionMultiplexer> SentinelConnectAsync(ConfigurationOptions configuration, TextWriter log = null)
+        {
+            SocketConnection.AssertDependencies();
+            return ConnectImplAsync(PrepareConfig(configuration, true), log);
```
NickCraver, line 1080:
```suggestion
            return ConnectImplAsync(PrepareConfig(configuration, sentinel: true), log);
```

src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -987,22 +1014,121 @@ private static ConnectionMultiplexer CreateMultiplexer(object configuration, Log
...
+
+        /// <summary>
+        /// Create a new ConnectionMultiplexer instance that connects to a sentinel server
+        /// </summary>
+        /// <param name="configuration">The configuration options to use for this multiplexer.</param>
+        /// <param name="log">The <see cref="TextWriter"/> to log to.</param>
+        public static ConnectionMultiplexer SentinelConnect(ConfigurationOptions configuration, TextWriter log = null)
+        {
+            SocketConnection.AssertDependencies();
+            return ConnectImpl(PrepareConfig(configuration, true), log);
```
NickCraver, line 1069:
```suggestion
            return ConnectImpl(PrepareConfig(configuration, sentinel: true), log);
```

src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -876,13 +882,20 @@ private static async Task<ConnectionMultiplexer> ConnectImplAsync(object configu
...
             SocketConnection.AssertDependencies();
-            return ConnectImplAsync(configuration, log);
+
+            bool sentinel = !String.IsNullOrEmpty(configuration.ServiceName);
+
+            if (!sentinel)
+                return await ConnectImplAsync(PrepareConfig(configuration), log).ForAwait();
+
+            var conn = await ConnectImplAsync(PrepareConfig(configuration, true), log).ForAwait();
+            return conn.GetSentinelMasterConnection(PrepareConfig(configuration), log);
```
NickCraver, line 895: I'm turn here, because this path _isn't_ async, but I don't really fancy adding a whole `GetSentinelMasterConnectionAsync` path either - @mgravell thoughts?

src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -987,22 +1014,121 @@ private static ConnectionMultiplexer CreateMultiplexer(object configuration, Log
         /// <param name="log">The <see cref="TextWriter"/> to log to.</param>
         public static ConnectionMultiplexer Connect(string configuration, TextWriter log = null)
         {
-            SocketConnection.AssertDependencies();
-            return ConnectImpl(configuration, log);
+            var config = PrepareConfig(configuration);
```
NickCraver, line 1017: Would be more consistent to inline this in the `Connect()` below

src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -987,22 +1014,121 @@ private static ConnectionMultiplexer CreateMultiplexer(object configuration, Log
...
         /// </summary>
-        /// <param name="configuration">The configurtion options to use for this multiplexer.</param>
+        /// <param name="configuration">The configuration options to use for this multiplexer.</param>
         /// <param name="log">The <see cref="TextWriter"/> to log to.</param>
         public static ConnectionMultiplexer Connect(ConfigurationOptions configuration, TextWriter log = null)
         {
             SocketConnection.AssertDependencies();
-            return ConnectImpl(configuration, log);
+
+            bool sentinel = !String.IsNullOrEmpty(configuration.ServiceName);
```
NickCraver, line 1030:
```suggestion
            bool sentinel = !string.IsNullOrEmpty(configuration.ServiceName);
```

src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -987,22 +1014,121 @@ private static ConnectionMultiplexer CreateMultiplexer(object configuration, Log
...
+
+            if (!sentinel)
```
NickCraver, line 1032: Same as above, let's invert please - can eliminate the `sentinel` variable and put a comment here as well.

src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -2236,7 +2363,8 @@ public ConnectionMultiplexer GetSentinelMasterConnection(ConfigurationOptions co
                 config.EndPoints.Add(newMasterEndPoint);
             }
 
-            ConnectionMultiplexer connection = Connect(config, log);
+            ConnectionMultiplexer connection = ConnectImpl(config, log);
+            // TODO: Verify ROLE is master as specified here: https://redis.io/topics/sentinel-clients
```
NickCraver, line 2367: What's the status on this?

NickCraver:
Oh one more note overall: do we _want_ the `SentinelConnect` or `SentinelConnectAsync` methods, given we not detect service name? I'm unclear from the conversation as to if these are really needed. I don't think effectively this will be used that much and _likely_ aren't worth the API addition - since so many calls are coming _from someone else_ (e.g. someone else's caching library, etc.) and they're all coming via `.Connect*()`.

Thoughts based on latest detection?

tests/RedisConfigs/Sentinel/redis-7010.conf
ejsmith: So the reason why I left those in there is twofold. 1. It was hard to figure out all the places that you have to set a password in a master/slave with sentinel setup. 2. The Windows port of Redis running in sentinel mode doesn’t work with a password so I had to turn it off in order for tests to pass. Also, having the password left in the code didn’t seem to negatively affect anything so I left it there so that’s it’s just a matter of commenting out all the password config entries to be able to test using a password in the setup.

tests/RedisConfigs/Sentinel/sentinel-26379.conf
ejsmith: Same reason as above.

tests/StackExchange.Redis.Tests/Sentinel.cs
ejsmith: Didn’t seem to affect anything negatively to leave it. Again, just say so an I will remove them all.

src/StackExchange.Redis/ConnectionMultiplexer.cs
ejsmith: Good catch.

src/StackExchange.Redis/ConnectionMultiplexer.cs
ejsmith: Ok, will do.

src/StackExchange.Redis/ConnectionMultiplexer.cs
ejsmith: Ok, will do.

src/StackExchange.Redis/ConnectionMultiplexer.cs
ejsmith: Yeah, that’s the same thing I was thinking. Didn’t seem worth having a whole new async copy of that method.

src/StackExchange.Redis/ConnectionMultiplexer.cs
ejsmith: It wasn’t doing this before and I did see that it’s kind of being verified by running an INFO call that gets the server mode.  But according to that link they recommend that you call ROLE after connect to verify the server is the right one. ROLE does not seem to be implemented and so I briefly looked at implementing it, but kind of just said never mind after I found out that it’s doing the INFO thing.

ejsmith:
Don’t worry about stepping on my toes. It’s your project, your rules. 😀

As far as the password stuff I explained my thoughts in the comments. If you don’t want them I will get them all removed.

Yeah, I’m on the fence with those as well, but they seemed like they provided some minor value in discoverability. Also, without the ConnectSentinel then it’s a bit harder to know how to connect directly to the sentinel server because you need to know that you’ve got to set the tiebreaker, command map config and port settings for it to work. On this topic though, I was just thinking about this and wondering if I should dig into the reason why not setting the tiebreaker config seems to be blowing things up connecting sentinel. Maybe if we could fix that issue and maybe even dynamically set the commandmap based on the detected server type? If we were able to do that then there wouldn’t be any weird issues with trying to connect to the sentinel server with a simple config that has server and port. Thoughts?

src/StackExchange.Redis/ConnectionMultiplexer.cs
ejsmith: Hmm... I know somewhere in the call stack it's setting up a subscription on the sentinel server to listen to events and I would think that would be holding onto a reference to that object, but I am not sure on that. I guess we need to figure out some way to test this and also to make it so that the 2 connections are linked and disposed together when the outer connection is disposed.

ejsmith:
@NickCraver I made most of the suggested changes. Waiting for more info before making any additional changes.

tests/RedisConfigs/Sentinel/redis-7010.conf
NickCraver: IMO, it serves more to confuse right now than it makes easy, so I'd vote to remove them. Right now the code config says you need a password, the redis config says you didn't, so the logic question anyone coming here is going to ask is "...why is this commented out?", and "why does this code still work?" - give the aim isn't to actually test password auth here (we're doing that separately), can we tidy and remove the confusion?

src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -987,22 +1014,122 @@ private static ConnectionMultiplexer CreateMultiplexer(object configuration, Log
...
         /// </summary>
-        /// <param name="configuration">The configurtion options to use for this multiplexer.</param>
+        /// <param name="configuration">The configuration options to use for this multiplexer.</param>
         /// <param name="log">The <see cref="TextWriter"/> to log to.</param>
         public static ConnectionMultiplexer Connect(ConfigurationOptions configuration, TextWriter log = null)
         {
             SocketConnection.AssertDependencies();
-            return ConnectImpl(configuration, log);
+
+            bool sentinel = !string.IsNullOrEmpty(configuration?.ServiceName);
```
NickCraver, line 1029: Thought; private method this to DRY up? e.g.:
```cs
if (IsSentinel(configuration))
```

src/StackExchange.Redis/ConnectionMultiplexer.cs
NickCraver: Gotcha - separately filed #1451 to get that fixed up, thanks!

src/StackExchange.Redis/ConnectionMultiplexer.cs
ejsmith: Done

ejsmith:
More updates from feedback. Reverted all the config / password stuff.

ejsmith:
@NickCraver @mgravell as it stands now it seems we have a few things to figure out with full sentinel support.

1. We need to figure out if the outer `ConnectionMultiplexer` is being properly disposed. It's possible that it is currently being disposed too soon and it's also possible that it's not being disposed at all. I need to dig into this and maybe figure out a way we can better tie the 2 together so they can be disposed at the same time.
2. We need to make sure that they server list is being fully populated when we discover the master/slaves from the sentinel cluster. It appears from what #1427 is saying that we are currently only populating the master server. This may be an easy fix. I will take a look.
3. We need to implement the ROLE command (#1451) and make sure we are fully following the sentinel guide from here: https://redis.io/topics/sentinel-clients

That being said. I would like to get this PR merged first and tackle these items after that. Sentinel support is currently already in the client. But there is one major issue keeping me and probably anyone else from using it and that is that the management of the server list is blowing up (#1430) and that you can't use it with other libraries because it couldn't be used with just a different connection string which this PR fixes.

ejsmith:
Ok, I just decided to go ahead and implement the items in the list.

1. Added code to make sure that the inner sentinel connection is disposed when the managed master connection is disposed.
2. Added any slaves to the list of endpoints and verified that you can run read commands against the slaves.
4. Implemented the ROLE command and verified that the server is a master role when connecting to sentinel master.

src/StackExchange.Redis/ConnectionMultiplexer.cs
ejsmith: I've updated this to keep a reference to the sentinel connection and to dispose it when the outer managed connection is disposed.

src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -2214,30 +2350,61 @@ public ConnectionMultiplexer GetSentinelMasterConnection(ConfigurationOptions co
...
+
+                // verify role is master according to:
+                // https://redis.io/topics/sentinel-clients
+                if (connection.GetServer(newMasterEndPoint).Role() == RedisLiterals.master)
+                {
+                    success = true;
+                    break;
+                }
+
+                Thread.Sleep(100);
```
NickCraver, line 2399: Having these sleeps in the multiplexer causes all sorts of issues on resumption we found in the first iteration of Sentinel (later removed in #1403), what's the goal here with them?

src/StackExchange.Redis/RedisServer.cs
```diff
@@ -819,6 +831,18 @@ public Task<EndPoint[]> SentinelGetSentinelAddressesAsync(string serviceName, Co
             return ExecuteAsync(msg, ResultProcessor.SentinelAddressesEndPoints);
         }
 
+        public EndPoint[] SentinelGetSlaveAddresses(string serviceName, CommandFlags flags = CommandFlags.None)
```
NickCraver, line 834: Is this the analog to `SentinelGetMasterAddressByName`? The naming seems off here, I'd expect `ByName` in here (honestly, we'd re-think all of these API additions as a whole and consistently if we had a time machine)

src/StackExchange.Redis/ConnectionMultiplexer.cs
NickCraver: Gotcha - I put some time to step through this with Marc live tomorrow - leaving other notes now!

src/StackExchange.Redis/RedisLiterals.cs
```diff
@@ -123,7 +123,9 @@ public static readonly RedisValue
             pubsub = "pubsub",
             replication = "replication",
             server = "server",
+            master = "master",
             slave = "slave",
+            sentinel = "sentinel",
```
NickCraver, line 128: nit: alpha order please!

ejsmith:
Yeah, whatever you want to help with is great. I was pretty sure you’d want to do more with the ROLE stuff. I guess I can remove that and break it out into a separate PR. I’d like to figure out a way to get part of this merged as soon as possible because we are still blocked by this. Any ideas how we can do that?

src/StackExchange.Redis/ConnectionMultiplexer.cs
ejsmith: I was just trying to follow the sentinel client best practices. It says to check role for master before doing anything else and to retry finding the master from sentinel and try again if it’s not master yet.

src/StackExchange.Redis/RedisServer.cs
ejsmith: No, it's analogous to the method right above it that I copied it from `SentinelMasters`. In redis it translates to `SENTINEL slaves mymaster` and `SENTINEL sentinels mymaster`. While `SentinelGetMasterAddressByName` translates to `SENTINEL get-master-addr-by-name mymaster`

src/StackExchange.Redis/Interfaces/IServer.cs
```diff
@@ -770,6 +786,24 @@ public partial interface IServer : IRedis
         /// <returns>a list of the sentinel ips and ports</returns>
         Task<EndPoint[]> SentinelGetSentinelAddressesAsync(string serviceName, CommandFlags flags = CommandFlags.None);
 
+        /// <summary>
+        /// Returns the ip and port numbers of all known Sentinel slaves
+        /// for the given service name.
+        /// </summary>
+        /// <param name="serviceName">the sentinel service name</param>
+        /// <param name="flags"></param>
```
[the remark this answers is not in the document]
ejsmith: Sorry, I copied this from another method.

ejsmith:
> I think this is looking overall good, but still some to do. I recommend the following (going to sync with @mgravell tomorrow though - curious on his thoughts):
> 
> * I'm not sure we need the `SentinelMasterConnect` methods - that's the same as calling connect with a service name - recommend we don't expose those at least since we're bound on a public API then

I still feel like it's good to have these separate, but I understand the desire to keep the API surface smaller. Your call. I'm good with whatever.

> * There are some naming inconsistencies on the slave methods, I wish we had an enum or something so we didn't have multiple APIs there but we've already shipped the master ones...the addition could be with an enum and plans to deprecate the master still though

Yeah, possibly could be an enum, but the other matching sentinel methods were already in there so I just made matching methods.

> * I think the `ROLE` API addition should be another PR since there's still a lot missing there - from the mocks and tests, etc. - it'd be better as a self-contained thing IMO. A recent example that includes all the pieces that need additions for a new command is #1291. If you'd like me to take that one and get eyes, happy to, but would want permission first!

Yeah, any help would be great. The `ROLE` command returns a lot more info and I think it should probably return it's own RoleInfo model or something a little nicer.

> * There's a spin loop with a `Thread.Sleep` in the connect path which generally leads to trouble and we saw some of that in the test runs with Sentinel on the first pass, resolved with their removal in #1403 - I'd like to find another approach there, or at the very least document heavily on the method why it's there so we have more context when fixing it. Given it's isolated to the Sentinel path, it's not a hard no there.

This one was just following the sentinel client best practices outlined here: https://redis.io/topics/sentinel-clients It says to connect and immediately make sure that the servers role returns master and if not try to discover the master again and reconnect.

> 
> Thanks for all the work on this so far - it's great and much appreciated, and please let me know where you'd like help (e.g. `ROLE`) and like hands off. I'd like to assist, but not step on toes.

src/StackExchange.Redis/RedisLiterals.cs
ejsmith: No problem. They didn't appear to be in strict alpha order previously so I just grouped like ones together.

src/StackExchange.Redis/ConnectionMultiplexer.cs
NickCraver: Discussed with Marc: Let's make these `SentinelMasterConnect` variants private for now (not adding to the API contract.

src/StackExchange.Redis/ConnectionMultiplexer.cs
NickCraver: @mgravell thoughts?

src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -2214,30 +2350,61 @@ public ConnectionMultiplexer GetSentinelMasterConnection(ConfigurationOptions co
...
+                }
+
+                foreach (var slaveEndPoint in slaveEndPoints)
+                    config.EndPoints.TryAdd(slaveEndPoint);
+
+                connection = ConnectImpl(config, log);
+
+                // verify role is master according to:
+                // https://redis.io/topics/sentinel-clients
+                if (connection.GetServer(newMasterEndPoint).Role() == RedisLiterals.master)
```
mgravell, line 2393: looking at the `ROLE` command on redis, it is *much* more complex and nuanced than the shape of the command you're throwing onto `IServer`; can I suggest *don't* add this to the public API right now - this needs a lot more consideration, and it is unrelated to this PR; you might want to just use `Execute`, which can invoke any command, and get the result back out from there

src/StackExchange.Redis/Interfaces/IServer.cs
```diff
@@ -638,6 +638,22 @@ public partial interface IServer : IRedis
...
+        /// <remarks>https://redis.io/commands/role</remarks>
+        string Role(CommandFlags flags = CommandFlags.None);
+
+        /// <summary>
+        /// The ROLE command returns the current server role which can be master, slave or sentinel.
+        /// </summary>
+        /// <param name="flags">The command flags to use.</param>
+        /// <returns>The server's current role.</returns>
+        /// <remarks>https://redis.io/commands/role</remarks>
+        Task<string> RoleAsync(CommandFlags flags = CommandFlags.None);
```
NickCraver, line 655: TODO: Remove from this PR (Marc's commenting on Execute)

src/StackExchange.Redis/Interfaces/IServer.cs
```diff
@@ -766,10 +782,28 @@ public partial interface IServer : IRedis
...
         Task<EndPoint[]> SentinelGetSentinelAddressesAsync(string serviceName, CommandFlags flags = CommandFlags.None);
 
+        /// <summary>
+        /// Returns the ip and port numbers of all known Sentinel slaves
+        /// for the given service name.
+        /// </summary>
+        /// <param name="serviceName">the sentinel service name</param>
+        /// <param name="flags">The command flags to use.</param>
+        /// <returns>a list of the slave ips and ports</returns>
+        EndPoint[] SentinelGetSlaveAddresses(string serviceName, CommandFlags flags = CommandFlags.None);
```
NickCraver, line 796: Shouldn't this be `SentinelGetSlaveAddressesByName`? Can't find previous comments, but analog seems to be those methods.

src/StackExchange.Redis/RedisServer.cs
```diff
@@ -543,6 +543,18 @@ public Task<DateTime> TimeAsync(CommandFlags flags = CommandFlags.None)
...
+        {
+            var msg = Message.Create(-1, flags, RedisCommand.ROLE);
+            return ExecuteSync(msg, ResultProcessor.Role);
+        }
+
+        public Task<string> RoleAsync(CommandFlags flags = CommandFlags.None)
+        {
+            var msg = Message.Create(-1, flags, RedisCommand.ROLE);
+            return ExecuteAsync(msg, ResultProcessor.Role);
+        }
```
NickCraver, line 556: TODO: Remove from this PR (entire file)

src/StackExchange.Redis/RedisServer.cs
NickCraver: On the API surface this is very confusing, it _looks_ like it's 1:1 with the underneath, since this is public API we should get it right. @mgravell thoughts?

src/StackExchange.Redis/ResultProcessor.cs
```diff
@@ -892,6 +898,22 @@ protected override bool SetResultCore(PhysicalConnection connection, Message mes
...
+                {
+                    case ResultType.MultiBulk:
+                        var arr = result.GetItems();
+                        var role = arr[0].GetString();
+                        SetResult(message, role);
+                        return true;
+                }
+                return false;
+            }
+        }
```
NickCraver, line 915: TODO: Remove Role bits from this PR

src/StackExchange.Redis/ConnectionMultiplexer.cs
mgravell: example usage:

``` c#
var role = (string)((RedisResult[])server.Execute("ROLE"))[0];
```

ejsmith:
Updated to remove the `Role` methods off of `IServer` and changed to loop looking for master for `connectTimeout` duration.

ejsmith:
Ok, so we have:
- `SentinelGetMasterAddressByName` (existing) = `SENTINEL get-master-addr-by-name mymaster`
- `SentinelGetSentinelAddresses` (existing) = `SENTINEL sentinels mymaster`
- `SentinelGetSlaveAddresses` = `SENTINEL slaves mymaster`

They seem consistent to me and they seem like they match the corresponding Redis commands. https://redis.io/topics/sentinel

src/StackExchange.Redis/Interfaces/IServer.cs
ejsmith: These were existing methods:
- SentinelGetMasterAddressByName = SENTINEL get-master-addr-by-name mymaster
- SentinelGetSentinelAddresses = SENTINEL sentinels mymaster

I just added this:
- SentinelGetSlaveAddresses = SENTINEL slaves mymaster

It seems like it is following the existing `SentinelGetSentinelAddresses` method to me, but I'm happy to change it.

rabberbock:
I know this has been opened for a while, is there an ETA for this? I have a sentinel setup on k8s and am wondering if this is planned for sometime soon or I should use a different client that supports Redis  Sentinel. Thanks!

PR #1433, opened 2020-04-15
Fix more than 99 times of elapsed time on Linux

pengweiqhca:
See [ValueStopwatch.cs](https://github.com/dotnet/runtime/blob/master/src/libraries/Common/src/Extensions/ValueStopwatch/ValueStopwatch.cs#L36)


mgravell:
I'm intrigued; the code looks valid, and is consistent with `ValueStopwatch`, so I'm not concerned about that - but do you perhaps have a citation talking about the actual linux issue here?

pengweiqhca:
I has created a issure https://github.com/StackExchange/StackExchange.Redis/issues/1434

mgravell:
looks good; intent is discussed in #1434; merging - many thanks!

PR #1471, opened 2020-05-22
Fix RedisResult.ToDictionary.  

nathanscottdaniels:
Previously it would only parse the first half of the array.  Also added test coverage for this method.


mgravell:
Not at a PC to check, but... From inspection only (not run), the old code should work fine. What error were you seeing?

nathanscottdaniels:
Only one half of the results array was parsed into a dictionary, because the for loop was set to stop when `i` reached arr.Length/2.

mgravell:
Ah yes, up on 225; yes, that's not good - thanks for resolving.

mgravell:
Merging, much appreciated

PR #1477, opened 2020-05-28
add last-delivered-id to StreamGroupInfo

AndyPook:
"last-delivered-id" was added to XINFO GROUPS in the 5.0 timeframe.
It can be useful as a proxy for stream lag when compared to StreamInfo,LastGeneratedId
However, the redis-doc wasn't updated (see https://github.com/antirez/redis-doc/pull/1318)

This PR simply adds the field to the StreamGroupInfo response.

### DIFF
--- a/src/StackExchange.Redis/ResultProcessor.cs
+++ b/src/StackExchange.Redis/ResultProcessor.cs
@@ -1566,18 +1566,23 @@ The coordinates as a two items x,y array (longitude,latitude).
                 //    4) (integer)2
                 //    5) pending
                 //    6) (integer)2
+                //    7) last-delivered-id
+                //    8) "1588152489012-0"
                 // 2) 1) name
                 //    2) "some-other-group"
                 //    3) consumers
                 //    4) (integer)1
                 //    5) pending
                 //    6) (integer)0
+                //    7) last - delivered - id
+                //    8) "1588152498034-0"
 
                 var arr = result.GetItems();
 
                 return new StreamGroupInfo(name: arr[1].AsRedisValue(),
                     consumerCount: (int)arr[3].AsRedisValue(),
-                    pendingMessageCount: (int)arr[5].AsRedisValue());
+                    pendingMessageCount: (int)arr[5].AsRedisValue(),
+                    lastDeliveredId: arr[7].AsRedisValue());
             }
         }
 
--- a/src/StackExchange.Redis/StreamGroupInfo.cs
+++ b/src/StackExchange.Redis/StreamGroupInfo.cs
@@ -6,11 +6,12 @@ namespace StackExchange.Redis
     /// </summary>
     public readonly struct StreamGroupInfo
     {
-        internal StreamGroupInfo(string name, int consumerCount, int pendingMessageCount)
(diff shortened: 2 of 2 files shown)


mgravell:
Thanks

AndyPook:
thanks

I was going to have a look at the positional stuff. redis-doc does say that clients shouldn't rely on ordering.
But if you're already on it

AndyPook:
Maybe something like
```csharp

        public static class SeqEx
        {
            public static RedisValue[] Get(this Sequence<RedisValue> values, params string[] names)
            {
                if (names.Length * 2 != values.Length)
                    throw new ArgumentException("values should be exactly double the length of names");

                var result = new RedisValue[names.Length];
                for (int i = 0; i < values.Length; i += 2)
                {
                    if (values[i] == names[i])
                        result[i] = values[i + 1];
                }

                return result;
            }
        }
```
The result will be guaranteed to be in the order of the names supplied. 
Or it could be added next to the `GetItems` method in RawResult?

The nice thing would be that none of the result types need to change. It's just that the ctor calls  in `ResultProcessor` would need their indexes changed to be sequential. A fairly big, if just tedious, change.

Just an idea :)

mgravell:
That pretty much just retains the positional version, but at least checks
the names. I'd rather not depend on the order.

AndyPook:
You are right, I'm an idiot :)

```csharp
        public static RedisValue Get(this RedisValue[] values, string name)
        {
            for (int i = 0; i < values.Length; i += 2)
            {
                if (values[i] == name)
                    return values[i + 1];
            }
            return RedisValue.EmptyString;
        }

        public static RedisValue[] GetValues(this RedisValue[] values, params string[] names)
        {
            if (names.Length * 2 != values.Length)
                throw new ArgumentException("values should be exactly double the length of names");

            var result = new RedisValue[names.Length];
            for (int i = 0; i < values.Length; i += 2)
            {
                for (int j = 0; j < names.Length; j++)
                {
                    if (names[j]!=null && values[i] == names[j])
                    {
                        result[j] = values[i + 1];
                        names[j] = null;
                    }
                }
            }

            return result;
        }
```
Get is the obvious approach.
GetValues works, in that the result is in the order given by the names.
I'm not sure I like it. The `null` bit is an attempt at making the loop faster. But maybe better than mapping to a dictionary?
Any approach is going to cost. But this is probably not going to be on the hi-perf path. So maybe it doesn't matter?

Although redis-doc says the order is undefined, from the code, it is constant for a given release.

mgravell:
Did you see the commit (I cross-reference it, it appears above) where I already implemented this at the `RawResult` level, so it doesn't need to create any `RedisValue` instances or the array?

AndyPook:
Ahh, sorry missed that, been spending too long in AzDevOps.
Cool, learned another something new today :)

As all these bits are internal, I've got a hacked up method using `.ExecuteAsync("XINFO" ...`
I am "not allowed" to use pre-release packages. When do you think this would get a nuget stable release?

mgravell:
I need to revisit a possible glitch in the mutex code before I can do that; have my hands busy with 16 other things at the moment, but: it is on my list!

PR #1492, opened 2020-06-09
Draft Only: net462 break repro

NickCraver:
Been debugging this tonight - we're firing off messages to the pipe but never getting any back on `net4x`...so yeah, confirmed something breaks with `System.IO.Pipelines` in 4.7.2 and it's subtle. Will continue in the morning.

cc @mgravell to eliminate dupe work - debuggable console app in the repro allows breakpoints, etc. It _does_ repro in DEBUG so fully observable.


NickCraver:
Fixes merged in - closing this one out

Issue #1438, opened 2020-04-18
Get & Set Failing After Error Cascade


JKurzer:
Aloha!

### The Problem
Under certain loads and after long up-times, we've begun seeing a curious error during addorset (Reproduced at the end).
This is accompanied by a much higher number of errors for Gets (Reproduced at the end) that are roughly identical, though we do use get far more than set.

### Additional
Once a client becomes symptomatic, it rapidly begins throwing these errors on most or all operations. Bouncing that client causes another client to become symptomatic in a similar way, which may be a red herring but was interesting enough to bear mention.

We currently use only one multiplexer, which seems like a possible cause, but before we refactor, I thought I'd check in. Our application is pretty much pure C#. We are seeing time-outs as well, during symptomatic periods. My assumption is that [StackExchange.Redis/ConnectionMultiplexer.cs#L2601](https://github.com/StackExchange/StackExchange.Redis/blob/master/src/StackExchange.Redis/ConnectionMultiplexer.cs#L2601) is the top level call in the multiplexer for the error, given that 2622 is part of error handling deferral. 

### Set Error:
```
REDACTED REDIS AddOrSet Failure on key REDACTED
    StackExchange.Redis.RedisConnectionException: Failed to write ---> System.InvalidOperationException: Concurrent reads or writes are not supported.
       at System.Runtime.ExceptionServices.ExceptionDispatchInfo.Throw()
       at System.IO.Pipelines.PipeCompletion.ThrowLatchedException()
       at System.IO.Pipelines.Pipe.GetFlushResult(FlushResult& result)
       at System.IO.Pipelines.Pipe.PrepareFlush(CompletionData& completionData, ValueTask`1& result, CancellationToken cancellationToken)
       at System.IO.Pipelines.Pipe.FlushAsync(CancellationToken cancellationToken)
       at System.IO.Pipelines.Pipe.DefaultPipeWriter.FlushAsync(CancellationToken cancellationToken)
       at Pipelines.Sockets.Unofficial.SocketConnection.WrappedWriter.FlushAsync(CancellationToken cancellationToken) in /_/src/Pipelines.Sockets.Unofficial/SocketConnection.cs:line 434
       at StackExchange.Redis.PhysicalConnection.FlushAsync(Boolean throwOnFailure) in /_/src/StackExchange.Redis/PhysicalConnection.cs:line 907
[...]

PR #1496, opened 2020-06-11
Temporary PR to help add debug code to diagnose concurrent read/write issue

Plasma:
This is a temporary PR to help add extra logging to the error discussed in https://github.com/StackExchange/StackExchange.Redis/issues/1438 and is supposed to be a temporary build.

This has been based off `v2.1.30` tag.


Plasma:
(Did not intend to make public PR, removed)

PR #1525, opened 2020-07-07
Fix sentinel default port when connecting with configuration string

ejsmith:
This makes it so that when using a configuration string to connect that we only parse the configuration string and not set any defaults on it until it is forwarded on to the connect methods that take `ConfigurationOptions`


niemyjski:
@mgravell Any chance we can get this reviewed and merged?

ejsmith:
@NickCraver @mgravell looks like you are getting ready to do another release. Think we can this small change merged? :-)

PR #1531, opened 2020-07-15
JRediSearch version 1.8 Sync Up

tombatron:
So here is my stab at another sync up with JRediSearch. 

There's one change that I didn't migrate over and it's their `toString` override on the `Document` class. They introduced `com.google.gson.Gson` in order to serialize the instance and I figured we wouldn't want to introduce a JSON serializer as a new dependency. 

Another thing that I noticed is the tests that involve "autosuggest" worked fine... once. Turns out that the "autosuggest" indexes aren't "regular" RediSearch indexes but presented more like standard Redis keys. You'll notice that in the tests that exercise the "autosuggest" functionality I added calls to `Db.KeyDelete(cl.IndexName);` instead of augmenting the `Reset` method because I figured it was just five or so tests. 

In any event the new functionality added here is:

- Support for `IF {condition}` when updating a document. 
- Executing batches of search queries (leveraging pipelining).
- Support for retrieving and setting RediSearch configuration at runtime.
- Support for index aliases. 
- Support for managing synonyms.

### DIFF
--- a/src/NRediSearch/AddOptions.cs
+++ b/src/NRediSearch/AddOptions.cs
@@ -27,6 +27,7 @@ namespace NRediSearch
         public string Language { get; set; }
         public bool NoSave { get; set; }
         public ReplacementPolicy ReplacePolicy { get; set; }
+        public string Filter { get; set; }
 
         /// <summary>
         /// Create a new DocumentOptions object. Methods can later be chained via a builder-like pattern
@@ -57,9 +58,11 @@ namespace NRediSearch
         /// Indicate the behavior for the existing document.
         /// </summary>
         /// <param name="mode">One of the replacement modes.</param>
-        public AddOptions SetReplacementPolicy(ReplacementPolicy mode)
+        /// <param name="filter">Updates the document only if a boolean expression applies to the document</param>
+        public AddOptions SetReplacementPolicy(ReplacementPolicy mode, string filter = null)
         {
             ReplacePolicy = mode;
+            Filter = filter;
             return this;
         }
     }
--- a/src/NRediSearch/Client.cs
+++ b/src/NRediSearch/Client.cs
@@ -1415,6 +1415,92 @@ namespace NRediSearch
             return (string)await _db.ExecuteAsync("FT.ALIASDEL", name).ConfigureAwait(false) == "OK";
         }
 
+        /// <summary>
+        /// Create a new synonyms group.
+        /// </summary>
+        /// <param name="terms">Terms to add to the synonym group.</param>
+        /// <returns>ID of the synonym group.</returns>
+        public long AddSynonym(params string[] terms)
+        {
+            var args = new object[terms.Length + 1];
+
+            args[0] = _boxedIndexName;
+
+            Array.Copy(terms, 0, args, 1, terms.Length);
+
+            return (long)DbSync.Execute("FT.SYNADD", args);
+        }
+
+        /// <summary>
+        /// Create a new synonyms group.
+        /// </summary>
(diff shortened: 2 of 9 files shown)


mgravell:
This looks very interesting! There's a lot here, and I'm going to have to go through it when I get a sec, but: on my list!

tombatron:
Sounds good Marc thanks!

NickCraver:
Hey Tom, I know this sat for a long time because we really aren't in a position to review it - as non-users of RediSearch or the Java version we just don't know what to do or what's correct past "yeah...the code works". I split out the repo into https://github.com/StackExchange/NRediSearch so that it could get maintainers that are knowledgeable about these things and make calls, not waiting on us. We'd need to transition the repo/library ownership to someone else but...is this of any interest? If not I totally understand (as we don't have the time to dedicate either), but to merge this PR we'd need to re-target it at the other repo. I kept this PR in mind when splitting, and it's based on the same git history so we _should_ be able to aim the branch there and quickly re-create the PR if you desire to go that route.

We haven't had time or expertise on this for some time...should have put a notice of that much sooner.

tombatron:
Alrighty, I managed to mostly apply my original patch to the new repository (there are a few issues but nothing huge). 

After doing that I figured I would bring NRediSearch back up to the latest version of JRediSearch once again only to find out that the maintainers (Redis, Inc) have deprecated JRediSearch and merged support for RediSearch directly into Jedis. 

In the time between opening my pull request for NRediSearch and the present, I had created a new package called "RediSearchClient" which is intended to target the 2.x+ version(s) (which is a huge departure from 1.x). But with the advent of a separate repository for NRediSearch I figured we could work towards once again reaching parity with JRediSearch (which is not deprecated). 

So now I'm kind of at a crossroads. 

I saw NRediSearch as a .NET available port of JRediSearch (along the same lines of nHiberate and Hiberate). But with JRediSearch being deprecated, does this make sense anymore?

NickCraver:
@tombatron That's a good question - since I updated this issue we also became aware of https://github.com/redis/redis-om-dotnet - so I'm leaning towards...maybe we should deprecate the repo, and point at your client and Redis OM? What are your thoughts? Having another client just because it _was_ there when it's been leap-frogged but the surface area isn't that huge to begin with, maybe we make the call now to direct people elsewhere? Or should we try and keep this alive for other reasons? I'm grasping at what value we really add maintaining another client here in the landscape, but not fully informed enough to be aware of the implications. Very curious where you're thinking.

tombatron:
tl;dr; I think the only reason to keep NRediSearch around would be to support EoL'd versions of RediSearch. Anyone running the latest version of RediSearch (2.2) and RedisJson (2.0) should probably use Redis OM. 

-------------

You know, I'd heard of Redis OM as well and I didn't consider that it might obviate the need for a separate RediSearch client. 

But skimming the source and documents, I can't imagine why I WOULDN'T want to leverage it instead of the client that I made (Looking specifically at the IQueryable support as well as a more friendly means of defining indexes). So I have some stuff to thinking about regarding the future of "RediSearchClient". 

I think having a client like NRediSearch stick around for users who decide not to upgrade beyond RediSearch 1.6 (which goes EoL in March) could be helpful... I suppose.

NickCraver:
@tombatron For those people - it's on NuGet so we're not taking away the status quo, just not continuing to fix anything by leaving it be...that's kinda how I lean here, updating the README to point people elsewhere. Thoughts?

NickCraver:
@tombatron sounds good - I shall close this out to wrap it up on this repo <3

Issue #1501, opened 2020-06-16
ServerSelectionStrategy set to standalone for a clustered cache


deepakverma:
Hi @mgravell ,
I think i might have found an issue where for a clustered cache mux can get into a bad state where requests start failing until mux is teared down or a reconnect occurs.
While debugging I observed that the cause is likely due to mux level serverselection strategy being standalone whereas serverendpoints have been detected as cluster. serverselect looks for         AnyConnected which fails to return any endpoint since endpoint.servertype(which is cluster) doesn't match serverselectstratergy.servertype(with is standalone). 
During my test, task faulted due to a connection drop when the ping was being sent. Connection got restored and reconfigure was called again but didn't proceed due to re-entrancy condition in reconfigureasync
(Reconfiguration was already in progress due to: connection failed, attempted to run for: connection failed)
I see in reconfigureAsync if the task is faulted it would result clustercount to be 0 which makes serverselectionstratergy to be set as standalone 
https://github.com/StackExchange/StackExchange.Redis/blob/master/src/StackExchange.Redis/ConnectionMultiplexer.cs#L1762 
https://github.com/StackExchange/StackExchange.Redis/blob/master/src/StackExchange.Redis/ConnectionMultiplexer.cs#L1852 (clustercount is 0)
Before proposing a fix, I wanted to check your opinion on this theory .

Thanks.

mgravell:
sounds interesting; it sounds like you were going to try a PR? if so: fine, otherwise I can add this to my backlog to look at (second in the queue, just under .NET Core latest failing)

dceravigupta:
I reported the same problem few months back and here is the PR which I submit to fix this problem. Probably we might want to reconsider that same again.

https://github.com/StackExchange/StackExchange.Redis/pull/1444

deepakverma:
@dceravigupta @mgravell 
I am proposing a fix by making it skip setting the strategy when servertype was undetermined. Default will still remain as standalone. But during a reconnect if the task to determine servertype is faulted, it won't reset the strategy to standalone. 
https://github.com/StackExchange/StackExchange.Redis/pull/1532

deepakverma:
ping

deepakverma:
Hi @mgravell , I have a pull request here #1532, can you review it. thanks.

kulku:
Hoping for a decision sooner from @mgravell. Thank you.

blackie1019:
Hi, @mgravell 

[...]

PR #1532, opened 2020-07-15
Don't set a default serverselection stratergy in case there was a fault getting the servertype

deepakverma:
#1501


joshsten:
@NickCraver Did this ever succeed to merge? I noticed one of the checks failing above. We are interested in this fix as well.

PR #1536, opened 2020-07-22
Adding support for cluster keyspace notifications to subscriber

### LINKED ISSUE #789: Cluster pub/sub: keyspace notification issues

ercangorgulu:
Adding support for cluster keyspace notifications to subscriber by allowing subscription to multiple nodes

Refactored code for previous pull request based on new changes
https://github.com/StackExchange/StackExchange.Redis/pull/813

Fixes: https://github.com/StackExchange/StackExchange.Redis/issues/789


shaunco:
@NickCraver - Any chance we can get this merged? Getting keyspace notification support would be amazing.

PR #1544, opened 2020-07-30
add support for Index Definition

gkorland:


src/NRediSearch/Client.cs
```diff
@@ -50,17 +50,96 @@ public enum IndexOptions
...
         {
             // This news up a enum which results in the 0 equivalent.
             // It's not used in the library and I'm guessing this isn't intentional.
             public static IndexOptions Default => new IndexOptions();
 
             private IndexOptions _options;
+            private IndexDefinition _definition;
             private string[] _stopwords;
-            public ConfiguredIndexOptions(IndexOptions options = IndexOptions.Default)
+            public ConfiguredIndexOptions(IndexOptions options = IndexOptions.Default, IndexDefinition definition = null)
```
mgravell, line 139: this is a binary-incompatible change; if possible, I would rather avoid hard breaks - could we do instead:

``` c#
public ConfiguredIndexOptions(IndexOptions options)
  =>ConfiguredIndexOptions(options, null);

public ConfiguredIndexOptions(IndexOptions options = IndexOptions.Default, IndexDefinition definition = null)
{...}
```

which will have the same API usage, but isn't a break

gkorland: :+1:

gkorland:
@mgravell the code is pretty much ready but the tests can't pass since the tests are running on `redislabs/redisearch:latest` docker container. But, some of the tests won't pass on `redislabs/redisearch:edge` since some of the APIs were removed in Search 2.0.
Is there a way to run the test with latest and search 2.0 and run only the relevant tests for each version.

mgravell:
Review request noted; should be able to look today

src/NRediSearch/Client.cs
```diff
@@ -50,19 +50,104 @@ public enum IndexOptions
...
         }
 
+        public sealed class IndexDefinition
+        {
+            public enum IndexType
+            {
+                /// <summary>
+                /// Used to indicates that the index should follow the keys of type Hash changes
+                /// </summary>
+                HASH
```
mgravell, line 60: minor naming nit; .NET standards would make this `Hash` (if you need `HASH` for some text purposes, that can be done via translation in any of a number of ways); note also that .NET enums are not checked - and any value (`42`, for example) can be passed in; you can tweak both of these in a simple way - will add example below (in the "add args" step)

src/NRediSearch/Client.cs
```diff
@@ -50,19 +50,104 @@ public enum IndexOptions
...
+                _language = language;
+                _scoreFiled = scoreFiled;
+                _score = score;
+                _payloadField = payloadField;
+            }
+
+            internal void SerializeRedisArgs(List<object> args)
+            {
+                args.Add("ON".Literal());
+                args.Add(_type.ToString("g"));
```
mgravell, line 90: to handle both casing and the unknown enums, consider:

``` c#
args.Add(_type switch {
    IndexType.Hash => "HASH".Literal(),
    _ => throw new InvalidOperationException($"Unexpected index kind: {_type}")
});
```

src/NRediSearch/Client.cs
```diff
@@ -50,19 +50,104 @@ public enum IndexOptions
...
+        {
+            public enum IndexType
+            {
+                /// <summary>
+                /// Used to indicates that the index should follow the keys of type Hash changes
+                /// </summary>
+                HASH
+            }
+
+            internal readonly IndexType _type = IndexType.HASH;
```
mgravell, line 63: do these fields really need to be `internal`? feels very unusual; I would *expect* `private` here; it isn't a blocker, just... feels odd

src/NRediSearch/Client.cs
```diff
@@ -84,17 +169,21 @@ public ConfiguredIndexOptions SetNoStopwords()
...
                     args.Add("STOPWORDS".Literal());
                     args.Add(_stopwords.Length.Boxed());
                     args.AddRange(_stopwords);
                 }
             }
 
-            internal static void SerializeRedisArgs(IndexOptions options, List<object> args)
+            internal static void SerializeRedisArgs(IndexOptions options, List<object> args, IndexDefinition definition)
             {
+                if (definition != null)
```
mgravell, line 183: this is absolutely 100% fine as written; just a C# language tip: this is semantically identical to

``` c#
definition?.SerializeRedisArgs(args);
```

src/NRediSearch/Client.cs
```diff
@@ -458,6 +547,7 @@ public Task<bool> ReplaceDocumentAsync(string docId, Dictionary<string, RedisVal
         /// <param name="score">the document's index score, between 0 and 1</param>
         /// <param name="replace">if set, and the document already exists, we reindex and update it</param>
         /// <returns>true on success</returns>
+        /// [Deprecated] Use IDatabase.HashSet instead.
```
mgravell, line 550: in C#, you can use `ObsoleteAttribute` for this, and it will be advertised to the caller, for example:

``` c#
/// <returns>true on success</returns>
[Obsolete("Use IDatabase.HashSet instead.")]
public bool AddHash(RedisKey docId, double score, bool replace)
```

Note that there is an optional boolean you can add, which determines whether it is a warning or a hard error. Since the old code remains, I'm guessing leaving it as the default (warning) is fine.

mgravell (changes requested):
Some minor feedback, but looking good; the suggested tweaks are minor - I'm happy to take them if you're busy

src/NRediSearch/Client.cs
gkorland: Thanks, I'll rename, there is no casing thing here, `Hash` will do.

src/NRediSearch/Client.cs
gkorland: There is no need to handle casing, both `Hash` and `HASH` will work.

src/NRediSearch/Client.cs
gkorland: I was just following the existing standard, the reset of the fields bellow are also marked as internal.

src/NRediSearch/Client.cs
gkorland: thanks

src/NRediSearch/Client.cs
gkorland: Should I remove the `[Deprecated]` from the docs? I see that in some places in the code we do use it

mgravell: if there is an `[Obsolete]`, then it is self-evident and doesn't **need** additional remarks, but those remarks won't do any harm either; either way is fine

src/NRediSearch/Client.cs
mgravell: fair enough :)

mgravell:
any chance you could `git merge main` / `git push`? there shouldn't be any nasty conflicts, but I've fixed up a ton of things that should help us see how it is behaving on the CI server.

Running locally, it now does a lot better than `main`, but I get 2 persistent fails and 1 intermittent fail; the persistent ones are:

- `ClientTest.TestGetSuggestionWithScore` - Missing required fields:  score not within range: 1.05830061435699
- `ClientTest.TestAddSuggestionGetSuggestionPayloadScores` - score not within range: 1.45960092544556

and the intermittent one (presumably timing related?) is:

- `ClientTest.Search` - `Document already exists`

(maybe disable concurrency?)

This is using:

``` txt
127.0.0.1:6385> module list
1) 1) "name"
   2) "search"
   3) "ver"
   4) (integer) 20001
```

mgravell:
What is odd here; the *test* is consistent with JRediSearch: https://github.com/RediSearch/JRediSearch/blob/master/src/main/java/io/redisearch/Suggestion.java#L97-L99 - however, I'm definitely seeing scores outside that range:

``` txt
127.0.0.1:6385> FT.SUGGET "ClientTests/ClientTest.cs:TestAddSuggestionGetSuggestionPayloadScores" "COU" MAX 5 WITHSCORES WITHPAYLOADS
 1) "COUNT NO PAYLOAD OR COUNT"
 2) "2.7106873989105225"
 3) (nil)
 4) "COUNT_ANOTHER"
 5) "0.30151134729385376"
 6) (nil)
 7) "COUNT"
 8) "0.11547005921602249"
 9) "My PAYLOAD is better"
10) "COUNT_ME TOO"
11) "0.063245557248592377"
12) "PAYLOADS ROCK "
```

so... I'm kinda confused as to what is *meant* to happen here

mgravell:
@gkorland to be clear: I 100% appreciate that the "new" failures are nothing whatsoever to do with index creation, so from the perspective of "what is this PR", I'm happy to merge without fixing that; however; being realistic: you're better placed than me to have an opinion on the "new" failures (again, air quotes intentional). Is this a bug in the 2.0 module? Or is the library wrong to enforce the range here? Any ideas? If not: we can still merge.

NickCraver:
Probably worth a hangout when I'm back in next week...the issue here with Redisearch v1 and v2 is very deep, the entire relationship structure has changed. I'll try and get v1 testing stable, but v2 is a whooooooole other thing.

NickCraver:
I got the same local 2-fails only, but that's trickery - need to `FLUSHDB` before the run to repro the 25 failures in CI, mostly around duplicate document names. In looking at how JRediSearch _isn't_ hitting these things, I see that it's effectively calling FLUSHDB before _every_ test (and sometimes inside as well). I'd put forth that it's not really testing multi-index scenarios or parallel friendly, and those are the items we're running into in the test runs on NRediSearch here.

Thoughts on how we best proceed here?

NickCraver:
As an example, adding a flush to the base constructor makes all tests pass by putting the expectations on par with JRediSearch, flushing before every test:
```cs
        protected RediSearchTestBase(ITestOutputHelper output)
        {
            muxer = GetWithFT(output);
            Output = output;
            Db = muxer.GetDatabase();
            var server = muxer.GetServer(muxer.GetEndPoints()[0]);
            server.FlushDatabase();
        }
```

I don't have strong feelings here - my notion is get us on par, and then discuss what _should_ happen on both of these. Thoughts?

NickCraver:
Discussed in DM a bit (Marc's mobile) - I think the build side of this is ready and now good for re-review.

mgravell (approved):

mgravell:
With the build etc changes, everything now passes; we need to think a little bit about how we actually deploy/release this. In particular, the 2.0 changes *look* to make it effectively incompatible with 1.x, so; if we merge this, we should try to advertise (release notes, semver, etc) that this is the client for 2.0 (and that anyone using a different server product should use "some earlier" build).

Unfortunately the versioning was previously mirrored to the SE-Redis versioning, so we can't just make this 2.0 - an error of planning, in hindsight. We *should* probably split the versions, though. We're currently at 2.1.30, so maybe split the versioning and make the 2.2.x for now, moving to 3.0 when some possible hypothetical RediSearch 3.0 is deployed? (then keep the major/minor in sync with the server, and use the 3rd part for the library).

I'm also still a little dubious about the `TestAddSuggestionGetSuggestionPayloadScores`,`TestGetSuggestionWithScore` behavior. It looks like this was *largely* caused by the data in the server being stomped because of timing, but what gives me cause to be concerned here is:

- the server *really did* return results with numbers > 1
- any timing things are just symptoms of concurrency, which we *expect* to happen in real world

To me this suggests that there is *either* a server bug (and the result > 1 should never have been returned), *or* the library (both `NRediSearch` and `JRediSearch`) is fundamentally incorrect to perform range checking on this number. I'm happy to take this as a side issue, and for now I've added [this change](https://github.com/gkorland/StackExchange.Redis/commit/f050f3f1741273c2f611a62e955ea95bd8b8580e) which means we will accept what the server says *regardless* of what the server says, which seems pragmatic to me.

Main remaining discussion is probably versioning and deployment?

mgravell:
(adding review back to @NickCraver specifically for the last commit)

NickCraver (approved):
Changes looks good - I don't have strong opinions about version, but we can either: a) split the repo out (maybe best long term), or b) put a `version.json` in the `src/NRediSearch/` directory (best short term) option.

It occurs to me that v2 is just _hugely_ breaking, and this doesn't really address all of it - there are several things in 2.0 that are deprecated and forwarded for the 2.x timeframe on the RediSearch side - if we do a major version break here for "use this for RediSearch 2.0+", IMO we should go through and address _all_ of those, and drop a few `[Obsolete]` as well.

gkorland:
@NickCraver @mgravell sorry I missed the all thread on the PR, I see that there no decision was taken on the PR

mgravell:
(discussing with Guy off-board)

AstronusX:
Any idea when this change is going release to nuget? I need the prefix functionality

mgravell:
It actually already is, but I have slipped on the release notes to finish making it public/visible. I'll try and do that today. But: try specifying version 2.2.11

Issue #1451, opened 2020-05-02
Feature request - add command: ROLE


NickCraver:
This command isn't yet in the lib and is useful for a few scenarios - came up in #1431.

Min version: 2.8.12
Docs: https://redis.io/commands/role

zmj:
I would like to take a shot at this! A couple questions:

* Result type: it seems like there's not much shared between the results for each node type. I was thinking to return a base Role type with some subtypes for master, replica, and sentinel?
* Tests: it seems like the use case for this command is mostly around configurations with sentinels? Is that a sensible place to put tests for this command?

mgravell:
Given that the shape changes so radically, inheritance is probably the best
option here, agreed. There is also an enum for the same purpose, but to be
honest I don't think we need the discrimiant once we have split it into
types. The intent is also perfectly valid in primary/replica scenarios, so
all 3 result types probably need testing. We should also consider what to
return when the type isn't recognised ... an exception? A RoleUnknown
subtype? I'd also recommend looking for the string "replica" in addition to
"slave", in case that ever swaps over.

Looking forward to it!

NickCraver:
Added in #1551, closing out!

PR #1551, opened 2020-08-15
Implement ROLE command

zmj:
This PR implements the ROLE command (#1451): https://redis.io/commands/role#sentinel-output . Redis implementation: https://github.com/redis/redis/blob/unstable/src/replication.c#L2672

Because the result format is pretty different for each instance type, we return a base Role type that the caller has to type check to get to the values specific to the instance type.

I've tested the master and replica cases locally. I haven't tested sentinel, but that one is the most straightforward.

### DIFF
--- a/src/StackExchange.Redis/Interfaces/IServer.cs
+++ b/src/StackExchange.Redis/Interfaces/IServer.cs
@@ -330,6 +330,7 @@ namespace StackExchange.Redis
         /// Get summary statistics associates with this server
         /// </summary>
         ServerCounters GetCounters();
+
         /// <summary>
         /// The INFO command returns information and statistics about the server in a format that is simple to parse by computers and easy to read by humans.
         /// </summary>
@@ -422,6 +423,18 @@ namespace StackExchange.Redis
         /// <param name="options">The options to use for this topology change.</param>
         /// <param name="log">The log to write output to.</param>
         void MakeMaster(ReplicationChangeOptions options, TextWriter log = null);
+
+        /// <summary>
+        /// Returns the role info for the current server.
+        /// </summary>
+        /// <remarks>https://redis.io/commands/role</remarks>
+        ServerRole RoleGet();
+
+        /// <summary>
+        /// Returns the role info for the current server.
+        /// </summary>
+        /// <remarks>https://redis.io/commands/role</remarks>
+        Task<ServerRole> RoleGetAsync();
 
         /// <summary>
         /// Explicitly request the database to persist the current state to disk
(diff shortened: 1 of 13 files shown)


src/StackExchange.Redis/Interfaces/IServer.cs
```diff
@@ -1042,14 +1055,5 @@ internal static class IServerExtensions
         /// </summary>
         /// <param name="server">The server to simulate failure on.</param>
         public static void SimulateConnectionFailure(this IServer server) => (server as RedisServer)?.SimulateConnectionFailure();
-
-        public static string Role(this IServer server)
```
zmj, line 1046: Removed the existing test helper and changed the usage to use the new code

src/StackExchange.Redis/Role.cs
```diff
@@ -0,0 +1,133 @@
...
+
+            /// <summary>
+            /// The port number of the master node for this replica.
+            /// </summary>
+            public int MasterPort { get; }
+
+            /// <summary>
+            /// This replica's replication state.
+            /// </summary>
+            public string State { get; }
```
zmj, line 92: This could be an enum, but returning the raw value seemed better for allowing the caller to do something useful with an unrecognized value.

tests/StackExchange.Redis.Tests/Roles.cs
```diff
@@ -0,0 +1,57 @@
...
+            Assert.Contains(master.Replicas, r =>
+                r.Ip == TestConfig.Current.ReplicaServer &&
+                r.Port == TestConfig.Current.ReplicaPort);
+        }
+
+        [Fact]
+        public void ReplicaRole()
+        {
+            var connString = $"{TestConfig.Current.ReplicaServerAndPort},allowAdmin=true";
+            using var muxer = ConnectionMultiplexer.Connect(connString);
```
zmj, line 37: Is it safe to assume there is always a replica in the test run?

zmj:
A few tests failed. This one looks like a bug: `StackExchange.Redis.RedisCommandException : This operation has been disabled in the command-map and cannot be used: ROLE`. I guess I need to add ROLE to one (or more) of the lists in CommandMap? Which ones are appropriate?

The others looked like network errors that were probably unrelated to this change.

mgravell:
On the failing test: look at public static CommandMap Sentinel

zmj:
MultiMaster.TestMultiNoTieBreak [5s 36ms] is still failing. Doesn't seem related to these changes. Any suggestions?

mgravell:
Ignore that one. I'll look into it.

mgravell (approved):

PR #1563, opened 2020-09-01
ForceReplication during failovers to reduce errors in clustered scenarios

aravindyeduvaka:
Multiple people had previously complained that whenever there is a cluster failover, they experience Connection Exceptions for at least close to a minute. 
During investigation, @joroda  found that this is due to the fact that config check is done roughly every 60 seconds by default and during failover it has to be done almost immediately to eliminate the errors. 

This fix shows significant improvement from the tests we did. it reduces errors to around 2 seconds for the 90th percentile case compared to around 56 seconds of errors for 90th percentile right now.

### DIFF
--- a/src/StackExchange.Redis/ServerEndPoint.cs
+++ b/src/StackExchange.Redis/ServerEndPoint.cs
@@ -540,6 +540,7 @@ namespace StackExchange.Redis
         internal bool CheckInfoReplication()
         {
             lastInfoReplicationCheckTicks = Environment.TickCount;
+            ForceReplicationCheck = false;
             PhysicalBridge bridge;
             if (version >= RedisFeatures.v2_8_0 && Multiplexer.CommandMap.IsAvailable(RedisCommand.INFO)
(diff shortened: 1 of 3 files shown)


src/StackExchange.Redis/ServerEndPoint.cs
```diff
@@ -530,6 +530,13 @@ internal int LastInfoReplicationCheckSecondsAgo
             get { return unchecked(Environment.TickCount - Thread.VolatileRead(ref lastInfoReplicationCheckTicks)) / 1000; }
         }
 
+        internal bool MarkedForForceReplicationCheck
+        {
+            // TODO: consider multi threading
```
NickCraver, line 535: Instead of a property here you can make this a `volatile bool` field :)

aravindyeduvaka: Sure. I will make the change.

aravindyeduvaka:
@NickCraver, @mgravell ping!

NickCraver (approved):
Looking good, thanks for the tweaks!

Issue #992, opened 2018-10-31
SortedSetRangeByValue not handling null values


srulandhs:
I'm attempting to use a Redis SortedSet in lexical compare mode. I am unable to get the StackExchange.Redis library to behave the same as when I issue the commands directly thur the redis console.

Versions
Redis: 4.0.0 via docker
Redis Desktop Mgr: 0.9.5
StackExchange.Redis: 2.0.513

Seed data
SortedSetAdd("testkey", "a:1000", 0, When.NotExists);
SortedSetAdd("testkey", "b:2000", 0, When.NotExists);
SortedSetAdd("testkey", "c:3000", 0, When.NotExists);
SortedSetAdd("testkey", "d:4000", 0, When.NotExists);
SortedSetAdd("testkey", "e:5000", 0, When.NotExists);
SortedSetAdd("testkey", "f:6000", 0, When.NotExists);

SortedSetRangeByValue("testkey", "a", RedisValue.Null)
yields: no matches

however
zrangebylex testkey [a +
yields
a:1000
b:2000
c:3000
d:0000
e:5000
f:6000

In looking through the code, I think that the issue may be in ReverseLimits()
The first line issues a start.CompareTo(stop). The null RedisValue seem to return true here which causes the min and max to be swapped.
Due to this swap it seems that the following is being executed:
zrangebylex testkey - [a

sircrowbar:
I can confirm this to also be the case. If I use SortedSetRangeByValue with a min set and a max set to RedisValue.Null, it will try and swap it so that the minimum is the "-" value and the max is the value I intend to be the minimum. Would love to see a fix for this!

nickchan9394:
Encounter this issue today as well. My case is searching value by lex with start string "05" and end string "05\xff".  but the start string "05" and the end string "05\xff" get swapped before the command sent to the database. After going through the source code, it turns out is the "RedisValue.Simplify" changed the "05"(string) to 5 (integer) but keep the "05\xff" as a string, and finally compare them both as strings, which result in the start string is greater than the end string, then the swapping happens. Would be grateful if anyone can look into this. Thanks.

NickCraver:
Fixed in #1573, thanks @tombatron!

PR #1573, opened 2020-09-22
Do not invert the range if the `max` value is `default(RedisValue)`

tombatron:
If the `stop` value is `default(RedisValue)` and is compared against a `start` value that isn't the result from `start.CompareTo(stop)` is `1`, which would reverse the range, which would cause the library to emit an incorrect query to Redis. 

I figured this change would have far fewer impacts than changing how `RedisValue.CompareTo` works. 

This should close issue #992


tombatron:
It looks like the tests that are failing are all related to NRediSearch:

```
StackExchange.Redis.RedisException : NRedisSearch module missing on 127.0.0.1:6385
```

mgravell (approved):

mgravell:
merged, thanks

Issue #1572, opened 2020-09-22
The edge case which causes TimeoutBeforeWrite when using transaction


devbv:
Hi guys, Our team have been using SE.Redis over three years on the production. Before I start, I just wanna say thanks to all contributors of this project. 🙏 

After we switched to SE.Redis version 2, we found some microservice servers were in trouble with TimeoutBeforeWrite errors.
We didn't have enough time to dig in at that time, so we just worked around by making a reconnection logic.
Recently, we build a new microservice server which uses transactions heavily and produces a lot of TimeoutBeforeWrite errors.
After the debugging, I found the edge case from using transaction causes TimeoutBeforeWrite.

The timeouts are caused by the combination of several natural behaviors.
1. Multiple ProcessBackLog() can be scheduled at the same time. (of course, the only one acquires lock and does processing)
2. GetMessages() of TransactionMessage includes Monitor.Wait to wait conditions completed.
3. Pipe sender, reader and ProcessBacklog() run on the same thread-pool.

Suppose, we have N workers(threads) and wait transaction conditions to be finished in GetMessages().
Normally, pipe sends and reads are scheduled on N-1 workers and GetMessages() finishes immediately.
But with heavy traffic, the situation that all N-1 workers run ProcessBackLog() and wait for lock acquire can be happened.
(a consumer makes the number of backlog 1->0 infinitely and many producers schedule new ProcessBacklog() after they make the number of backlog 0->1.)
As GetMessages() is running with lock, all ProcessBackLog() doesn't release thread.
As there's no thread to send and read through Pipe, Monitor.Wait in GetMessages() eventually reaches to timeout.
Furthermore, when it happens, most messages in backlog just set as TimeoutBeforeWrite because we spend 5 seconds(default timeout) in GetMessages().
```
The timeout was reached before the message could be written to the output buffer, and it was not sent, command=EXEC, timeout: 5000, outbound: 0KiB, inbound: 0KiB
```

To reproduce this situation easier, change a worker count less than default value(10). However, I could reproduce the case even with 16 workers.

My temporal workaround for this issue is to use dotnet thread pool. After we change the option to use dotnet thread pool, our production servers become stable. It seems that hill-climbing of dotnet thread pool is helpful in somehow.
But you know, it's not an ideal solution and I believe you guys could solve this issue fundamentally.
(+ I also submitted simple PR to fix it.)

Here's testing code that I used for debugging.

```
namespace TestConsole
{
    internal static class Program
    {
[...]

PR #1574, opened 2020-09-23
Prevent too many ProcessBacklog() from being scheduled

devbv:
This PR prevents the situation that a consumer makes the number of backlog 1->0 infinitely and many producers schedule new ProcessBacklog() after they make the number of backlog 0->1.
The situation could lead a starvation of thread pool and a lot of timeout when we use transaction heavily. (#1572)

### DIFF
--- a/src/StackExchange.Redis/PhysicalBridge.cs
+++ b/src/StackExchange.Redis/PhysicalBridge.cs
@@ -850,14 +850,20 @@ namespace StackExchange.Redis
                 // so now we are the writer; write some things!
                 Message message;
                 var timeout = TimeoutMilliseconds;
-                while(true)
+                var messages = new Queue<Message>();
+
+                _backlogStatus = BacklogStatus.CheckingForWork;
+                lock (_backlog)
                 {
-                    _backlogStatus = BacklogStatus.CheckingForWork;
-                    lock (_backlog)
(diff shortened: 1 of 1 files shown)


devbv:
@mgravell Hello, Could you take a look for this MR or issue #1572 ?

arsnyder16:
@mgravell I am also seeing these errors in my logs. They are a semi related to the issue solved with #1585 in that in these timeout scenarios seems to cascade and build upon each other, and the library struggles to recover.  

I would also be interested in seeing your opinion on getting this change in as well

devbv:
image
Additional info for better understanding. This is stacktrace of threads when the timeout is occurred. You can see all worker threads are stopped at MutexSlim.TakeWithTimeout called by PhysicalBridge.ProcessBacklog()

NickCraver:
Merging main in here for build check changes

mgravell:
This is a great find, and could explain a lot, but I'm not sure that the *approach* is the best way of solving it - I'm playing with an alternative strategy in #1612 - but: even if I merge that, I'll still give credit for the research here; much appreciated!

mgravell:
Have merged #1612, but credited in release notes:

> - fix backlog contention issue (#1612, see also #1574 via devbv)

Issue #1450, opened 2020-04-29
LPUSHX/RPUSHX support for multiple values


Badabum:
Currently there are `ListLeftPush/ListRightPush` methods with 2 signatures: 
1. Accepting single value and condition, commandoptions: 
`(string key, RedisValue value, When, ComandOptions)`
2. Accepting key, multiple values and command options:
 `(string key, RedisValue[] value, CommandOptions options)`

According to Redis docs - starting from v4.0 the `LPUSHX/RPUSHX` accepts multiple elements. Is there any chance support will be added to the library?

mgravell:
Yep, this checkout out; so, I guess we need:

``` c#
// new API
long ListLeftPush(RedisKey key, RedisValue[] values, When when = When.Always, CommandFlags flags = CommandFlags.None);
// remove default parameter on existing API
long ListLeftPush(RedisKey key, RedisValue[] values, CommandFlags flags);
```

- and the core implementation (should be trivial)
- and the database wrapper implementation
- and the same for async
- and tests

Any takers?

Badabum:
I could make a try :)

dmytrohridin:
Hey @mgravell, trying to implement this but do not understand why we should remove default parameter `CommandFlags` from  `long ListLeftPush(RedisKey key, RedisValue[] values, CommandFlags flags = CommandFlags.None);`
Does it mean that not all `CommandFlags` value should be acceptable?
Thank you

mgravell:
> why we should remove default parameter 

Context? What are we talking about here?

dmytrohridin:
>  // remove default parameter on existing API
long ListLeftPush(RedisKey key, RedisValue[] values, CommandFlags flags);
About this. You mentioned this earlier in this issue

mgravell:
No, this has no crossover whatsoever with any flags. What we're trying to do here is make the API backwards compatible but prefer the new API, nothing more.

kijanawoodard:
Looks like this one is done given the merge?

NickCraver:
Yup, closing this out :)

PR #1577, opened 2020-10-03
Implement api for rpushx/lpushx with multiple values

dmytrohridin:
Issue: https://github.com/StackExchange/StackExchange.Redis/issues/1450
Add methods for rpushx/lpushx with multiple values
Do not delete old to save backward compatibility

### DIFF
--- a/src/StackExchange.Redis/Interfaces/IDatabase.cs
+++ b/src/StackExchange.Redis/Interfaces/IDatabase.cs
@@ -652,6 +652,18 @@ namespace StackExchange.Redis
         long ListLeftPush(RedisKey key, RedisValue value, When when = When.Always, CommandFlags flags = CommandFlags.None);
 
         /// <summary>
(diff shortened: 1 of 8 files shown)


mgravell:
This is looking really good, very nice job. I especially like the use of `LLEN` when there are zero elements, to preserve the result semantics.

IMO there's still a little bit missing here - specifically some integration tests that show it actually working on a real server (I see some unit tests checking the wrapper, that's great, thanks). In particular, I'd like to see validation of the behaviour of (cross-joined):

- zero vs one vs many items added
- "when" usage with key exists vs key doesn't exist

dmytrohridin:
Got it. Will add tests today later

dmytrohridin:
@mgravell added more tests :)

mgravell (approved):

mgravell:
merged, thanks

PR #1585, opened 2020-10-14
fix for mutex slim being released too early during flush operation

### LINKED ISSUE #1438: Get & Set Failing After Error Cascade

arsnyder16:
Fixes #1438 

When flushing during FlushSync we are specifying a timeout to the Task, but when this timeout gets hit the underlying task is still running. The thread that invoked FlushSync continues and releases the MutexSlim that is protecting the io pipe allowing another thread to obtain the MutexSlim and try to run operations on the pipe. 

To fix use the cancellation token that can be passed to PipeWriter.FlushAsync. 

An alternative to this is to not specify a timeout. There are 2 additional locations that flush the pipe neither of those locations specify a timeout they just simply await the Task.    Generally the timeouts are driven by how long an operation waits to obtain the MutexSlim

### DIFF
--- a/src/StackExchange.Redis/PhysicalConnection.cs
+++ b/src/StackExchange.Redis/PhysicalConnection.cs
@@ -892,6 +892,7 @@ namespace StackExchange.Redis
                         {
                             ThrowTimeout();
                         }
+                        throw;
                     }
                 }
                 return flush.Result;
--- a/src/StackExchange.Redis/PhysicalConnection.cs
+++ b/src/StackExchange.Redis/PhysicalConnection.cs
@@ -875,11 +875,23 @@ namespace StackExchange.Redis
         [Obsolete("this is an anti-pattern; work to reduce reliance on this is in progress")]
         internal WriteResult FlushSync(bool throwOnFailure, int millisecondsTimeout)
         {
-            var flush = FlushAsync(throwOnFailure);
+            CancellationTokenSource source = new CancellationTokenSource();
+            source.CancelAfter(TimeSpan.FromMilliseconds(millisecondsTimeout));
+            var flush = FlushAsync(throwOnFailure, source.Token);
             if (!flush.IsCompletedSuccessfully)
             {
-                // here lies the evil
-                if (!flush.AsTask().Wait(millisecondsTimeout)) ThrowTimeout();
+                try
+                {
+                    // here lies the evil
+                    flush.AsTask().Wait();
+                }
+                catch (AggregateException ex)
+                {
+                    if (ex.InnerExceptions.Any(e => e is TaskCanceledException))
+                    {
+                        ThrowTimeout();
+                    }
+                }
             }
             return flush.Result;
 
@@ -891,7 +903,7 @@ namespace StackExchange.Redis
                 throw new TimeoutException("timeout while synchronously flushing");
             }
         }
-        internal ValueTask<WriteResult> FlushAsync(bool throwOnFailure)
+        internal ValueTask<WriteResult> FlushAsync(bool throwOnFailure, CancellationToken cancellationToken = default)
         {
             var tmp = _ioPipe?.Output;
             if (tmp == null) return new ValueTask<WriteResult>(WriteResult.NoConnectionAvailable);
@@ -903,7 +915,7 @@ namespace StackExchange.Redis
                 long flushBytes = -1;
                 if (_ioPipe is SocketConnection sc) flushBytes = sc.GetCounters().BytesWaitingToBeSent;
 #endif
-                var flush = tmp.FlushAsync();
+                var flush = tmp.FlushAsync(cancellationToken);
                 if (!flush.IsCompletedSuccessfully) return FlushAsync_Awaited(this, flush, throwOnFailure
 #if DEBUG
                     , startFlush, flushBytes
--- a/src/StackExchange.Redis/PhysicalConnection.cs
+++ b/src/StackExchange.Redis/PhysicalConnection.cs
@@ -875,26 +875,27 @@ namespace StackExchange.Redis
         [Obsolete("this is an anti-pattern; work to reduce reliance on this is in progress")]
         internal WriteResult FlushSync(bool throwOnFailure, int millisecondsTimeout)
         {
-            CancellationTokenSource source = new CancellationTokenSource();
-            source.CancelAfter(TimeSpan.FromMilliseconds(millisecondsTimeout));
-            var flush = FlushAsync(throwOnFailure, source.Token);
-            if (!flush.IsCompletedSuccessfully)
+            using (var source = new CancellationTokenSource())
             {
-                try
+                source.CancelAfter(TimeSpan.FromMilliseconds(millisecondsTimeout));
+                var flush = FlushAsync(throwOnFailure, source.Token);
+                if (!flush.IsCompletedSuccessfully)
                 {
-                    // here lies the evil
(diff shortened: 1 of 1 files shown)


arsnyder16:
Fixes StackExchange/StackExchange.Redis#1438

arsnyder16:
Looks like the CI checks are failing for some unrelated issue, seems like the other PRs have the same issue

mgravell:
Normally, time constraints are limiting how much I can get done on PRs etc,
but this is very interesting. I'm going to carve out some of tomorrow
morning to look at this, because if you're right (and I'm assuming you are,
haven't looked deep at the code yet) it could be really useful! Thanks.

arsnyder16:
Thanks, i would be interested in hearing your thoughts, since i am not overly familiar with the library. Feel free to adjust it in another way, this seemed to make the most sense to me.

src/StackExchange.Redis/PhysicalConnection.cs
```diff
@@ -875,11 +875,23 @@ internal static int WriteRaw(Span<byte> span, long value, bool withLengthPrefix
         [Obsolete("this is an anti-pattern; work to reduce reliance on this is in progress")]
         internal WriteResult FlushSync(bool throwOnFailure, int millisecondsTimeout)
         {
-            var flush = FlushAsync(throwOnFailure);
+            CancellationTokenSource source = new CancellationTokenSource();
```
Plasma, line 878: We need to wrap this in a using/dispose block

sassywarsat:
I can't speak to the exact fix, but the described behavior matches exactly the symptoms we've seen provoke the issue.

sassywarsat:
Okay, I've walked through the precise fixes. Normally, I'd like to see a continuation bound for on cancel but that's not really the model or style of the lib, I think?

src/StackExchange.Redis/PhysicalConnection.cs
arsnyder16: good catch. fixed.

Plasma:
I agree the timeout can cause the lock to be released (incorrectly) while the task is actually making progress in the background. so I think this change is correct.

NickCraver:
Merging main in here for updated build checks

JKurzer:
Do we have a timeline for accepting the PR?

mgravell:
Sorry, I had planned to already look at it, but things got ... busy. I hope to look tomorrow (although I appreciate that I've said that before)

JKurzer:
No worries, we're in a stable-ish state on an older version, and life is pretty complicated for all of us right now.

mgravell:
I concur with the problem, but I'm very dubious of allocating a CTS with scheduled timeout *every* call; I wonder instead if we could allocate and reuse a CTS, and only schedule the timeout (dooming it for reuse) when we get a non-sync response; something like this: https://github.com/StackExchange/StackExchange.Redis/commit/e1e7189d452bcf2527defa9c6714459d02bdcfc5

thoughts?

mgravell:
merged with the CTS tweak as proposed

PR #1598, opened 2020-10-24
NRediSearch build fixes

NickCraver:
Fundamentally, JRediSearch is doing this, and bases test assumptions on it. To get back to a decent/matching state, we need to do the same. It also brings in the hash failure check from https://github.com/RediSearch/JRediSearch/blob/f1d7e2ed64a18e99b31b3b7fe48bb972c00e3999/src/test/java/io/redisearch/client/ClientTest.java#L403-L408


NickCraver:
Getting this in to get other PRs in shape. Looks like AppVeyor Linux builds are generally broken so I'll look at removing AppVeyor and getting package pushes into Actions as a next step.

PR #1612, opened 2020-11-13
Async backlog

mgravell:
See also #1574; rather than buffering (which helps, but just kicks the problem down a little), this moves `ProcessBacklog` to be truly async; it also checks for work *before* attempting to acquire the lock, to short-circuit high contention scenarios

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -1,6 +1,6 @@
 # Release Notes
 
-## 2.2.0
+## 2.2.2
 
 - add .NET 5 target
 - fix mutex race condition (#1585 via arsnyder16)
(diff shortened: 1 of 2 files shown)


NickCraver (approved):
:+1: Much cleaner too

CatalinAdlerDF:
@mgravell Just for curiosity, why are you using `lock` inside an async method?

IgorIgorevich94:
> @mgravell Just for curiosity, why are you using `lock` inside an async method instead of an async-friendly construct?

I would assume that there is no point in introducing additional `async` overhead here, since there is no `IO` or `Network` stuff happening inside the `lock` code block.

PR #1621, opened 2020-11-18
Adding a devcontainer

m0sa:
so that people can play around with this repo via GitHub Codespaces

image

image

image


.github/.github.csproj
```diff
@@ -1,5 +0,0 @@
-﻿<Project Sdk="Microsoft.Build.NoTargets/1.0.94">
-  <PropertyGroup>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
-  </PropertyGroup>
-</Project>
```
NickCraver, line 5: This file may not matter to you in VS Code, but over here in full IDE land we do use it =P

m0sa:
Something is up with Sentinel tests, they don't seem to be respecting the values from TestConfig.json:

image

TestConfig.Current.SentinelServer does have contain the correct value (`"redis"`) here:

https://github.com/StackExchange/StackExchange.Redis/blob/159ce663b02256e2cf90d59cf6cb5b7d9c71b868/tests/StackExchange.Redis.Tests/Sentinel.cs#L26-L32

m0sa:
I also get omnisharp fails in github codespaces (works fine locally / wsl2 -> remote container):

image

m0sa:
woohoo, intellisense seems to light up with the 3.1 SDK installed:

image

NickCraver (approved):
Getting v1 in, so we can play and tweak - thanks for this!!

PR #1624, opened 2020-11-24
add support for parsing and formatting unix domain sockets

mgravell:
- marked by a leading `!` in the configuration string, i.e. `!/foo` connects to `/foo` over a unix domain socket
- only available on .builds for NET Core 3.1 and above; in theory we could also support .NET Standard 2.1, but I kinda don't want to add that as a TFM

(the actual work to *connect* via unix domain sockets has worked for a long time - `Socket` deals with it)

### DIFF
--- a/src/StackExchange.Redis/Format.cs
+++ b/src/StackExchange.Redis/Format.cs
@@ -3,6 +3,7 @@ using System.Buffers;
 using System.Buffers.Text;
 using System.Globalization;
 using System.Net;
+using System.Net.Sockets;
 using System.Runtime.InteropServices;
 using System.Text;
 
@@ -78,17 +79,21 @@ namespace StackExchange.Redis
 
         internal static string ToString(EndPoint endpoint)
         {
-            if (endpoint is DnsEndPoint dns)
+            switch (endpoint)
             {
-                if (dns.Port == 0) return dns.Host;
-                return dns.Host + ":" + Format.ToString(dns.Port);
+                case DnsEndPoint dns:
+                    if (dns.Port == 0) return dns.Host;
+                    return dns.Host + ":" + Format.ToString(dns.Port);
+                case IPEndPoint ip:
+                    if (ip.Port == 0) return ip.Address.ToString();
+                    return ip.Address + ":" + Format.ToString(ip.Port);
+#if UNIX_SOCKET
+                case UnixDomainSocketEndPoint uds:
+                    return "!" + uds.ToString();
+#endif
+                default:
+                    return endpoint?.ToString() ?? "";
             }
-            if (endpoint is IPEndPoint ip)
-            {
-                if (ip.Port == 0) return ip.Address.ToString();
-                return ip.Address + ":" + Format.ToString(ip.Port);
-            }
-            return endpoint?.ToString() ?? "";
         }
 
         internal static string ToStringHostOnly(EndPoint endpoint)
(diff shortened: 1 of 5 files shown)


src/StackExchange.Redis/StackExchange.Redis.csproj
```diff
@@ -10,6 +10,7 @@
     <SignAssembly>true</SignAssembly>
     <PublicSign Condition=" '$(OS)' != 'Windows_NT' ">true</PublicSign>
     <DefineConstants Condition="'$(TargetFramework)' != 'net461'">$(DefineConstants);VECTOR_SAFE</DefineConstants>
+    <DefineConstants Condition="'$(TargetFramework)' != 'net461' and '$(TargetFramework)' != 'net472' and '$(TargetFramework)' != 'netstandard2.0'">$(DefineConstants);UNIX_SOCKET</DefineConstants>
```
NickCraver, line 13: Recommend we define `NETCOREAPP3_1_OR_GREATER` here, which is landing in a future .NET 5.0 SDK (servicing release) we can remove: https://github.com/dotnet/designs/pull/164/files. This let's us cleanup nicely later.

tests/StackExchange.Redis.Tests/StackExchange.Redis.Tests.csproj
```diff
@@ -30,6 +30,6 @@
 
   <ItemGroup Condition=" '$(TargetFramework)' == 'net462' ">
     <Reference Include="Microsoft.CSharp" />
-    <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="4.7.1" />
+    <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="5.0.0" />
```
NickCraver, line 33: Is this fixing a build error? Definitely one of the danger noodles, so just checking...not worried since it's us and not the lib.

src/StackExchange.Redis/Format.cs
```diff
@@ -78,17 +79,21 @@ internal static string ToString(object value)
...
             {
-                if (ip.Port == 0) return ip.Address.ToString();
-                return ip.Address + ":" + Format.ToString(ip.Port);
+                case DnsEndPoint dns:
+                    if (dns.Port == 0) return dns.Host;
+                    return dns.Host + ":" + Format.ToString(dns.Port);
+                case IPEndPoint ip:
+                    if (ip.Port == 0) return ip.Address.ToString();
+                    return ip.Address + ":" + Format.ToString(ip.Port);
+#if UNIX_SOCKET
```
NickCraver, line 90: Suggest:
```suggestion
#if NETCOREAPP3_1_OR_GREATER
```

src/StackExchange.Redis/Format.cs
```diff
@@ -233,6 +238,16 @@ internal static EndPoint TryParseEndPoint(string addressWithPort)
             string portPart = null;
             if (string.IsNullOrEmpty(addressWithPort)) return null;
 
+            if (addressWithPort[0]=='!')
+            {
+                if (addressWithPort.Length == 1) return null;
+
+#if UNIX_SOCKET
```
NickCraver, line 245:
```suggestion
#if NETCOREAPP3_1_OR_GREATER
```

NickCraver (changes requested):
LGTM functionally, but suggested some quick tweaks to make cleanup super easy in a follow-up as .NET ships some upcoming bits.

tests/StackExchange.Redis.Tests/StackExchange.Redis.Tests.csproj
mgravell: just cleaning up package updates; only impacts tests, and I'd rather know that the tests work with "current"

NickCraver: kk, gotcha

PR #1639, opened 2020-12-09
Method for XREAD BLOCK added.

adnanshaukat469:
A method has been added to read the stream with the blocking request, I have added two new method with name StreamReadBlocking and StreamReadBlockingAsync. Use the position as '$' to get the new messages.


mgravell:
This can not and will not be considered unless/until we implement the pooled connection mode. Any blocking operation is death to a multiplexer. It is fundamentally incompatible until then. This is discussed more here: https://stackexchange.github.io/StackExchange.Redis/PipelinesMultiplexers.html#multiplexing

mgravell:
Closing out pending blocking support

PR #1645, opened 2020-12-15
Update index.md for broken links to other md files

grammyleung:
Not good for online browsing between docs.


mgravell:
The links are correct as is; they are meant to be used *here*: https://stackexchange.github.io/StackExchange.Redis/

PR #1663, opened 2021-01-10
redis URI support added

### LINKED ISSUE #1590: Feature request: support redis:// scheme uri
https://www.iana.org/assignments/uri-schemes/prov/redis

> Example: redis://user:secret@localhost:6379/0?foo=bar&qux=baz

Jalalx:
this should resolve #1590

### DIFF
--- a/src/StackExchange.Redis/ConfigurationOptions.cs
+++ b/src/StackExchange.Redis/ConfigurationOptions.cs
@@ -148,6 +148,11 @@ namespace StackExchange.Redis
         private IReconnectRetryPolicy reconnectRetryPolicy;
 
         /// <summary>
+        /// Gets a value that indicates if the configuration is built from the Redis URI scheme.
+        /// </summary>
+        public bool IsFromUri { get; private set; } = false;
+
+        /// <summary>
         /// A LocalCertificateSelectionCallback delegate responsible for selecting the certificate used for authentication; note
         /// that this cannot be specified in the configuration-string.
         /// </summary>
@@ -443,6 +448,7 @@ namespace StackExchange.Redis
         {
             var options = new ConfigurationOptions
             {
+                IsFromUri = IsFromUri,
                 ClientName = ClientName,
                 ServiceName = ServiceName,
                 keepAlive = keepAlive,
@@ -485,8 +491,10 @@ namespace StackExchange.Redis
         /// </summary>
         public void SetDefaultPorts()
         {
-            EndPoints.SetDefaultPorts(Ssl ? 6380 : 6379);
+            EndPoints.SetDefaultPorts(GetDefaultPort());
         }
+
+        private int GetDefaultPort() => Ssl ? 6380 : 6379;
 
         /// <summary>
         /// Sets default config settings required for sentinel usage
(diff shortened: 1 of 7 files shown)


Jalalx:
@mgravell  @NickCraver  There's a failing test that I guess has nothing to do with my changes. Do I need to do anything?

src/StackExchange.Redis/ConfigurationOptions.cs
```diff
@@ -653,109 +653,79 @@ private void DoParse(string configuration, bool ignoreUnknown)
 
             Clear();
 
-            // break it down by commas
-            var arr = configuration.Split(StringSplits.Comma);
-            Dictionary<string, string> map = null;
-            foreach (var paddedOption in arr)
+            if (configuration.TrimStart().StartsWith("redis://", StringComparison.InvariantCultureIgnoreCase))
```
mgravell, line 656: or `rediss://`, which should enable TLS

src/StackExchange.Redis/ConfigurationOptions.cs
```diff
@@ -769,19 +739,153 @@ private void DoParse(string configuration, bool ignoreUnknown)
...
+                    return false;
             }
+
+            return true;
+        }
+
+        private static Dictionary<string, HashSet<string>> ParseQueryString(string qs)
+        {
+            var dict = new Dictionary<string, HashSet<string>>();
+            foreach (var p in qs.Trim().Split(StringSplits.Ampersand))
```
mgravell, line 855: value, and probably the key, need unescaping, presumably; consider %-encoded parameters

src/StackExchange.Redis/ConfigurationOptions.cs
```diff
@@ -769,19 +739,153 @@ private void DoParse(string configuration, bool ignoreUnknown)
...
+                    value = pair[1];
+                }
+                else
+                {
+                    // ignore next possible value(s).
+                }
+
+                if (!string.IsNullOrWhiteSpace(key))
+                {
+                    if (!dict.ContainsKey(key))
```
mgravell, line 877: this is a complex one; technically query string values are ordered, and hash-set is not; an query-string values (for the same key) could be repeated, but they can't in a hash-set; whether that matters is contextual

src/StackExchange.Redis/StringSplits.cs
```diff
@@ -4,6 +4,9 @@ internal static class StringSplits
     {
         public static readonly char[]
             Space = { ' ' },
-            Comma = { ',' };
+            Comma = { ',' },
```
mgravell, line 7: purely observation, but in recent TFMs we *finally* have an API that resolves to `string.Split('x')` without creating an array; sadly, this doesn't help us since we multi-target :(

mgravell:
This is great work and is heading in the right direction; added a few comments, but great to see this moving!

mgravell:
also, we've recently merged in unix socket support; that would probably need mapping to ... `unix://` ?

src/StackExchange.Redis/ConfigurationOptions.cs
Jalalx: fixed

src/StackExchange.Redis/ConfigurationOptions.cs
```diff
@@ -653,109 +653,79 @@ private void DoParse(string configuration, bool ignoreUnknown)
...
-            foreach (var paddedOption in arr)
+            if (configuration.TrimStart().StartsWith("redis://", StringComparison.InvariantCultureIgnoreCase))
             {
-                var option = paddedOption.Trim();
+                if (!Uri.TryCreate(configuration, UriKind.Absolute, out Uri uri))
+                {
+                    throw new ArgumentException("is not valid", nameof(configuration));
+                }
+
+                var ep = Format.TryParseEndPoint(uri.Authority);
```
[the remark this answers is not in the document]
Jalalx: done

src/StackExchange.Redis/ConfigurationOptions.cs
Jalalx: done

src/StackExchange.Redis/ConfigurationOptions.cs
Jalalx: In case of `rediss://` it will set `Ssl = true` and `SslProtocols = Tls12 | Tls13`. Is it OK?

src/StackExchange.Redis/ConfigurationOptions.cs
Jalalx: I fixed it by implementing the `QueryStringDictionary` that is based on `Dictionary<string, List<string>>`

Jalalx:
@mgravell  I fixed the code review issues and added the `ToString()` support but We still need to discuss the multi-host support.

Jalalx:
@mgravell Can you explain about mapping to `unix://`?

NickCraver:
I apologize for not addressing this one sooner - we've talked about it at various points but never commented on the PR. After discussing the pros and cons, I think ultimately: this isn't something we want to support. The problems with introducing support are:
1. URI format doesn't handle multiple servers (our connection string does)
2. URI doesn't support our options which may conflict with URI params or have parsing fun
3. This adds complexity into the connection code and the server endpoint matching code (e.g. Sentinel failures here).

If we solve 1, we create ambiguity problems with 2, or we have some mashup of hybrid support. But the big point is 3: it adds a lot of complexity without significant gains (very few users have asked for this). I know as a user it affects the minority use case that requires complexity not worth it just sucks, but that's the situation here. From a project standpoint the incompatibilities and the lengths we have to go to in making that work and supporting it aren't a good tradeoff for this case.

I really do appreciate your effort to make this work - it required actually going through it to see the complexity and remaining issues.

Issue #1659, opened 2021-01-05
Cluster auto discovery doesn't update node relations after a master failover if only a single node is passed in ConfigurationOptions


dreijer:
I've been testing the StackExchange.Redis library against a Redis cluster with three shards. My topology consists of three physical nodes each running three Redis instances, i.e. one master and two replicas (for the other two shards) on each physical node, resulting in a total of nine Redis instances. Each of the physical nodes has its own public IP address and each Redis instance.

When I connect to my Redis cluster, I use a FQDN pointing to a traffic manager in Azure that just picks one of the three physical nodes for the initial auto discovery request and the library then connects directly to each of the nine Redis instances via its public IP address. The library then performs auto discovery by default every 60 seconds against the FQDN endpoint that I provided in `ConfigurationOptions`. In other words, I don't provide the library with all nine Redis endpoints -- they're auto discovered from the FQDN that I provide.

I seem to have discovered what appears to be a bug, or perhaps an unsupported scenario, after a failover of the master with this setup. Before the failover, if I attempt to read a value and pass `CommandFlags.PreferReplica`, the library properly round robins between the two replicas for the desired hash slot. However, after I manually fail over the master and the library has done auto discovery, reads now only go to the master. In fact, looking closer at `ServerSelectionStrategy`, the master `ServerEndpoint` [doesn't actually _have_ any Replicas](https://github.com/StackExchange/StackExchange.Redis/blob/6409b9d1f2afd01a7eab9498a525360d1ca2b752/src/StackExchange.Redis/ServerSelectionStrategy.cs#L254)...

After debugging through the library, it looks like the auto discovery logic doesn't correctly update the cluster config and [the `ServerEndpoint.Replicas` array](https://github.com/StackExchange/StackExchange.Redis/blob/6409b9d1f2afd01a7eab9498a525360d1ca2b752/src/StackExchange.Redis/ServerEndPoint.cs#L29) on the newly selected master remains empty.

This appears to be because the auto discovery is done against the endpoint that was passed in `ConfigurationOptions`, i.e. the FQDN, which isn't actually used by the Redis cluster itself. That is, when the response is processed by `ClusterNodesProcessor`, `ServerEndpoint.SetClusterConfiguration` is only called on the FQDN `ServerEndpoint` instance, which is a no-op since the endpoint won't match any of the nodes in the cluster (remember: they all have public IP addresses and were auto discovered through the initial bootstrap):
[...]

PR #1684, opened 2021-02-10
  Update Cluster AutoDiscovery after failover

laurauzcategui:
The following commits address: https://github.com/StackExchange/StackExchange.Redis/issues/1659

  Where basically after manual failover the replicas endpoint list is
not refreshed.

  ServerSelectionStrategy.cs

  Updated:
   - SetClusterConfiguration to call new method UpdateNodeRelations with
the configuration
  Added:
   - UpdateNodeRelations to have the same update logic that it was
before in SetClusterConfiguration so it could be used in
ConnectionMultiplexer

   ConnectionMultiplexer.cs
   Updated:
   - GetEndpointsFromClusterNodes, to call UpdateNodeRelations and
refresh the configuration before returning the endpoints list


NickCraver (approved):
Verified the symptoms locally and explains one test condition - pulling in!

Thank you!!

NickCraver:
@TimLovellSmith @philon-msft can disregard here, we're good to go on next release, was able to (accidentally) cause the case this is resolving.

Issue #1442, opened 2020-04-22
Xamarin.Forms android can not resolve reference: `System.Diagnostics.PerformanceCounter`, referenced by `StackExchange.Redis`


tomasfabian:
Hi,
I have a Xamarin.Forms.Droid build problem:

_Severity	Code	Description	Project	File	Line	Suppression State
Error		Can not resolve reference: `System.Diagnostics.PerformanceCounter`, referenced by `StackExchange.Redis`. Please add a NuGet package or assembly reference for `System.Diagnostics.PerformanceCounter`, or remove the reference to `StackExchange.Redis`.	Xamarin.Sample.Android_			
			
I also tried to include System.Diagnostics.PerformanceCounter package:
```
    <PackageReference Include="StackExchange.Redis">
      <Version>2.1.30</Version>
    </PackageReference>
    <PackageReference Include="System.Diagnostics.PerformanceCounter">
      <Version>4.7.0</Version>
    </PackageReference>
```

Is it possible to use StackExchange.Redis with Xamarin.Forms apps?

Thank you, Tomas

Filipowicz251:
I have the same issue. Adding NuGet package doesn't solve anything

Filipowicz251:
Most probably this is correlated with nuget package reference and this bug especially: https://github.com/xamarin/xamarin-android/issues/2674

Going back to packages.config fix the issue (i just read it - i will right now try it;) )

andrev121:
> I have the same issue. Adding NuGet package doesn't solve anything
> 
> [UPDATE]
> 
> Most probably this is correlated with nuget package reference and this bug especially: [xamarin/xamarin-android#2674](https://github.com/xamarin/xamarin-android/issues/2674)
> 
> Going back to packages.config fix the issue (i just read it - i will right now try it;) )

Tried install both of packages in shared and droid projects. Placed packages.config with reference on PerfomanceCounter in root of droid project. Nothing worked. How to fix this issue? Does redis unusable with forms?

SouJack:
See pullrequest #1702

SHuygh:
@mgravell Would it be possible to remove the reference to the performance counter based on @SouJack PR for MonoAndroid10  and Xamarin.iOS10 (cfr. https://docs.microsoft.com/nl-be/xamarin/cross-platform/app-fundamentals/nuget-manual) to be able to use these within Xamarin Forms and Xamarin Native projects.

mgravell:
[...]

PR #1702, opened 2021-03-10
Solve problems #1442

SouJack:
A small change to remove the dependency on the "System.Diagnostics.PerformanceCounter" package for non-Windows binary projects.
For #1442

### DIFF
--- a/src/StackExchange.Redis/PerfCounterHelper.cs
+++ b/src/StackExchange.Redis/PerfCounterHelper.cs
@@ -9,13 +9,15 @@ namespace StackExchange.Redis
     internal static class PerfCounterHelper
     {
         private static readonly object staticLock = new object();
+#if !XAMARIN
         private static volatile PerformanceCounter _cpu;
+#endif
         private static volatile bool _disabled = !RuntimeInformation.IsOSPlatform(OSPlatform.Windows);
 
         public static bool TryGetSystemCPU(out float value)
         {
             value = -1;
-
+#if !XAMARIN
             try
             {
                 if (!_disabled && _cpu == null)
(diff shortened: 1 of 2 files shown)


mgravell:
With the definition

``` xml
<DefineConstants Condition="'$(TargetFramework)' != 'net461' and '$(TargetFramework)' != 'net472'">$(DefineConstants);XAMARIN</DefineConstants>
```

This isn't really "am I Xamarin?", but is instead: "am I anything except .NET Framework", and since .NET Framework is only really retained for legacy convenience, it is really "am I anything?"

I would prefer to be much more selective in this issue, if possible - perhaps checking the runtime environment to isolate places where we know it will fail, rather than just assuming "not .NET Framework? it'll fail". Perhaps via one of the `RuntimeInformation` APIs?

SouJack:
In fact, it is not really xamarin. But it has been tested and is being used at runtime. The solution can be more elaborate, I agree, but, fact that the solution for those who want to use in XF works correctly both Android and iOS. And of course it still works normally in .Net Core since the Nuget package itself does not contain the separation for xamarin runtimes.
It is a suggested course to follow for a final solution.

SouJack:
That's correct, but by your own code and comments, if you can't get access to the object, you return false, what I did was remove it, since for Android and iOS the package doesn't exist. As I said earlier, this is only a suggestion following the logic of writing in the class in question.
For me it is working perfectly and I see that there are people with the same problem, until a definitive suggestion comes out I think it can be useful.

mgravell:
Ultimately, the problem is that you didn't remove it for Android and iOS -
you removed it for everyone. It is frustrating, because ultimately this is
a packaging failure in an external package: it should *just work* (even if
that means it doesn't work at runtime, but in a known way), but... sigh.

PR #1709, opened 2021-03-18
corrects wording of HashIncrement/HashDecrement summaries

f2calv:
I experienced an unexpected behaviour when using HashIncrementAsync on a hashset field that was a string.
I was using FireAndForget which initially hid my error... however I was relying on the wording of the intellisense which stated;
```
If field does not exist **or holds a string that cannot be interpreted as integer**, the value is set to 0 before the operation is performed.
```

So I (wrongly) assumed my below test would work and "abc" would be reset to "0" and then incremented by 1.
```csharp
[Fact]
public async Task dtHashSet()
{
    await _redisCacheSvc.db.HashSetAsync("prefix:key", new HashEntry[] { new HashEntry("hashField", "abc") });
    var before = await _redisCacheSvc.db.HashGetAsync("prefix:key", "hashField");
    Assert.Equal("abc", before);
    var after = await _redisCacheSvc.db.HashIncrementAsync("prefix:key", "hashField", 1);
    Assert.Equal(1, after);
}
```
Maybe I missed something obvious here?
But either way thanks for a great library :)


mgravell:
thanks, appreciated

PR #1710, opened 2021-03-18
OnManagedConnectionFailed thread/memory leak fix

alexSatov:
Hi. I faced a problem with 'ConnectionMultiplexer': when it loses connection with Redis cluster, 'OnManagedConnectionFailed' event handler creates timer, which tries to 'SwitchMaster'. Timer ticks every second, but there is PLINQ query in 'GetConfiguredMasterForService' method, that leads to constantly threads starting (threads start faster than they exit). And if disconnect time is long enough, I get thread/memory leak in service.
PR contains hotfix of a problem with timer (timer starts next iteration only after previous iteration complete). But maybe you should take a closer look at PLINQ queries (they seem superfluous).
Also, I commented `TextWriter.Null` creation in 'SwitchMaster', because it seems like useless memory allocations (you handle nullable LogProxy).

### DIFF
--- a/src/StackExchange.Redis/ConnectionMultiplexer.cs
+++ b/src/StackExchange.Redis/ConnectionMultiplexer.cs
@@ -2439,11 +2439,9 @@ namespace StackExchange.Redis
         {
             ConnectionMultiplexer connection = (ConnectionMultiplexer)sender;
 
-            if (connection.sentinelMasterReconnectTimer != null)
-            {
-                connection.sentinelMasterReconnectTimer.Dispose();
-                connection.sentinelMasterReconnectTimer = null;
-            }
+            var oldTimer = Interlocked.Exchange(ref connection.sentinelMasterReconnectTimer, null);
+            oldTimer?.Dispose();
+
             try
             {
 
@@ -2486,7 +2484,7 @@ namespace StackExchange.Redis
             // or if we miss the published master change
             if (connection.sentinelMasterReconnectTimer == null)
             {
-                connection.sentinelMasterReconnectTimer = new Timer((_) =>
+                connection.sentinelMasterReconnectTimer = new Timer(_ =>
                 {
                     try
                     {
@@ -2495,9 +2493,12 @@ namespace StackExchange.Redis
                     }
                     catch (Exception)
                     {
-
                     }
-                }, null, TimeSpan.FromSeconds(0), TimeSpan.FromSeconds(1));
+                    finally
+                    {
+                        connection.sentinelMasterReconnectTimer?.Change(TimeSpan.FromSeconds(1), Timeout.InfiniteTimeSpan);
+                    }
+                }, null, TimeSpan.Zero, Timeout.InfiniteTimeSpan);
             }
         }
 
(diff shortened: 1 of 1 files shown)


eduardobr:
@alexSatov, if the timer iterations are overlapping, wouldn't we still have the problem after this change?
Simply denying reentrancy isn't what we need? (like this example with [Interlocked](https://docs.microsoft.com/en-us/dotnet/api/system.threading.interlocked?view=net-5.0))

alexSatov:
As I wrote, this change does not allow timer iterations to overlap. Arguments (0, 0) guarantee a single iteration, so I see no point in "resource lock".

eduardobr:
You're right about not overlapping - my bad, now I see the original interval was also changed to 0.
But in case SwitchMaster fails now it will always retry without any delay, right? In that case, this timer without interval is the same as a `while(!succeeded)` without a timer.
But I have no idea if removing the 1 sec delay wouldn't overwhelm the CPU in this scenario, so I'll refrain to comment.

vait:
@eduardobr , you can add a delay it doesn't matter. In this change a working timer will start a new iteration only after previous one has been completed.

src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -2535,7 +2538,8 @@ internal void OnManagedConnectionFailed(object sender, ConnectionFailedEventArgs
         /// <param name="log">Log to write to, if any</param>
         internal void SwitchMaster(EndPoint switchBlame, ConnectionMultiplexer connection, TextWriter log = null)
         {
-            if (log == null) log = TextWriter.Null;
+            // useless memory allocations?
+            // if (log == null) log = TextWriter.Null;
```
NickCraver, line 2542: Can we revert this section please, scoping the change to only one thing here?

src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -2495,9 +2495,12 @@ internal void OnManagedConnectionFailed(object sender, ConnectionFailedEventArgs
                     }
                     catch (Exception)
                     {
-
                     }
-                }, null, TimeSpan.FromSeconds(0), TimeSpan.FromSeconds(1));
+                    finally
+                    {
+                        connection.sentinelMasterReconnectTimer?.Change(TimeSpan.FromSeconds(1), TimeSpan.FromSeconds(0));
```
NickCraver, line 2501: If we're disabling period pulling, the second arg should be `Timeout.InfiniteTimeSpan` here :)

vait: Yep! -1 is enough

src/StackExchange.Redis/ConnectionMultiplexer.cs
vait: I've found one more mistake. Alex will send new commit next week.

NickCraver (approved):
Thanks for the fix here! Confirmed working :+1:

PR #1713, opened 2021-03-28
Allow ReadOnlyMemory<byte> as Lua Script parameter type

eduardobr:
While `ReadOnlyMemory<byte>` is convertible to `RedisValue`, the type is not allowed as a Lua script parameter.
I can see the restriction doesn't seem to be necessary and the type could work as expected on Lua scripts.

How to reproduce:
```
1. const string Script = "return @ident";
2. var prepared = LuaScript.Prepare(Script);
3. var db = conn.GetDatabase();
4. 
5. var val = prepared.Evaluate(db, new { ident = new ReadOnlyMemory<byte>(new byte[] { 4, 5, 6 }) });
6. Assert.True(new byte[] { 4, 5, 6 }.SequenceEqual((byte[])val));
```

Here line 5 will fail with
```
System.ArgumentException : ps
    Parameter name: Expected [ident] on [<>f__AnonymousType3`1[[System.ReadOnlyMemory`1[[System.Byte, mscorlib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089]], System.Memory, Version=4.0.1.1, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51]]] to be convertable to a RedisValue
```

This PR adds `ReadOnlyMemory<byte>` as a valid script parameter type convertable to `RedisValue`

### DIFF
--- a/src/StackExchange.Redis/LuaScript.cs
+++ b/src/StackExchange.Redis/LuaScript.cs
@@ -103,7 +103,7 @@ namespace StackExchange.Redis
 
                 var psType = ps.GetType();
                 var mapper = (Func<object, RedisKey?, ScriptParameterMapper.ScriptParameters>)ParameterMappers[psType];
-                if (ps != null && mapper == null)
+                if (mapper == null)
                 {
                     lock (ParameterMappers)
                     {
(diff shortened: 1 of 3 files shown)


mgravell (approved):
Looks good. I wonder whether we should also allow Memory-byte and (lower priority) ArraySegment-byte, but ... not urgent, perhaps

mgravell (changes requested):
On reflection, I think one thing that is important is to ensure that the *basic* (not named parameter) scripting API has the same feature set. It may be worth adding a test that uses vanilla ScriptEvaluate

eduardobr:
> 
> 
> On reflection, I think one thing that is important is to ensure that the _basic_ (not named parameter) scripting API has the same feature set. It may be worth adding a test that uses vanilla ScriptEvaluate

Added a version that tests ScriptEvaluate from a raw string

src/StackExchange.Redis/LuaScript.cs
```diff
@@ -114,10 +114,10 @@ internal void ExtractParameters(object ps, RedisKey? keyPrefix, out RedisKey[] k
                             {
                                 if (missingMember != null)
                                 {
-                                    throw new ArgumentException("ps", "Expected [" + missingMember + "] to be a field or gettable property on [" + psType.FullName + "]");
+                                    throw new ArgumentException("Expected [" + missingMember + "] to be a field or gettable property on [" + psType.FullName + "]", nameof(ps));
```
mgravell, line 117: nice catch

PR #1714, opened 2021-03-28
Start measuring Ping latency closer to sending request and remove Stopwatch instance

eduardobr:
Inspired on ValueStopWatch: https://github.com/dotnet/extensions/blob/master/src/Shared/src/ValueStopwatch/ValueStopwatch.cs, I'm removing the instance of Stopwatch from TimerMessage.

There's a change of behavior here:
We're not starting latency measurement when message is instantiated, but right before it is written to network.

### DIFF
--- a/src/StackExchange.Redis/ResultProcessor.cs
+++ b/src/StackExchange.Redis/ResultProcessor.cs
@@ -329,6 +329,8 @@ namespace StackExchange.Redis
 
         public sealed class TimingProcessor : ResultProcessor<TimeSpan>
         {
+            private static readonly double TimestampToTicks = TimeSpan.TicksPerSecond / (double)Stopwatch.Frequency;
+
             public static TimerMessage CreateMessage(int db, CommandFlags flags, RedisCommand command, RedisValue value = default(RedisValue))
             {
                 return new TimerMessage(db, flags, command, value);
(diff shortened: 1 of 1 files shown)


NickCraver:
Pushed up naming separate to more easily see changes here!

NickCraver (commented):
As I understand it, a subtle but important difference here is in system clock changes. For example if a system clock is off and get corrected by an NTP sync (this happens pretty often), it'll introduce incorrect timings, since 2 absolute timestamps, as opposed to `Stopwatch` (which is relative) will not be relatively correct. It could show too much time or even negative time, when corrections occur.

As an example, let's say a command starts just after midnight:
```
[System time: 00:00:05.500] Command starts (`timingMessage.CreatedTimestamp` is set)
[System time: 00:00:06.000] NTP time corrects, it's actually [00:00:04] (clock drift was 2 seconds)
[System time: 00:00:05.000] Command finishes
```

...according to profiling, this command took `-500ms`, because that's that the diff in the 2 timestamps told us.

I agree on potential efficiency here, but I think we are trading correctness, which is a hard sell when it comes to profiling.

Thoughts? Suggestions?

eduardobr:
Hi Nick,
I was just checking the source of StopWatch, we can see that the same static GetTimestamp is used internally when Start is called and when we get Elapsed property of the running StopWatch.
Considering this, the change here calculates time exactly same way. It moved from StopWatch.Start() to StopWatch.GetTimestamp().
Unless I’m missing some other bit?

Other than this, what I really wanted when I was looking this code is to understand when the Ping() latency measurement starts. This is because in some projects I’m working I can see the latency calculated being much higher than the actual trips to redis when measured from outside the application. I suspect the reason was because the application was running in containers with extremely low CPU (0.2 cpu core) and the actual code path to run a Ping() call was slower than the actual roundtrip to Redis when application was busy processing other thing in parallel.
So I wanted to have a measure of latency that started as late as possible before reaching the network from the application and as early as possible when the response comes.
But I think that requires a larger refactoring to achieve and needs to be announced so people don’t think suddenly their network became faster for some reason.
I would be happy to try something if you think all of this makes sense.

NickCraver (approved):
Looking good - thank you!

PR #1732, opened 2021-05-19
PhysicalBridge ConcurrentQueue refactor 

TimLovellSmith:
A 'data-structure-refactor' that changes the `_backlog` queue to support writing concurrently with reading, which can reduce both read and write thread blocking.

I originally wanted to do this as a way to reduce contention between threads adding to the backlog - which was secondary and mainly happens due to pre-existing contention on the write lock (which I haven't tried to eliminate). This was motivated by having too-many-threads logs that I saw while testing some other changes I'm testing for `PhysicalBridge`. 

I haven't been able to fully reduce contention between competing reads and competing writes, but by reducing contention between reads and writes, I think therefore this now requires lower max worker threads in general, and makes thread-blocking less likely in total, and worker thread usage lower overall.

The CLR's `ConcurrentQueue` is used because it is mostly non-blocking, and I was hoping it would eliminate needs for `lock()` statement on writes in enqueue scenarios, and even reads.

However, I found that for overall correctness of the backlog management, we are still required to use `lock()` statement on it in deqeueing scenarios, so that we can avoid e.g. dequeueing a message which was not the one we saw with `peek()` (for timeout processing).

And also in order prevent races where `StartProcessingBacklog()` could fail to be called when the queue is not empty, we end up needing to enhance the logic for checking whether to start processing the backlog with some extra synchronization.

[Note, this change is complex and will need some more testing than I have done so far before merge, though it passes the 'build' tests for me pretty well so far, but I think its ready for some review and feedback.]

### DIFF
--- a/src/StackExchange.Redis/PhysicalBridge.cs
+++ b/src/StackExchange.Redis/PhysicalBridge.cs
@@ -27,6 +27,7 @@ namespace StackExchange.Redis
         private readonly long[] profileLog = new long[ProfileLogSamples];
 
         private readonly ConcurrentQueue<Message> _backlog = new ConcurrentQueue<Message>();
+        private int _backlogProcessorIsRunning = 0;
 
         private int activeWriters = 0;
         private int beating;
@@ -136,11 +137,8 @@ namespace StackExchange.Redis
                 // you can go in the queue, but we won't be starting
                 // a worker, because the handshake has not completed
                 message.SetEnqueued(null);
-                lock (_backlog)
-                {
-                    message.SetBacklogState(_backlog.Count, null);
-                    _backlog.Enqueue(message);
-                }
+                message.SetBacklogState(_backlog.Count, null);
+                _backlog.Enqueue(message);
                 return WriteResult.Success; // we'll take it...
             }
             else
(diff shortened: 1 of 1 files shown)


src/StackExchange.Redis/PhysicalBridge.cs
```diff
@@ -750,45 +735,66 @@ internal WriteResult WriteMessageTakingWriteLockSync(PhysicalConnection physical
...
+                    // and rerun the exact same checks as above
+                    if (!_backlog.TryPeek(out message)) break;
+                    if (message.IsInternalCall) break;
+                    if (!message.HasAsyncTimedOut(now, timeout, out var _)) break;
 
-                    if (!message.HasAsyncTimedOut(now, timeout, out var _)) break; // not a timeout - we can stop looking
-                    _backlog.Dequeue(); // consume it for real
+                    if (!_backlog.TryDequeue(out var message2) || (message != message2)) // consume it for real
+                    {
+                        throw new Exception("Thread safety bug detected! A queue message disappeared while we had the backlog lock");
```
NickCraver, line 794: Let's use `RedisException` here please, for handling downstream :)

src/StackExchange.Redis/PhysicalBridge.cs
```diff
@@ -750,45 +735,66 @@ internal WriteResult WriteMessageTakingWriteLockSync(PhysicalConnection physical
...
+                    }
 
+                    // TODO: PERF, consider doing this outside the lock, or just taking the lock once and looping
```
NickCraver, line 797: Noting for discussion on Tuesday!

src/StackExchange.Redis/PhysicalBridge.cs
```diff
@@ -916,6 +920,23 @@ private async Task ProcessBacklogAsync()
             finally
             {   
                 token.Dispose();
+
+                // Do this in finally block, so that thread aborts can't convince us the backlog processor is running forever
+                if (Interlocked.CompareExchange(ref _backlogProcessorIsRunning, 0, 1) != 1)
+                {
+                    throw new Exception("Bug detection, couldn't indicate shutdown of backlog processor");
```
NickCraver, line 927:
```suggestion
                    throw new RedisException("Bug detected: couldn't indicate shutdown of backlog processor");
```

src/StackExchange.Redis/PhysicalBridge.cs
```diff
@@ -916,6 +920,23 @@ private async Task ProcessBacklogAsync()
...
+                }
+
+                // Now that nobody is processing the backlog, we should consider starting a new backlog processor
+                // in case a new message came in after we ended this loop.
+                if (!_backlog.IsEmpty)
+                {
+                    // Check for faults? To prevent unlimited tasks spawning because I keep throwing and hitting this condition? :-/
+                    if (_backlogStatus != BacklogStatus.Faulted)
+                    {
+                        StartBacklogProcessor();
```
NickCraver, line 937: Probably worth a comment here that this won't stack overflow eventually, but only because of the `Task.Run` implementation nature decoupling the stacks - IMO, maybe not an obvious thing at first glance.

NickCraver (commented):
Looks good for discussion next week! I left some quick comments ahead of call, no big changes just a few tidy items.

mgravell (approved):

NickCraver:
Review status: discussed in call - just a few requested tweaks above then we'll get this merged in.

src/StackExchange.Redis/PhysicalBridge.cs
TimLovellSmith: Whether perf matter here was a bit of a puzzle for me to be honest. I'm not sure if perf is the main concern for CheckBacklogForTimeouts, actually, unless there *are* timeouts, and then maybe its goodness that it holds the lock and processes as many timeouts as possible without letting anybody else enqueue them. 

However, now that I look at it again, I wonder how safe calling SetExceptionAndComplete while holding the lock is. Do continuations get to run on this thread? Could it ever cause deadlocks?

NickCraver (approved):
Tweaks look good, thank you!

PR #1735, opened 2021-05-23
Improving ConfigCheckSeconds for a faster failover detection

deepakverma:
Fixes info command storm on a disconnect
After a master node disconnect performs ConfigCheckSeconds on replica nodes with an exponential backoff

### DIFF
--- a/src/StackExchange.Redis/ServerEndPoint.cs
+++ b/src/StackExchange.Redis/ServerEndPoint.cs
@@ -56,7 +56,7 @@ namespace StackExchange.Redis
             databases = 0;
             writeEverySeconds = config.KeepAlive > 0 ? config.KeepAlive : 60;
             serverType = ServerType.Standalone;
-
+            ConfigCheckSeconds = Multiplexer.RawConfig.ConfigCheckSeconds;
             // overrides for twemproxy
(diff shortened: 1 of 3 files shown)


src/StackExchange.Redis/ServerEndPoint.cs
```diff
@@ -556,7 +556,27 @@ internal bool CheckInfoReplication()
...
 
         private int lastInfoReplicationCheckTicks;
-        internal volatile bool ForceReplicationCheck;
+        internal volatile int ConfigCheckSeconds;
+        [ThreadStatic]
+        private static Random r;
+
+
+        // Forces frequent replication check starting from 1 second upto max ConfigCheckSeconds with an exponential increment
+        internal void ForceExponentiallyReplicationCheck()
```
NickCraver, line 565: Can we rename this for clarity please?
```suggestion
        internal void ForceExponentialBackoffReplicationCheck()
```

NickCraver (changes requested):
I think this one's good - would like a name tweak for clarity though please!

philon-msft (approved):

NickCraver (approved):
:+1: Looking good, thanks for the tweaks!

Issue #1744, opened 2021-06-03
AbortOnConnectFail=false message even though AbortOnConnectFail is already false?


genne:
I'm confused by this message - we get it even though `AbortOnConnectFail` is false - shouldn't the if case be inverted?
https://github.com/StackExchange/StackExchange.Redis/blob/f4c69fa21f93e6f9e0f067cf07294baab77de558/src/StackExchange.Redis/ExceptionFactory.cs#L347

(Also mentioned in this issue: #1287)

philon-msft:
@genne I agree, thank you for bringing this up. Please see my linked PR.

NickCraver:
Fixed in #1747, thanks @philon-msft!

Issue #1287, opened 2019-11-22
The 'To create a disconnected multiplexer, disable AbortOnConnectFail' message is confusing


TimLovellSmith:
I just want to give some feedback that this error message is confusing people:

"No connection is available to service this operation: GET 0C2F1DC9-FEED-4AE6-91F2-6FE736613BB3_MUSTCHANGEPASSWORD; It was not possible to connect to the redis server(s). To create a disconnected multiplexer, disable AbortOnConnectFail. ConnectTimeout; IOCP: (Busy=0,Free=1000,Min=1,Max=1000), WORKER: (Busy=1,Free=32766,Min=100,Max=32767), Local-CPU: n/a"

The reason I think its confusing is that 'disable AbortOnConnectFail' sounds like you should set that 'AbortOnConnectFail=false' to _disable_ it. But AFAICS thats not really what the error message means, since you will only see that error message if it was _already false_:

```if (!muxer.RawConfig.AbortOnConnectFail) sb.Append(" To create a disconnected multiplexer, disable AbortOnConnectFail.");```

TimLovellSmith:
One other thing I find unclear about the message is it is unclear whether it is prescriptive. Is it meant to be teaching people that creating disconnected multiplexers is a really good thing to do? Or is it just saying that 'if you use a different config from what you have now, you can see different behavior'? And if the latter, why is it important?

NickCraver:
I agree the phrasing of this is bad. We'll get something clearer in and I'll update here.

NickCraver:
How about:
> "Error connecting right now. To allow this multiplexer to continue retrying until it's able to connect, use abortConnect=false in your connection string or AbortOnConnectFail=false; in your code."

cc @mgravell - thoughts on the above phrasing?

PR #1747, opened 2021-06-03
Invert check for AbortOnConnectFail advisory message

philon-msft:
Adjust logic to display suggestion for setting AbortOnConnectFail=false only when it's currently true.

Mentioned in these issues:
#1744
#1287


NickCraver (approved):
Well crap, that was backwards - good call!

Issue #1748, opened 2021-06-04
Tests reliability - fail with IsConnected = false, but IsConnecting = true and no exception, perhaps due to blocked threads


TimLovellSmith:
Prologue: I'm looking at test reliability and trying to identify issues can work on. Here is a note on one.

As an example SO10504853.LoopLotsOfTrivialStuff blows up. It has its own ClientName and I believe is meant to have its own connection multiplexer... but I think its running into thread starvation, due to other tests which block the threads and/or XUnit synchronization context.

One such test is trying to create connections that will never work:

```
            using (var conn = ConnectionMultiplexer.Connect($"doesnot.exist.ds.{Guid.NewGuid():N}.com:6500,abortConnect=false,connectTimeout=1000", Writer))
            {
```

Which, using a sync API, will block the thread for a good long time.
>
 	System.Private.CoreLib.dll!System.Threading.ManualResetEventSlim.Wait(int millisecondsTimeout, System.Threading.CancellationToken cancellationToken)	Unknown
 	System.Private.CoreLib.dll!System.Threading.Tasks.Task.SpinThenBlockingWait(int millisecondsTimeout, System.Threading.CancellationToken cancellationToken)	Unknown
 	System.Private.CoreLib.dll!System.Threading.Tasks.Task.InternalWaitCore(int millisecondsTimeout, System.Threading.CancellationToken cancellationToken)	Unknown
 	System.Private.CoreLib.dll!System.Threading.Tasks.Task<bool>.GetResultCore(bool waitCompletionNotification)	Unknown
 	System.Private.CoreLib.dll!System.Threading.Tasks.Task<bool>.Result.get()	Unknown
 	[Waiting on Async Operation, double-click or press enter to view Async Call Stacks]	
 	StackExchange.Redis.dll!StackExchange.Redis.ConnectionMultiplexer.ConnectImpl(StackExchange.Redis.ConfigurationOptions configuration, System.IO.TextWriter log) Line 1163	C#
 	StackExchange.Redis.dll!StackExchange.Redis.ConnectionMultiplexer.Connect(StackExchange.Redis.ConfigurationOptions configuration, System.IO.TextWriter log) Line 1032	C#
 	StackExchange.Redis.dll!StackExchange.Redis.ConnectionMultiplexer.Connect(string configuration, System.IO.TextWriter log) Line 1015	C#
[...]

PR #1749, opened 2021-06-05
Proposed fix for #1719

TimLovellSmith:
After struggling quite a while with why a simple fix for #1719 wouldn't pass tests, and making some tweaks to improve test reliability which I'll PR separately, I eventually realized its all because StackExchange.Redis currently *relies* on out of order message delivery by PhysicalBridge in certain scenarios. A particular set of tests which would reliably fail because of that were the ones doing Tiebreaker with a secure server and needing to authenticate. Instead of being able to fetch tie tiebreaker key, they would fail with NOAUTH.

So this is here the simplest fix I can propose for #1719 which doesn't break the existing ways of doing things. But before you merge... I wonder if there might be an even better factoring. Because to be honest, I am a little nervous about two aspects:

1. the blanket categorization that ALL internal calls should be delivered out of order, as soon as they can get the lock. Mainly because there might be a set of existing bugs related to internal calls getting unexpectedly reordered, and if so, this approach will not fix them! And I couldn't be sure that there won't also later turn out to be cases where ordering of internal calls versus user operations matters .

2. I hypothesize that because this particular fix leaves the policy asALWAYS_USE_BACKLOG_IF_CANNOT_GET_SYNC_LOCK = true, then this means that internal calls are still in a _race_ to acquire the lock with other writers, which could be a source of existing reliability bugs (and test issues). I.e. due to a race acquiring the write lock, internal calls could rarely *fail to jump the queue in the usual way*, resulting in failures similar to the ones seen in the tiebreaker scenario tests. So it might be better to set `ALWAYS_USE_BACKLOG_IF_CANNOT_GET_SYNC_LOCK = !message.IsInternalCall` - but what do you think?

Issue #1719, opened 2021-04-16
Publish messages out-of-order without waiting the result


vwxyzh:
[Defect code](https://github.com/StackExchange/StackExchange.Redis/blob/main/src/StackExchange.Redis/PhysicalBridge.cs#L959):
```cs
token = _singleWriterMutex.TryWait(options: WaitOptions.NoDelay);
if (!token.Success)
{
   if (PushToBacklog(message, onlyIfExists: !ALWAYS_USE_BACKLOG_IF_CANNOT_GET_SYNC_LOCK))
      return new ValueTask<WriteResult>(WriteResult.Success); // queued counts as success
   ...
}
...
if (!flush.IsCompletedSuccessfully)
{
  ...
  return CompleteWriteAndReleaseLockAsync(token, flush, message, lockTaken);
}
```

Scenario:
Thread A:
```cs
await subscriber.PublishAsync("channel-1", 1);
```
Thread B:
```cs
await Task.WhenAll(subscriber.PublishAsync("channel-2", 2), subscriber.PublishAsync("channel-2", 3));
```
1. Thread A try to publish message `1`
2. Thread A takes the lock(_singleWriterMutex) first.
3. Thread B try to publish message `2`
4. Thread B cannot take the lock, send to backlog.
5. Thread B create the backlog, and start the background processing.
6. Thread B returns, since `queued counts as success`.
7. Thread A finish the write message `1`, release the lock.
8. Thread B try to publish message `3`
9. Thread B takes the lock.
10. Thread B finish the write message `3`, release the lock.
11. The backlog takes the lock.
12. The backlog write message `2`.

Then, finally we get following result:
Channel-1: `1`
Channel-2: `3, 2`

TimLovellSmith:
@vwxyzh If you want the messages to arrive in order, you would need to send them in order, so why are you doing 

```
await Task.WhenAll(subscriber.PublishAsync("channel-2", 2), subscriber.PublishAsync("channel-2", 3));
```
Which has no ordering guarantee on task execution, since tasks get distributed across thread pool threads?

Shouldn't you instead do
```
await subscriber.PublishAsync("channel-2", 2);
await subscriber.PublishAsync("channel-2", 3);
```
?

vwxyzh:
@TimLovellSmith that `await` will wait for redis server ack, it will be slower than expected.
And it will keep order when client in idle mode (do not use the backlog) or in busy mode (all messages queued in backlog),
[...]

PR #1750, opened 2021-06-07
For Issue1748: Thread.Sleep fixes

TimLovellSmith:
This is a partial fix for Issue #1748. While debugging the issue further, I found that one cause of test worker threads being blocked for a long time was bad old `Thread.Sleep()`.

This fix eliminates all the `Thread.Sleep()`s from test code.

(With this fix there is only one remaining sleep call now in the product code, in GetSentinelMasterConnection(), which is a public synchronous API. It would probably be a good idea to deprecate that API and encourage people to use an Async API instead.)

### DIFF
--- a/tests/NRediSearch.Test/ClientTests/AggregationBuilderTests.cs
+++ b/tests/NRediSearch.Test/ClientTests/AggregationBuilderTests.cs
@@ -1,4 +1,5 @@
 ﻿using System.Threading;
+using System.Threading.Tasks;
 using NRediSearch.Aggregation;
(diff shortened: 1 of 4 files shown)


TimLovellSmith:
The failing test on this one, I suspect is flakey due to issue #1719. Which should be fixed by PR #1749

NickCraver (approved):
We can't deprecate the one sync path there (for example even at Stack Overflow, it's in a static constructor path, ultimately), but agree on the fixes here :+1:

PR #1751, opened 2021-06-07
For test reliability and 1748:- don't parallelize Moq tests.

TimLovellSmith:
This PR is also a *partial* fix for #1748.

I found that another common cause of too many blocked worker threads while running tests is the code for dynamic generation of mock object Types. This code hits contention due to exclusive locking i.e. the test threads block waiting for each other, all to take turns being the one to then use a hefty amount of CPU to generate some types dynamically. (And once one thread gets unblocks, it then goes busy generating another type, and continues to block all the other Moq threads...)

The proposed solution is to ensure we only run one Moq-dependent test at a time, so that the Moq-dependent tests don't actually have to wait to acquire the lock.


NickCraver (approved):
Good find here!

PR #1752, opened 2021-06-07
Increase timeouts for ConfigVerifyReceiveConfigChangeBroadcast test

TimLovellSmith:
This is not a deeply-analyzed change, but based on one test run on my machine, I suspect these timeouts are probably just a little too short to be reliable while running tests in parallel.


TimLovellSmith:
The same test also seems to be flakey in the CI runs, e.g. https://github.com/StackExchange/StackExchange.Redis/pull/1751/checks?check_run_id=2767064082

NickCraver (approved):
:+1: I think we're really onto it with command order in #1749, but can't hurt here

PR #1754, opened 2021-06-07
For test reliability - use TaskCreationOptions LongRunning hint for 'failed connection' tests

TimLovellSmith:
Also for #1748. Using TaskCreationOptions.LongRunning can give us a temporary boost in thread numbers for those tests which want to block the thread in failed connection scenarios (these tend to block the thread for a while, which can lead to connectivity issues from exhausting the worker thread pool for other tests).


NickCraver (approved):
This makes a lot of sense, good call on isolating these off!

PR #1755, opened 2021-06-07
Auto retry message on connection failure

deepakverma:
PR to Auto Retry requests on a connection failure:

### Requirement
During a network blip with the current implementation Se.Redis multiplexer marks requests as failed for any inflight or yet to be sent requests.
The feature is to have a SE.Redis inbuilt retriable mechanism. Se.Redis based on a retry manager will queue the requests and play them back once connection has been restored.


NickCraver:
@deepakverma if you merge main in here, should help with test stability - can you when time allows please?

deepakverma:
> @deepakverma if you merge main in here, should help with test stability - can you when time allows please?

Great, I have merged & see at least one of the flaky tests I had failures is green now :)

NickCraver:
Will try and dig in tonight or tomorrow, but I do see `SSL.Issue883_Exhaustive` is for some reason failing across all environments - something's up there.

deepakverma:
> Will try and dig in tonight or tomorrow, but I do see `SSL.Issue883_Exhaustive` is for some reason failing across all environments - something's up there.

Thanks, test is fixed now.

src/StackExchange.Redis/PhysicalBridge.cs
```diff
@@ -884,7 +884,7 @@ private async Task ProcessBacklogAsync()
                             if (maxFlush >= 0) ex.Data["Redis-MaxFlush"] = maxFlush.ToString() + "ms, " + (physical?.MaxFlushBytes ?? -1).ToString();
                             if (_maxLockDuration >= 0) ex.Data["Redis-MaxLockDuration"] = _maxLockDuration;
 #endif
-                            message.SetExceptionAndComplete(ex, this);
+                            message.SetExceptionAndComplete(ex, this, onConnectionRestoreRetry: false);
```
NickCraver, line 887: Note for potential enum re-use, rather than true/false here. We have a few different ways to describe the retry depending on the point in the stack from config through execution, let's see what we can do to unify.

src/StackExchange.Redis/Message.cs
```diff
@@ -704,11 +757,9 @@ internal void SetRequestSent()
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         internal void SetWriteTime()
         {
-            if ((Flags & NeedsAsyncTimeoutCheckFlag) != 0)
-            {
-                _writeTickCount = Environment.TickCount; // note this might be reset if we resend a message, cluster-moved etc; I'm OK with that
-            }
+            _writeTickCount = Environment.TickCount; // note this might be reset if we resend a message, cluster-moved etc; I'm OK with that
```
NickCraver, line 760: This will incur more overhead every call - do we need to raise it to every time?

src/StackExchange.Redis/Message.cs
```diff
@@ -621,14 +630,58 @@ internal bool ComputeResult(PhysicalConnection connection, in RawResult result)
...
         }
 
-        internal virtual void SetExceptionAndComplete(Exception exception, PhysicalBridge bridge)
+        private bool TryEnqueueForConnectionRestoreRetry(Exception exception, PhysicalBridge bridge, bool pushMessageToRetryManager = false)
+        {
+            if (pushMessageToRetryManager)
+            {
+                try
+                {
+                    return bridge.Multiplexer.messageRetryManager.PushMessageForRetry(this);
```
NickCraver, line 665: Need null checks here

src/StackExchange.Redis/Message.cs
```diff
@@ -621,14 +630,58 @@ internal bool ComputeResult(PhysicalConnection connection, in RawResult result)
...
+                        Flags |= CommandFlags.OnConnectionRestoreAlwaysRetry;
+                        break;
+                    case OnConnectionRestore.RetryIfNotYetSent:
+                        Flags |= CommandFlags.OnConnectionRestoreRetryIfNotYetSent;
+                        break;
+                    case OnConnectionRestore.NoRetry:
+                    default:
+                        Flags |= CommandFlags.OnConnectionRestoreNoRetry;
+                        break;
+                }
```
NickCraver, line 649: Can switch expression here, though the bigger win would be to unify the enum and eliminate the conversion.

src/StackExchange.Redis/Message.cs
```diff
@@ -222,6 +227,10 @@ internal void SetScriptUnavailable()
 
         public bool IsFireAndForget => (Flags & CommandFlags.FireAndForget) != 0;
         public bool IsInternalCall => (Flags & InternalCallFlag) != 0;
+        public bool IsOnConnectionRestoreRetryIfNotYetSent => (Flags & CommandFlags.OnConnectionRestoreRetryIfNotYetSent) != 0;
+        public bool IsOnConnectionRestoreAlwaysRetry => (Flags & CommandFlags.OnConnectionRestoreAlwaysRetry) != 0;
+        public bool IsOnConnectionRestoreNoRetry => (Flags & CommandFlags.OnConnectionRestoreNoRetry) != 0;
+        private bool IsOnConnectionRestoreFlagSet => IsOnConnectionRestoreAlwaysRetry || IsOnConnectionRestoreRetryIfNotYetSent || IsOnConnectionRestoreNoRetry;
```
NickCraver, line 233: This approach adds `n` properties for `n` retry notions, let's chat and see if we can remove that cardinality on the central Message component.

src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -2772,7 +2786,15 @@ private static async Task<T> ExecuteAsyncImpl_Awaited<T>(ConnectionMultiplexer @
...
-                ThrowFailed(tcs, ex);
+                if (@this.ShouldRetryOnConnectionRestore(message, ex))
+                {
+                    if(!@this.messageRetryManager.PushMessageForRetry(message))
+                        ThrowFailed(tcs, ex);
+                }
+                else
+                {
+                    ThrowFailed(tcs, ex);
+                }
```
NickCraver, line 2797: Can simplify like below for an easier time understanding.

src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -2760,7 +2766,15 @@ internal Task<T> ExecuteAsyncImpl<T>(Message message, ResultProcessor<T> process
...
-                    ThrowFailed(tcs, ex);
+                    if (ShouldRetryOnConnectionRestore(message, ex))
+                    {
+                        if (!messageRetryManager.PushMessageForRetry(message))
+                            ThrowFailed(tcs, ex);
+                    }
+                    else
+                    {
+                        ThrowFailed(tcs, ex);
+                    }
```
NickCraver, line 2777: Same as below.

NickCraver (changes requested):
Added initial thoughts for discussion - I'm unsure about the queues but wanted review in to chat with Marc on Monday.

NickCraver:
Had some more thoughts after sleeping on it: there are inherent problems that this being global yields, for example we probably don't want to re-issue `REPLICAOF` commands and things like that to the server, or `SHUTDOWN` for that matter. Which things should be retried needs to be determined, and either flagged or have some other approach.

There's one other point of confusion I thought of: we do retry a command when we get a `MOVED` response on cluster already today - let's just be cognizant in the final story that `CommandFlags.NoRedirect` remains clearly different.

src/StackExchange.Redis/ConfigurationOptions.cs
```diff
@@ -90,7 +119,9 @@ internal const string
                 TieBreaker = "tiebreaker",
                 Version = "version",
                 WriteBuffer = "writeBuffer",
-                CheckCertificateRevocation = "checkCertificateRevocation";
+                CheckCertificateRevocation = "checkCertificateRevocation",
+                CommandRetryOnConnectionRestore = "CommandRetryOnConnectionRestore",
+                RetryQueueLengthOnConnectionRestore = "RetryQueueLengthOnConnectionRestore";
```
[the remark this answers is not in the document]
deepakverma: I had come up with this naming based on the current implementation where retry mechanism functionality doesn't retry on timeout etc. I agree that  in future Retry can be extended to ensure that a command has been executed Atleastonce and dropping  "onconnectionrestore" makes sense.

deepakverma:
> Had some more thoughts after sleeping on it: there are inherent problems that this being global yields, for example we probably don't want to re-issue `REPLICAOF` commands and things like that to the server, or `SHUTDOWN` for that matter. Which things should be retried needs to be determined, and either flagged or have some other approach.
> 
> There's one other point of confusion I thought of: we do retry a command when we get a `MOVED` response on cluster already today - let's just be cognizant in the final story that `CommandFlags.NoRedirect` remains clearly different.
Good point, Shutdown, replicaof or any other similar command shouldn't be retried (generally cloud providers disable this command) 
for noredirect flag, I have removed the reset of no redirect flag in messageretry so that it's being honored during retry

src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -2834,7 +2860,16 @@ internal T ExecuteSyncImpl<T>(Message message, ResultProcessor<T> processor, Ser
...
+                        var exResult = GetException(result, message, server);
+                        if (ShouldRetryOnConnectionRestore(message, exResult))
+                        {
+                            if(!messageRetryManager.PushMessageForRetry(message))
+                                throw exResult;
+                        }
+                        else
+                        {
+                            throw exResult;
+                        }
```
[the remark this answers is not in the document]
deepakverma: I know it's less readable but I need to make sure if messageRetryManager.PushMessageForRetry fails it should throw as well.

src/StackExchange.Redis/Message.cs
deepakverma: I found there are few internal teams that are using commandflags per request. So I went with Marc's suggestion to have commandretry as commandflag with a validation to allow retry options on it.

src/StackExchange.Redis/Message.cs
deepakverma: After unifying config option command retry with command flags this is improved.

deepakverma:
@NickCraver I took care of the failing test, can you please review it and let me know if you have any other feedback. thanks.

src/StackExchange.Redis/Message.cs
deepakverma: It was earlier being used every time for an async request.  https://github.com/StackExchange/StackExchange.Redis/blob/56ac9a21174684f30b8aa333f8069fdca77ed4e7/src/StackExchange.Redis/Message.cs#L749

I had to modify it so that sync requests also has the write time set. So that I can leverage it to figure out how long a sync message has been waiting, and if it's long enough time it out from the retry queue
            int millisecondsTaken = unchecked(Environment.TickCount - message.GetWriteTime());

src/StackExchange.Redis/PhysicalBridge.cs
deepakverma: thanks for the feedback here, simplified it in the next iteration

deepakverma:
ah, I will be taking a look at how to resolve conflicts here tomm morning.

deepakverma:
I have resolved the conflicts. CommandFlags unit test is not liking the addition of the new enums, where it's failing with the duplicate enum value for prefer*. @mgravell , do you have any recommendations on how to fix it? thanks.

TimLovellSmith:
@deepakverma Apparently I failed at github - I had some old comments which were sitting around in a 'review' that I had not 'published'. Were you able to see them until now? How about now? Sorry for the ones which are outdated.

NickCraver:
I have a PR against this for simplifications we talked about at https://github.com/deepakverma/StackExchange.Redis/pull/5 but it's against a fork repo so a bit hard to see...and we can't change the source branch for this PR, but opening a new one loses all discussion and great points & suggestions above. I'm not sure on best path here since @deepakverma's out this week, @TimLovellSmith should we work on those changes against my amendment branch?

I want to talk through some items like not being in love with the `Func<ConnectionMultiplexer, CommandRetryPolicy>`, maybe we can come up with better there but hopefully I illustrated why we need to decouple the instance itself from config options (which can be re-used). We could for example make _another_ interface that...basically has a single method that's the func, or something else. Let's definitely talk through on Tuesday, but trying to get as many suggestions in as possible for a full call :)

src/StackExchange.Redis/CommandRetry/MessageRetryQueue.cs
```diff
@@ -0,0 +1,175 @@
...
+        private async Task ProcessRetryQueueAsync()
+        {
+            Message message = null;
+            while (true)
+            {
+                message = null;
+                Exception failedEndpointex = null;
+                lock (queue)
+                {
+                    if (queue.Count == 0) break; // all done
```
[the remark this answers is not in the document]
NickCraver: This one isn't available in older frameworks :(

src/StackExchange.Redis/CommandRetry/MessageRetryQueue.cs
```diff
@@ -0,0 +1,175 @@
...
+            }
+        }
+
+        
+
+        internal void CheckRetryQueueForTimeouts() // check the head of the backlog queue, consuming anything that looks dead
+        {
+            lock (queue)
+            {
+                var now = Environment.TickCount;
```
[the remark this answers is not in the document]
NickCraver: It's indeed unused, but the current implementation could end up accessing `Environment.TickCount` `n` times which isn't awesome - think we should refactor that a bit.

src/StackExchange.Redis/CommandRetry/MessageRetryQueue.cs
```diff
@@ -0,0 +1,175 @@
...
+        }
+
+        private void DrainQueue(Exception ex)
+        {
+            Message message;
+            lock (queue)
+            {
+                while (queue.Count != 0)
+                {
+                    message = queue.Dequeue();
```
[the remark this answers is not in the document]
NickCraver: Same < `netstandard2.0` issue here :(

NickCraver:
Collapsing into https://github.com/StackExchange/StackExchange.Redis/pull/1856 to get us on a branch in this repo to collab on as discussed on call!

PR #1765, opened 2021-06-15
Micro optimization - repeated dictionary lookups

TimLovellSmith:
FWLIW. Not really a hot path, but why not save some electrons.


NickCraver (approved):
:+1:

PR #1772, opened 2021-06-19
ReconfigureAsync: fix re-entrancy issue

NickCraver:
When this was refactored long ago, we went from a counter starting at 1 (which handled the first run) to a string reason (which defaults to empty). Combined with the `||` clause format, this subtly introduced a re-entrancy bug where a subsequent run can enter during the first run (where first was true, so no exchange for reason occurred).

This was causing several Sentinel issues due to on connect handlers setting config as nodes returned and ultimately the `Set<T>` triggering a `ReconfigureIfNeeded`....and actually running, while the original config was still in progress. This led to all sorts of race oddness especially if one endpoint was any kind of significantly different in timing than another.

Note: the other changes are only labels on the bool arguments for clarity at the call sites.

### DIFF
--- a/src/StackExchange.Redis/ConnectionMultiplexer.cs
+++ b/src/StackExchange.Redis/ConnectionMultiplexer.cs
@@ -511,7 +511,7 @@ namespace StackExchange.Redis
             {
                 Interlocked.Exchange(ref activeConfigCause, null);
             }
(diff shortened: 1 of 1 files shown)


NickCraver:
cc @TimLovellSmith on this fix - not the only Sentinel issue but it was one of the NoConnectionAvailable sources, because a re-entrancy race improperly set both nodes as master yielding an unselectable replica in the DemandReplica path. In reality, this probably caused far more production connection problems.

PR #1778, opened 2021-06-21
Resend - Fix1719

TimLovellSmith:
Since I can't reopen the old PR. Proposed fix for #1719


TimLovellSmith:
Ah no, that doesn't look right.

PR #1786, opened 2021-06-27
Add == & != operators for ChannelMessaege and RedisFeatures

NickCraver:
Best practice cleanup, since we're overriding `.Equals()`

### DIFF
--- a/src/StackExchange.Redis/ChannelMessageQueue.cs
+++ b/src/StackExchange.Redis/ChannelMessageQueue.cs
@@ -57,8 +57,7 @@ namespace StackExchange.Redis
         /// <summary>
         /// Checks if 2 messages are not .Equal()
         /// </summary>
-        public static bool operator !=(ChannelMessage left, ChannelMessage right) => !(left == right);
-    }
+        public static bool operator !=(ChannelMessage left, ChannelMessage right) => !left.Equals(right);
 
     /// <summary>
     /// Represents a message queue of ordered pub/sub notifications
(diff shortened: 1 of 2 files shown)


src/StackExchange.Redis/ChannelMessageQueue.cs
```diff
@@ -48,6 +48,16 @@ internal ChannelMessage(ChannelMessageQueue queue, in RedisChannel channel, in R
...
+
+        /// <summary>
+        /// Checks if 2 messages are .Equal()
+        /// </summary>
+        public static bool operator ==(ChannelMessage left, ChannelMessage right) => left.Equals(right);
+
+        /// <summary>
+        /// Checks if 2 messages are not .Equal()
+        /// </summary>
+        public static bool operator !=(ChannelMessage left, ChannelMessage right) => !(left == right);
```
mgravell, line 60: feels like an unnecessary indirection here; OK, it'll probable be inlined etc, but why not `!left.Equals(right)` ?

src/StackExchange.Redis/RedisFeatures.cs
```diff
@@ -259,9 +259,20 @@ orderby prop.Name
...
+
+        /// <summary>
+        /// Checks if 2 RedisFeatures are .Equal()
+        /// </summary>
+        public static bool operator ==(RedisFeatures left, RedisFeatures right) => left.Equals(right);
+
+        /// <summary>
+        /// Checks if 2 RedisFeatures are not .Equal()
+        /// </summary>
+        public static bool operator !=(RedisFeatures left, RedisFeatures right) => !(left == right);
```
mgravell, line 276: ditto

PR #1796, opened 2021-07-01
Add .NET version and timestamp on default logger rows

philon-msft:
Prefix each entry with a timestamp that looks like:

```
18:18:47.9261: Connecting (async) on .NET Core 3.1.16
18:18:47.9263: cachename.redis.cache.windows.net:6380/Interactive: Connecting...
18:18:47.9463: cachename.redis.cache.windows.net:6380: BeginConnectAsync
18:18:47.9483: 1 unique nodes specified
```

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -1,6 +1,7 @@
 # Release Notes
 
 ## Unreleased
+- logging additions (.NET Version and timestamps) for better debugging (#1796 via philon-msft)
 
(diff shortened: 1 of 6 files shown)


NickCraver:
This likely dupes up timing where we do it for tests today - need to adjust there but then good to go :+1: I'll try and take a peek tonight, but away from desk a lot today.

NickCraver (approved):
:+1:

PR #1807, opened 2021-07-19
Stream Length Transaction Conditions

AlphaGremlin:
Adding some new condition overloads for XLEN, useful for deleting empty streams in a safe method. It's a very simple change, one line to the length condition implementation plus the overloads.

Tests have been added and pass, though I don't have the infrastructure to run the failover/cluster/etc tests. I can include the .trx file if that's important.

I would have liked to add more conditions, like one that calls XPENDING to examine the number of pending messages, but that would need a new condition implementation with different messages and more complicated code to examine the results. Maybe I'll do it in another pull request, if I find the time.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 - logging additions (.NET Version and timestamps) for better debugging (#1796 via philon-msft)
+- add: `Condition` API (transactions) now supports `StreamLengthEqual` and variants (#1807 via AlphaGremlin)
 
 ## 2.2.62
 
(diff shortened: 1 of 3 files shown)


mgravell (approved):
All looks good to me; any chance you can add a pending entry in the release notes section (following the existing style)?

AlphaGremlin:
Done

NickCraver (changes requested):
This breaks the Windows server build because it doesn't support streams - can we please add the skip condition here? See https://github.com/StackExchange/StackExchange.Redis/blob/main/tests/StackExchange.Redis.Tests/Streams.cs#L38 for the example#1813 
Should be good after that!

AlphaGremlin:
I actually encountered that problem when testing, though I solved it by installing Redis 6.2 via WSL2. Skipping works too.
Annoying that the Windows port isn't maintained anymore, WSL2 isn't exactly memory efficient. I need that RAM for when Visual Studio starts hogging it all!

mgravell:
@NickCraver we could consider using the Memurai Developer Edition on our Windows build servers? it is an msi install, and the main feature limitation is that the free version needs to be restarted every 10 days; if our CI build lasts 10 days, we have bigger problems!

NickCraver (approved):
:+1:

PR #1808, opened 2021-07-19
Support on json index

AvitalFineRedis:


src/NRediSearch/FieldName.cs
```diff
@@ -0,0 +1,46 @@
...
+            return new FieldName(name);
+        }
+
+        public FieldName As(string attribute) {
+            this.attribute = attribute;
+            return this;
+        }
+
+        public static FieldName[] convert(params string[] names) {
+        if (names == null) return null;
```
AviAvni, line 39: indent

AviAvni (approved):

AviAvni (approved):

mgravell:
Merged, thanks; I've also added a fix to avoid it being a breaking change, here: https://github.com/StackExchange/StackExchange.Redis/pull/1837

I'm at a point when we should probably talk about whether we can just hand NRediSearch to Redis Inc - it isn't useful to have "us" (meaning: main library maintainers) as a bottleneck here, I don't think

Issue #1371, opened 2020-02-28
Documentation for scripting does not mention automatic EVALSHA after first EVAL call


ilya-spv:
The documentation page here:

https://github.com/StackExchange/StackExchange.Redis/blob/master/docs/Scripting.md

says this:

> To avoid retransmitting the Lua script to redis each time it is evaluated, LuaScript objects can be converted into LoadedLuaScripts via LuaScript.Load(IServer). LoadedLuaScripts are evaluated with the EVALSHA, and referred to by hash.

I have tested a bit and it seems that the client does send EVAL for the first time and uses EVALSHA afterwards itself to optimize the process. I think it is worth mentioning that it already behaves the way the documentation seem to suggest developers should implement themselves.

mgravell:
PRs are open... :)

ilya-spv:
I'd like to, but I don't know the implementation details good enough to be comfortable with editing the documentation used by so many people :)

I'd like to thank you for the library though, it works very well out of the box!

Issue #1246, opened 2019-10-04
[Question] Do LoadedLuaScripts get cached on all redis endpoints?


bynho:
Hi I have been trying to find out whether we need to load lua scripts for all endpoints in order for it to be cached against all of them,  is it needed?  I have found this question on stackoverflow but no answers: https://stackoverflow.com/questions/31531637/storing-lua-scripts-with-stackexchange-redis

mgravell:
Lua is per-server, yes. The standard API on the library should track where
it thinks it is cached, and deal with it automatically.

PR #1812, opened 2021-07-20
Script transmission in the scripting doku

Chriz76:
See https://github.com/StackExchange/StackExchange.Redis/issues/1371
and
https://github.com/StackExchange/StackExchange.Redis/issues/1246


NickCraver (approved):
:+1: Thanks!

PR #1845, opened 2021-08-24
Add a backlog status that tracks non-activated backlogs

mgravell:
Part of #1847, giving better debug information when this happens.


NickCraver (approved):
:+1:

PR #1850, opened 2021-08-29
Add support for count argument to ListLeftPop, ListLeftPopAsync, ListRightPop, and ListRightPopAsync

jjfmarket:
Add support for count argument to ListLeftPop, ListLeftPopAsync, ListRightPop, and ListRightPopAsync

Optional Count argument was added in Redis 6.2


mgravell (approved):
Looks great. Can you add something to the release notes md, too?

Also: thanks for not.falling into the trap of adding an optional arg to the existing method. You'd be surprised how often we need to explain about that.

jjfmarket:
Added a note to the release notes 👍🏻

NickCraver:
@jjfmarket I've been on a work break here - sorry for the slow response but your effort is very much appreciated <3

PR #1854, opened 2021-09-01
Use a dedicated thread, rather than the thread-pool, for the backlog processor

mgravell:
Pain point in the last release was a spiral of death triggered (or at least: exacerbated) by the backlog processor

### DIFF
--- a/src/StackExchange.Redis/PhysicalBridge.cs
+++ b/src/StackExchange.Redis/PhysicalBridge.cs
@@ -770,7 +770,14 @@ namespace StackExchange.Redis
                 _backlogProcessorRequestedTime = Environment.TickCount;
 #endif
                 _backlogStatus = BacklogStatus.Activating;
-                Task.Run(ProcessBacklogAsync);
+
+                // start the backlog processor; this is a bit unorthadox, as you would *expect* this to just
+                // be Task.Run; that would work fine when healthy, but when we're falling on our face, it is
+                // easy to get into a thread-pool-starvation "spiral of death" if we rely on the thread-pool
+                // to unblock the thread-pool when there could be sync-over-async callers. Note that in reality,
+                // the initial "enough" of the back-log processor is typically sync, which means that the thread
+                // we start is actually useful, despite thinking "but that will just go async and back to the pool"
+                new Thread(s => ((PhysicalBridge)s).ProcessBacklogAsync().RedisFireAndForget()).Start(this);
             }
         }
 #if DEBUG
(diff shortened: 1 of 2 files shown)


philon-msft (approved):

NickCraver (approved):
Looking good :+1:

src/StackExchange.Redis/PhysicalBridge.cs
```diff
@@ -770,7 +770,14 @@ private void StartBacklogProcessor()
...
                 _backlogStatus = BacklogStatus.Activating;
-                Task.Run(ProcessBacklogAsync);
+
+                // start the backlog processor; this is a bit unorthadox, as you would *expect* this to just
+                // be Task.Run; that would work fine when healthy, but when we're falling on our face, it is
+                // easy to get into a thread-pool-starvation "spiral of death" if we rely on the thread-pool
+                // to unblock the thread-pool when there could be sync-over-async callers. Note that in reality,
+                // the initial "enough" of the back-log processor is typically sync, which means that the thread
+                // we start is actually useful, despite thinking "but that will just go async and back to the pool"
+                new Thread(s => ((PhysicalBridge)s).ProcessBacklogAsync().RedisFireAndForget()).Start(this);
```
WeihanLi, line 780: Would it better to use `Task.Factory.StartNew(ProcessBacklogAsync, TaskCreationOptions.LongRunning)`

mgravell: That's a great question. Let me investigate.

TimLovellSmith: Pragmatically I would vote threads just because I think I know how they work... Task.Factory has been full of pits of failure in my very limited experience with it...

TimLovellSmith (approved):

src/StackExchange.Redis/PhysicalBridge.cs
WeihanLi: Just a question, not sure if it's better, a test may be needed to determine.
And with `Thread`, maybe declare it as a background thread is more meaningful.

mgravell:
That (test) is a tricky one, as the scenario where it matters is "we're
desperately unwell"

src/StackExchange.Redis/PhysicalBridge.cs
TimLovellSmith: A background thread sounds good - should be consistent with old threadpool behavior.

mgravell: yup, it should be background for the same app-lifetime impact (in particular); tweaking

NickCraver (approved):

Issue #1847, opened 2021-08-24
Message Backlog: Task pool exhaustion scenario with synchronous calls


NickCraver:
#### Problem
We had a rolling outage at Stack Overflow today caused by a few issues. After troubleshooting through memory dumps, we found our cause: cascading failure ultimately due to sync-over-async at the pipeline writer level. What was a network event of some sort causing redis to timeout (could have been a large payload on the wire or an actual network issue), [the backlog](https://github.com/StackExchange/StackExchange.Redis/blob/89812faf3fe19cda7ed085238b28adae58931dd1/src/StackExchange.Redis/PhysicalBridge.cs#L29) was activated on the `PhysicalBridge`. However, with the changes in #1779, the issue was exacerbated further for synchronous scenarios.

Due to the ultimately sync-over-async nature of the socket IO under the covers, what we have is any synchronous API callers above in the stack are awaiting their pipeline write to complete (...and reach Redis, and come back, but not relevant here yet) before moving on. What happens is the write itself is async, but the thread waiting on that operation (on the sync side of the fence) is still using a thread pool thread. When we combine this fact with more aggressive backlogs in an attempt to preserve ordering and lose less data in #1779, we reach a thread pool starvation state of synchronous waiters on commands that would have otherwise [timed out directly](https://github.com/StackExchange/StackExchange.Redis/blob/89812faf3fe19cda7ed085238b28adae58931dd1/src/StackExchange.Redis/PhysicalBridge.cs#L650-L658) due to a failure to write to the pipeline in time.

The result looks like this on the pool side:
![image](https://user-images.githubusercontent.com/454813/130669465-d402143d-ce6d-403d-b64f-b3b4b4e38a13.png)

[...]

PR #1779, opened 2021-06-21
Fixes "messages get sent out-of-order when a backlog is created by another thread"

TimLovellSmith:
Fixes #1719.

PR #1856, opened 2021-09-04
Auto retry message on connection failure

NickCraver:
A move of #1755 into the repo so we can collab better (as discussed!).

Here we're getting command retries (at first for reconnects) into the primary library for users that want to enable this.


src/StackExchange.Redis/CommandRetry/MessageRetryQueue.cs
```diff
@@ -0,0 +1,167 @@
...
+            bool startProcessor = false;
+            lock (_queue)
+            {
+                startProcessor = _queue.Count > 0;
+            }
+            if (startProcessor)
+            {
+                if (_runRetryLoopAsync)
+                {
+                    var task = Task.Run(ProcessRetryQueueAsync);
```
NickCraver, line 54: Based on our last experience here, need to switch this to thread kickoff /cc @mgravell @deepakverma

src/StackExchange.Redis/CommandRetry/MessageRetryQueue.cs
```diff
@@ -0,0 +1,167 @@
...
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        internal void StartRetryQueueProcessor()
+        {
+            bool startProcessor = false;
+            lock (_queue)
+            {
+                startProcessor = _queue.Count > 0;
+            }
+            if (startProcessor)
```
NickCraver, line 50: Can simplify to:
```cs
if(_queue.Count == 0)
{
    return;
}
```

src/StackExchange.Redis/CommandRetry/MessageRetryQueue.cs
```diff
@@ -0,0 +1,172 @@
...
+namespace StackExchange.Redis
+{
+    internal class MessageRetryQueue : IDisposable
+    {
+        private readonly Queue<Message> _queue = new Queue<Message>();
+        private readonly IMessageRetryHelper _messageRetryHelper;
+        private readonly int? _maxRetryQueueLength;
+        private readonly bool _runRetryLoopAsync;
+
+        internal MessageRetryQueue(IMessageRetryHelper messageRetryHelper, int? maxRetryQueueLength = null, bool runRetryLoopAsync = true)
```
NickCraver, line 15: @deepakverma Trying to figure out the scenario in which `runRetryLoopAsync` would be `false` (triggering a `.Wait()` below), can you shed more light on that scenario/path?

src/StackExchange.Redis/CommandRetry/CommandRetryPolicy.cs
```diff
@@ -0,0 +1,56 @@
...
+        protected CommandRetryPolicy(ConnectionMultiplexer muxer)
+        {
+        }
+
+        internal abstract bool TryQueue(Message message, Exception ex);
+
+        /// <summary>
+        /// Returns the current length of the retry queue.
+        /// </summary>
+        public abstract int CurrentQueueLength { get; }
```
NickCraver, line 23: For exception messages, probably want to `CurrentlyProcessing` (bool) for exception messages

NickCraver:
Had a sync call on this right now, wanted to summarize and get agreement, in no particular order:

1. I think the changes in #1779 create an issue we failed to realize: the handshake itself will get queued behind other message in the backlog. This means especially in authenticated scenarios we'll cycle failure by issuing commands before the AUTH is gotten to in the pipe.
2. The `Task`-based retry flush is subject to the same problems as #1847, so we should switch to a `Thread` there.
3. There's a big interaction question with how the retry policy works (at the multiplexer level) and how the per-bridge retry policy works (at the bridge level). As it stands, upon failure [we'd retry _all_ commands](https://github.com/StackExchange/StackExchange.Redis/blob/7e5c219354d5e039604bf066ec0b54754b8addb5/src/StackExchange.Redis/PhysicalBridge.cs#L702) via the `PhysicalBridge` backlog upon reconnect, but importantly those come back [with a status of successful](https://github.com/StackExchange/StackExchange.Redis/blob/7e5c219354d5e039604bf066ec0b54754b8addb5/src/StackExchange.Redis/PhysicalBridge.cs#L704). This means they won't be touched by the retry policy, which through they succeeded.
4. The above is complicated, but ultimately the reason the backlog exists was for the `AUTH` command ordering in the handshake phase of a connection. This is to keep the [configuration](https://github.com/StackExchange/StackExchange.Redis/blob/7e5c219354d5e039604bf066ec0b54754b8addb5/src/StackExchange.Redis/ServerEndPoint.cs#L301) and [handshake](https://github.com/StackExchange/StackExchange.Redis/blob/7e5c219354d5e039604bf066ec0b54754b8addb5/src/StackExchange.Redis/ServerEndPoint.cs#L821) phases of a connection sequence in order. Instead, what if we removed the `PhysicalBridge` backlog completely?
   - This needs investigation on if we can, but if all the retry mechanics were _only_ the `CommandRetryPolicy` (and our built-in policy is effectively "always"), then there'd be no backlog at the bridge layer. Commands issued directly down there (e.g. handshakes) would fail, like we want, and no retry (like we want)
   - `Message` would need to grow a reference to the `Server` it was queued for _if_ it was queued to a specific server (most things are not). If a `Server is present, we'd want to retry on that specific server (like with a `REPLICAOF` or `KEYS` command you _really_ don't want to arbitrarily redirect).
     - We can do this in [`PrepareToPushMessageToBridge<T>`](https://github.com/StackExchange/StackExchange.Redis/blob/7e5c219354d5e039604bf066ec0b54754b8addb5/src/StackExchange.Redis/ConnectionMultiplexer.cs#L2218) where we were either passed an explicit server or chose one in the first place.
   - Overall, it'd be a huge simplification: system commands issued beneath the multiplexer would not be subject to retry or any mechanics thereof the user specified. e.g. `WriteDirectOrQueueFireAndForgetAsync` would not be subject to the policy or any backlogging at all.
     - However, I'm not sure this works reliably, it's possible we subtly depend on the backlog today to complete connections in ways we didn't fully reason about in the call. Need to heads down on this for a while and try things.

5. This is a big change, and if the above is correct, we've got a regression in the 2.x line. Since we know this is a big change to the default retry mechanics, the proposal is to start on the 3.x branch line and aim this PR at that, allowing us to do alpha releases _alongside_ 2.x, and get out of the critical path. We'll want to hammer these new approaches well and I think their appropriate status is pre-release anyhow.


For number 1, if @TimLovellSmith agrees with the observations here, we should revert that change from the 2.x line and get a release out there with the few other misc fixes in queue. Can others please take a look at the above and provide thoughts? If we're agreed I'll start hacking at this on days off - if we're agreed on the regression we could get a new cut out there quickly and move the intended behavior into the 3.x line.

cc @mgravell @philon-msft @carldc @deepakverma

TimLovellSmith:
@NickCraver 
>the handshake itself will get queued behind other message in the backlog. This means especially in authenticated scenarios we'll cycle failure by issuing commands before the AUTH is gotten to in the pipe.

This is interesting. Rather than completely reverting the out-of-order changes, it suggests to me that we should
A) have a way to bypass the backlog completely, used for commands that are run from internally to re-establish the connection, and maintain the connection state, or 
B) we just need to live with it as by design, and allow unauthenticated commands to fail for the sake of 'consistency' versus whenever the auth command was fired etc.

Sounds like you prefer A rather than B.
I think I'm OK with that in the case where we're talking about AUTH commands issued by the multiplexer based on automatic connection management from configuration -- but I can't help feeling a little doubt about whether this is also the behavior someone issuing their own AUTH commands manually would expect, if anyone does that.

TimLovellSmith:
A follow-on thought... its sort of like how auto-retry is not really the right retry strategy for everything. Like what? Well, we already have this 'CanRetry' concept, which is an acknowledgement that some commands aren't suitable for auto-retrying, especially internal connection state management commands.

```
        internal bool CanRetry(Message message, Exception ex)
        {
            if ((message.Flags & CommandFlags.NoRetry) != 0
                || ((message.Flags & CommandFlags.RetryIfNotSent) != 0 && message.Status == CommandStatus.Sent)
                || message.IsAdmin
                || message.IsInternalCall
                || !(ex is RedisException)
                || !ShouldRetry(message.Status))
            {
                return false;
            }

            return true;
        }
```

You might or might not believe that internal management commands would like to have their order preserved _relative to each other_.... I wonder which?

Does this call for separate backlogs?

NickCraver:
@TimLovellSmith I think if we remove the message backlog at the physical bridge layer all together, the handshake bits simply won't be subject to the retry policy, since they're lower and writing to the pipe further down. I _think_ this will be okay because they're in-order and we're now awaiting each command as it's written down (where as before we had the races). So instead of separate backlogs, connection handshake stuff just isn't backlogged at all, it goes direct. When it completes, the `ComamndRetryPolicy` should kick in and begin flushing. Does that approach make sense?

TimLovellSmith:
@NickCraver I wonder if I am understanding it. Is the proposal to remove the backlog altogether? Or is it just to bypass the backlog and retry stuff for handshakes?

My understanding is that a fairly important function of the backlog is also to reduce lock contention for writers by ensuring that async writes don't have to block to acquire the lock on the socket. I.e. performance optimization.

NickCraver:
> Is the proposal to remove the backlog altogether? 

Yep, this!

> My understanding is that a fairly important function of the backlog is also to reduce lock contention for writers by ensuring that async writes don't have to block to acquire the lock on the socket. I.e. performance optimization.

Hmmm I wouldn't say that's a critical function of the backlog - it was originally created to order `AUTH` correctly in the handshake (which as you helped me debug, wasn't even quite right). If we're backlogged or retrying at all, the amount of time we're spending flushing and sending to Redis should completely dwarf backlog contention - if it's not dwarfing it, we probably have a spiral of death situation with anyone running that hot after a disconnect.

What order things come from callers down to the socket when there is a retry queue though: that's a question - either we disregard ordering completely and let them have it out on the lock, or we buffer into the queue similar to the backlog today. Thoughts?

TimLovellSmith:
The problem I'm thinking of is indeed with the apps that run very hot. Normally they would be fine (looks good enough to put into prod...), but I have seen a couple times in the past clues in the dumps that when suddenly usage spikes up and lock contention on the write lock starts to be one of their big problems/contributing factors.
The backlog (theoretically) acts as a kind of pressure release / catchup optimization, where once triggered, it basically just has to acquire the lock once to write out all the stuff and be all caught up on writes again without lots of synchronization or switching overheads.

TimLovellSmith:
PS I am not really sure whether AutoRetry is intended to break guarantees on command ordering... 

Guarantees might be achievable, but only by saying there's one backlog for everything, everything stays in it until successfully retried, and so no commands can get sent until all commands in front of them already are known to have succeeded, which is probably going to do terrible things to throughput.

So I kinda assume AutoRetry breaks ordering guarantees.

NickCraver:
If we think that's an issue (agreeing it is if you're seeing it in those logs), we could provide a similar mechanism for the retry policy to take an exclusive lock during the flush perhaps. You think that'd address the best of both worlds?

TimLovellSmith:
I believe the part which is the beneficial optimization is the part where lock contention is avoided for ordinary writes, and I don't know any good way to keep that AND have async APIs, and have no reordering bugs... except a backlog.

So maybe we can reconsider the problem from scratch? Is the problem that AUTH commands get queued behind other new writes that are incoming while we are reconnecting? If that is it, that seems like it was probably always a problem with backlog - regardless of autoretry.

Another way of explaining the same problem - we actually DON'T want consistent order of messages. or autoretry for things like AUTH commands while we are reconnecting. We want these commands to be special and send immediately.

The proposed AutoRetry already treats such commands specially. By which I mean it doesn't automatically retry them. My question is therefore why not have the PhysicalBridge treat these special commands specially too? Just have them be completely opted out of consistent delivery order and bypass the backlog, and the problem is solved, isn't it?

NickCraver:
> I believe the part which is the beneficial optimization is the part where lock contention is avoided for ordinary writes, and I don't know any good way to keep that AND have async APIs, and have no reordering bugs... except a backlog.

I think this is where we may be talking past each other. The idea in my mind is that the `CommandRetryPolicy` _is_ the backlog...the only backlog, in this proposal. Otherwise, we basically have 2 levels of backlog which is hard to even reason about.

NickCraver:
@TimLovellSmith @mgravell I'm getting changes for the above in #1857, but note I have NOT removed the `PhysicalBridge` backlog, because there's even more fun there digging in. Any chance we could get together early next week and talk through? Changes thus far are cleanup, simplification, status additions, Thread-based flushing, additional tests, and a move to `FailedCommand` for users to more easily create policies here.

Issue #1838, opened 2021-08-19
NRediSearch Does not support Sortable Tags in Index Definition Schema Object


dcoopertd:
When creating a new schema like this:

`var schema = new Schema().AddTagField("TagField", "|") ;`

There is no option to make it sortable.  For example there is no `AddSortableTagField()` method like there is for text fields.  According to https://oss.redis.com/redisearch/1.6/Commands/#ftcreate , tag fields can clearly be sortable.

dcoopertd:
I also wanted to point out:  Notice the separator must go before sortable or a syntax error will occur:

`FT.CREATE idx ON HASH PREFIX 1 foo:bar SCHEMA  Tag1 TAG SEPARATOR "|"  SORTABLE`

PR #1862, opened 2021-09-17
Adding sortable tag field support & UNF support

slorello89:
Adding the ability to add sorted tag fields in indices per #1838

Note: When serializing the arguments RediSearch cares about the order of the separator and sortable arguments, hence after calling `base.SerializeRedisArgs`, if the index is sortable and removes the sortable flag, and then adds it back after the separator flag is added.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -2,6 +2,7 @@
 
 ## Unreleased
 - NRediSearch: Support on json index (#1808 via AvitalFineRedis)
+- NRediSearch: Support sortable TagFields and unNormalizedForm for Tag & Text Fields (#1862 via slorello89 & AvitalFineRedis)
 - fix potential errors getting socket bytes (#1836 via NickCraver)
 - logging additions (.NET Version and timestamps) for better debugging (#1796 via philon-msft)
 - add: `Condition` API (transactions) now supports `StreamLengthEqual` and variants (#1807 via AlphaGremlin)
--- a/src/NRediSearch/Schema.cs
+++ b/src/NRediSearch/Schema.cs
@@ -220,12 +220,12 @@ namespace NRediSearch
         {
             public string Separator { get; }
 
-            internal TagField(string name, string separator = ",") : base(name, FieldType.Tag, false)
+            internal TagField(string name, string separator = ",", bool sortable = false) : base(name, FieldType.Tag, sortable)
             {
                 Separator = separator;
             }
 
-            internal TagField(FieldName name, string separator = ",") : base(name, FieldType.Tag, false)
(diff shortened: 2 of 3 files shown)


dcoopertd:
While you are in there, could support for the new UNF be added as well? https://github.com/RediSearch/RediSearch/pull/2188

AvitalFineRedis:
@dcoopertd  https://github.com/slorello89/StackExchange.Redis/pull/1

mgravell:
The main problem I can see here is the breaking change on `TextField`'s ctor and the `AddSortableTextField` method which is going to cause `MissingMethodException` on anything that takes a transient dependency (i.e. not rebuilt explicitly). We *usually* try hard to avoid hard breaks, with the preferred method being: to add overloads rather than optional args.

For example:

``` c#
- public TextField(string name, double weight = 1.0, bool sortable = false, bool noStem = false, bool noIndex = false)
- : base(name, FieldType.FullText, sortable, noIndex)
+ public TextField(string name, double weight = 1.0, bool sortable = false, bool noStem = false, bool noIndex = false, bool unf = false)
+ : base(name, FieldType.FullText, sortable, noIndex, unf)
```

could be achieved instead via:

``` c#
public TextField(string name, double weight, bool sortable, bool noStem, bool noIndex)
  : this(name, weight, sortable, noStem, noIndex, false) {}
public TextField(string name, double weight = 1.0, bool sortable = false, bool noStem = false, bool noIndex = false, bool unf = false)
```

- by adding the overload, we don't break existing code (noting that adding/removing default parameter values **does not** cause a `MissingMethodException`)
- by removing the parameter defaults, we ensure that most rebuilt new code goes to the *new* method, and doesn't pay any indirection overhead

Likewise:

``` c#
- public Schema AddSortableTextField(string name, double weight = 1.0)
+ public Schema AddSortableTextField(string name, double weight = 1.0, bool unf = false)
```

could be 

``` c#
public Schema AddSortableTextField(string name, double weight)
  => AddSortableTextField(name, weight, false);
public Schema AddSortableTextField(string name, double weight = 1.0, bool unf = false)
```

mgravell:
If we can fix ^^^, and add something to releasenotes.md, we should be set?

mgravell:
oh: additional thought: what is `unf`? the fact that I need to ask suggests that it needs to be clearer; perhaps `unNormalizedForm` would be clearer than `unf`?

I *also* wonder whether we're getting into the territory where a `[Flags]` enum might be a better choice than a myriad of `bool` values?

slorello89:
> The main problem I can see here is the breaking change on `TextField`'s ctor and the `AddSortableTextField` method which is going to cause `MissingMethodException` on anything that takes a transient dependency (i.e. not rebuilt explicitly). We _usually_ try hard to avoid hard breaks, with the preferred method being: to add overloads rather than optional args.

TIL -> thanks for pointing that out.

PR #1865, opened 2021-09-23
Add support for listening to Azure Maintenance Events

amsoedal:
Adding an automatic subscription to the AzureRedisEvents pubsub channel for Azure caches. This channel notifies clients of upcoming maintenance and failover events. 

By exposing these events, users will be able to know about maintenance ahead of time, and can implement their own logic (e.g. diverting traffic from the cache to another database for the duration of the maintenance event) in response, with the goal of minimizing downtime and disrupted connections. We also automatically refresh our view of the topology of the cluster in response to certain events.

Here are some of the possible notifications:

```
// This event gets fired ~20s before maintenance begins
NodeMaintenanceStarting,

// This event gets fired when maintenance is imminent (<5s)
NodeMaintenanceStart,

// Indicates that the node maintenance operation is over
NodeMaintenanceEnded,

// Indicates that a replica has been promoted to primary
NodeMaintenanceFailover
```


NickCraver:
Hey there! Thanks for getting this in :) 

I'll try and get eyes on this week, but wanted to scan quick and add some top level thoughts:
1. I think we can remove the config option here - we're detecting if it's Azure Redis already so we can use that to determine whether to hook up to events or not.
2. I think we can simplify a lot of the objects here, e.g. `MaintenanceNotificationListener` being simplified to a single method like `AzureMaintenanceEvent.AddListener(ConnectionMultiplexer muxer, string cause)` (or whatever overload makes sense).
   - Also applies to `ReconfigureAfterMaintenance` - we can decide whether parsing the event is needed or not
3. I think we can simplify the looping on the pipe-delimited message to be span-based for an easier-to-understand iterator.
4. In testing, doing inline `Assert.True()` instead of `if...return false` statements in `ValidateAzureMaintenanceEvent` will make for easier testing - we'll see exactly which check failed by line number in test output that way. If it's false, it'll blow up and not continue the method so that's alright, IMO.

amsoedal:
@NickCraver Thanks so much for the quick reply! I have a couple questions/comments:

1) Philo mentioned that we'll automatically subscribe to these events for anyone who is using the Azure.StackExchange.Redis extension package. However since subscribing to the pubsub creates another connection, we want this to be opt-in only in order not to burden folks who might already be short on connections. Thoughts?

3) Could you point me to a place in the code or a doc that illustrates what span-based iteration looks like? Not sure I'm familiar with that type/concept! 

Working on addressing the other feedback now :)

NickCraver:
Oooh gotcha - alrighty maybe a mismatch on comms here between me and @philon-msft - let me clarify and see if we agree.

On 1: I'm of the opinion we can go ahead and opt everyone into this in the current package if we think we're connecting to Azure. We could make the method for doing so explicitly public so that any non-detectable/custom endpoints are easily opted in for anyone in that situation. I don't think we have to wait for a consumer package move. It won't create another connection than it does today because we already light up the subscriber connection unconditionally on the ServerEndPoint - ultimately [you're looking here](https://github.com/StackExchange/StackExchange.Redis/blob/56ac9a21174684f30b8aa333f8069fdca77ed4e7/src/StackExchange.Redis/ConnectionMultiplexer.cs#L1653) - called when we first connect.

On 2: I took a stab at illustrating this and...meh, it's kind of a wash. Here's what I ended up with:
```cs
/// <summary>
/// Azure node maintenance event details.
/// </summary>
public class AzureMaintenanceEvent
{
	public AzureMaintenanceEvent(ReadOnlySpan<char> message)
	{
		try
		{
			while (message.Length > 0)
			{
				if (message[0] == '|')
				{
					message = message.Slice(1);
					continue;
				}

				// Grab the next pair
				var key = message.Slice(0, message.IndexOf('|'));
				message = message.Slice(key.Length + 1);

				var valueEnd = message.IndexOf('|');
				var value = valueEnd > -1 ? message.Slice(0, valueEnd) : message;
				message = message.Slice(value.Length);

				if (key.Length > 0 && value.Length > 0)
				{
					switch (key)
					{
						case var _ when key.SequenceEqual(nameof(NotificationType)):
							NotificationType = value.ToString();
							break;
						case var _ when key.SequenceEqual(nameof(StartTimeUtc)) && DateTime.TryParse(value, out DateTime startTime):
							StartTimeUtc = DateTime.SpecifyKind(startTime, DateTimeKind.Utc);
							break;
						case var _ when key.SequenceEqual(nameof(IsReplica)) && bool.TryParse(value, out var isReplica):
							IsReplica = isReplica;
							break;
						case var _ when key.SequenceEqual(nameof(IPAddress)) && IPAddress.TryParse(value, out var ipAddress):
							IpAddress = ipAddress;
							break;
						case var _ when key.SequenceEqual(nameof(SSLPort)) && Int32.TryParse(value, out var port):
							SSLPort = port;
							break;
						case var _ when key.SequenceEqual(nameof(NonSSLPort)) && Int32.TryParse(value, out var nonsslport):
							NonSSLPort = nonsslport;
							break;
						default:
							break;
					}
				}
			}
		}
		catch { }
	}

	/// <summary>
	/// The type of event.
	/// </summary>
	public readonly string NotificationType;

	/// <summary>
	/// The start time of the event.
	/// </summary>
	public readonly DateTime? StartTimeUtc;

	/// <summary>
	/// Whether the event is for a replica node.
	/// </summary>
	public readonly bool IsReplica;

	/// <summary>
	/// IPAddress of the node event is intended for
	/// </summary>
	public readonly IPAddress IpAddress;

	/// <summary>
	/// SSL port
	/// </summary>
	public readonly int SSLPort;

	/// <summary>
	/// Non-SSL Port
	/// </summary>
	public readonly int NonSSLPort;

	/// <summary>
	/// Returns a string representing the maintenance event with all of its properties.
	/// </summary>
	public override string ToString() =>
		$"{nameof(NotificationType)}|{NotificationType}|" +
			$"{nameof(StartTimeUtc)}|{StartTimeUtc:s}|" +
			$"{nameof(IsReplica)}|{IsReplica}|" +
			$"{nameof(IpAddress)}|{IpAddress}|" +
			$"{nameof(SSLPort)}|{SSLPort}|" +
			$"{nameof(NonSSLPort)}|{NonSSLPort}";
}
```

However, after doing that I don't really know that there's any reason to even parse it for our use case, so maybe we just record the string itself (for logs/events/whatever) and move along? If the only practical use of the event parsing is to have specifics we're turning back into a string for debugging...maybe we don't do any work on it at all. If we can log it as-is and that's equally readable as any `.ToString()` implementation: how about we simplify? We should be aiming to trigger the same reconfigure either way.

What are y'all thinking on these fronts?

amsoedal:
@NickCraver thanks for explaining!

I removed the configuration option since we're not sparing any connections by having it around, as you pointed out. I also cleaned up the tests and condensed the classes into an AzureMaintenanceEvent.AddListenerAsync() method as suggested, and it's so much cleaner now! 😃

The only thing I had some concerns about is the parsing -- I'm not sure if I understood your response correctly (please do correct me if I'm wrong) -- I think we do want to parse the message into a class, since the user might want to do some additional logic based on which event comes back and when the start time is. As for whether we keep the pipe-based iteration or not, I'm happy to switch to using the span method if it's more intuitive!

src/StackExchange.Redis/AzureMaintenanceEvent.cs
```diff
@@ -0,0 +1,143 @@
...
+using System.Net;
+using System.Threading.Tasks;
+using static StackExchange.Redis.ConnectionMultiplexer;
+
+namespace StackExchange.Redis
+{
+    /// <summary>
+    /// Azure node maintenance event details
+    /// </summary>
+    public class AzureMaintenanceEvent
```
NickCraver, line 11: For now, if we do remove the events, let's make this internal to start with so we can change the surface area if needed in a non-breaking way :)

src/StackExchange.Redis/AzureMaintenanceEvent.cs
```diff
@@ -0,0 +1,143 @@
...
+                    logProxy?.WriteLine("Failed to GetSubscriber for AzureRedisEvents");
+                    return;
+                }
+
+                await sub.SubscribeAsync("AzureRedisEvents", (channel, message) =>
+                {
+                    var newMessage = new AzureMaintenanceEvent(message);
+                    multiplexer.InvokeServerMaintenanceEvent(newMessage);
+
+                    if (StringComparer.OrdinalIgnoreCase.Equals(newMessage.NotificationType, "NodeMaintenanceEnded") || StringComparer.OrdinalIgnoreCase.Equals(newMessage.NotificationType, "NodeMaintenanceFailover"))
```
NickCraver, line 82: I was confused by examples and tests - can we depend on casing here _not_ changing? It'd make things cleaner, and would be nice if they were consistent.

src/StackExchange.Redis/AzureMaintenanceEvent.cs
```diff
@@ -0,0 +1,143 @@
...
+                    {
+                        multiplexer.ReconfigureAsync(first: false, reconfigureAll: true, log: logProxy, blame: null, cause: "server maintenance").Wait();
```
NickCraver, line 84: In the cause here, I recommend passing `$"Azure Event: {newMessage.NotificationType}" so we can tell which one it was in any debugger.

src/StackExchange.Redis/AzureMaintenanceEvent.cs
```diff
@@ -0,0 +1,143 @@
...
+                                NonSSLPort = nonsslport;
+                                break;
+
+                            default:
+                                break;
+                        }
+                    }
+                }
+            }
+            catch { }
```
NickCraver, line 63: I'm worried we'll silently throw here and not have any debuggability, but we don't really have an out for it today. One proposal (not for this PR) was to have a rolling log for debugging. Can we please add a comment here for later in the catch? e.g. `// TODO: Append to rolling debug log when it's present`

src/StackExchange.Redis/AzureMaintenanceEvent.cs
```diff
@@ -0,0 +1,143 @@
...
+            try
+            {
+                var sub = multiplexer.GetSubscriber();
+                if (sub == null)
+                {
+                    logProxy?.WriteLine("Failed to GetSubscriber for AzureRedisEvents");
+                    return;
+                }
+
+                await sub.SubscribeAsync("AzureRedisEvents", (channel, message) =>
```
NickCraver, line 77: Minor suggestion: recommend we make this channel name (AzureRedisEvents) a private constant up top, so that the primary entry point for this whole deal isn't down in and a bit easier for others to spot later.

NickCraver (commented):
Thanks for the changes! I left some more comments for refinement, but I'm a little fuzzy on if/why we need this to be public and may be missing info there, so take comments with a grain of salt. I see the mismatch on "do we need to parse" after the latest changes - if we expose, I agree we'd need to parse.

Tried to address everything in comments but happy to chat further - looking great!

src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -1277,7 +1290,7 @@ internal ServerEndPoint GetServerEndPoint(EndPoint endpoint, LogProxy log = null
 
         internal readonly CommandMap CommandMap;
 
-        private ConnectionMultiplexer(ConfigurationOptions configuration)
+        private ConnectionMultiplexer(ConfigurationOptions configuration, LogProxy logProxy)
```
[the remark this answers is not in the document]
amsoedal: Good catch!

src/StackExchange.Redis/AzureMaintenanceEvent.cs
amsoedal: Absolutely! I was unsure whether or not to add a comment here. Thanks for clarifying

src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -1187,6 +1199,7 @@ private static ConnectionMultiplexer ConnectImpl(ConfigurationOptions configurat
                         // Initialize the Sentinel handlers
                         muxer.InitializeSentinel(logProxy);
                     }
+                    AzureMaintenanceEvent.AddListenerAsync(muxer, logProxy).Wait();
```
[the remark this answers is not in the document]
amsoedal: Good call. Fixed

src/StackExchange.Redis/AzureMaintenanceEvent.cs
amsoedal: Yeah, we control the content of the messages so it makes sense to expect them not to change. Fixed

src/StackExchange.Redis/AzureMaintenanceEvent.cs
```diff
@@ -0,0 +1,143 @@
...
+        {
+            RawMessage = message;
+            try
+            {
+                var info = message?.Split('|');
+                for (int i = 0; i < info?.Length / 2; i++)
+                {
+                    string key = null, value = null;
+                    if (2 * i < info.Length) { key = info[2 * i].Trim(); }
+                    if (2 * i + 1 < info.Length) { value = info[2 * i + 1].Trim(); }
```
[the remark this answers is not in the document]
amsoedal: Fixed. I had to add some type conversions here and there (e.g. ToCharArray(), ToString()). But splitting the message is a bit clearer :)

amsoedal:
@NickCraver thanks for the excellent review! I've replied to most of the comments, and updated this PR's description with more context (specifically re: parsing and exposing events). Let me know if it makes sense, and hopefully we can sync up if there's anything else we need to align on! :)

src/StackExchange.Redis/ConfigurationOptions.cs
```diff
@@ -792,21 +792,21 @@ private void DoParse(string configuration, bool ignoreUnknown)
 
         internal bool IsAzureEndpoint()
```
philon-msft, line 793: Is there an easy way to make this `Lazy<>`, so it executes only once per `ConnectionMultiplexer` instance?

src/StackExchange.Redis/ConfigurationOptions.cs
```diff
@@ -792,21 +792,21 @@ private void DoParse(string configuration, bool ignoreUnknown)
...
         {
+            List<string> azureRedisHosts = new List<string> {
+                ".redis.cache.windows.net",
+                ".redis.cache.chinacloudapi.cn",
+                ".redis.cache.usgovcloudapi.net",
+                ".redis.cache.cloudapi.de",
+                ".redisenterprise.cache.azure.net"
+            };
+
             foreach (var ep in EndPoints)
```
philon-msft, line 803: Can we replace this `foreach` with `Linq` too?

src/StackExchange.Redis/ConfigurationOptions.cs
mgravell: This only happens during init/connect; I wouldn't be concerned about memoizing anything here - that would seem to add memory overhead for something that will never (or very rarely) be used again

src/StackExchange.Redis/ConfigurationOptions.cs
mgravell: I'm always a little cautious of blindly applying LINQ; you tend to end up with unexpected closures; see `dnsEp` in `Any`, for example. I'm not saying "never use LINQ", but: if the existing code works fine and is clear: I wouldn't reach for LINQ, either

src/StackExchange.Redis/AzureMaintenanceEvent.cs
```diff
@@ -0,0 +1,153 @@
...
+                {
+                    logProxy?.WriteLine("Failed to GetSubscriber for AzureRedisEvents");
+                    return;
+                }
+
+                await sub.SubscribeAsync(PubSubChannelName, (channel, message) =>
+                {
+                    var newMessage = new AzureMaintenanceEvent(message);
+                    multiplexer.InvokeServerMaintenanceEvent(newMessage);
+
```
mgravell, line 92: suggestion: add a hyperlink to the azure documentation of what maintenance events exist, and what they mean?

src/StackExchange.Redis/ConfigurationOptions.cs
amsoedal: Makes sense. Will resolve

src/StackExchange.Redis/AzureMaintenanceEvent.cs
amsoedal: Added a link in the comment at the top of the file!

src/StackExchange.Redis/AzureMaintenanceEvent.cs
```diff
@@ -0,0 +1,153 @@
...
+        internal AzureMaintenanceEvent(string azureEvent)
+        {
+            if (azureEvent == null)
+            {
+                return;
+            }
+
+            // The message consists of key-value pairs delimted by pipes. For example, a message might look like:
+            // NotificationType|NodeMaintenanceStarting|StartTimeUtc|2021-09-23T12:34:19|IsReplica|False|IpAddress|13.67.42.199|SSLPort|15001|NonSSLPort|13001
+            var message = new ReadOnlySpan<char>(azureEvent.ToCharArray());
```
mgravell, line 24: suggest: `var message = azureEvent.AsSpan();` - the `.ToCharArray()` forces an extra allocation; `.AsSpan()` creates the ROS directly over the existing `string`

src/StackExchange.Redis/AzureMaintenanceEvent.cs
```diff
@@ -0,0 +1,153 @@
...
+                while (message.Length > 0)
+                {
+                    if (message[0] == '|')
+                    {
+                        message = message.Slice(1);
+                        continue;
+                    }
+
+                    // Grab the next pair
+                    var key = message.Slice(0, message.IndexOf('|'));
```
mgravell, line 36: this line presumes that there *is* another pair - you should, IMO, be very guarded against malformed messages here; what happens if there *is no more `|`*, and `message.IndexOf('|')` returns negative?

src/StackExchange.Redis/AzureMaintenanceEvent.cs
```diff
@@ -0,0 +1,153 @@
...
+
+                    var valueEnd = message.IndexOf('|');
+                    var value = valueEnd > -1 ? message.Slice(0, valueEnd) : message;
+                    message = message.Slice(value.Length);
+
+                    if (key.Length > 0 && value.Length > 0)
+                    {
+                        switch (key)
+                        {
+                            case var _ when key.SequenceEqual(nameof(NotificationType).ToCharArray()):
```
mgravell, line 47: these `.ToCharArray()` allocations are unnecessary; on modern frameworks (from netcoreapp3.0 onwards, I believe) there is an implicit conversion from `string` to `ReadOnlySpan<char>`, but since we have down-level TFMs in scope, you might need to use `nameof(NotificationType).AsSpan()`

(applies to all similar in this `switch`)

src/StackExchange.Redis/AzureMaintenanceEvent.cs
```diff
@@ -0,0 +1,153 @@
...
+                                NotificationType = value.ToString();
+                                break;
+                            case var _ when key.SequenceEqual(nameof(StartTimeInUTC).ToCharArray()) && DateTime.TryParse(value.ToString(), out DateTime startTime):
```
mgravell, line 50: These `.ToString()` are avoidable; on modern TFMs, there are span overloads on the types directly, but before that: consider `System.Buffers.Text.Utf8Parser.TryParse`, which has a range of APIs; also - and this is a very important one: think i18n/l10n: I *would* say "explicitly specify invariant", but I believe if using `Utf8Parser`: it is assumed - you should probably check on that, though

src/StackExchange.Redis/AzureMaintenanceEvent.cs
```diff
@@ -0,0 +1,153 @@
...
+                                StartTimeInUTC = DateTime.SpecifyKind(startTime, DateTimeKind.Utc);
+                                break;
+                            case var _ when key.SequenceEqual(nameof(IsReplica).ToCharArray()) && bool.TryParse(value.ToString(), out var isReplica):
+                                IsReplica = isReplica;
+                                break;
+                            case var _ when key.SequenceEqual(nameof(IPAddress).ToCharArray()) && IPAddress.TryParse(value.ToString(), out var ipAddress):
```
mgravell, line 56: note: no overload in `Utf8Parser` here, so this might be a good candidate for:

``` c#
#if NETSTANDARD2_1_OR_GREATER || NETCOREAPP3_0_OR_GREATER
    && IPAddress.TryParse(value, out var ipAddress):
#else
    && IPAddress.TryParse(value.ToString(), out var ipAddress):
#endif

File: src/StackExchange.Redis/AzureMaintenanceEvent.cs [hunk 17]
```diff
@@ -0,0 +1,153 @@
...
+            try
+            {
+                var sub = multiplexer.GetSubscriber();
+                if (sub == null)
+                {
+                    logProxy?.WriteLine("Failed to GetSubscriber for AzureRedisEvents");
+                    return;
+                }
+
+                await sub.SubscribeAsync(PubSubChannelName, (channel, message) =>
```
mgravell, line 88: it probably won't matter, but I wonder if this `await` is needed - i.e. could remove the `async` and just `return` here; I suppose that would potentially change how/when `catch (Exception e)` is seen, though...

src/StackExchange.Redis/AzureMaintenanceEvent.cs
```diff
@@ -0,0 +1,153 @@
...
+            catch (Exception e)
+            {
+                logProxy?.WriteLine($"Encountered exception: {e}");
+            }
+        }
+
+        /// <summary>
+        /// Raw message received from the server
+        /// </summary>
+        public readonly string RawMessage;
```
mgravell, line 108: should be get-only props (all) - i.e. `public string RawMessage {get;}` - not public fields

src/StackExchange.Redis/AzureMaintenanceEvent.cs
```diff
@@ -0,0 +1,153 @@
...
+
+        /// <summary>
+        /// indicates the event type
+        /// </summary>
+        public readonly string NotificationType;
+
+        /// <summary>
+        /// indicates the start time of the event
+        /// </summary>
+        public readonly DateTime? StartTimeInUTC;
```
mgravell, line 118: Suggest `StartTimeUtc` (the "in" feels superfluous, and applying same casing convention as `DateTime.UtcNow`)

src/StackExchange.Redis/AzureMaintenanceEvent.cs
```diff
@@ -0,0 +1,153 @@
...
+
+        /// <summary>
+        /// non-ssl port
+        /// </summary>
+        public readonly int NonSSLPort;
+
+        /// <summary>
+        /// Returns a string representing the maintenance event with all of its properties
+        /// </summary>
+        public override string ToString()
```
mgravell, line 143: I wonder if this should just return `RawMessage` ?

src/StackExchange.Redis/ConfigurationOptions.cs
```diff
@@ -790,22 +790,25 @@ private void DoParse(string configuration, bool ignoreUnknown)
...
             {
                 if (ep is DnsEndPoint dnsEp)
                 {
-                    int firstDot = dnsEp.Host.IndexOf('.');
-                    if (firstDot >= 0)
+                    foreach (var host in azureRedisHosts)
                     {
-                        switch (dnsEp.Host.Substring(firstDot).ToLowerInvariant())
+                        if (dnsEp.Host.ToLowerInvariant().EndsWith(host))
```
mgravell, line 809: this means we're applying the `ToLowerInvariant` N times, for `azureRedisHosts` size N; seems avoidable?

src/StackExchange.Redis/AzureMaintenanceEvent.cs
```diff
@@ -0,0 +1,153 @@
...
+                if (sub == null)
+                {
+                    logProxy?.WriteLine("Failed to GetSubscriber for AzureRedisEvents");
+                    return;
+                }
+
+                await sub.SubscribeAsync(PubSubChannelName, (channel, message) =>
+                {
+                    var newMessage = new AzureMaintenanceEvent(message);
+                    multiplexer.InvokeServerMaintenanceEvent(newMessage);
```
mgravell, line 91: we can avoid the `AzureMaintenanceEvent` alloc when no subscriber, if we just pass down `message`, i.e. 

``` c#
multiplexer.InvokeServerMaintenanceEvent(message);
// ...
internal void InvokeServerMaintenanceEvent(string message)
  => AzureServerMaintenanceEvent?.Invoke(this, new AzureMaintenanceEvent(message));
```

the args on the right here are only evaluated if the `?.` check passes, i.e. we have listeners

mgravell (changes requested):
did a review pass; heading in the right direction, but quite a few suggestions; please please feel free to discuss any of them (not trying to be a dictator here)

mgravell:
oh, can we update releasenotes.md too please, in the unreleased section? thanks

mgravell:
The DateTime parse is of particular concern to me; I wonder if the format should be explicit/"exact" here (i.e. `DateTime.TryParseExact`) - or is it sufficient to use the `'u'` pattern (which is not the default) on `Utf8Parser`? Does `'u'` act the same [as `"u"`?](https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-date-and-time-format-strings#UniversalSortable) (open question; I don't know the answer)

src/StackExchange.Redis/AzureMaintenanceEvent.cs
amsoedal: Thanks Marc! I haven't worked with spans too much so the pointers are appreciated :)

src/StackExchange.Redis/ConfigurationOptions.cs
```diff
@@ -804,9 +804,12 @@ internal bool IsAzureEndpoint()
             {
                 if (ep is DnsEndPoint dnsEp)
                 {
-                    if (azureRedisHosts.Any(s => dnsEp.Host.ToLowerInvariant().EndsWith(s)))
+                    foreach (var host in azureRedisHosts)
```
mgravell, line 807: nice unroll; just as a "ooh, while we're there, this could be better" - would this work with `dnsEp.Host.EndsWith(host, StringComparison.InvariantCultureIgnoreCase)` ?

amsoedal: You got it!

src/StackExchange.Redis/AzureMaintenanceEvent.cs
amsoedal: Makes sense. We have some unit tests testing for messages like this, although I can add some specific checks so we don't always go straight to the "catch" block and handle them explicitly

src/StackExchange.Redis/AzureMaintenanceEvent.cs
amsoedal: If it's alright with you, I'll probably keep this as-is since this code only gets called during Connect. (I'm also not sure how it will impact the execution of the catch block)

src/StackExchange.Redis/AzureMaintenanceEvent.cs
amsoedal: 👍 will update

src/StackExchange.Redis/AzureMaintenanceEvent.cs
amsoedal: Makes sense -- I updated the property name. Unfortunately this means we have to treat this property differently in the switch block since we're stuck with the `StartTimeInUTC` name from the server side, but let me know if it looks ok to you.

src/StackExchange.Redis/AzureMaintenanceEvent.cs
amsoedal: Even when there's no subscriber, we currently want the AzureMaintenanceEvent handy so that we can read what type of event it is and decide whether to reconfigure or not. Let me know if I understood you correctly :)

src/StackExchange.Redis/AzureMaintenanceEvent.cs
amsoedal: I tried applying the Utf8Parser, but it deals with ReadOnlySpan<byte> instead of ReadOnlySpan<char>, which is what we're working with here. Adding the extra conversion between the two seems like the opposite of what we're trying to do here? Also [the documentation](https://docs.microsoft.com/en-us/dotnet/api/system.buffers.text.utf8parser.tryparse?view=net-5.0) doesn't mention anything about cultures, so I think it makes sense to opt for the DateTime.TryParseExact() option for StartTimeUtc so we can specify it explicitly, as per your other comment. 
I ended up using the preprocessor directive you mentioned in another comment to get rid of all the .ToString() conversions for the newer frameworks.

amsoedal: > The DateTime parse is of particular concern to me; I wonder if the format should be explicit/"exact" here (i.e. `DateTime.TryParseExact`) - or is it sufficient to use the `'u'` pattern (which is not the default) on `Utf8Parser`? Does `'u'` act the same [as `"u"`?](https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-date-and-time-format-strings#UniversalSortable) (open question; I don't know the answer)

Regarding this comment -- on the Azure side we're sending these events with the "s" (sortable) format, but according to VS 'u' looks like this: 

image

And 'U' looks like: 

image

src/StackExchange.Redis/AzureMaintenanceEvent.cs
mgravell: unless I'm missing it, all we do is `AzureServerMaintenanceEvent?.Invoke` with the event; I'm just trying to avoid the overhead if `AzureServerMaintenanceEvent` turns out to be `null`, which we can reasonably presume is the default and most common scenario; do we do anything else special in the case when no external consumer has touched the event?

src/StackExchange.Redis/AzureMaintenanceEvent.cs
mgravell: fair enough; whatever works :) My key point: we need to make sure it is culture-insensitive, and we should *probably* validate that in a test by changing (and restoring in  a `finally`) the thread's culture and ui-culture to something random and suitably awkward

src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -892,6 +901,11 @@ private static async Task<ConnectionMultiplexer> ConnectImplAsync(ConfigurationO
                         // Initialize the Sentinel handlers
                         muxer.InitializeSentinel(logProxy);
                     }
+
+                    if (configuration.IsAzureEndpoint())
```
mgravell, line 905: `&& CommandMap.IsAvailable(RedisCommand.SUBSCRIBE)` ?

src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -1187,6 +1201,10 @@ private static ConnectionMultiplexer ConnectImpl(ConfigurationOptions configurat
                         // Initialize the Sentinel handlers
                         muxer.InitializeSentinel(logProxy);
                     }
+                    if (configuration.IsAzureEndpoint())
```
mgravell, line 1204: `&& CommandMap.IsAvailable(RedisCommand.SUBSCRIBE)` ?

NickCraver:
Hey @amsoedal! I figured we'd see you on call today but missed out on syncing up - figured it was easier to talk through some tweaks to make this more extensible to other providers later and not lock ourselves in on the new public types with a few changes. I have the proposals in craver/1865-tweaks but definitely don't want to stomp on PR and credit here. I've granted you write access to this repo - could you please move your branch here and open a new PR (I think all current comments are good to lose). Then we can open a PR from `craver/1865-tweaks` as a suggested PR to that branch :) Curious on thoughts here, and told @philon-msft I'm happy to hop on a call tomorrow to go through and finish this off if sync helps :)

PR #1876, opened 2021-10-05
Add support for listening to Azure Maintenance Events

amsoedal:
Adding an automatic subscription to the AzureRedisEvents pubsub channel for Azure caches. This channel notifies clients of upcoming maintenance and failover events.

By exposing these events, users will be able to know about maintenance ahead of time, and can implement their own logic (e.g. diverting traffic from the cache to another database for the duration of the maintenance event) in response, with the goal of minimizing downtime and disrupted connections. We also automatically refresh our view of the topology of the cluster in response to certain events.

Here are some of the possible notifications:

```
// This event gets fired ~20s before maintenance begins
NodeMaintenanceStarting,

// This event gets fired when maintenance is imminent (<5s)
NodeMaintenanceStart,

// Indicates that the node maintenance operation is over
NodeMaintenanceEnded,

// Indicates that a replica has been promoted to primary
NodeMaintenanceFailover
```

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -7,6 +7,7 @@
 - logging additions (.NET Version and timestamps) for better debugging (#1796 via philon-msft)
 - add: `Condition` API (transactions) now supports `StreamLengthEqual` and variants (#1807 via AlphaGremlin)
 - fix potential task/thread exhaustion from the backlog processor (#1854 via mgravell)
+- add support for listening to Azure Maintenance Events (#1865 via amsoedal)
 
 ## 2.2.62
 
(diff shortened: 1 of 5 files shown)


amsoedal:
Hi @NickCraver, I closed https://github.com/StackExchange/StackExchange.Redis/pull/1865 and opened this one instead! Also working on addressing remaining comments from @mgravell. Big thanks to both of you!

NickCraver:
Thanks @amsoedal! I opened #1877 with suggestions to fixes here (poking ASAP so we don't overlap) - can you please look when you have a chance?

NickCraver:
I merged in an unrelated PR on the search side - updating to get those failing test runs off the radar.

PR #1888, opened 2021-10-13
Updated the Configuration.md to verbalize ECHO permission require

blackicec:
My ACL user was never granted the ECHO permission since my logic never had a use case for it. When I first configured StackExchange.Redis, I kept getting a connection exception, but I have a nodejs implementation that uses the node-redis npm package to connect to my server just fine. I looked through your logs and source code and found what I was missing.

"...Error: NOPERM this user has no permissions to run the 'echo' command or its subcommand [serverName]:6379: OnConnectedAsync completed (Disconnected)..."

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -1,7 +1,7 @@
 ﻿Configuration
 ===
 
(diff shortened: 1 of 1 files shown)


blackicec:
Agreed. I've added in the version 6 wording to make it more clear. Thank you.

PR #1896, opened 2021-10-25
Change default reconnection policy to Exponential

lolodi:
Change default reconnection policy to Exponential
Update the Exponential policy to take the base as parameter to allow for different growth rates

Description:
To speed up reconnections, and also avoid reconnection storms, we are changing the default reconnection policy to ExponentialRetry and reducing the deltaBackOff from the previous default of 5s to 2.5s. 
This allows more frequent retries early on, when a reconnection is more probable, and less frequent retries later on when it has already been impossible for some time.
Another plus for the ExponentialRetry is the additional randomness added to the intervals so that we avoid reconnection storms.
Here is a comparison of the two retry policies.
image
image

And here is a real world example:
image
The randomized exponential (orange) is within the the minimum delta (yellow) and the maximum exponential (gray).
The retry happens when the timeout (blue) is greater than the retry limit (orange).


src/StackExchange.Redis/ConfigurationOptions.cs
```diff
@@ -334,7 +334,7 @@ public bool PreserveAsyncOrder
         /// <summary>
         /// The retry policy to be used for connection reconnects
         /// </summary>
-        public IReconnectRetryPolicy ReconnectRetryPolicy { get { return reconnectRetryPolicy ??= new LinearRetry(ConnectTimeout); } set { reconnectRetryPolicy = value; } }
```
philon-msft, line 337: >??= [](http://example.com/codeflow?start=94&length=3)

should remain `??=`

philon-msft:
> 

we can revert all changes in this file now

---
Refers to: src/StackExchange.Redis/ExponentialRetry.cs:2 in 033a09b. [](commit_id = 033a09bae5ef14f6de3c35dc312d3377af1e826d, deletion_comment = False)

philon-msft (approved):

NickCraver:
This was a ways behind - merging main in so that we're good to go!

Issue #1864, opened 2021-09-19
Backlog, Backpressure, and Retry - Overview and Planning


NickCraver:
In #1856 we’ve been trying to access what’s needed in a retry policy to help handle interruption cases users see today in a more graceful and hopefully often more transparent way. However, in exploring the realities of this at all layers, it’s far more nuanced than has ever been considered in this library before. This issue is to round up discovery and thinking of these fronts.

I am _not_ discussing specific implementation of solutions here, more rounding up “what problems are we trying to solve?” and the details of that question. The most important step of any solution is understanding the problem - and the group working on this thus far think we’re about to that point: understanding the intricacies of what’s going on. This issue is to detail all that info and discuss/clarify anything further.

Note: I'm using the term "command" generically here - this is concretely a `Message` in our library.

### Overview/Context
The fundamental architecture of StackExchange.Redis is a pipelined approach. For a single `ServerEndPoint`, all commands are in a conveyor-belt style pipeline. One command can block others if it takes a long time to send or receive, and when a connection failure happens where it happened, when it happened, what it affects, and what’s behind that are all interesting combinations of things.

The overall hierarchy of the structure in StackExchange.Redis is:
```
- ConnectionMultiplexer
  - ServerEndPoint
    - PhysicalBridge
      - PhysicalConnection
        - IDuplexPipe
        - Socket (used in the pipe as well)
```

When a command is sent through the multiplexer, it may specify a particular endpoint (rare) or auto-select an endpoint (common) based on preference. This can have some steering by `CommandFlags` (e.g. `CommandFlags.DemandReplica`).

A server endpoint could not be connected in a few ways:
1. It hasn’t connected yet (initial connection)
2. It has disconnected and we don’t realize it yet
3. It was disconnected and hasn’t reconnected yet (failure)

It’s worth noting that the first case (not connected yet) is the simplest and generally well handled and seems solid today. It’s simple to reason about everything involved and I think Tim and I solved the remaining issues there. Reconnects though, there be our dragons.

Importantly, we don’t always know when a socket dies. It can take anywhere from milliseconds to minutes to determine this happens. This could be for a few primary reasons:
1. Network error, we lost the socket connection for some reason
2. Unknown failover of the server (not graceful, reactive)
3. Known failover of the server (graceful, proactive)

[...]

PR #1912, opened 2021-11-16
v2.5 work: BacklogPolicy

NickCraver:
This is a work in progress of the Backlog Policy bits to help address #1864. The overall change here is to make the message pathway backlog commands when an endpoint is down. It does not hand off to another viable endpoint in this iteration (future plans there). If the policy is set to `FailFast`, we should get the old behavior not queueing anything. In the new path:
- Handshake commands go straight to the IO pipe (the only way this would fail is if we lost a pipe _during a handshake_, in which case things are ill and let's backoff until the next heartbeat anyway).
- If connected and there is no backlog queue, commands go into the pipe
- If disconnected or there is a backlog queue, non-internal and non-handshake commands go into the backlog, so that they are played to the pipe in order

TODO List:
- [x] Implement simplified backlog queue (3 queues -> 1, no handoffs)
- [x] Implement backlog policy and config
- [x] Add initial testing
- [x] Find intermittent async hang in local test runner (test: `QueuesAndFlushesAfterReconnecting`) - this may be boneheaded or a fundamental flaw
- [x] Add test for sync commands (e.g. `.Ping()`) and that they recover appropriately


src/StackExchange.Redis/PhysicalBridge.cs
```diff
@@ -864,20 +979,21 @@ private async Task ProcessBacklogAsync()
                 int failureCount = 0;
 #endif
                 _backlogStatus = BacklogStatus.Starting;
+
                 while (true)
                 {
                     // check whether the backlog is empty *before* even trying to get the lock
-                    if (_backlog.IsEmpty) return; // nothing to do
+                    if (backlog.IsEmpty) return; // nothing to do
```
philon-msft, line 987: If we stick with a 'single backlog' design, can we continue using `_backlog` rather than passing `backlog` into this method?

src/StackExchange.Redis/PhysicalBridge.cs
```diff
@@ -776,34 +841,40 @@ private bool PushToBacklog(Message message, bool onlyIfExists)
...
         private volatile int _backlogProcessorRequestedTime;
 #endif
 
-        private void CheckBacklogForTimeouts() // check the head of the backlog queue, consuming anything that looks dead
+        /// <summary>
+        /// Crawls from the head of the backlog queue, consuming anything that should have timed out
+        /// and pruning it accoordingly (these messages will get timeout exceptions).
+        /// </summary>
+        private void CheckBacklogsForTimeouts()
         {
```
philon-msft, line 878: name nit: `CheckBacklogForTimeouts` ('backlog' singular), if we're going to have one backlog?

src/StackExchange.Redis/PhysicalBridge.cs
```diff
@@ -852,9 +925,51 @@ internal enum BacklogStatus : byte
             SettingIdle,
             Faulted,
         }
+
         private volatile BacklogStatus _backlogStatus;
-        private async Task ProcessBacklogAsync()
+        private async Task ProcessBacklogsAsync()
+        {
```
philon-msft, line 931: name nit: `ProcessBacklogAsync` ('backlog' singular)?

src/StackExchange.Redis/PhysicalConnection.cs
```diff
@@ -68,6 +68,7 @@ internal void GetBytes(out long sent, out long received)
         }
 
         private IDuplexPipe _ioPipe;
+        internal bool HasOuputPipe => _ioPipe?.Output != null;
```
philon-msft, line 71: >HasOuputPipe [](http://example.com/codeflow?start=22&length=12)

spelling nit: `HasOutputPipe` (missing a 't')

tests/StackExchange.Redis.Tests/BacklogTests.cs
```diff
@@ -0,0 +1,158 @@
+﻿using System;
+using System.Threading.Tasks;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace StackExchange.Redis.Tests
+{
+    public class BacklogTests : TestBase
+    {
```
philon-msft, line 9: Is there a good wait to test that handshake commands fail fast, even with the default BacklogPolicy?

src/StackExchange.Redis/PhysicalBridge.cs
```diff
@@ -914,7 +1030,7 @@ private async Task ProcessBacklogAsync()
 #endif
                             message.SetExceptionAndComplete(ex, this);
                         }
-                        else
+                        else if (physical?.HasOuputPipe == true)
                         {
```
philon-msft, line 1034: Do we need to re-queue the message if `!physical.HasOutputPipe`?

tests/StackExchange.Redis.Tests/BacklogTests.cs
NickCraver: Since handshake commands won't go through this they shouldn't be subject to the policy in general - at least that's the intent, but if you see otherwise I need to fix!

src/StackExchange.Redis/PhysicalBridge.cs
NickCraver: On this one I want to leave it to be clear we're handling multiple backlogs, in total - e.g. when we flush cross-connection. May be semantically incorrect sometimes, but making you think is good - I'll add comments to help here.

src/StackExchange.Redis/PhysicalBridge.cs
NickCraver: This is a bit of prep for possible vNext - where we feed _another_ bridge's backlog in to process it ;)

NickCraver:
Pointed this on top of the tie breaker changes to eliminate the PR diff. I'm debating the `isHandshake` method changes vs. a method flag in the write path there - will wait for @mgravell to chime in tomorrow. It wouldn't save a lot, but it would be slightly cleaner I think and only accessed as an additional on _when we're already in trouble_, so meh on cost.

NickCraver:
Note the last build on Windows hung here for a very long time - we still have a gremlin in play with this approach.

NickCraver:
Alrighty I think we have a legit problem with SubscriptionsSurviveConnectionFailureAsync which reproduces quite a lot now (not 100%, and rarely locally, but often in the build server).

We've seen it in the past and wrote it off as flaky but I think I've legit improved the _chances_ of an existing race/failure. This may be unrelated to the PR here in terms of what's broken, but we can finally find/fix it. @mgravell if you get eyes earlier, `SubscriptionsSurviveConnectionFailureAsync` is where the fun is - can run only that test local and get it to repro ~10% of the time here. I'll poke in the morning and we can take a peek. Based on some issues filed in the past...I think this is a legit failure to figure out.

NickCraver:
Noteworthy: the resubscribe (see `ResendSubscriptions`) is a `msg.SetInternalCall();`, so I wonder if we're discarding it in a race somewhere I'm not seeing. We don't want to error for a user, but we _do_ want to queue and send a resubscription. In theory though since we're in `OnFullyEstablished`, we're actively connected and good to go. We should only be getting calls to that method with the subscription bridge and being called when it's up and running (unless `sub.Ping();` isn't going on the bridge we think it is...which glancing at it end-of-day here...is possible). We may have been pinging the wrong (interactive) connection the entire time.

More tomorrow, just wanted to add thoughts there.

src/StackExchange.Redis/RedisServer.cs
```diff
@@ -575,7 +575,7 @@ internal override Task<T> ExecuteAsync<T>(Message message, ResultProcessor<T> pr
         {   // inject our expected server automatically
             if (server == null) server = this.server;
             FixFlags(message, server);
-            if (!server.IsConnected)
+            if (!server.IsConnected && !multiplexer.RawConfig.BacklogPolicy.QueueWhileDisconnected)
```
TimLovellSmith, line 578: This particular change felt surprising to me. FireAndForget would only be FireAndForget if we're connected? Sort of makes sense, but.... just from the most cursory read of this logic, feels like there is a risk that it could lead to _blocking_ behavior which might be a surprise to people using fire and forget because they want nonblocking behavior?

src/StackExchange.Redis/PhysicalConnection.cs
```diff
@@ -649,7 +649,8 @@ internal void OnBridgeHeartbeat()
                     var timeout = bridge.Multiplexer.AsyncTimeoutMilliseconds;
                     foreach (var msg in _writtenAwaitingResponse)
                     {
-                        if (msg.HasAsyncTimedOut(now, timeout, out var elapsed))
+                        // We only handle async timeouts here, synchronous timeouts are handled upstream.
```
TimLovellSmith, line 652: The direction of 'upstream' isn't that conventionally clear to me. Does it mean before OnBridgeHeartbeat would be called (since upstream is the 'source' of the data flowing)? Or does it just mean 'somewhere else in the class which normally calls into this?'

src/StackExchange.Redis/Message.cs
```diff
@@ -662,21 +656,17 @@ internal void SetWriteTime()
...
-            {
-                millisecondsTaken = unchecked(now - _writeTickCount); // note: we can't just check "if sent < cutoff" because of wrap-aro
-                if (millisecondsTaken >= timeoutMilliseconds)
-                {
-                    Flags &= ~NeedsAsyncTimeoutCheckFlag; // note: we don't remove it from the queue - still might need to marry it up; but: it is toast
-                    return true;
-                }
-            }
-            else
+            millisecondsTaken = unchecked(now - _writeTickCount); // note: we can't just check "if sent < cutoff" because of wrap-aro
```
TimLovellSmith, line 666: did 'wrap-around' get truncated?

src/StackExchange.Redis/Message.cs
```diff
@@ -60,8 +58,7 @@ internal abstract class Message : ICompletable
 
         private const CommandFlags AskingFlag = (CommandFlags)32,
                                    ScriptUnavailableFlag = (CommandFlags)256,
-                                   NeedsAsyncTimeoutCheckFlag = (CommandFlags)1024,
-                                   DemandSubscriptionConnection = (CommandFlags)2048;
+                                   DemandSubscriptionConnection = (CommandFlags)1024;
```
TimLovellSmith, line 61: Nit, I habitually avoid changing the numeric value of enums. Why take any risk of something thats hard to debug?

src/StackExchange.Redis/PhysicalConnection.cs
NickCraver: Adding an additional comment here to clarify:
```cs
// Those sync timeouts happen in ConnectionMultiplexer.ExecuteSyncImpl() via Monitor.Wait.
```

src/StackExchange.Redis/Message.cs
NickCraver: haha yes, this is moved from previous code and I've fixed (and lost it) across branches 3 times now...I'm feeling good about number 4 though.

src/StackExchange.Redis/Message.cs
NickCraver: Since these are all internal to-the-build it's more of a tidy, but I also have no strong feelings here - will migrate back!

src/StackExchange.Redis/RedisServer.cs
NickCraver: We need to do _something_ here to not throw but I agree those 2 cases aren't quite right - good catch! I think this should be adjusted slightly in that we only want to wrap in the "throw to bail" in both this and ExecuteSync below - will adjust/test and ping for re-review here shortly :)

NickCraver: @TimLovellSmith alrighty I think this is as it should be now, can you please give eyes/thoughts on it again when time allows?

src/StackExchange.Redis/BacklogPolicy.cs
```diff
@@ -0,0 +1,43 @@
+﻿namespace StackExchange.Redis
+{
+    /// <summary>
+    /// The backlog policy to use for commands. This policy comes into effect when a connection is unhealthy or unavailable.
+    /// The policy can choose to backlog commands and wait to try them (within their timeout) against a connection when it comes up,
+    /// or it could choose to fail fast and throw ASAP. Different apps desire different behaviors with backpressure and how to handle
+    /// large amounts of load, so this is configurable to optimize the happy path but avoid spiral-of-death queue scenarios for others.
+    /// </summary>
+    public class BacklogPolicy
```
mgravell, line 9: is this designed for inheritance? if not, recommend `sealed`

mgravell (approved):

src/StackExchange.Redis/ServerSelectionStrategy.cs
```diff
@@ -114,19 +114,19 @@ public ServerEndPoint Select(Message message)
                     if (slot == MultipleSlots) throw ExceptionFactory.MultiSlot(multiplexer.IncludeDetailInExceptions, message);
                     break;
             }
-            return Select(slot, message.Command, message.Flags);
+            return Select(slot, message.Command, message.Flags, allowDisconnected);
         }
 
-        public ServerEndPoint Select(RedisCommand command, in RedisKey key, CommandFlags flags)
+        public ServerEndPoint Select(RedisCommand command, in RedisKey key, CommandFlags flags, bool allowDisconnected = false)
```
philon-msft, line 120: Eliminate the `= false` defaults, so callers are always explicit about it?

philon-msft (approved):

src/StackExchange.Redis/ConfigurationOptions.cs
```diff
@@ -146,6 +146,8 @@ public static string TryNormalize(string value)
 
         private IReconnectRetryPolicy reconnectRetryPolicy;
 
+        private BacklogPolicy backlogPolicy;
```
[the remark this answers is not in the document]
NickCraver: `.Default` is a static ref as a fallback, so no creation happening. The problem with initing explicitly is you get fun cloning consequences later :)

We could look at hardening the value to the field to eliminate the fallback in a few paths but don't think it'll change perf too much either way.

src/StackExchange.Redis/ServerSelectionStrategy.cs
NickCraver: I was torn on this - wanted to make it obvious that _by default_ it should be `false` (I'd love an analyzer that requires putting the argument name everywhere it's used). No strong feelings here, but other approaches weren't as clean, e.g. flags on a message had overhead and don't work where we're passing just a command for things like server selection, so stuck with true/false down the pipe.

deepakverma:
This looks great and thank you for incorporating it as a single queue implementation :)

src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -2169,6 +2169,12 @@ private bool PrepareToPushMessageToBridge<T>(Message message, ResultProcessor<T>
             {
                 // Infer a server automatically
                 server = SelectServer(message);
+
+                // If we didn't find one successfully, and we're allowed, queue for any viable server
```
deepakverma, line 2173: If I am understand it correctly, I think it will be interesting to test this logic with a clustered cache. In case of a failure, the logic will select any viable server. This viable server might return a moved exception and the message will get redirected to the right endpoint. It should be fine if eventually it gets to the right endpoint. Does that sound correct?

NickCraver: Correct! Let's make sure in tests though - I'll add some cluster tests around this to the PR.

NickCraver: Added additional testing in 2c2b0ea51c406fecd765f851a96acbbce5f06f2d!

PR #1914, opened 2021-11-17
Change result type/value in the xml tag of the method documentation

testfirstcoder:
Change result type/value in the documentation for the method `HashSet`: `1` => `True` and `0` => `False`


mgravell (approved):

mgravell:
Thanks, appreciated

PR #1915, opened 2021-11-22
Update Server.md

rkalasky:
Adding GCP offering to list of cloud providers


mgravell:
My one reservation here is that we have, to date, been unable to validate Memorystore, and I know of at least one "it doesn't work" (with the library) remark. Does it work for you?

mgravell (approved):

rkalasky:
Memorystore + StackExchange.Redis is working great in production for Buildertrend.

PR #1921, opened 2021-12-01
Ensure valid timeout settings in ExponentialRetry

gliljas:
Setting a timeout from the connectionstring can cause `deltaBackOffMilliseconds `to be lower than `maxDeltaBackOffMilliseconds`, which will cause an `ArgumentOutOfRangeException `('minValue' cannot be greater than maxValue.) in `ShouldRetry`.


NickCraver (approved):
Thanks for the additional checks here!

Issue #1926, opened 2021-12-12
Exception when using KeysAsync over a huge database


alexandrutatagit:
Hi!
I have an issue: I want to check all the keys for a specific condition, once per day. In the last days, on PROD, the size of the DB has increased to 50M. I am doing the following: 
await foreach (var item in server.KeysAsync(_db)){
...
}
I have updated the connection string to use a 10 min timeout:
allowAdmin=true,connectTimeout=600000,syncTimeout=600000,asyncTimeout=600000,keepAlive=600,$INFO=,$PING=,name=...

But still after 100 sec I get this error:

[...]

PR #1929, opened 2021-12-14
Default versions: assume v2.8 as a minimum (SCAN over KEYS) and 4.0+ in Azure

NickCraver:
This more reflects reality today and if we cannot detect a server version (e.g. config has INFO disabled as in #1926) then we'll do smarter things like use SCAN over KEYS.


mgravell (approved):

philon-msft (approved):

philon-msft (approved):

NickCraver:
hahahaha, it's a tiebreaker test failing - unrelated but hilarious, the other should help here :)

PR #1930, opened 2021-12-14
Initialize ScriptEvalMessage with the correct command

martinpotter:
Currently the constructor for ScriptEvalMessage that takes a SHA1 hash, and sends the `EVALSHA` command, initializes the base Message class's command to `RedisCommand.EVAL` instead of `RedisCommand.EVALSHA`.

### DIFF
--- a/src/StackExchange.Redis/RedisDatabase.cs
+++ b/src/StackExchange.Redis/RedisDatabase.cs
@@ -3807,7 +3807,7 @@ namespace StackExchange.Redis
             }
 
             public ScriptEvalMessage(int db, CommandFlags flags, byte[] hash, RedisKey[] keys, RedisValue[] values)
-                : this(db, flags, RedisCommand.EVAL, null, hash, keys, values)
+                : this(db, flags, RedisCommand.EVALSHA, null, hash, keys, values)
             {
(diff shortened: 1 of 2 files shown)


mgravell:
could you please also add a line to releasenotes.md (you'll see the pattern); if you're feeling *especially* generous, it would also be great to add a test here to show it working, but: if not, I can go back and add some here

martinpotter:
> wow; looks like we missed a critical test here

As far as I can tell, this really is only an issue for things like profiling (OpenTelemetry, NR, etc) that log the Command since `ScriptEvalMessage` overrides `WriteImpl` and [correctly writes] (https://github.com/StackExchange/StackExchange.Redis/blob/ddf9341d2fc580215118789e2f8b2603ad1f9e3c/src/StackExchange.Redis/RedisDatabase.cs#L3863-L3867) `RedisCommand.EVALSHA` to the connection.

mgravell:
@martinpotter that probably makes it even more useful to have a test, honestly :)

NickCraver:
Missed the force push here, thanks for the test update! Kicking off builds :)

NickCraver (approved):
Looking good, thanks for the catch and the fix!

PR #1931, opened 2021-12-14
Tiebreakers: move into the handshake

NickCraver:
Currently the way we handshake is to get everything configured, wait for a tracer to complete, and then issue the tiebreakers to all servers if they are in play. This complicates a few things with respect to timings, duplication, and write paths being a one-off for tie breakers, which I tripped on hard in #1912.

In this, we instead move the tie breaker fetch as part of AutoConfigure as a fire-and-forget-process-the-result-later setup with a dedicated processor. This all happens before the tracer fires moving us to the next connection phase (added comments) so we should be safe. It should reduce both complexity and overall connection times proportional to endpoint latency (since we wait for completion right now).

What needs adding here is tests with us disabling commands like INFO, GET, etc. and ensuring things still behave as we want. In the overall, the tie breaker is slightly less isolated but _should_ be happening in the same order and with the same exception if any - no net result change is intended there with respect to how we do or don't error along the way. But we never want a connection to fail _because of a tiebreaker_ and I think that warrants a few tests:

- [ ] Disable `INFO` and see if we can connect
- [ ] Disable `GET` and see if we can connect
- [ ] Store some invalid TieBreaker and see if we can connect (e.g. make it a hash instead of a string)
...and maybe others?

### DIFF
--- a/src/StackExchange.Redis/ConnectionMultiplexer.cs
+++ b/src/StackExchange.Redis/ConnectionMultiplexer.cs
@@ -1719,20 +1719,16 @@ namespace StackExchange.Redis
                     }
                     int standaloneCount = 0, clusterCount = 0, sentinelCount = 0;
                     var endpoints = RawConfig.EndPoints;
-                    log?.WriteLine($"{endpoints.Count} unique nodes specified");
+                    bool useTieBreakers = !string.IsNullOrWhiteSpace(RawConfig.TieBreaker);
+                    log?.WriteLine($"{endpoints.Count} unique nodes specified ({(useTieBreakers ? "with" : "without")} tiebreaker)");
 
                     if (endpoints.Count == 0)
                     {
                         throw new InvalidOperationException("No nodes to consider");
                     }
-#pragma warning disable CS0618
-                    const CommandFlags flags = CommandFlags.NoRedirect | CommandFlags.HighPriority;
-#pragma warning restore CS0618
                     List<ServerEndPoint> masters = new List<ServerEndPoint>(endpoints.Count);
-                    bool useTieBreakers = !string.IsNullOrWhiteSpace(RawConfig.TieBreaker);
 
                     ServerEndPoint[] servers = null;
-                    Task<string>[] tieBreakers = null;
                     bool encounteredConnectedClusterServer = false;
                     Stopwatch watch = null;
 
@@ -1748,7 +1744,6 @@ namespace StackExchange.Redis
                         if (endpoints == null) break;
 
                         var available = new Task<string>[endpoints.Count];
-                        tieBreakers = useTieBreakers ? new Task<string>[endpoints.Count] : null;
                         servers = new ServerEndPoint[available.Length];
 
                         RedisKey tieBreakerKey = useTieBreakers ? (RedisKey)RawConfig.TieBreaker : default(RedisKey);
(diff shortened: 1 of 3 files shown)


src/StackExchange.Redis/ResultProcessor.cs
```diff
@@ -2068,6 +2069,34 @@ protected override bool SetResultCore(PhysicalConnection connection, Message mes
...
+            protected override bool SetResultCore(PhysicalConnection connection, Message message, in RawResult result)
+            {
+                switch (result.Type)
+                {
+                    case ResultType.SimpleString:
+                    case ResultType.BulkString:
+                        var tieBreaker = result.GetString();
+                        SetResult(message, tieBreaker);
+
+                        var bridge = connection.BridgeCouldBeNull;
```
mgravell, line 2083: remove var and just use `connection.BridgeCouldBeNull?.ServerEndPoint` ?

src/StackExchange.Redis/ResultProcessor.cs
```diff
@@ -2068,6 +2069,34 @@ protected override bool SetResultCore(PhysicalConnection connection, Message mes
...
+                        try
+                        {
+                            var endpoint = bridge?.ServerEndPoint;
+                            if (endpoint != null)
```
mgravell, line 2087: nit: `is not null`

src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -2050,44 +2029,26 @@ private void ResetAllNonConnected()
...
-                log?.WriteLine("Waiting for tiebreakers...");
-                await WaitAllIgnoreErrorsAsync("tiebreaker", tieBreakers, Math.Max(timeoutMs, 200), log).ForAwait();
-                for (int i = 0; i < tieBreakers.Length; i++)
+                for (int i = 0; i < servers.Length; i++)
                 {
-                    var ep = servers[i].EndPoint;
-                    var status = tieBreakers[i].Status;
-                    switch (status)
+                    var server = servers[i];
+                    string serverResult = server.TieBreakerResult;
```
mgravell, line 2041: should we nuke this here? or not worth the additional race conditions that would introduce for overlapped/staggered connects; meh, probably not worth the complexity risk - just wanted to share a thought

mgravell (approved):
some comments, but: 👍

src/StackExchange.Redis/ConnectionMultiplexer.cs
NickCraver: hmmm, I like the thinking - seems like we'd want to change where it's stored so it's transient in a connection state object or some such and not stored on the server but as a lookup that has the lifetime of the connect handshake perhaps?

src/StackExchange.Redis/ResultProcessor.cs
NickCraver: If we're getting cheeky, we're going all the way:
```cs
if (connection.BridgeCouldBeNull?.ServerEndPoint is ServerEndPoint endpoint)
```

mgravell (approved):

NickCraver:
Follow-up: going to look at moving this into a "handshake state" object we pass around to AutoConfigure and into the result processor, but that's a bigger refactor so merging here then revisiting.

NickCraver:
@mgravell Tweaked code here a bit (to not issue tie breakers when GET is disabled) and added tests, ready for eyes when back in action!

mgravell (approved):

PR #1936, opened 2022-01-03
Update streams readme for consistency

ansario:
There is inconsistency between the usage of `events_stream` and `event_stream` in the documentation for Getting Started with Streams. If a user copies the sample code from the first part of the page to add to the stream and then the consumer group related code from the latter part of the page to consume, they may not notice the difference in the name of the stream and wonder why they are not receiving messages.


NickCraver (approved):
👍 Looking good - thank you!

PR #1939, opened 2022-01-03
v2.5 work: .NET 6.0: add build defaulting to the thread pool

NickCraver:
This adds a `net6.0` build (and test run) that uses the default thread pool for pipe scheduling rather than our scheduler. Namely due to 2 major changes since this was introduced:
1. Thread theft has long since been resolved (never an issue in .NET Core+)
2. Sync-over-async is an app problem, but should be much better in .NET 6.0+ which has changes specifically for this in thread pool growth (see https://devblogs.microsoft.com/dotnet/performance-improvements-in-net-6/#threading)

Due to this combo of changes, we want to see if this behaves much better in the next alpha. The advantages are we move from 10 threads by default shared between connections to no dedicated threads for the scheduler. It has the following benefits:
1. While these threads were mostly idle (waiting for data), people saw them in profiler traces and attribute them as working/CPU-consuming though that's not what a profiler is really saying.
2. The default of 10 threads is a best guess, but the world varies widely. Some users are deploying many connections on 20-100 core VMs and it's a bad default limiting them (by default - not if configured). On the other end of the spectrum, a _lot_ of people run small 1-4 core VMs or containers and the default size is bigger than needed.

Instead of the above tradeoffs meant to unblock users, want to simplify, let the main thread pool scale, and hope the default is a net win for most or all users. If a consumer application has a _ton_ of sync-over-async (e.g. 100,000 tasks queued suddenly), this may be a net negative change for them - that's why we want to alpha and see how this behaves with workloads we may not have predicted.

### DIFF
--- a/src/StackExchange.Redis/ConnectionMultiplexer.ReaderWriter.cs
+++ b/src/StackExchange.Redis/ConnectionMultiplexer.ReaderWriter.cs
@@ -1,4 +1,6 @@
-﻿namespace StackExchange.Redis
+﻿using System;
+
+namespace StackExchange.Redis
 {
     public partial class ConnectionMultiplexer
     {
@@ -6,7 +8,7 @@
 
         partial void OnCreateReaderWriter(ConfigurationOptions configuration)
         {
-            SocketManager = configuration.SocketManager ?? SocketManager.Shared;
+            SocketManager = configuration.SocketManager ?? GetDefaultSocketManager();
         }
 
         partial void OnCloseReaderWriter()
@@ -14,5 +16,23 @@
             SocketManager = null;
         }
         partial void OnWriterCreated();
+
+        private static readonly Version SharedThreadpoolMinVersion = new(6, 0, 0);
+
+        /// <summary>
+        /// .NET 6.0+ has changes to sync-over-async stalls in the .NET primary thread pool
+        /// If we're in that environment, by default remove the overhead of our own threadpool
+        /// This will eliminate some context-switching overhead and better-size threads on both large
+        /// and small environments, from 16 core machines to single core VMs where the default 10 threads
+        /// isn't an ideal situation.
+        /// </summary>
+        internal static SocketManager GetDefaultSocketManager()
+        {
+#if NET6_0_OR_GREATER
+            return SocketManager.ThreadPool;
+#else
+            return SocketManager.Shared;
+#endif
+        }
     }
 }
--- a/src/StackExchange.Redis/ConnectionMultiplexer.ReaderWriter.cs
+++ b/src/StackExchange.Redis/ConnectionMultiplexer.ReaderWriter.cs
@@ -17,8 +17,6 @@ namespace StackExchange.Redis
         }
         partial void OnWriterCreated();
 
-        private static readonly Version SharedThreadpoolMinVersion = new(6, 0, 0);
-
         /// <summary>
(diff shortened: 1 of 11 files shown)


tests/StackExchange.Redis.Tests/PubSub.cs
```diff
@@ -643,7 +643,7 @@ public async Task Issue38()
             }
         }
 
-        internal static Task AllowReasonableTimeToPublishAndProcess() => Task.Delay(100);
+        internal static Task AllowReasonableTimeToPublishAndProcess() => Task.Delay(500);
```
NickCraver, line 646: Note: this isn't related, but is an intermittent test failure (unrelated to this PR at all) under load due to the minimal wait for publish observations vs. over large tests. Sneaking a fix in here.

tests/StackExchange.Redis.Tests/TestBase.cs
```diff
@@ -361,14 +361,7 @@ protected IServer GetAnyMaster(IConnectionMultiplexer muxer)
...
         public static string Me([CallerFilePath] string filePath = null, [CallerMemberName] string caller = null) =>
-#if NET472
-            "net472-"
-#elif NETCOREAPP3_1
-            "netcoreapp3.1-"
-#else
-            "unknown-"
-#endif
-         + Path.GetFileNameWithoutExtension(filePath) + "-" + caller;
+            Environment.Version.ToString() + Path.GetFileNameWithoutExtension(filePath) + "-" + caller;
```
NickCraver, line 364: This isn't a 1:1 replacement, but the goal is to simple make it unique per version - let's go simpler and never update this with TFMs again.

version.json
```diff
@@ -1,5 +1,5 @@
 {
-  "version": "2.2",
+  "version": "2.5-prerelease",
```
NickCraver, line 2: Note: this change also exists in #1912 - whichever lands first wins the v2.5-prelease versioning kick.

src/StackExchange.Redis/ConnectionMultiplexer.ReaderWriter.cs
```diff
@@ -1,18 +1,38 @@
...
+            SocketManager = configuration.SocketManager ?? GetDefaultSocketManager();
         }
 
         partial void OnCloseReaderWriter()
         {
             SocketManager = null;
         }
         partial void OnWriterCreated();
+
+        private static readonly Version SharedThreadpoolMinVersion = new(6, 0, 0);
```
mgravell, line 20: never used?

mgravell (approved):

NickCraver:
@mgravell ahahaha, `TestPublishWithSubscribers` - yeah it's the same story here. I don't think we need to block this PR on it but looks like ultimately the same subscriber cause we're already on in the (sometimes) failing test on Ubuntu.

vchirikov:
I faced with huge thread creation (`Microsoft-Windows-DotNETRuntime/Thread/Creating`) + lock contention (`Microsoft-Windows-DotNETRuntime/Contention/Start`) inside StackExchange.Redis ( Pipelines.Sockets.Unofficial)

image

I don't have enough time to investigate this, but I think that `SocketManager.ThreadPool` could help. @NickCraver / @mgravell could you publish this new version to the nuget? (as prerelease will be ok too)

PR #1945, opened 2022-01-10
Changes for #1912

NickCraver:
Reducing diff from #1912 on bits we can simplify & merge in sooner.


NickCraver:
@mgravell this is to reduce diff and make the history on #1912 more straightforward - note the only test failing is our friend: `PubSub.SubscriptionsSurviveConnectionFailureAsync`, so ignoring that here.

mgravell (approved):

Issue #1835, opened 2021-08-17
Pub sub in cluster mode, not seeing the replicas service traffic


greyhoundgames:
I had just figured out that my main redis cluster(AWS) was not using its replicas because I had failed to use 
CommandFlags.PreferReplica during reads. After adding this, the metrics data showed the traffic split nicely between the replicas. Great!
I went to my other redis cluster I have which is a chat app and I did the same thing on it. However in this case it was a pub\sub where as the first cluster was doing string gets.

I added the PreferReplica in the subscribe. However here the traffic does not seem to have been spread out evenly. Is there a problem with replicas when using pub sub?

Code used to subscribe:
ISubscriber subscriber = RedisHelpers.ConnectionChat.GetSubscriber();
subscriber.Subscribe(user.room + ChatConstants.MESSAGES_SUB, chatCallback, CommandFlags.PreferReplica);

SDK version: Nuget 2.2.62
AWS redis Engine Version Compatibility: Version compatibility of the engine that will be run on your nodes 6.0.5
Cluster config 1 shard, 3 nodes, 2 of which are replicas.

NickCraver:
Indeed there is! We found this later (recently) following up on issues. Specifically there's a bug in sharding around a `default(RedisKey)` path that @mgravell is looking into as part of #1912. Once a fix goes in that PR (or is broken into another), we'll update here.

NickCraver:
This is drastically improved in #1947 and will be in the 2.5 release - thank you for the detail! It'll still prefer the first replica but this should be an improvement over current. This preference is kept in mind across reconnects now as well.

Issue #1830, opened 2021-08-15
Default keep alive behavior only works for the Interactive ConnectionType


adraut:
The scenario that was discovered is as follows:
Redis 5.0.2 with the timeout at the default of zero, the tcp-keepalive at the default of 300 and maxclients at the default of 10000.
StackExchange.Redis 2.2.4 with abortConnect=true (the default) and keepAlive=-1 (the default).

Once connected the [`ServerEndPoint` calls `AutoConfigureAsync`, but only for the `Interactive` `ConnectionType`](https://github.com/StackExchange/StackExchange.Redis/blob/56ac9a21174684f30b8aa333f8069fdca77ed4e7/src/StackExchange.Redis/ServerEndPoint.cs#L872) so only the `PhysicalBridge` for the `Interactive` `ConnectionType` will have it's `WriteEverySeconds` value set to a value greater than zero.

The problem with setup is that if something severs the subscription connection and the Redis server still thinks the client's subscription connection is connected it will never be disconnected/closed on the Redis server and it will eventually hit the maxclients value and stop accepting new connections.

I do not think this is a bug with the client as this is easy to fix by configuring everything properly.  At minimum the Redis server should have a timeout value greater than zero.  Even better is to set the `KeepAlive` in the `ConfigurationOptions` to something greater than zero.

What would be nice is if this behavior of the `ConnectionMultiplexer` was documented because it is not accurate enough in my opinion.  Currently it says:
>Time (seconds) at which to send a message to help keep sockets alive (60 sec default)

I think it should state that the default only applies to the `Interactive` `ConnectionType` and that the value must be greater than zero for it to run the keep alive process for `Subscription` `ConnectionType`s.

mgravell:
Sounds like a big to me; the default should still apply to the subscription
connection.

NickCraver:
There was another issue we found here with the subscriptions in which the keepalive was being sent over the wrong connection on Redis server 3.0+ due to the `PING` command and bridge selection paths. @mgravell is taking a look at this as part of #1912 where we came across the issue in debugging. Worth checking into this as part of that work (happy to help - tagging so we see it).

NickCraver:
This is fixed in #1958 and will land in the 2.5 release, thank you for the report!

adraut:
Awesome!  Thanks guys.

Issue #1586, opened 2020-10-15
Subscription unsubscribes unexpectedly


winseros:
Greetings.

We face a situation when sometimes subscriptions get unsubscribed on their own will.

The application code looks simple:

```
_redis = ConnectionMultiplexer.Connect(_redisConnectionString);
_subscriber = _redis.GetSubscriber();
_subscriber.Subscribe("inbox-channel").OnMessage(OnMessage);
```

The `ConnectionMultiplexer` log has nothing suspicious:

```
Connecting 172.22.0.2:6379/Interactive...
BeginConnect: 172.22.0.2:6379
Connected Interactive/172.22.0.2:6379
Server handshake
Authenticating (user/password)
Setting client name: SvcInboxApp
Auto-configure...
Sending critical tracer: Interactive/172.22.0.2:6379
Writing to Interactive/172.22.0.2:6379: ECHO
Flushing outbound buffer
Starting read
Response from Interactive/172.22.0.2:6379 / ECHO: BulkString: 16 bytes
1 unique nodes specified
Writing to Interactive/172.22.0.2:6379: PING
Requesting tie-break from 172.22.0.2:6379 > __Booksleeve_TieBreak...
Writing to Interactive/172.22.0.2:6379: GET __Booksleeve_TieBreak
Allowing endpoints 00:00:40 to respond...
Awaiting task completion, IOCP: (Busy=0,Free=1000,Min=2,Max=1000), WORKER: (Busy=3,Free=32764,Min=2,Max=32767)
Response from Interactive/172.22.0.2:6379 / PING: SimpleString: PONG
Response from Interactive/172.22.0.2:6379 / GET __Booksleeve_TieBreak: (null)
All tasks completed cleanly, IOCP: (Busy=0,Free=1000,Min=2,Max=1000), WORKER: (Busy=3,Free=32764,Min=2,Max=32767)
172.22.0.2:6379 returned with success
Waiting for tiebreakers...
All tasks are already complete
172.22.0.2:6379 had no tiebreaker set
Single master detected: 172.22.0.2:6379
172.22.0.2:6379: Standalone v2.0.0, master; keep-alive: 00:01:00; int: ConnectedEstablished; sub: ConnectedEstablishing
172.22.0.2:6379: int ops=11, qu=0, qs=0, qc=0, wr=0, socks=1; sub ops=4, qu=0, qs=4, qc=0, wr=0, socks=1
[...]

Issue #1110, opened 2019-04-04
Subscriber unsubscribes after subscribing on reconnect


shellicar:
## Issue
Subscriber very occasionally unsubscribes after subscribing after the subscriber connection is restored from a hard disconnect (port being closed).
This seems to be related to issue: #273
But I can still reproduce this.

## Version
Reproduced with the following versions.
1.2.6
2.0.519
2.0.588

## Behaviour
using redis-monitor on the server, the following is observed.
```
1554338345.603361 [0 127.0.0.1:10441] "SUBSCRIBE" "__Booksleeve_MasterChanged"
1554338345.620982 [0 127.0.0.1:10441] "SUBSCRIBE" "Channel1"
1554338345.621008 [0 127.0.0.1:10441] "SUBSCRIBE" "Channel2"
1554338345.629399 [0 127.0.0.1:10441] "UNSUBSCRIBE" "Channel1"
1554338345.629421 [0 127.0.0.1:10441] "UNSUBSCRIBE" "Channel2"
```

## Expected/Normal Behaviour
```
1554338954.604720 [0 127.0.0.1:11370] "SUBSCRIBE" "__Booksleeve_MasterChanged"
1554338954.609184 [0 127.0.0.1:11370] "UNSUBSCRIBE" "Channel1"
1554338954.609228 [0 127.0.0.1:11370] "UNSUBSCRIBE" "Channel2"
1554338954.609439 [0 127.0.0.1:11370] "SUBSCRIBE" "Channel1"
1554338954.609456 [0 127.0.0.1:11370] "SUBSCRIBE" "Channel2"
```
OR
```
1554338956.418663 [0 127.0.0.1:11372] "SUBSCRIBE" "__Booksleeve_MasterChanged"
1554338956.419116 [0 127.0.0.1:11372] "SUBSCRIBE" "Channel1"
1554338956.419133 [0 127.0.0.1:11372] "SUBSCRIBE" "Channel2"
```

shellicar:
Any information on this?
I should add that the internal subscribers dictionary still thinks its subscribed.

This makes the library unusable for us as the subscriptions are unreliable.

mgravell:
I haven't had chance to look (all support here comes from our personal
time, etc), but I agree it looks incorrect. My suspicion is that it
*thinks* it is unsubscribing from the old primary - most likely what I need
to do here is track the endpoint currently subscribed, and make sure we
include that in all routing decisions.

[...]

PR #1947, opened 2022-01-10
Pub/Sub fixes for subscribe/re-subscribe

NickCraver:
We're working on pub/sub - breaking it out explicitly from #1912. This relates to several issues and in general handling resubscriptions on reconnect.

Issues: #1110, #1586, #1830 #1835

There are a few things in play we're investigating:
- [x] Subscription heartbeat not going over the subscription connection (due to `PING` and `GetBridge`)
- [x] Subscriptions not reconnecting at all (or potentially doing to and unsubscribing according to some issues)
- [x] Subscriptions always going to a single cluster node (due to `default(RedisKey)`)

Overall this set of changes:
- Completely restructures how RedisSubscriber works
  - No more `PendingSubscriptionState` (`Subscription` has the needed bits to reconnect)
  - Cleaner method topology (in `RedisSubscriber`, rather than `Subscriber`, `RedisSubscriber`, and `ConnectionMultiplexer`)
    - By placing these on `RedisSubscriber`, we can cleanly use `ExecuteSync/Async` bits, get proper profiling, etc.
  - Proper sync/async split (rather than `Wait()` in sync paths)
- Changes how subscriptions work
  - The `Subscription` object is added to the `ConnectionMultiplexer` tracking immediately, but the command itself actually goes to the server and back (unless FireAndForget) before returning for proper ordering like other commands.
  - No more `Task.Run()` loop - we now ensure reconnects as part of the handshake
  - Subscriptions are marked as not having a server the moment a disconnect is fired
    - Question: Should we have a throttle around this for massive numbers of connections, or async it?
- Changes how connecting works
  - The connection completion handler will now fire when the _second_ bridge/connection completes, this means we won't have `interactive` connected but `subscription` in an unknown state - both are connected before we fire the handler meaning the moment we come back from connect, subscriptions are in business.
- Moves to a `ConcurrentDictionary` since we only need limited locking around this and we only have it once per multiplexer.
  - TODO: This needs eyes, we could shift it - implementation changed along the way where this isn't a critical detail
- Fixes the `TrackSubscriptionsProcessor` - this was never setting the result but didn't notice in 8 years because downstream code never cared.
  - Note: each `Subscription` has a processor instance (with minimal state) because when the subscription command comes back _then_ we need to decide if it successfully registered (if it didn't, we need to maintain it has no successful server)
- `ConnectionMultiplexer` grew a `DefaultSubscriber` for running some commands without lots of method duplication, e.g. ensuring servers are connected.
- Overrides `GetHashSlot` on `CommandChannelBase` with the new `RedisChannel`-based methods so that operates correctly

Not directly related changes which helped here:
- Better profiler helpers for tests and profiler logging in them
- Re-enables a few `PubSub` tests that were unreliable before...but correctly so.

TODO: I'd like to add a few more test scenarios here:
- [x] Simple Subscribe/Publish/await Until/check pattern to ensure back-to-back subscribe/publish works well
- [x] Cluster connection failure and subscriptions moving to another node

To consider:
- [x] Subscription await loop from EnsureSubscriptionsAsync and connection impact on large reconnect situations
   - In a reconnect case, this is background and only the nodes affected have any latency...but still.
- [ ] TODOs in code around variadic commands, e.g. re-subscribing with far fewer commands by using `SUBSCRIBE <key1> <key2>...`
   - In cluster, we'd have to batch per slot...or just go for the first available node
   - ...but if we go for the first available node, the semantics of `IsConnected` are slightly off in the not connected (`CurrentServer is null`) case, because we'd say we're connected to where it _would_ go even though that'd be non-deterministic without hashslot batching. I think this is really minor and shouldn't affect our decision.
- [x] `ConcurrentDictionary` vs. returning to locks around a `Dictionary`
   - ...but if we have to lock on firing consumption of handlers anyway, concurrency overhead is probably a wash.

### DIFF
--- a/src/StackExchange.Redis/RedisSubscriber.cs
+++ b/src/StackExchange.Redis/RedisSubscriber.cs
@@ -13,6 +13,14 @@ namespace StackExchange.Redis
     {
         private readonly Dictionary<RedisChannel, Subscription> subscriptions = new Dictionary<RedisChannel, Subscription>();
 
+        internal int GetSubscriptionsCount()
+        {
+            lock (subscriptions)
+            {
+                return subscriptions.Count;
+            }
+        }
+
         internal static void CompleteAsWorker(ICompletable completable)
         {
             if (completable != null) ThreadPool.QueueUserWorkItem(s_CompleteAsWorker, completable);
@@ -166,11 +174,10 @@ namespace StackExchange.Redis
 
         internal bool SubscriberConnected(in RedisChannel channel = default(RedisChannel))
         {
-            var server = GetSubscribedServer(channel);
-            if (server != null) return server.IsConnected;
+            // TODO: default(RedisKey) is incorrect here - should shard based on the channel in cluster
+            var server = GetSubscribedServer(channel) ?? SelectServer(RedisCommand.SUBSCRIBE, CommandFlags.DemandMaster, default(RedisKey));
 
-            server = SelectServer(RedisCommand.SUBSCRIBE, CommandFlags.DemandMaster, default(RedisKey));
-            return server?.IsConnected == true;
+            return server?.IsConnected == true && server.IsSubscriberConnected;
         }
 
         internal long ValidateSubscriptions()
@@ -221,6 +228,7 @@ namespace StackExchange.Redis
             [System.Diagnostics.CodeAnalysis.SuppressMessage("Usage", "RCS1210:Return completed task instead of returning null.", Justification = "Intentional for efficient success check")]
             public Task SubscribeToServer(ConnectionMultiplexer multiplexer, in RedisChannel channel, CommandFlags flags, object asyncState, bool internalCall)
             {
+                // TODO: default(RedisKey) is incorrect here - should shard based on the channel in cluster
                 var selected = multiplexer.SelectServer(RedisCommand.SUBSCRIBE, flags, default(RedisKey));
                 var bridge = selected?.GetBridge(ConnectionType.Subscription, true);
                 if (bridge == null) return null;
@@ -299,14 +307,13 @@ namespace StackExchange.Redis
 
             internal void Resubscribe(in RedisChannel channel, ServerEndPoint server)
             {
-                if (server != null && Interlocked.CompareExchange(ref owner, server, server) == server)
+                // Only re-subscribe to the original server
+                if (server != null && GetOwner() == server)
                 {
                     var cmd = channel.IsPatternBased ? RedisCommand.PSUBSCRIBE : RedisCommand.SUBSCRIBE;
                     var msg = Message.Create(-1, CommandFlags.FireAndForget, cmd, channel);
                     msg.SetInternalCall();
-#pragma warning disable CS0618
-                    server.WriteDirectFireAndForgetSync(msg, ResultProcessor.TrackSubscriptions);
-#pragma warning restore CS0618
+                    server.Multiplexer.ExecuteSyncImpl(msg, ResultProcessor.TrackSubscriptions, server);
                 }
             }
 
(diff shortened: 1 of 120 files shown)


NickCraver:
This is getting there - refactor to make it a lot simpler (I hope) is worth it - e.g. no Subscriber -> Multiplexer -> Subscription path for a single op, the method structure is simpler now (we may be able to de-dupe more).

The current test race problems are mainly around not having a connection to send the command over. While the backlog policy would actually resolve this situation by queueing and sending on the subscriber bridge, the fundamental issue today is we do not wait on subscriber connections to complete when establishing a connection - everything is based on interactive connections.

We could slow down ReconfigureAsync to awaiting wait on subscribers as well, or find another approach to waiting to send the first subscriptions until that connection is ready.

NickCraver:
Added tests for Primary/Replica and expected behavior of the various Demand/Prefer `CommandFlags` - looking good!

src/StackExchange.Redis/RedisSubscriber.cs
```diff
@@ -35,273 +55,178 @@ internal bool GetSubscriberCounts(in RedisChannel channel, out int handlers, out
...
-                    {
-                        remove = sub.Remove(handler, queue);
-                    }
-                    if (remove)
-                    {
-                        subscriptions.Remove(channel);
-                        task = sub.UnsubscribeFromServer(channel, flags, asyncState, false);
-                    }
-                }
+                DefaultSubscriber.EnsureSubscribedToServer(pair.Value, pair.Key, flags, true);
```
NickCraver, line 115: This fires `n` times for `n` subscriptions, but is called with `FireAndForget` so is not O(n) blocking at least.

src/StackExchange.Redis/RedisSubscriber.cs
```diff
@@ -35,273 +55,178 @@ internal bool GetSubscriberCounts(in RedisChannel channel, out int handlers, out
...
         {
-            if (server == null) return;
-            lock (subscriptions)
+            // TODO: Evaluate performance here, this isn't good for a large number of subscriptions.
+            // It's probable we want to fire and forget `n` here, recording how many are going to try to reconnect?
+            long count = 0;
+            foreach (var pair in subscriptions)
             {
-                foreach (var pair in subscriptions)
+                if (await DefaultSubscriber.EnsureSubscribedToServerAsync(pair.Value, pair.Key, flags, true))
```
NickCraver, line 129: This is currently *not* called as `FireAndForget` upstream in the connection pipe and is a `O(n)` issue. We can do `FireAndForget` all the way down, but there's a subtle issue of a default task state for `<bool>` being false and the count would always be zero - would need to change implementation somehow to get a real count out of it for the connection log trace.

andre-ss6:
Hey, maybe it's better to open a issue about this, but seeing that the PR still open, perhaps you can take a look at this before merging? 😅

It seems that re-subscription doesn't work specifically when calling `GetSubscriber` again after a disconnect _and_ when running in cluster mode. 

I've written this rudimentary test, and I've ran it with a build of your branch:

```cs
public static async Task Main(string[] args)
        {
            string connectionString = null;
            if (args.Length == 0 || string.IsNullOrWhiteSpace(args[0]))
            {
                connectionString = "localhost:56379,abortConnect=false,connectTimeout=15000,connectRetry=3";
            }

            var redis = ConnectionMultiplexer.Connect(connectionString);
            var subscriber = redis.GetSubscriber();

            var tcs = new TaskCompletionSource<bool>();
            var channel = new RedisChannel("channelTest", RedisChannel.PatternMode.Auto);
            Action<RedisChannel, RedisValue> handler = (_, v) => Debug.WriteLine(v);
            await subscriber.SubscribeAsync(channel, handler);

            // Stop the redis service, for example, or use a proxy
            Console.WriteLine("Please jam the connection and then press any key...");
            Console.ReadKey();

            Exception exception = null;
            try
            {
                await redis.GetDatabase().StringGetAsync("foo");
            }
            catch (RedisConnectionException ex)
            {
                exception = ex;
            }

            if (exception is null)
            {
                throw new Exception("Redis was not offline.");
            }

            // Start it back
            Console.WriteLine("Please turn Redis back up and then press any key...");
            Console.ReadKey();

            try
            {
                await redis.GetDatabase().StringGetAsync("foo");
            }
            catch (RedisConnectionException)
            {
                Console.WriteLine("Redis seems not to be back online."); throw;
            }

            var newSubscriber = redis.GetSubscriber();
            var result = await newSubscriber.PublishAsync(channel, "messageTest");

            if (result != 1)
            {
                throw new Exception($"Expected 1 client to receive the message. Actual number of clients: {result}");
            }

            Console.WriteLine("Test passed succesfully");
        }
```

This will fail as-is (given that your redis is in cluster mode), but will pass if you change 

```cs
await newSubscriber.PublishAsync(channel, "messageTest");
```

to

```cs
await subscriber.PublishAsync(channel, "messageTest");
```

It makes sense that the old subscriber instance keeps working, but that seems rather pointless; The reconnection/resubscription process should be transparent. If the caller is required to know whether or not a disconnect has happened since it last called `GetSubscriber`, then there is no point in handling re-subscription.

In my company we had to written a lot of code ourselves to handle re-connects and re-subscriptions in past versions of the library; We were hoping that we could stop using that code after upgrading to 2.0, but we found this edge case which is blocking us.

Do you think you can take a look?

NickCraver:
@andre-ss6 Of course! I think there may be a confusing aspect of PublishAsync here though, in Redis that's how many subscribers were _on that server_, not _on that cluster_. So for example you can get a 0 back, but still get the subscription from the other node. It's going from Publisher- > Node A -> Node B -> Subscriber.  This tripped me up in tests too, and is probably worth of a remark on the XML docs for this.

NickCraver:
@andre-ss6 Check out the `ClusterNodeSubscriptionFailover` test in the changeset here to get an example of what I mean - in that we're checking _that we got the message_, rather than the publisher count which on a cluster turns out is...kinda meaningless :(

andre-ss6:
@NickCraver  Oh I see! Thanks a lot for the clarification! I was going to actually test against the callback, instead of the return of `PublishAsync`, but was too lazy to do it 😅. 

I will re-do the tests later then. Thanks again!

NickCraver:
@andre-ss6 Thanks for the poke here, I do appreciate the eyes :) I'm changing all the `<returns>` on these methods now to better indicate exactly what we both hit:
```cs
        /// <returns>
        /// The number of clients that received the message *on the destination server*,
        /// note that this doesn't mean much in a cluster as clients can get the message through other nodes.
        /// </returns>
```

NickCraver:
See [`169a173` (#1947)](https://github.com/StackExchange/StackExchange.Redis/pull/1947/commits/169a173e9371c265b27f50ac9a8979cb4ea8f80e) for what I'm thinking around large numbers of subs - moving ReconfigureAsync to FireAndForget for sub re-establish.

andre-ss6:
Hey @NickCraver, thanks a lot for the help and for your time! I re-did the tests and they worked! However, not how I thought they would 😅, as they also worked on an old version of the library (1.2.6). Since it worked, however, I guess this is now a bit off-topic for the PR, so I sent you a message on gitter, if you have time to clarify some questions. But in any way, you already helped a lot. Thanks again!

tests/RedisConfigs/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM redis:5
+FROM redis:6.2.6
```
mgravell, line 1: +1; this gives us a few other unrelated things we can use

mgravell (approved):
this looks great and very well considered; unable to find anything other than pettiness, so: nice job sir

NickCraver:
@mgravell If there's little things to tweak as well, happy to! (same for all the PRs), think we're about in a good state so more than happy to polish anywhere

PR #1958, opened 2022-01-20
Pub/Sub: default to 3.0, fix PING, fix server selection in cluster, and cleanup

NickCraver:
In prep for changes to how we handle subscriptions internally, this does several things:
- Upgrades default Redis server assumption to 3.x
- Routes PING on Subscription keepalives over the subscription bridge appropriately
- Fixes cluster sharding from default(RedisKey) to shared logic for RedisChannel as well (both in byte[] form)
- General code cleanup in the area (getting a lot of DEBUG/VERBOSE noise into isolated files)

### DIFF
--- a/src/StackExchange.Redis/ConfigurationOptions.cs
+++ b/src/StackExchange.Redis/ConfigurationOptions.cs
@@ -306,7 +306,7 @@ namespace StackExchange.Redis
         /// </summary>
         public Version DefaultVersion
         {
-            get => defaultVersion ?? (IsAzureEndpoint() ? RedisFeatures.v4_0_0 : RedisFeatures.v2_8_0);
+            get => defaultVersion ?? (IsAzureEndpoint() ? RedisFeatures.v4_0_0 : RedisFeatures.v3_0_0);
             set => defaultVersion = value;
         }
 
--- a/src/StackExchange.Redis/ConnectionMultiplexer.Verbose.cs
+++ b/src/StackExchange.Redis/ConnectionMultiplexer.Verbose.cs
@@ -1,0 +1,59 @@
+﻿using System;
+using System.Diagnostics;
+using System.Net;
+using System.Runtime.CompilerServices;
+
+namespace StackExchange.Redis
+{
+    public partial class ConnectionMultiplexer
+    {
+        internal event Action<string, Exception, string> MessageFaulted;
+        internal event Action<bool> Closing;
+        internal event Action<string> PreTransactionExec, TransactionLog, InfoMessage;
+        internal event Action<EndPoint, ConnectionType> Connecting;
+        internal event Action<EndPoint, ConnectionType> Resurrecting;
+
+        partial void OnTrace(string message, string category);
+        static partial void OnTraceWithoutContext(string message, string category);
+
+        [Conditional("VERBOSE")]
+        internal void Trace(string message, [CallerMemberName] string category = null) => OnTrace(message, category);
+
+        [Conditional("VERBOSE")]
+        internal void Trace(bool condition, string message, [CallerMemberName] string category = null)
+        {
+            if (condition) OnTrace(message, category);
+        }
+
+        [Conditional("VERBOSE")]
+        internal static void TraceWithoutContext(string message, [CallerMemberName] string category = null) => OnTraceWithoutContext(message, category);
+
+        [Conditional("VERBOSE")]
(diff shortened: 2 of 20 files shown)


src/StackExchange.Redis/ConfigurationOptions.cs
```diff
@@ -306,7 +306,7 @@ public int ConnectTimeout
         /// </summary>
         public Version DefaultVersion
         {
-            get => defaultVersion ?? (IsAzureEndpoint() ? RedisFeatures.v4_0_0 : RedisFeatures.v2_8_0);
+            get => defaultVersion ?? (IsAzureEndpoint() ? RedisFeatures.v4_0_0 : RedisFeatures.v3_0_0);
```
NickCraver, line 309: So that we go with PING off the bat

src/StackExchange.Redis/ConnectionMultiplexer.Verbose.cs
```diff
@@ -0,0 +1,59 @@
+﻿using System;
```
NickCraver, line 1: These are method moves - no functional changes

src/StackExchange.Redis/ExceptionFactory.cs
```diff
@@ -312,7 +312,7 @@ ServerEndPoint server
             // Add server data, if we have it
             if (server != null && message != null)
             {
-                var bs = server.GetBridgeStatus(message.Command);
+                var bs = server.GetBridgeStatus(message.IsForSubscriptionBridge ? ConnectionType.Subscription: ConnectionType.Interactive);
```
NickCraver, line 315: Turns out we were also getting the wrong error stats, for the command we sent to the wrong place - fixing that.

NickCraver:
Note: pub/sub failing tests are _because subscriptions don't do awesome things_, and we're not masking that - the tests are correct, overall implementation and our `Task.Run` race on the subscription loop is bad (to be fixed after this prep).

NickCraver:
@mgravell ^ this is the cheese move (tests appropriately failing)

src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -2174,16 +2147,14 @@ internal void UpdateClusterRange(ClusterConfiguration configuration)
...
 
-        internal ServerEndPoint SelectServer(RedisCommand command, CommandFlags flags, in RedisKey key)
-        {
-            return ServerSelectionStrategy.Select(command, key, flags);
-        }
+        internal ServerEndPoint SelectServer(RedisCommand command, CommandFlags flags, in RedisKey key) =>
+            ServerSelectionStrategy.Select(command, key, flags);
+
+        internal ServerEndPoint SelectServer(RedisCommand command, CommandFlags flags, in RedisChannel channel) =>
+            ServerSelectionStrategy.Select(command, channel, flags);
```
mgravell, line 2157: is this an implicit conversion? if so, we need to check whether that is allocatey

src/StackExchange.Redis/ServerSelectionStrategy.cs
```diff
@@ -58,19 +58,26 @@ public ServerSelectionStrategy(ConnectionMultiplexer multiplexer)
...
-            => ServerType == ServerType.Standalone ? NoSlot : GetClusterSlot(key);
+            => ServerType == ServerType.Standalone || key.IsNull ? NoSlot : GetClusterSlot((byte[])key);
 
-        private static unsafe int GetClusterSlot(in RedisKey key)
+        /// <summary>
+        /// Computes the hash-slot that would be used by the given channel.
+        /// </summary>
+        /// <param name="channel">The <see cref="RedisChannel"/> to determine a slot ID for.</param>
+        public int HashSlot(in RedisChannel channel)
+            => ServerType == ServerType.Standalone || channel.IsNull ? NoSlot : GetClusterSlot((byte[])channel);
```
mgravell, line 72: TODO Marc: consider this in my "don't use `byte[]` for `GetClusterSlot` rework"

mgravell (approved):
looks good; some side thoughts

src/StackExchange.Redis/ConnectionMultiplexer.cs
NickCraver: No conversion - new overload!

PR #1959, opened 2022-01-22
Pub/Sub subscription processor: actually set the result

NickCraver:
This hasn't worked in over 8 years, but due to other things upstream ignoring the result and pretending success, it didn't matter.

I imagine only @mgravell will understand why this was confusing to debug for hours or find it hilarious, but yep: that was broken.


NickCraver:
Result should also be called in the false case to not encounter an infinite wait - just folding this into #1947.

PR #1964, opened 2022-01-25
Exceptions: Add threadpool stats in .NET Core+

NickCraver:
I swear I already did this somewhere but it got lost, this adds additional thread pool stats exposed in .NET Core 3.1+. We'll get for example `ThreadPool.PendingWorkItemCount` on exception messages to better indicate app contention.

### DIFF
--- a/src/StackExchange.Redis/PerfCounterHelper.cs
+++ b/src/StackExchange.Redis/PerfCounterHelper.cs
@@ -58,7 +58,7 @@ namespace StackExchange.Redis
         {
             GetThreadPoolStats(out string iocp, out string worker, out string workItems);
             var cpu = includePerformanceCounters ? GetSystemCpuPercent() : "n/a";
-            return $"IOCP: {iocp}, WORKER: {worker}, POOL: {workItems}, Local-CPU: {cpu}";
+            return $"IOCP: {iocp}, WORKER: {worker}, POOL: {workItems ?? "n/a"}, Local-CPU: {cpu}";
         }
 
         internal static string GetSystemCpuPercent() =>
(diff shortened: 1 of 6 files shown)


src/StackExchange.Redis/PerfCounterHelper.cs
```diff
@@ -77,6 +77,13 @@ internal static int GetThreadPoolStats(out string iocp, out string worker)
 
             iocp = $"(Busy={busyIoThreads},Free={freeIoThreads},Min={minIoThreads},Max={maxIoThreads})";
             worker = $"(Busy={busyWorkerThreads},Free={freeWorkerThreads},Min={minWorkerThreads},Max={maxWorkerThreads})";
+
+#if NETCOREAPP
+            workItems = $"(Threads={ThreadPool.ThreadCount},QueuedItems={ThreadPool.PendingWorkItemCount},CompletedItems={ThreadPool.CompletedWorkItemCount})";
```
mgravell, line 82: how often is this called? should I care about the `string` alloc?

mgravell (approved):

NickCraver:
@mgravell allocated in the exception path - I don't think it's a major item but would like to go through and end-to-end builder these paths once existing PRs are in (several touch it).

src/StackExchange.Redis/PerfCounterHelper.cs
NickCraver: I'm not a big fan of how existing works - was light touch, let's get existing PRs touching it in, then I'll add a bench and do a builder pass and such on the exception path, no reason we can't be civilized with the GC!

PR #1971, opened 2022-01-31
Update LICENSE

MarnixCroes:
Update LICENSE to 2022


NickCraver:
Hey there! Fortunately, copyrights do not require an end date so this is not required. In fact, there are nothing but downsides to specifying a range: old copies of the code, yearly commit updates needed, etc. so leaving it off is intentional :)

PR #2001, opened 2022-02-19
Misc connection/subscription fixes

NickCraver:
In investigating an issue in #1989, I found a few gaps. Overall:
1. Twemproxy has an out of date CommandMap, which propagated to Envoy.
2. We were expecting both interactive and subscription connections to complete to complete the async handler...but we shouldn't because subscriptions may be disabled.
3. RedisSubscriber changes on the sync path weren't validating the message (asserting the command map has it enabled).

This fixes all of the above and adds another test considering all 3.


rkarthick (approved):
LGTM

philon-msft (approved):

PR #2003, opened 2022-02-19
Implement GET option of SET command

martinekvili:
Redis 6.2.0 has deprecated the [GETSET](https://redis.io/commands/getset) command and in turn added the `GET` option to the [SET](https://redis.io/commands/set) command.

_Note: I am not 100% happy with the `StringSetAndGet` method name. However, I am quite torn, as it would be nice to call it `StringGetSet` to better show the intent and the deprecation of the old method. On the other hand, I quite like how this driver follows the names of the underlying Redis commands quite closely. The current name tries to follow that convention._


src/StackExchange.Redis/Interfaces/IDatabase.cs
```diff
@@ -2202,6 +2202,18 @@ public interface IDatabase : IRedis, IDatabaseAsync
...
+        /// <summary>
+        /// Atomically sets key to value and returns the old value stored at key.
+        /// </summary>
+        /// <param name="key">The key of the string.</param>
+        /// <param name="value">The value to set.</param>
+        /// <param name="expiry">The expiry to set.</param>
+        /// <param name="when">Which condition to set the value under (defaults to always).</param>
+        /// <param name="flags">The flags to use for this operation.</param>
+        /// <returns>The old value stored at key, or nil when key did not exist.</returns>
+        /// <remarks>https://redis.io/commands/set</remarks>
```
NickCraver, line 2214:
```suggestion
        /// <summary>
        /// Atomically sets key to value and returns the previous value (if any) stored at <paramref name="key"/>.
        /// </summary>
        /// <param name="key">The key of the string.</param>
        /// <param name="value">The value to set.</param>
        /// <param name="expiry">The expiry to set.</param>
        /// <param name="when">Which condition to set the value under (defaults to <see cref="When.Always"/>).</param>
        /// <param name="flags">The flags to use for this operation.</param>
        /// <returns>The old value stored at <paramref name="key"/>, or nil when key did not exist.</returns>
        /// <remarks>https://redis.io/commands/getset</remarks>
```

src/StackExchange.Redis/Interfaces/IDatabaseAsync.cs
```diff
@@ -2155,6 +2155,18 @@ public interface IDatabaseAsync : IRedisAsync
...
+        /// <summary>
+        /// Atomically sets key to value and returns the old value stored at key.
+        /// </summary>
+        /// <param name="key">The key of the string.</param>
+        /// <param name="value">The value to set.</param>
+        /// <param name="expiry">The expiry to set.</param>
+        /// <param name="when">Which condition to set the value under (defaults to always).</param>
+        /// <param name="flags">The flags to use for this operation.</param>
+        /// <returns>The old value stored at key, or nil when key did not exist.</returns>
+        /// <remarks>https://redis.io/commands/set</remarks>
```
NickCraver, line 2167:
```suggestion
        /// <summary>
        /// Atomically sets key to value and returns the previous value (if any) stored at <paramref name="key"/>.
        /// </summary>
        /// <param name="key">The key of the string.</param>
        /// <param name="value">The value to set.</param>
        /// <param name="expiry">The expiry to set.</param>
        /// <param name="when">Which condition to set the value under (defaults to <see cref="When.Always"/>).</param>
        /// <param name="flags">The flags to use for this operation.</param>
        /// <returns>The old value stored at <paramref name="key"/>, or nil when key did not exist.</returns>
        /// <remarks>https://redis.io/commands/getset</remarks>
```

NickCraver (changes requested):
I see some quick doc fixes but this is looking excellent, thank you! I need to poke a bit more with fresh eyes but I think we're good - 👍 from me on current naming.

src/StackExchange.Redis/Interfaces/IDatabase.cs
martinekvili: Thank you very much for checking this out!

I did the requested changes with one exception, since I think it might be due a slight confusion. I have left the link in the remarks as <https://redis.io/commands/set>, since that is the link that points to the `SET` method which in version **6.2.0** has the `GET` option — this is what this method uses. On the other hand, <https://redis.io/commands/getset> points to the `GETSET` method which has been deprecated in **6.2.0**.

Please let me know if I misunderstood your reasons / intent, and I'll happily change the link as well!

NickCraver: Oh no you're absolutely right - I think we should probably add that distinction to the remarks though!

NickCraver: Looking good, thanks!

NickCraver (approved):
I've touched this so should get extra eyes /cc @mgravell

NickCraver:
@martinekvili thanks for this!!

PR #2009, opened 2022-02-24
#2008 Follow-ups

NickCraver:
A few tweaks to the changes #2008 for disposing and normalization.

### DIFF
--- a/src/StackExchange.Redis/PhysicalBridge.cs
+++ b/src/StackExchange.Redis/PhysicalBridge.cs
@@ -113,6 +113,7 @@ namespace StackExchange.Redis
         public void Dispose()
         {
             isDisposed = true;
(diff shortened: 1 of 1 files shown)


src/StackExchange.Redis/PhysicalBridge.cs
```diff
@@ -113,6 +113,7 @@ public enum State : byte
         public void Dispose()
         {
             isDisposed = true;
+            _backlogAutoReset?.Dispose();
```
kevin-montrose, line 116: Why `?.` - this is `readonly`, don't see how it could end up `null`.

kevin-montrose: I suppose in the absence of nullable reference types that's reasonable.

PR #2015, opened 2022-02-25
Backlog locking fixes - more #2008 follow-up

NickCraver:
In troubleshooting these 2 tests, I realized what's happening: a really dumb placement mistake in #2008. Now, instead of locking _inside the damn lock_, it loops outside a bit cleaner and higher up. Performance wins are the same but it's a lot sander and doesn't block both the backlog and the writer for another 5 seconds. Now _only_ the thread lingers and it'll try to get the lock when running another pass, if it gets any in the next 5 seconds.


philon-msft (approved):

NickCraver:
@philon-msft @mgravell This should fix the inadvertent post-backlog (in-lock) stall from #2008's gains...while maintaining those wins.

philon-msft (approved):

PR #2018, opened 2022-03-02
Update Timeouts.md

ja-nick:
The abbreviations table claims that 'v' represents the version of Redis but I'm pretty sure it actually shows the version of the `StackExchange.Redis` library currently in use.


NickCraver (approved):
Good clarification, thank you!

PR #2019, opened 2022-03-03
Connection logging: enhancements!

NickCraver:
This tweaks logging a bit to make it easier to parse and switches timings to `ValueStopwatch` while I'm in here adding one. Helps investigate things like #2017.

### DIFF
--- a/src/StackExchange.Redis/ConnectionMultiplexer.cs
+++ b/src/StackExchange.Redis/ConnectionMultiplexer.cs
@@ -1971,7 +1971,7 @@ namespace StackExchange.Redis
                     {
                         log?.WriteLine("Resetting failing connections to retry...");
                         ResetAllNonConnected();
-                        log?.WriteLine($"  Retrying; attempts left: {attemptsLeft}...");
+                        log?.WriteLine($"  Retrying - attempts left: {attemptsLeft}...");
                     }
                     //WTF("?: " + attempts);
                 } while (first && !healthy && attemptsLeft > 0);
--- a/src/StackExchange.Redis/ConnectionMultiplexer.cs
+++ b/src/StackExchange.Redis/ConnectionMultiplexer.cs
@@ -669,7 +669,7 @@ namespace StackExchange.Redis
         {
             if (tasks == null) throw new ArgumentNullException(nameof(tasks));
             if (tasks.Length == 0) return true;
-            var watch = Stopwatch.StartNew();
+            var watch = ValueStopwatch.StartNew();
             try
             {
                 // If no error, great
@@ -683,7 +683,7 @@ namespace StackExchange.Redis
                 var task = tasks[i];
                 if (!task.IsCanceled && !task.IsCompleted && !task.IsFaulted)
                 {
-                    var remaining = timeout - checked((int)watch.ElapsedMilliseconds);
+                    var remaining = timeout - watch.ElapsedMilliseconds;
                     if (remaining <= 0) return false;
                     try
                     {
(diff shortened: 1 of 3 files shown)


src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -998,6 +1001,16 @@ public void WriteLine(string message = null)
                     }
                 }
             }
+            public void WriteLine(string prefix, string message)
```
philon-msft, line 1004: Consider an `int indent` instead of `string prefix`?

src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -1772,10 +1789,11 @@ internal async Task<bool> ReconfigureAsync(bool first, bool reconfigureAll, LogP
                             }
                         }
 
+                        log?.WriteLine($"Endpoint summary:");
                         // Log current state after await
                         foreach (var server in servers)
                         {
-                            log?.WriteLine($"{Format.ToString(server.EndPoint)}: Endpoint is {server.ConnectionState}");
+                            log?.WriteLine($"  {Format.ToString(server.EndPoint)}: Endpoint is {server.ConnectionState}");
```
philon-msft, line 1796: Use indent/prefix instead of hardcoded spaces?

src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -1790,21 +1808,21 @@ internal async Task<bool> ReconfigureAsync(bool first, bool reconfigureAll, LogP
                                 var aex = task.Exception;
                                 foreach (var ex in aex.InnerExceptions)
                                 {
-                                    log?.WriteLine($"{Format.ToString(server)}: Faulted: {ex.Message}");
+                                    log?.WriteLine($"  {Format.ToString(server)}: Faulted: {ex.Message}");
```
philon-msft, line 1811: Are these `Format.ToString()` calls necessary? It looks like `ServerEndPoint` already overrides ToString to do that

src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -912,6 +913,8 @@ private static async Task<ConnectionMultiplexer> ConnectImplAsync(ConfigurationO
 
                     await Maintenance.ServerMaintenanceEvent.AddListenersAsync(muxer, logProxy).ForAwait();
 
+                    logProxy?.WriteLine($"Total connect time: {sw.ElapsedMilliseconds:n0} ms");
```
philon-msft, line 916: Is there any value in knowing how long it took to _fail_ to connect? If so, you could move these "Total connect time" lines into the `finally`

philon-msft (approved):

src/StackExchange.Redis/ConnectionMultiplexer.cs
NickCraver: We could, but overall it'd be doing a string creation and allocation that way so kept it simple :)

src/StackExchange.Redis/ConnectionMultiplexer.cs
NickCraver: Good question - I would like to change these _everywhere_ but going consistent for this PR and want to do those all at once :)

src/StackExchange.Redis/ConnectionMultiplexer.cs
NickCraver: I debated that as well - I feel like that might be confusing to give a total time when it just blew sky high for no reason - in those cases we care more about the log of what went sky high or what was the last thing logged, rather than the time...I think.

PR #2023, opened 2022-03-06
Fix for #1988 - respect SELECT missing from command map

### LINKED ISSUE #1988: SELECT after every EXEC / non-standard command
👋 I was debugging an issue today, a customer saw an abnormal number of SELECTS being sent to their Redis instance, which was pretty odd given they weren't passing any arguments into `GetDatabase`. After a bit of digging, I found [this](https://github.com/StackExchange/StackExchange.Redis/blob/main/src/StackExchange.Redis/PhysicalBridge.cs#L1419) bit of code from the libraries earliest days, which pointed me to the fact that whenever they execute a transaction, a `SELECT` is called immediately afterward.

I then noticed that above that bit, the `UNKNOWN` command also falls through to the same condition. This means that if you execute any Redis Command via the `Execute`/`ExecuteAsync` API (which is the pattern in NRediSearch, Redis.OM, NReJSON, etc. . .) that the multiplexer will immediately call `SELECT` afterward.

I suppose my question is in 2 parts

1: What was the origination of this behavior for `EXEC`? My guess would be that it's because if you call a `SELECT` in the middle of a transaction that it will change the client's selected database. But I don't think there's a way in which you can call an arbitrary command on a transaction is there? My other guess would be that since you are loaning a copy of the database to the transaction you need to be double sure that you set the selected database is correct afterward, but for some reason that doesn't ring correctly to me.

2: Would it be possible to provide either a way to extend the `RedisCommand` enum to allow expanded command sets, or alternatively perhaps provide a configuration item to prevent `UNKNOWN` commands from encountering this? 

I'd of course be happy to open any PRs required myself.

Thanks!

NickCraver:
This resolves #1988 by respecting when the `SELECT` command has been disabled. It's memoized which seems  bit extreme here but that's because we're inside the lock and every op matters. Measured locally to ensure no regressions.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -1,5 +1,9 @@
 # Release Notes
 
+## Unreleased
+
+- Fix [#1988](https://github.com/StackExchange/StackExchange.Redis/issues/1988): Don't issue `SELECT` commands if explicitly disabled ([#2023 by NickCraver](https://github.com/StackExchange/StackExchange.Redis/pull/2023))
+
 ## 2.5.43
 
 - Adds: Bounds checking for `ExponentialRetry` backoff policy ([#1921 by gliljas](https://github.com/StackExchange/StackExchange.Redis/pull/1921)) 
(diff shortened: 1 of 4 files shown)


NickCraver:
cc @slorello89

mgravell (approved):

PR #2031, opened 2022-03-10
Add ConfigurationOptions.BeforeSocketConnect

### LINKED ISSUE #1472: IP address binding
We need to bind to a specific IP address when creating a connection to our Redis instance. Not doing so means that Redis.StackExchange's behavior is dependant on something else (i.e. OS behavior); in our case, we seem to be picking which IP to connect from _mostly_ at random, which means that sometimes we won't be able to connect due to picking an interface without internet connectivity.

As far as I can tell, there is no option to bind to a specific address in the ConnectionMultiplexer's options.

NickCraver:
This adds a new config `Action` that allows modification of a socket after creating it and before connecting. It's passing the endpoint and connection type as well so if necessary per-connection options can be set.

Resolves #1472.

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -100,7 +100,11 @@ The `ConfigurationOptions` object has a wide range of properties, all of which a
 
 Additional code-only options:
 - ReconnectRetryPolicy (`IReconnectRetryPolicy`) - Default: `ReconnectRetryPolicy = ExponentialRetry(ConnectTimeout / 2);`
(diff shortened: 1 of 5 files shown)


philon-msft:
Is there a good spot in docs to mention this?

NickCraver:
@philon-msft oh yeah the config docs should get some love - good call!

NickCraver:
@philon-msft done :)

philon-msft (approved):

Issue #1698, opened 2021-03-02
Authentication failure when different passwords are used for sentinel and redis


rmkm4:
I am receiving below error when sentinel and redis are configured with different passwords. Is there a workaround to overcome this error ?

`It was not possible to connect to the redis server(s). There was an authentication failure; check that passwords (or client certificates) are configured correctly. AuthenticationFailure (None, last-recv: 34) on 10.33.190.74:6379/Interactive, Flushed/ComputeResult, last: ECHO, origin: SetResult, outstanding: 0, last-read: 0s ago, last-write: 0s ago, keep-alive: 60s, state: ConnectedEstablishing, mgr: 8 of 10 available, last-heartbeat: never, global: 0s ago, v: 2.2.4.27433`

NickCraver:
I don't think we support this right now - can you elaborate on your use case for needing 2 passwords, for the 2 connection layers?

minhnguyenvan95:
Hi @NickCraver I have same issue too, I don't know how to config Sentinel Password, We have 2 different password 1 for Sentinel, 1 for Redis

capcom923:
Hi @NickCraver I have the same issue. Do we have any progress?

BTW, Java already support this feature https://github.com/spring-projects/spring-boot/issues/21353

So, please help to priority this feature. Don't let us lose this game C# vs Java :)

Thanks,

NickCraver:
I was looking at this, but it's not as trivial to change as it appears - needs more thought. See the problem is we don't know for sure it's a Sentinel _until we connect_, so inverse to normal reasoning (at least for me) of "SetinelUser, SentinelPassword" settings, you really have new settings for the _child_ servers that sentinels return...or something. The relationship is very odd as a result of connecting and subsequently connecting downstream, needs more thought.

minhnguyenvan95:
thumbs up

kyao233:
I encountered the same issue, we didn't set password for sentinels, but redis instances are set password. 

spring:
  cache:
    cache-names: mycache
    type: redis
  redis:
    password: password1
    sentinel:
      master: mymaster
      password: password2
      nodes: myredis01:26379,myredis02:26379,myredis03:2637

if I set spring.redis.sentinel to null, this property will be set to spring.redis.password.

christianerbsmehl:
[...]

PR #2033, opened 2022-03-10
Add support for a credentials provider.

shachlanAmazon:
This allows the client to support servers that authenticates using ephemeral passwords, that are only usable for a limited amount of time.

If a server authenticates using ephemeral passwords, such as IAM or HashiCorp Vault, the set `Password` would become eventually unusable. Even if the user modified the `Password` property externally, since the client manages authentication internally at opaque points in time, the user would have to update the password regularly, regardless of whether it was actually required or not.

This change allows the user to define and inject a credentials provider, so that new credentials will be generated when needed.

### DIFF
--- a/src/StackExchange.Redis/ConfigurationOptions.cs
+++ b/src/StackExchange.Redis/ConfigurationOptions.cs
@@ -9,6 +9,7 @@ using System.Security.Cryptography.X509Certificates;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
+using StackExchange.Redis.Interfaces;
 using static StackExchange.Redis.ConnectionMultiplexer;
 
 namespace StackExchange.Redis
@@ -340,12 +341,20 @@ namespace StackExchange.Redis
         /// <summary>
         /// The user to use to authenticate with the server.
         /// </summary>
-        public string User { get; set; }
+        public string User { get => CredentialsProvider?.getUser(); set
+            {
+                CredentialsProvider = new ConstantCredentialsProvider(value, Password);
+            }
+        }
 
         /// <summary>
         /// The password to use to authenticate with the server.
         /// </summary>
-        public string Password { get; set; }
+        public string Password { get => CredentialsProvider?.getPassword(); set
+            {
+                CredentialsProvider = new ConstantCredentialsProvider(User, value);
+            }
+        }
 
         /// <summary>
         /// Specifies whether asynchronous operations should be invoked in a way that guarantees their original delivery order.
@@ -490,6 +499,11 @@ namespace StackExchange.Redis
         }
 
         /// <summary>
+        /// A provider for authentication credentials.
+        /// </summary>
+        public ICredentialsProvider CredentialsProvider { get; set; }
+
+        /// <summary>
         /// Parse the configuration from a comma-delimited configuration string.
         /// </summary>
         /// <param name="configuration">The configuration string to parse.</param>
@@ -531,8 +545,7 @@ namespace StackExchange.Redis
                 allowAdmin = allowAdmin,
                 defaultVersion = defaultVersion,
                 connectTimeout = connectTimeout,
-                User = User,
-                Password = Password,
+                CredentialsProvider = CredentialsProvider,
                 tieBreaker = tieBreaker,
                 writeBuffer = writeBuffer,
                 ssl = ssl,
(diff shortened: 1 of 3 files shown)


NickCraver:
Hey there!
This is something we want to add (and soon), but not specific to credentials. The plan here is to generically allowing changing of config at runtime. There are 2 ways I plan on doing this:
1. Adding user/password to the `DefaultOptionsProvider`, allowing anyone to inherit and override that way if wanted.
2. Providing the ability to run a `Action<ConfigurationOptions>` to change the running config of the multiplexer, hopefully with a subscription model so that some aggregate somewhere doesn't hold on to multiplexers.

I do appreciate the contribution here, and we do see the use case, but the solution is a very different approach we're going to take - this is something I hope to get to in the next few weeks (hopefully sooner).

shachlanAmazon:
Great! Happy to hear that I'm aiming in the right direction :)

The `DefaultOptionsProvider` sounds like it would work great for this. Running an `Action<ConfigurationOptions>` sounds like a push-based model, while a push-based model better suits the scenarios I described.

If there's anything I can do to help with adding these abilities to `DefaultOptionsProvider`, please say so. From a brief look it seems like just adding getters & setters would be sufficient. If so, I'd be happy to do that.

NickCraver:
The `DefaultOptionsProvider` will work for _a_ case of this which is why I want to add this, but it doesn't work for the _common_ case. The provider is only for _fallback_ values where they're not explicitly set (e.g. in a connection string). Most people (read: all, pretty much) _are_ setting them that way today. This means a fallback doesn't work, but may be viable in the future. I'm happy to ping when the rotation bits are in - I already have a branch local it may just be a few nights before I get to think through it.

Rotating is easy, how to _not_ hold onto multiplexers is the question mark and I need to explore there a bit.

shachlanAmazon:
yes, please ping me. Thanks :)

NickCraver:
We have a design for this in mind after discussions yesterday - going to close this out. In short: ConfigurationOptions will snapshot _some_ things (like endpoints) on the multiplexer but the rest will be observed _live_ from the `ConfigurationOptions` passed in. So basically: hold on to the options you pass in, and modify User/Pass there, or any other mechanism/piping in place to accomplish that. It's something I'll poke at after NRTs since there would be quite a bit of merge conflicting atm.

shachlanAmazon:
As I mentioned, this means that you use a push-based model, which isn't a good solution to this problem. It would be much better if there was some mechanism to generated and pass new credentials only when they're needed, instead of requiring the user to constantly update the options, regardless of necessity.

shachlanAmazon:
In my PR, if I wanted credentials to be requested on demand, I would've set a credentials provider where the implementation of the getter of the username/password properties generated the credentials on request.
If I understand your suggestion, then I'll need to pass a `ConfigurationOptions` and update it externally, without being aware of when the getters are being called. Since `ConfigurationOptions` is sealed, I can't inherit from it and override the getters. So, yes - I still have access to the setters, and can change the values, but if I don't know when the getters are called, I can't change the values on demand. Specifically, I can't change the values based on their getters being called, before the getters return the last set value. So, the "based on anything at all" part isn't correct, AFAIS.  The credentials provider / default options provider would've enabled that, by allowing the user to create an object that would've created new credentials on a call to the credentials' properties getters.

It's been a while since I've worked with .net, so I might be missing a mechanism here. Am I missing something?

NickCraver:
@shachlanAmazon This PR didn't do that either - if a username or password was set it was that value _no matter what_. Nothing changes there. If it'd fallback to a provider (`DefaultOptionsProvider`, which I mentioned above), that _is_ something you can update and have the getters do anything you want. There's no functionality difference vs. what's in this PR, only a simpler approach with fewer classes.

shachlanAmazon:
> This PR didn't do that either - if a username or password was set it was that value no matter what.

That's not correct. This PR allowed the user to implement a credentials provider that generated credentials on demand, and then on reconnection attempts the new credentials were used. I know, because I've implemented one in a test branch, which was tested with a Redis Server that required new credentials after a set amount of time. I know you don't want the dependency on the AWS SDK, so I didn't add it to this PR.

But as I mentioned before, the fallback to a `DefaultOptionsProvider` for credentials would work well with my scenario. I understood from [this comment](https://github.com/StackExchange/StackExchange.Redis/pull/2033#issuecomment-1069314767) that it wasn't on the table anymore, but if it's still possible, I'm all for that. I don't care about this PR, I only care about the ability to refresh credentials without an external trigger, however it's implemented.

NickCraver:
Both are still in the plan - there are many issues we have consider for things like this. For example, if we had this in a public interface, I'd have to break it immediately to fix #1698 whereas we can much more simply add props to the base classes today. All of this factors into the plan, but we don't write all that up - quicker to do it in the limited time we have to throw at it.

shachlanAmazon:
Fair enough, and I do appreciate the time you put into this library, both in code and in these discussions. Again, if there's anything I can do to help, please don't hesitate to ask.

PR #2040, opened 2022-03-16
Test Suite: Enable Nullable Reference Types

NickCraver:
I want to enable NRTs on the main library, because in testing doing so I found a lot of interesting issues we _should_ have caught especially in result processors and unhandled cases. I figure the best way to do that is to first enable NRTs on our test suite to get the experience similar to what our users would when we shipped a NRT-enabled library version. Either we do or we don't, but having the tests NRT aware is a new win and let's us explore what that diff would look like.


NickCraver:
@philon-msft @mgravell This is *tests only*, so in further PRs we can get a feel for "how much would this suck". Either way, net win here and good to have this in `main` to have decent behavior discussions afterwards.

PR #2051, opened 2022-03-22
ILogger for ConnectionMultiplexer

NickCraver:
ILogger support - since we support `Action<ConfigurationOptions>` this can be added today e.g. with `o => o.LoggerFactory = myLoggerFactory`:
```cs
var muxer = ConnectionMultiplexer.Connect("localhost", o => o.LoggerFactory = myLoggerFactory);
```
...or we could add sibling overloads to the existing `TextWriter`. I'd love to not create a ^2 matrix every time we do this, though. Starting simpler for that reason. Note: this is on top of #2050 for simplicity, will roll against `main` when merged.

TODOs here:
- [x] General sanity check
- [x] Overloads decision
- [x] Tests of this implementation
- [x] Seeing which additional things need logging, we should make a list and check it here.
- [ ] How do we make these calls the most efficient?
- [ ] Do we eat the overhead for structured logging?


src/StackExchange.Redis/ConfigurationOptions.cs
```diff
@@ -410,6 +413,15 @@ public int KeepAlive
             set => keepAlive = value;
         }
 
+        /// <summary>
+        /// The <see cref="ILogger"/> to use for connection events.
+        /// </summary>
+        public ILogger Logger
+        {
+            get => logger;
```
NickCraver, line 421: Should add to the default options as well

src/StackExchange.Redis/ConfigurationOptions.cs
```diff
@@ -429,6 +432,15 @@ public int KeepAlive
             set => keepAlive = value;
         }
 
+        /// <summary>
+        /// The <see cref="ILogger"/> to use for connection events.
+        /// </summary>
+        public ILogger? Logger
```
mgravell, line 438: add `Logger = Logger,` to `Clone`

src/StackExchange.Redis/ResultProcessor.cs
```diff
@@ -221,7 +221,7 @@ public virtual bool SetResult(PhysicalConnection connection, Message message, in
             {
                 try
                 {
-                    logging.Log?.WriteLine($"Response from {bridge?.Name} / {message.CommandAndKey}: {result}");
+                    logging.Log?.LogInfo($"Response from {bridge?.Name} / {message.CommandAndKey}: {result}");
```
mgravell, line 224: how do we feel about allocations and log-levels? there are the stateful callback APIs that avoid allocs / formats when the log level turns out to be too low, although these take a very different shape, i.e.
```
log.Log(LogLevel.Information, default, (name, id), null, static (state, _) => $"Thing {state.name} of {state.id}");
```

(can be simplified quite a bit by using local extension methods)

if we don't care, that's also fine, as long as deliberate

petrroll: (not sure if possible due to supporting older build SDKs but: https://learn.microsoft.com/en-us/dotnet/core/extensions/logger-message-generator)

petrroll, line 224: Apart from using stateful callback APIs that avoid allocations, providing Logging API with pre-interpolated strings effectively disables structured logging (https://learn.microsoft.com/en-us/dotnet/core/extensions/logging?tabs=command-line#log-message-template or aka.ms/r9 (internal only) and search structured logging). 

There's a push (internally, at least in some orgs) towards structured logging so if it was possible to support that, it'd be doubly awesome :).

petrroll: Ofc, the change 'd need to be made in LogProxy and threaded through all layers.

src/StackExchange.Redis/ConfigurationOptions.cs
NickCraver: Aye, need to add to DefaultOptions too, I think - was confirming with @philon-msft on design but I think 👍 to both it is

src/StackExchange.Redis/ResultProcessor.cs
NickCraver: There are a few ways to make this faster on allocations - totally cool with any of them my thinking is:
1. It should be analyzer enforced if possible
2. It should work everywhere possible.

There exists today the signatures like [LogInformation(ILogger, String, Object[])](https://learn.microsoft.com/en-us/dotnet/api/microsoft.extensions.logging.loggerextensions.loginformation?view=dotnet-plat-ext-6.0#microsoft-extensions-logging-loggerextensions-loginformation(microsoft-extensions-logging-ilogger-system-string-system-object())) and friends, but if we're formatting that message at all it's the same performance _plus_ the object array in the end to get structured logging.

I was really hoping before tackling this that the interpolated string handler goodness in .NET 6+ (https://devblogs.microsoft.com/dotnet/string-interpolation-in-c-10-and-net-6/) would have landed in logging extension methods but alas this set of APIs didn't happen. I chatted with @stephentoub on this quite a while back and did poke at it, but the fundamental problem you arrive at is "that's not the shape of ILogger" in the end, so if that's your interface someone is calling the formatter in all the cases I could try to be more efficient in. It's entirely possible a smarter API design that I was capable of would still be a win here though, and if so: I'm all game for it.

In the example above my worry cases would be "touching ValueTuple again with a 1000 ft pole" and exceedingly high format count cases to maintain like exception messages today with something like 30-50+ args in play.

I'll keep hacking at making all the basics correct here, but if we can come up with optimal paths to go back through: yeah, absolutely. Do you wanna try a spike against this branch and see what we can do for options? Absolutely wanting to make this more efficient if we reasonably can.

NickCraver: This is true, but it's also more expensive at every call site (due to the allocation of the `object[]` array at the end). Ideally we could enable structure logging without the overhead :-/ I'm not against structured logging - tons of uses, but it's odd to me that logging is such a performance hog across so much and there isn't an easier path that's also efficient here. Hoping we continue to see that improve in the coming .NET releases if we can push a bit on logging APIs. `ILogger`'s shape makes that hard though, when realizing wins through the pipe.

petrroll: It's more expensive at every call site _when the logging actually goes trough_, i.e. for levels that are enabled. But for log-calls that don't end up being used (loglevel, ...), you don't pay for interpolating the strict, which I'd guess would be roughly similar to the cost or higher than allocating object[] array.

NickCraver: If you have a logger _at all_, it's almost certainly to go through here. We're already null checking against that, so we've solved the first problem and moved on to the second. In our case, because of the elvis approach, it' universally more expensive :)

src/StackExchange.Redis/ConfigurationOptions.cs
NickCraver: Alrighty, added these in last push!

src/StackExchange.Redis/ResultProcessor.cs
petrroll: Ah right, brain-fart on my side 😅. The array needs to be allocated always. In that case I see how it's a performance negative.

stebet: Do the Source Generated logger messages help here? I admit I haven't looked at them properly. https://learn.microsoft.com/en-us/dotnet/core/extensions/logger-message-generator

src/StackExchange.Redis/LogProxy.cs
```diff
@@ -1,50 +1,77 @@
...
-    private TextWriter? _log;
-    internal static Action<string> NullWriter = _ => { };
 
-    public object SyncLock => this;
-    private LogProxy(TextWriter log) => _log = log;
-    public void WriteLine()
+    public void LogInfo(string? message = null)
     {
+        var msg = $"{DateTime.UtcNow:HH:mm:ss.ffff}: {message}";
         if (_log != null) // note: double-checked
```
WeihanLi, line 39: Maybe we could consider exacting a method to write with `_log`, maybe

``` csharp
private void WriteForTextWriter(string? message)
{
  if (_log is null) return;
  lock (SyncLock)
  {
    if (message is null)
    {
      _log?.WriteLine();
    }
    else
    {
      _log?.WriteLine(message);
    }
  }
}
```

NickCraver: Thanks for the suggestion! In this case I'm not sure it cleans it up much, because we're calling a different overload each time (trace/info/error/etc.) and the only 2 that stare a signature are Trace & Info each of which would take a action for the proper trace/info method underneath. If we could use it for more it might be worth breaking out, but not sure we'd save much here...better to be consistent across all methods I think :)

NickCraver:
@philon-msft I reversed the `LoggerFactory` approach here because it doesn't really solve any problems...but makes several others. For the wrapping question I think we do the wrap of the `ILogger` when configuring with Azure options in the upstream package, like this from an example test (needs null guards and such, but you same idea):
```cs
    [Fact]
    public async Task WrappedLogger()
    {
        var options = ConfigurationOptions.Parse(GetConfiguration());
        options.Logger = NullLogger.Instance;
        var wrapped = new TestWrapperLogger(options.Logger);
        options.Logger = wrapped;

        using var conn = await ConnectionMultiplexer.ConnectAsync(options);
        Assert.True(wrapped.LogCount > 0);
    }

    public class TestWrapperLogger : ILogger
    {
        public int LogCount = 0;
        public ILogger _inner { get; }

        public TestWrapperLogger(ILogger toWrap) => _inner = toWrap;

        public IDisposable BeginScope<TState>(TState state) => _inner.BeginScope(state);
        public bool IsEnabled(LogLevel logLevel) => _inner.IsEnabled(logLevel);
        public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception? exception, Func<TState, Exception?, string> formatter)
        {
            Interlocked.Increment(ref LogCount);
            _inner.Log(logLevel, eventId, state, exception, formatter);
        }
    }
```

This also allows opt-out for those who want it. Note: consumers could re-use options so in any wrap you'd want to check type of the existing logger and handler null, something else, and *my wrapper already* accordingly, to not create cascade wrappers.

Thoughts?

davidfowl:
Is there any reason this isn't an ILoggerFactory? It would allow the client to use several logging categories instead of logging everything through a single ILogger instance.

NickCraver:
> Is there any reason this isn't an ILoggerFactory? It would allow the client to use several logging categories instead of logging everything through a single ILogger instance.

We went back and forth here, ultimately I don't think we have more than 1 log category and example usage was far easier in non-DI scenarios which happens a lot since we're in the position of being a static thing held onto (manually) especially in older frameworks. But we also couldn't find a lot of other art in existing Azure SDKs either as examples, what we could find had no logger or was split between the options. Thoughts?

> Guard the logging calls with IsEnabled checks if you are computing data to log (and if the level is rarely enabled)

Definitely planned, wasn't part of the original move (none of this logging is high volume) - likely moving to extensions or something to not have that `if` wrap everywhere (are there any build-in extensions methods for that type of behavior, where the only calculation is on the line, e.g. `log.InfoIfEnabled(....)` in the base packages? I wasn't aware of anything there but generally handy since formatting is the work, if any, involved for our primary case.

> Don't use the build int LogXXX methods, use the source generator

I'm not familiar with the current state of generators, is there something outside of `LoggerMessageAttribute` that works into older frameworks that we support? I'd love to optimize everywhere we reasonably can, even if these aren't high volume.

davidfowl:
> But we also couldn't find a lot of other art in existing Azure SDKs either as examples, what we could find had no logger or was split between the options. Thoughts?

Working on getting this documented. It should be an ILoggerFactory. The library is responsible for creating and owning the categories (which map to loggers). You can change today from 1 to N, but that’s up to your library. Still the top level API should take a factory.

davidfowl:
> likely moving to extensions or something to not have that if wrap everywhere (are there any build-in extensions methods for that type of behavior, where the only calculation is on the line, e.g. log.InfoIfEnabled(....) in the base packages? I wasn't aware of anything there but generally handy since formatting is the work, if any, involved for our primary case.

Not for high performance code. This client should be using the most optimal path. No closures, no delegates, only the fast path (it statements).

davidfowl:
> I'm not familiar with the current state of generators, is there something outside of LoggerMessageAttribute that works into older frameworks that we support? I'd love to optimize everywhere we reasonably can, even if these aren't high volume.

yep it works everywhere

NickCraver:
@davidfowl If you have docs for this I'd love to look - not living in an ILogger world daily right now and things like creating any non-null logger factory for examples and such were very painful (maybe because we couldn't find the docs). If I can see much simpler examples on usages _not depending on DI_, that would change my mind a ton about the approach. We just couldn't find decent examples either way (either from lack of docs or SEO), pointers would be very much appreciated <3

NickCraver:
Thanks @IEvangelist! What I didn't find obvious was unit testing here. For example making a logger and having the factory return it (so I could check what's logged and ensure it's correct). Is there a built-in way to do this cleanly or do you need to make your own logger and factory implementation each time?

The same question stands for any users trying to get a simple string out the end in debugging something. Unless there's a built-in wrapper for ILogger -> ILoggerFactory (for the simple case of using an existing logger or a custom one, again without DI), don't they have to boilerplate both?

The .AddConsole() examples are great for simplicity, but library authors need things more like this for guidance IMO. It's possible all this exists but I wasn't able to find it and ended up writing implementations each time…which also led to ILogger being more practical and simpler, since we don't need many.

Thus far I've found users tend to think of the library as 1 thing (SQL, Cosmos, Redis, etc.) and I need to be able to reason about what that output looks like, so I'm not sure I'm actually sold on getting more complex than ILogger for libraries specifically, or for us at least.

Thoughts?

IEvangelist:
> Is there a built-in way to do this cleanly or do you need to make your own logger and factory implementation each time?

I'm not familiar with any existing built-in way of doing this. But I also don't think it's too complex either. You should be able to easily create a test logger factory that returns a test logger. I believe this could be mocked too.

> The same question stands for any users trying to get a simple string out the end in debugging something. Unless there's a built-in wrapper for ILogger -> ILoggerFactory (for the simple case of using an existing logger or a custom one, again without DI), don't they have to boilerplate both?

As for boilerplating both, in the case of unit testability—that's not too big of a concern, is it? I mean if you're concerned about the logs output, create a test logger factory and test logger, and read the output, no?

NickCraver:
@IEvangelist Yeah I guess we can mock each every time, but: if we're only ever making 1 (for the multiplexer)...what is the positive of taking an ILoggerFactory over an ILogger? I'm not understanding what the win is for the additional mocks/complexity. If we were making `n`, yeah sure I totally get it - when we are making 1...why do that?

davidfowl:
Today we don't ship any fakes for the logger factory, but we're shipping it in .NET 8 (https://github.com/dotnet/extensions/tree/main/src/Libraries/Microsoft.Extensions.Telemetry.Testing/Logging). Today lots of teams have their own impl.

NickCraver:
@davidfowl Any thoughts on this given we only have the 1 logger in play? We have no desire to expose internals to a user of the library, they see/use ConnectionMultiplexer and that's where the logger lives, so my preference is to stick to `ILogger` unless we have a compelling reason outside the multiple-loggers case - thoughts?

IEvangelist:
@NickCraver if it's only ever going to be a single logger for just the `ConnectionMultiplexer` why not `ILogger<ConnectionMultiplexer>` so that it has a category?

NickCraver:
@IEvangelist yeah that's absolutely fine, it can be an `ILogger` or an `ILogger<T>`, no big difference there in mocking or usability IMO

philon-msft:
One thought: If we take an ILoggerFactory, then we can CreateLogger() a separate ILogger with a different category for logging from extension packages like MS.Azure.SERedis vs. the ConnectionMultiplexer's ILogger.

NickCraver:
@philon-msft That's a great point, things _not us_ may be making other loggers. I just pushed up a migration back to `ILoggerFactory` to support this (we can probably simplify tests further), let's test this out and critique.

@davidfowl @IEvangelist For this PR/iteration, we're explicitly not concerned with the attribute-based logging (these are not hot paths), but we can optimize in a follow-up pass and see how messy that gets. Overall though I'd really expect the generation to happen from string interpolation here long-term, is that planned? I heard it talked about but not sure where it stands (cc @stephentoub)

davidfowl:
My thoughts haven’t changed. The API should take a factory and use a single logger internally (if that desired)

davidfowl:
More info here: https://github.com/dotnet/designs/blob/main/accepted/2021/logging-generator.md#supporting-string-interpolation-builder-in-logging

NickCraver:
> More info here: [dotnet/designs@`main`/accepted/2021/logging-generator.md#supporting-string-interpolation-builder-in-logging](https://github.com/dotnet/designs/blob/main/accepted/2021/logging-generator.md?rgh-link-date=2023-08-12T22%3A48%3A24Z#supporting-string-interpolation-builder-in-logging)

@davidfowl thanks! I'm still unclear of the status (is it still a proposal/design, or landed?). Overall: we're going to go with the current non-optimal (but also non-hot path) approach to get this in, but open to optimizing it more later. I think we should also get some proper event IDs going and refine a bit overall there too to be a decent example - that is definitely something I'd like as well querying Kusto daily.

The method-per-line boilerplate approach on some of the current methods is definitely what I'm most averse too, that's a lot of maintenance overhead to optimize cold paths. Proposals to remove that and allow an inline call to be optimize (as it is with existing StringBuilder usage) is very appealing. If it's still in proposal, happy to help make the case for it.

davidfowl:
> The method-per-line boilerplate approach on some of the current methods is definitely what I'm most averse too, that's a lot of maintenance overhead to optimize cold paths. Proposals to remove that and allow an inline call to be optimize (as it is with existing StringBuilder usage) is very appealing. If it's still in proposal, happy to help make the case for it.

You should treat logs with a similar level of rigor as public API. I think the source generator helps you manage that. Sure, it's more boilerplate but it's actually very important for libraries to be more careful with how they manage event ids and log levels. Same goes for metrics and other diagnostics information.

NickCraver:
I'm not sure I can agree with that in all areas - we find new things to log and add all the time from scenarios people haven't hit before. Happy to chat more on this, but if you look at the history of our logging here that's been the case for _most_ additions, and adding more useful logs for example in the log of connecting to the server is going to get additions and changes not nearly as strict as a public API. Exception messages are also an extreme version of this - we add a ton of information necessary to help debug there, because all we get in an issue as library authors is often _just that message_, so it's intentionally pretty crammed full of stats - I wouldn't consider that anywhere near as stable as public API (which we take very seriously).

I think we may have 2 different things in mind here. For things like events, exceptions, etc. I do not think the boilerplate or methods or events are onerous - I think it's fine. In the places we're say logging 50-100 lines of a connection logs: that's kinda crazy levels of overhead to me. What are your thoughts on that area? That's where I don't see the current advice fitting well (but is also probably a pretty rare use case). What implementation do you picture there as being ideal?

mgravell (approved):
I like it

philon-msft (approved):

docs/Configuration.md
```diff
@@ -100,6 +100,8 @@ The `ConfigurationOptions` object has a wide range of properties, all of which a
 | setlib={bool}          | `SetClientLibrary`     | `true`                       | Whether to attempt to use `CLIENT SETINFO` to set the library name/version on the connection              |
 
 Additional code-only options:
+- Logger (`ILogger`) - Default: `null`
```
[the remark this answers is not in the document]
NickCraver: Fixed!

AlexanderKot:
Hello all
When using StackExchange.Redis with 
[StackExchange.Redis.Extensions RedisConnectionPoolManager](https://github.com/imperugo/StackExchange.Redis.Extensions/blob/faae0f46236a2ac0cf699d6a3c7796ee9c67acc9/src/core/StackExchange.Redis.Extensions.Core/Implementations/RedisConnectionPoolManager.cs)
It is not possible to use new LoggerFactory feature, because RedisConnectionPoolManager create redis connection inside, and initialized by RedisConfiguration class. And RedisConfiguration cannot deserialise LoggerFactory from string representation.

Is this issue for StackExchange.Redis project or for StackExchange.Redis.Extensions?

NickCraver:
@AlexanderKot Thats definitely an issue for them - we have zero control over that package :)

AlexanderKot:
Thank you
I was not attentive enough and decided that RedisConfiguration for your code base - i was wrong, it is also from Extensions

AlexanderKot:
But it is not so trivial
RedisConfiguration inside use 
newOptions = ConfigurationOptions.Parse(ConnectionString);  //This options will be used for connection
ConfigurationOptions from StackExchange.Redis codebase.
It do clone of LoggerFactory, but cannot initialize it in Parse.
So question seems opened

pairbit:
Nooo what have you done. Why did you make a dependency on Microsoft.Extensions.Logging.Abstractions
I cried when I saw this

davidfowl:
That's a fine dependency.

NickCraver:
> Nooo what have you done. Why did you make a dependency on Microsoft.Extensions.Logging.Abstractions??? I cried when I saw this :(((

To log stuff, mainly...

Note that as I'm writing this that package has 3.7 billion downloads, which means this isn't a net new dependency for likely the vast majority of users.

pairbit:
right now I made a fork without the dependency Microsoft.Extensions.Logging.Abstractions. The fact is that Microsoft.Extensions.Logging.LoggerExtensions are very bad, they are prohibited from being used in our projects.

Please create your own IRedisLogger interface without params object[]. Also replace the ILoggerFactory type with the Func<Type,IRedisLogger> delegate.

NickCraver:
@pairbit Are we using the params overload somewhere I'm not seeing? We very much go out of our way to keep things efficient, so if you have a specific thing to point to I'm all ears. "Please break your API because I don't like this thing" isn't about efficiency though - this is something that's extremely pervasive in the ecosystem and allows most devs to discover issues more easily - that's worth a dependency to me. Every dev having to write basically the same adapter or include 2 packages doesn't ultimately remove a package (including this package) or enhance the user experience.

For what it's worth, I would love to see codegen improvements for those other areas of the logging interfaces (as was done with StringBuilder), and sincerely hope that happens.

pairbit:
@NickCraver You don't use methods with an objects parameter. 
I described this on the ticket. 
Your library managed without this dependency for 9 years. 
And I still don’t understand why you had to add a whole dependency for the sake of one Log method, which is not the main one in your library.

I understand that my request is probably too petty.
Thank you for your time, my task was to try to convince you.

src/StackExchange.Redis/EndPointCollection.cs
```diff
@@ -208,12 +209,12 @@ internal async Task ResolveEndPointsAsync(ConnectionMultiplexer multiplexer, Log
...
                         else
                         {
-                            log?.WriteLine($"Using DNS to resolve '{dns.Host}'...");
+                            log?.LogInformation($"Using DNS to resolve '{dns.Host}'...");
                             var ips = await Dns.GetHostAddressesAsync(dns.Host).ObserveErrors().ForAwait();
                             if (ips.Length == 1)
                             {
                                 ip = ips[0];
-                                log?.WriteLine($"'{dns.Host}' => {ip}");
+                                log?.LogInformation($"'{dns.Host}' => {ip}");
```
niemyjski, line 217: Seems like a lot of these logging levels should have been LogDebug or LogTrace

NickCraver: Based on what? This is a core part of connecting to the server, knowing which IP we connected to is definitely information level in my mind. I'm open to adjusting these of course, but don't agree with the specific thing commented on here. I think most people would default to only-errors for StackExchange.Redis, but anything beneath that like connecting when we're a client with a persistent connection is, IMO, information level.

niemyjski: I agree it could be important and can see if your only logging / interaction is with this library you might prefer it. However, most apps don't need this kind of logging and just adds to the noise. as the default log level is info 

But if you are having connectivity issues you are probably going down to the debug or trace levels anyways. Based on default out of the box log levels. Also, for most people this information is not needed and might even be a possible security concern leaking information about services (log collection tends to have different roles and policies).

Issue #2047, opened 2022-03-18
Support for ZRANGESTORE


danielgerlag:
Are there any plans to implement support for ZRANGESTORE?

https://redis.io/commands/zrangestore

NickCraver:
No immediate plans at the moment but seems on the simpler side since it's a new command (e.g. unlike `ZRANGE`/`ZREVRANGE` it's an `int` return) - happy to take a PR or I'll try and get to this in the queue.

One thing that concerns me is that the redis.io prompt doesn't have it working which I haven't seen before /cc @slorello89: 
<img width="336" alt="image" src="https://user-images.githubusercontent.com/454813/159126184-a1b5e458-f134-4369-aedc-d78821167ebd.png">

danielgerlag:
I think that prompt is backed by v6.0. This command is new in v6.2.

NickCraver:
Gotcha, thanks. Given how many commands are added in v6.2.0, I want to see if we can change that - as a library author it really helps in playing with things quickly.

slorello89:
Hey @NickCraver , it looks like it's not an issue with the version or the command or anything, just a bit of weirdness in redis-io. 

TL;DR: There's a missing line in a config file in the reids-io app that's causing that weirdness

[...]

PR #2052, opened 2022-03-22
Adding ZRangeStore

slorello89:
Adds ZRangeStore for #2047 

APIs added:

* SortedSetRangeAndStore
* SortedSetRangeAndStoreAsync

Rather than funneling everything through a single method, I left the interface to be reflective of the current ZRange interface in `IDatabase`/`IDatabaseAsync`.

Added a new Message type:

`SortedSetRangeAndStoreCommandMessage` to handle the `ZRANGESTORE` as it didn't quite fit any of the other message types. I left this in `RedisDatabase.cs` as this follows the pattern of `SortedSetCombineAndStoreCommandMessage`, if you'd like to evict it to somewhere else I've no issue doing that.

API Quirks

`ZRANGESTORE` with the `REV` flag is a bit odd, all it effectively does is reverse the ordering of the start/stop arguments, basically if `REV` is set `start` must be greater than `stop`, I did not add any sort of smarts to effect this reversal by default as again this would not be the expected behavior of the older APIs, happy to effect the swap if you think that API would be more desirous.


NickCraver:
Awesome - will try and review this tonight! Some quick pass feedback: 
- Let's rename `SortedSetOrderBy` to `SortedSetOrder` (to not release "By").
- There's a few `throw new NotImplementedException();` which should be `.Inner` calls
- I'm not sure about the method split (need to look harder - my instinct say we collapse these but will experiment - for example `SortAndStore*` has a type overload). The reason those were separate methods were because they had different return types which isn't the case here. The more I look through writing this...definitely thinking 1 method (per sync/async) taking the "by" argument - thoughts?
- Is there a case where the return type wouldn't be `long` for these? If not, that's a more friendly API (or did you spot a consistency issue there?)

I'll try and give a thorough review tonight but wanted to give feedback ASAP too. This looks like a great effort to get the API in and match the norms, hugely appreciated <3

NickCraver:
@slorello89 Oh I see now (thanks for the detail!) - yes for certain that's why those were separate. I'd recommend we more closely match Redis here on these new methods by combining and look at folding those old style together in a future major release.

slorello89:
@NickCraver - sounds good, I'll fold them together, honestly will make the API much cleaner. Did I break the thing in appveyor? sounds like it's probably separate from what I did. Probably should skip all my code since it's running on windows lol.

image

NickCraver:
Aye ignore - that's not you :)

slorello89:
@NickCraver - one final question, for erroneous combinations of arguments, should we catch the error's pre-flight?

E.g. using the inclusive/exclusive ranges for ZRANGE for ranks is not valid, same with using LIMIT/offset (skip/take). 

My philosophy is so long as the errors thrown by the server clear enough I would typically leave such validations to the server, not sure where you land on this:

```
127.0.0.1:6379> ZRANGESTORE destKey sourceKey 0 (-1
(error) ERR value is not an integer or out of range
127.0.0.1:6379> ZRANGESTORE destKey source 0 -1 LIMIT 0 10
(error) ERR syntax error, LIMIT is only supported in combination with either BYSCORE or BYLEX
```

your call (I sort of don't think either of those errors will be extremely helpful to someone debugging so we're probably better off throwing argument errors).

NickCraver:
> one final question, for erroneous combinations of arguments, should we catch the error's pre-flight?

I could go either way on this - if it's a few lines to put an `ArgumentException` in, I lean towards doing it on because it then doesn't couple our args/names/etc. to server names which can be confusing as things shift over time and we maintain compatibility. No strong feelings - we can always add :)

tests/StackExchange.Redis.Tests/SortedSets.cs
```diff
@@ -146,5 +176,468 @@ public async Task SortedSetPopMulti_Zero_Async()
...
+
+            var db = conn.GetDatabase();
+            var me = Me();
+            var sourceKey = $"{me}:ZSetSource";
+            var destinationKey = $"{me}:ZSetDestination";
+
+            db.KeyDelete(new RedisKey[] {sourceKey, destinationKey}, CommandFlags.FireAndForget);
+            db.SortedSetAdd(sourceKey, lexEntries, CommandFlags.FireAndForget);
+            var exception = Assert.Throws<ArgumentException>(()=>db.SortedSetRangeAndStore(destinationKey, sourceKey,0,-1, take:5));
+            Assert.Equal($"take argument is not valid when sortedSetOrder is ByRank you may want to try setting the SortedSetOrder to ByLex or ByScore{Environment.NewLine}Parameter name: take", exception.Message);
```
NickCraver, line 623: This may be built on one OS and run on another - if new lines are in tests you'll want to use the optional param:
```cs
Assert.Equal("", "", ignoreLineEndingDifferences: true);
``

Comment (slorello89) [reply tests/StackExchange.Redis.Tests/SortedSets.cs]: I figured out what it was, it's not an OS thing, it's a Framework vs core thing lol. framework's Argument 

good ol' framework uses environment newline: https://github.com/microsoft/referencesource/blob/master/mscorlib/system/argumentexception.cs#L70

whereas core runtime uses a space:
https://github.com/dotnet/runtime/blob/main/src/libraries/System.Private.CoreLib/src/System/ArgumentException.cs#L86

lol

Comment (slorello89) [reply tests/StackExchange.Redis.Tests/SortedSets.cs]: I just said, use the param name, good enough lol.

File: src/StackExchange.Redis/Enums/SortedSetOrder.cs [hunk 2]
```diff
@@ -0,0 +1,32 @@
...
+    ByLex
+}
+
+internal static class SortedSetOrderByExtensions
+{
+    public static RedisValue GetLiteral(this SortedSetOrder sortedSetOrder) => sortedSetOrder switch
+    {
+        SortedSetOrder.ByLex => RedisLiterals.BYLEX,
+        SortedSetOrder.ByScore => RedisLiterals.BYSCORE,
+        _ => RedisValue.EmptyString
```
NickCraver, line 30: Recommend `RedisValue.Null` here, and downstream always call `.GetLiteral()` and check `! .IsNull` on it (prop on `RedisValue`). This makes maintenance a bit easier. Right now these 2 if/switches have to correlate.

src/StackExchange.Redis/Interfaces/IDatabase.cs
```diff
@@ -1395,6 +1395,24 @@ public interface IDatabase : IRedis, IDatabaseAsync
         /// <remarks>https://redis.io/commands/zrevrange</remarks>
         RedisValue[] SortedSetRangeByRank(RedisKey key, long start = 0, long stop = -1, Order order = Order.Ascending, CommandFlags flags = CommandFlags.None);
 
+        /// <summary>
+        /// Takes the specified range of elements in the sorted set of the source key and stores
+        /// them in a new sorted set at the destination key.
+        /// </summary>
+        /// <param name="destinationKey">Where the new set will be stored</param>
+        /// <param name="sourceKey">The sorted set to take the range from</param>
```
NickCraver, line 1403:
```suggestion
        /// <param name="destinationKey">Where the resulting set will be stored.</param>
        /// <param name="sourceKey">The sorted set to take the range from.</param>
```

src/StackExchange.Redis/Interfaces/IDatabase.cs
```diff
@@ -1395,6 +1395,24 @@ public interface IDatabase : IRedis, IDatabaseAsync
...
+        /// <param name="start">The starting point in the Sorted Set, if using BYLEX this should be a string.</param>
+        /// <param name="stop">The stopping point in the range of the sorted set, if using BYLEX this should be a string.</param>
+        /// <param name="sortedSetOrder">The ordering criteria to use for the range. Choices are ByRank, ByScore, and ByLex.</param>
+        /// <param name="exclude">Whether to consider score or lex ranges exclusively or not (e.g. scores greater than 3 or greater than or equal to 3)</param>
+        /// <param name="order">The direction to consider the start and stop in, if Ascending, the start must be smaller than the stop, if Descending Stop smaller than start</param>
+        /// <param name="skip">Elements into the sorted set to skip. Should only be used </param>
+        /// <param name="take">Number of elements to pull into the new set.</param>
+        /// <param name="flags">The flags to use for this operation</param>
```
NickCraver, line 1411:
```suggestion
        /// <param name="exclude">Whether to exclude <paramref name="start"/> and <paramref name="stop"/> from the range check (defaults to both inclusive).</param>
        /// <param name="order">The direction to consider the <paramref name="start"/> and <paramref name="stop"/> in. If <see cref="Order.Ascending"/>, the <paramref name="start"/> must be smaller than the <paramref name="stop"/>. If <see cref="Order.Descending "/>, <paramref name="stop"/> must be smaller than <paramref name="start"/>.</param>
        /// <param name="skip">The number of elements into the sorted set to skip. Note: this iterates after sorting so incurs O(n) cost for large values.</param>
        /// <param name="take">The maximum number of elements to pull into the new (<paramref name="destinationKey"/>) set.</param>
        /// <param name="flags">The flags to use for this operation.</param>
```

src/StackExchange.Redis/Interfaces/IDatabase.cs
```diff
@@ -1395,6 +1395,24 @@ public interface IDatabase : IRedis, IDatabaseAsync
...
+        /// <remarks>https://redis.io/commands/zrangestore</remarks>
+        /// <returns>The cardinality of the newly created sorted set</returns>
```
NickCraver, line 1413:
```suggestion
        /// <returns>The cardinality of (number of elements in) the newly created sorted set.</returns>
```

src/StackExchange.Redis/Interfaces/IDatabaseAsync.cs
```diff
@@ -1360,6 +1360,35 @@ public interface IDatabaseAsync : IRedisAsync
...
+        /// <param name="start">The starting point in the Sorted Set, if using BYLEX this should be a string.</param>
+        /// <param name="stop">The stopping point in the range of the sorted set, if using BYLEX this should be a string.</param>
+        /// <param name="sortedSetOrder">The ordering criteria to use for the range. Choices are ByRank, ByScore, and ByLex.</param>
+        /// <param name="exclude">Whether to consider score or lex ranges exclusively or not (e.g. scores greater than 3 or greater than or equal to 3)</param>
+        /// <param name="order">The direction to consider the start and stop in, if Ascending, the start must be smaller than the stop, if Descending Stop smaller than start</param>
+        /// <param name="skip">Elements into the sorted set to skip. Should only be used </param>
+        /// <param name="take">Number of elements to pull into the new set.</param>
+        /// <param name="flags">The flags to use for this operation</param>
+        /// <remarks>https://redis.io/commands/zrangestore</remarks>
+        /// <returns>The cardinality of the newly created sorted set</returns>
```
NickCraver, line 1378: Copy from above please!

src/StackExchange.Redis/Interfaces/IDatabase.cs
```diff
@@ -1395,6 +1395,24 @@ public interface IDatabase : IRedis, IDatabaseAsync
         /// <remarks>https://redis.io/commands/zrevrange</remarks>
         RedisValue[] SortedSetRangeByRank(RedisKey key, long start = 0, long stop = -1, Order order = Order.Ascending, CommandFlags flags = CommandFlags.None);
 
+        /// <summary>
+        /// Takes the specified range of elements in the sorted set of the source key and stores
+        /// them in a new sorted set at the destination key.
```
NickCraver, line 1400:
```suggestion
        /// Takes the specified range of elements in the sorted set of the <paramref name="sourceKey"/> and stores
        /// them in a new sorted set at the <paramref name="destinationKey"/>.
```
Also: let's reverse the order of these args please - IMO source -> destination is a much more natural way to read a C# API, despite the args in Redis being in reverse of this.

src/StackExchange.Redis/KeyspaceIsolation/DatabaseWrapper.cs
```diff
@@ -576,6 +576,14 @@ public RedisValue[] SortedSetRangeByRank(RedisKey key, long start = 0, long stop
             return Inner.SortedSetRangeByRank(ToInner(key), start, stop, order, flags);
         }
 
+        public long SortedSetRangeAndStore(RedisKey destinationKey, RedisKey sourceKey, RedisValue start,
+            RedisValue stop,
+            SortedSetOrder sortedSetOrder = SortedSetOrder.ByRank, Exclude exclude = Exclude.None,
+            Order order = Order.Ascending, long skip = 0, long take = 0, CommandFlags flags = CommandFlags.None)
```
NickCraver, line 582: Global suggestions here: reverse source/dest (more in line with most APIs/.NET-isms), let's go clean with the line breaks, and make `take` default to `null` instead of `0` since that's more accurate for "unlimited" rather than using a sentinel value. Skip does makes sense as "skipping 0", though.
```suggestion
        public long SortedSetRangeAndStore(
        	RedisKey sourceKey, 
        	RedisKey destinationKey, 
        	RedisValue start, 
        	RedisValue stop,
            SortedSetOrder sortedSetOrder = SortedSetOrder.ByRank, 
            Exclude exclude = Exclude.None,
            Order order = Order.Ascending, 
            long skip = 0, 
            long? take = null, 
            CommandFlags flags = CommandFlags.None)
```

src/StackExchange.Redis/KeyspaceIsolation/WrapperBase.cs
```diff
@@ -564,6 +564,13 @@ public Task<RedisValue[]> SortedSetRangeByRankAsync(RedisKey key, long start = 0
             return Inner.SortedSetRangeByRankAsync(ToInner(key), start, stop, order, flags);
         }
 
+        public Task<long> SortedSetRangeAndStoreAsync(RedisKey destinationKey, RedisKey sourceKey, RedisValue start, RedisValue stop,
+            SortedSetOrder sortedSetOrder = SortedSetOrder.ByRank, Exclude exclude = Exclude.None,
+            Order order = Order.Ascending, long skip = 0, long take = 0, CommandFlags flags = CommandFlags.None)
```
NickCraver, line 569: (same as above on args!)

src/StackExchange.Redis/PublicAPI.Unshipped.txt
```diff
@@ -1 +1,7 @@
...
+StackExchange.Redis.IDatabaseAsync.SortedSetRangeAndStoreAsync(StackExchange.Redis.RedisKey destinationKey, StackExchange.Redis.RedisKey sourceKey, StackExchange.Redis.RedisValue start, StackExchange.Redis.RedisValue stop, StackExchange.Redis.SortedSetOrder sortedSetOrder = StackExchange.Redis.SortedSetOrder.ByRank, StackExchange.Redis.Exclude exclude = StackExchange.Redis.Exclude.None, StackExchange.Redis.Order order = StackExchange.Redis.Order.Ascending, long skip = 0, long take = 0, StackExchange.Redis.CommandFlags flags = StackExchange.Redis.CommandFlags.None) -> System.Threading.Tasks.Task<long>
+StackExchange.Redis.RedisFeatures.SortedSetRangeStore.get -> bool
+StackExchange.Redis.SortedSetOrder
+StackExchange.Redis.SortedSetOrder.ByLex = 2 -> StackExchange.Redis.SortedSetOrder
+StackExchange.Redis.SortedSetOrder.ByRank = 0 -> StackExchange.Redis.SortedSetOrder
+StackExchange.Redis.SortedSetOrder.ByScore = 1 -> StackExchange.Redis.SortedSetOrder
```
NickCraver, line 7: Let's move these to be in the .Shipped version in their respective positions please (happy to push the meta stuff like this, just don't wanna step on toes)

src/StackExchange.Redis/RedisDatabase.cs
```diff
@@ -1631,12 +1631,32 @@ public RedisValue[] SortedSetRangeByRank(RedisKey key, long start = 0, long stop
...
 
+        public long SortedSetRangeAndStore(RedisKey destinationKey, RedisKey sourceKey, RedisValue start,
+            RedisValue stop,
+            SortedSetOrder sortedSetOrder = SortedSetOrder.ByRank, Exclude exclude = Exclude.None,
+            Order order = Order.Ascending, long skip = 0, long take = 0, CommandFlags flags = CommandFlags.None)
+        {
+            var msg = CreateSortedSetRangeStoreMessage(Database, flags, destinationKey, sourceKey, start, stop,
+                sortedSetOrder, order, exclude, skip, take);
+            return ExecuteSync(msg, ResultProcessor.Int64);
+        }
```
NickCraver, line 1642:
```suggestion
        public long SortedSetRangeAndStore(RedisKey sourceKey, RedisKey destinationKey, RedisValue start, RedisValue stop,
            SortedSetOrder sortedSetOrder = SortedSetOrder.ByRank, Exclude exclude = Exclude.None,
            Order order = Order.Ascending, long skip = 0, long take = 0, CommandFlags flags = CommandFlags.None)
        {
            var msg = CreateSortedSetRangeStoreMessage(Database, flags, sourceKey, destinationKey, start, stop, sortedSetOrder, order, exclude, skip, take);
            return ExecuteSync(msg, ResultProcessor.Int64);
        }
```

src/StackExchange.Redis/RedisDatabase.cs
```diff
@@ -3980,6 +4000,83 @@ protected override RedisValue[] Parse(in RawResult result, out int count)
...
+                }
+            }
+
+            var arguments = new List<RedisValue>();
+            arguments.Add(exclude == Exclude.Both || exclude == Exclude.Start ? $"({start}" : sortedSetOrder == SortedSetOrder.ByLex ? $"[{start}" : start);
+            arguments.Add(exclude == Exclude.Both || exclude == Exclude.Stop ? $"({stop}" : sortedSetOrder == SortedSetOrder.ByLex ?  $"[{stop}" : stop);
+
+            if (sortedSetOrder != SortedSetOrder.ByRank)
+            {
+                arguments.Add(sortedSetOrder.GetLiteral());
```
NickCraver, line 4035: Instead, let's call `.GetLiteral()` and check for `.IsNull` on our `if` here - extra line, but cleaner and decoupled.

src/StackExchange.Redis/RedisDatabase.cs
```diff
@@ -3980,6 +4000,83 @@ protected override RedisValue[] Parse(in RawResult result, out int count)
             }
         }
 
+        private static SortedSetRangeAndStoreCommandMessage CreateSortedSetRangeStoreMessage(
+            int db,
+            CommandFlags flags,
+            RedisKey destinationKey,
+            RedisKey sourceKey,
+            RedisValue start,
+            RedisValue stop,
```
NickCraver, line 4009: If we're converting the `(`/`[` prefixes internally, can these be `long` instead through all overloads or am I missing a case/use?

src/StackExchange.Redis/RedisLiterals.cs
```diff
@@ -49,6 +49,8 @@ public static readonly RedisValue
             AND = "AND",
             BEFORE = "BEFORE",
             BY = "BY",
+            BYSCORE="BYSCORE",
```
NickCraver, line 52:
```suggestion
            BYSCORE = "BYSCORE",
```

src/StackExchange.Redis/RedisLiterals.cs
```diff
@@ -84,6 +86,7 @@ public static readonly RedisValue
             PING = "PING",
             PURGE = "PURGE",
             PX = "PX",
+            REV = "REV",
```
NickCraver, line 89: Alpha order on all these please :)

src/StackExchange.Redis/RedisDatabase.cs
```diff
@@ -3980,6 +4000,83 @@ protected override RedisValue[] Parse(in RawResult result, out int count)
...
+            {
+                arguments.Add(RedisLiterals.REV);
+            }
+
+            if (take > 0)
+            {
+                arguments.Add(RedisLiterals.LIMIT);
+                arguments.Add(skip);
+                arguments.Add(take);
+            }
```
NickCraver, line 4048: I'm happy to take a pass at this but basically we want to avoid the `List` and array allocations here, likely by using an existing `Message` type. We may need to add a 7th on `CommandKeyValueValueValueValueValueMessage`. If you want, I can take the efficiency part of that on and push those bits :)

NickCraver (changes requested):
Went through and tried to give a thorough review - major kudos on a plethora of tests <3 - happy to take these tweaks, hands off, or anywhere in-between - please let me know! Thanks a ton for everything so far here.

src/StackExchange.Redis/RedisDatabase.cs
slorello89: Sure thing, that's how I initially had it, I figured an enum-comparison would be slightly more efficient than the null check.

src/StackExchange.Redis/RedisDatabase.cs
slorello89: I don't mind doing this, what's the correct pattern here?

I guess it would be `CommandKeyValueValueValueValueValueValueValueValueMessage`, set any unused value to null and just not write the nulls out?

slorello89: as there's the command, 2 keys, and at most 7 literals?

src/StackExchange.Redis/KeyspaceIsolation/DatabaseWrapper.cs
slorello89: dest->source is the order of the parameters in the command, that said I've made the mistake more than once of ordering it source->dest so as long as you're ok with that inconsistency, sure thing 😂

slorello89:
@NickCraver

I went through and made the updates you requested, One thing I'd point you at is all the classes I added to `Message.cs`, basically tuplish extensions of the Message class so that we could fit all the different argument combinations for `ZRANGESTORE`, there's one that's superfluous `CommandKeyKeyValueValueValueValueValueMessage` since there's no combination of arguments for `ZRANGESTORE` that has 5 non-key arguments, but as I was filling in the ones before and after it I figured I might as well leave it (can tare it out if you prefer). 

Also not 100% sure I did the unshipped->shipped conversion correctly, I think there's a script somewhere that automagically ports them for you, but it was less effort to just alphabetize copy/paste 😂 so if that's off I'd ask you to just do the port for that bit of meta-data.

src/StackExchange.Redis/Message.cs
```diff
@@ -1138,6 +1163,219 @@ protected override void WriteImpl(PhysicalConnection physical)
...
+                physical.WriteBulkString(value2);
+                physical.WriteBulkString(value3);
+                physical.WriteBulkString(value4);
+                physical.WriteBulkString(value5);
+            }
+
+            public override int ArgCount => 8;
+        }
+
+        private sealed class CommandKeyKeyValueValueValueValueValueValueValueMessage : CommandKeyBase
```
NickCraver, line 1336: It's possible we've jumped the shark...but the garbage collector will thank us.

src/StackExchange.Redis/RedisDatabase.cs
```diff
@@ -3980,6 +4014,74 @@ protected override RedisValue[] Parse(in RawResult result, out int count)
...
+                }
+
+                if (order == Order.Ascending)
+                {
+                    return Message.Create(db, flags, RedisCommand.ZRANGESTORE, destinationKey, sourceKey, start, stop);
+                }
+                else
+                {
+                    return Message.Create(db, flags, RedisCommand.ZRANGESTORE, destinationKey, sourceKey, start, stop, RedisLiterals.REV);
+                }
```
NickCraver, line 4049: TODO: let's make this a switch expression with a `_ =>` argument exception for order. (can do tonight)

src/StackExchange.Redis/RedisDatabase.cs
```diff
@@ -3980,6 +4014,74 @@ protected override RedisValue[] Parse(in RawResult result, out int count)
...
+                {
+                    return Message.Create(db, flags, RedisCommand.ZRANGESTORE, destinationKey, sourceKey, formattedStart,
+                        formattedStop, sortedSetOrder.GetLiteral(), RedisLiterals.LIMIT, skip, take);
+                }
+                else
+                {
+                    return Message.Create(db, flags, RedisCommand.ZRANGESTORE, destinationKey, sourceKey, formattedStart,
+                        formattedStop, sortedSetOrder.GetLiteral());
+                }
+            }
```
NickCraver, line 4082: TODO: can so more switch expressions here to simplicity - will poke tonight!

NickCraver:
@slorello89 Sorry I've been slammed the last 2 days. The changes are just fantastic, thanks a ton - I'd like to push just a few style/doc tweaks tonight then merge this in :) I should be able to get to it this evening!

src/StackExchange.Redis/RedisDatabase.cs
```diff
@@ -1631,12 +1631,46 @@ public RedisValue[] SortedSetRangeByRank(RedisKey key, long start = 0, long stop
...
+            SortedSetOrder sortedSetOrder = SortedSetOrder.ByRank,
+            Exclude exclude = Exclude.None,
+            Order order = Order.Ascending,
+            long skip = 0,
+            long? take = null,
+            CommandFlags flags = CommandFlags.None)
+        {
+            var msg = CreateSortedSetRangeStoreMessage(Database, flags, sourceKey, destinationKey, start, stop,
+                sortedSetOrder, order, exclude, skip, take);
+            return await ExecuteAsync(msg, ResultProcessor.Int64);
```
NickCraver, line 1671: Oh note for future - no async/await needed here, saves a state machine :)

NickCraver (approved):
Looking awesome - thanks a ton for doing this and going all out on fitting in the existing codebase <3

PR #2061, opened 2022-04-03
Support BITFIELD and BITFIELD_RO

Avital-Fine:
Adds support to 2 redis commands:
[BITFIELD](https://redis.io/commands/bitfield/)
[BITFIELD_RO](https://redis.io/commands/bitfield_ro/)


slorello89 (changes requested):
I think there are still some open questions on the API that will affect this PR, I'd suggest changing this to WIP while we get consensus.

Issue #1520, opened 2020-07-05
Cluster slot change may lead to EXECABORT errors, app restart fixes issue


Plasma:
Hey,

Using SE.Redis v2.1.0.

We added a new shard to our redis cluster, and mostly things were uneventful, except for our backend worker machines receiving `EXECABORT Transaction discarded because of previous errors` for a certain code path.

This code path uses transactions, and was working fine with 2 shards prior. After adding the 3rd shard, these errors seemed to be constant for a code path.

A restart of our worker application immediately fixed the issue.

I'm thinking perhaps the transaction code did not detect/know to refresh the cluster slots map?

nloyolag:
Hey Plasma

I'd like to bump this issue since I experienced the exact same problem recently. One theory that a colleague cooked up is that inside the transaction we are getting MOVED errors. Transactions will wrap this error as EXECABORT and the StackExchange.Redis client will therefore be unable to understand what to do. If it received the MOVED error plainly it would perhaps just point to the new shard, but I have not tested if this repros without transactions.

NickCraver:
If this is still happening in the latest version, can you please post the full message on the errors you're seeing? This situation _should_ be handled appropriately in the latest version.

hamish-omny:
I'll answer for Andrew. We are just about to upgrade to `2.2.88` and may need to add a shard next week so will try provide an update after that

dusty9023:
@NickCraver, I wanted to add a little extra information as we're seeing this exact same problem on our side as well.

Notes: 
- We are using the latest version of the library - 2.5.43.
- We are using AWS' MemoryDB Service (Redis 6.2 Engine)

For our multiplexer we register the following for logging:
```
multiplexer.ErrorMessage += (sender, ev) => logger.Warn("ErrorMessage: " + ev.Message + " " + ev.EndPoint);
```

When resizing our shard, we start seeing the follow kinds of logs from the ErrorMessage event:
```
2022-03-31 21:41:59,262 [295] Level=WARN  ErrorMessage: EXECABORT Transaction discarded because of previous errors. Unspecified/<NODE>:6379
```


We then see a RedisServerException that is thrown from the `StackExchange.Redis.RedisTransaction.Execute`.
```
2022-03-31 21:41:59,263 [183]  ... Error: 
StackExchange.Redis.RedisServerException: EXECABORT Transaction discarded because of previous errors.
[...]

PR #2062, opened 2022-04-03
Fix for #1520: Add support for retry on transactions for resharding

dusty9023:
A proposed solution for #1520. I ask that the proposed change be thoroughly reviewed by those who are experienced with the repo, and the Redis Transaction code in the library.

**Investigation:**
In order to investigate, I took one of the many transaction errors we saw when increasing/decreasing the number of shards in our cluster, and used the associated keys + nodes to reply the the exact same transaction using the debugger.

The error for the transaction EXEC was the same as what we saw in our logs:
```
EXECABORT Transaction discarded because of previous errors.
```
image

Looking at the InnerOperations of the TransactionMessage, I was surprised to find that the ResultBox was null, and the result was already lost by the time the SetResult() for transaction message was called.
image

I eventually tracked the cause down to base Message class' Complete() method. This removed the resultbox to ensure it couldn't be called twice, but it also mean't that the result of the InnerOperations result aren't available when the transaction result is being processed.
image

This required additional changes to ensure that the result of the InnerOperations were available to the TransactionMessage when it was processing the result, but that they were also cleaned up when the TransactionMessage itself was complete.

Digging into the ResultProcessor, also showed that it RELIED on the RawResult error message following a very particular format in order to detect, and thus retry, messages where the hashslot had been moved. 

Based on these facts, the RawResult error message had to be edited to include:
- prefix the "MOVED (HASHSLOT) (endpoint) " to the entire raw result error if we determine that the ONLY reason that the transaction failed was due to a single hashslot being moved.
- Also aggregate all the InnerOperation errors/fault messages, and append them to the existing RawResult from the EXEC call.

This required changes to the TransactionMessage and QueueMessage's lifecycle (Complete() method), relying heavily on the TransactionMessage Complete() to mark the QueueMessages actually completed.

**Important Notes:**
- I attempted to write unit tests for these changes that didn't rely on a cluster, but most of the nested classes involved are private. Changing those to internal, also revealed that many of the dependencies, such as ServerEndPoint, PhysicalBridge, PhysicalConnection... are all sealed and cannot be mocked using MOQ.
- In order to test these changes, you require a REDIS cluster that has at least two primaries. One that had keys written to it originally, then the hashslot with that key is moved to the other primary.
- I've also added/updated a functional test to cover this case.
- I noticed the "src/StackExchange.Redis/PublicAPI.Unshipped.txt", and "src/StackExchange.Redis/PublicAPI.Shipped.txt", but I'm not aware of what process you follow for moving an API from being considered Unshipped to Shipped.

Due to the challenges in testing, I resorted to the following for testing:
- Taking a transaction that had failed in our pre-production cluster due to re-sharding, and then retrying it again using these changes.
- After proving the solution worked on a smaller scale, we deployed the change to our pre-production environment, and scaled our cluster (resharding) from 25 to 30 shards, while having active traffic on the cluster.

cluster | CLUSTER | April 3 2022 17:50:50 UTC | Scaling out cluster ---- from 25 shards to 30 shards
cluster | CLUSTER | April 3 2022 18:13:00 UTC | Added node n  in availability zone <AZ>
...
cluster | CLUSTE> | April 3 2022 18:13:00 UTC | Added node n+1  in availability zone <AZ>
cluster | CLUSTER | April 3 2022 18:14:39 UTC | Migrating slots from shard 0001 to 0026 to rebalance slots
...
cluster | CLUSTER| April 3 2022 18:15:40 UTC | Moved a total of 109 slots out of 109 slots from shard 0001 to shard 0026
....

During this time we saw no errors in our logs.

### DIFF
--- a/src/StackExchange.Redis/RedisTransaction.cs
+++ b/src/StackExchange.Redis/RedisTransaction.cs
@@ -58,6 +58,12 @@ namespace StackExchange.Redis
             return base.ExecuteAsync(msg, proc); // need base to avoid our local wrapping override
         }
 
+        internal bool ExecuteInternal(CommandFlags flags, ServerEndPoint endpoint = null)
+        {
+            var msg = CreateMessage(flags, out ResultProcessor<bool> proc);
+            return base.ExecuteSync(msg, proc, endpoint); // need base to avoid our local "not supported" override
+        }
+
         internal override Task<T> ExecuteAsync<T>(Message message, ResultProcessor<T> processor, ServerEndPoint server = null)
         {
             if (message == null) return CompletedTask<T>.Default(asyncState);
--- a/src/StackExchange.Redis/RedisTransaction.cs
+++ b/src/StackExchange.Redis/RedisTransaction.cs
@@ -1,5 +1,7 @@
 ﻿using System;
+using System.Buffers;
 using System.Collections.Generic;
+using System.Linq;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
@@ -144,6 +146,33 @@ namespace StackExchange.Redis
             {
                 message.SetNoRedirect();
                 Wrapped = message;
+            }
+
+            // for transactions, the inner operations should only be marked completed when the the final EXEC has been processed
+            /// <summary>
+            /// For queued messages (InnerOperations) in a the transaction, we cannot actually mark it complete until the exec
+            /// function has returned with the state of the transaction.
+            ///
+            /// Calling the base complete resets the ResultBox, 
+            /// </summary>
+            public override void Complete()
+            {
+                // still need to activate continuations for GetMessages(),
+                // which might be waiting for the last innerOperation to
+                // complete.
+                ResultBox?.ActivateContinuations();
+            }
+
+            /// <summary>
+            /// This is called when the transaction has been complete (Exec), and marks the operations as complete, updates performance,
+            /// and clears the ResultBox.
+            ///
+            /// It also triggers the ActivateContinuations() for a second time.
+            /// 
+            /// </summary>
+            public void TransactionComplete()
+            {
+                base.Complete();
             }
 
             public bool WasQueued
@@ -428,6 +457,21 @@ namespace StackExchange.Redis
                 }
                 return result;
             }
+
+            /// <summary>
+            /// When the transaction completed, the innerOperations needs to be informed that the transaction has been completed.
+            /// </summary>
+            public override void Complete()
+            {
+                // let all the inneroperations know that the transaction is complete
+                foreach (var msg in InnerOperations)
+                {
+                    msg.TransactionComplete();
+                }
+
+                // continue marking this message as complete.
+                base.Complete();
+            }
         }
 
         private class TransactionProcessor : ResultProcessor<bool>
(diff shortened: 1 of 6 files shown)


src/StackExchange.Redis/RedisTransaction.cs
```diff
@@ -56,6 +58,12 @@ public Task<bool> ExecuteAsync(CommandFlags flags)
             return base.ExecuteAsync(msg, proc); // need base to avoid our local wrapping override
         }
 
+        internal bool ExecuteInternal(CommandFlags flags, ServerEndPoint endpoint = null)
```
dusty9023, line 61: Not a huge fan of this, but I also didn't want to change the public API of this for the sake of making a functional test work.

tests/StackExchange.Redis.Tests/Cluster.cs
```diff
@@ -188,7 +188,69 @@ static string StringGet(IServer server, RedisKey key, CommandFlags flags = Comma
                     string e = StringGet(conn.GetServer(node.EndPoint), key);
                     Assert.Equal(value, e); // wrong replica, allow redirect
 
-                    var ex = Assert.Throws<RedisServerException>(() => StringGet(conn.GetServer(node.EndPoint), key, CommandFlags.NoRedirect));
+                    var ex = Assert.Throws<RedisHashslotMigratedAndNoRedirectException>(() => StringGet(conn.GetServer(node.EndPoint), key, CommandFlags.NoRedirect));
```
dusty9023, line 191: I'm not sure how much of an impact switching from RedisServerException to RedisHashslotMigratedAndNoRedirectException will have; it's a pretty specific criteria and error to cause this.

One strategy that could be used to mitigate the impact would be to switch RedisHashslotMigratedAndNoRedirectException be a subclass of RedisServerException (requires unsealing the class).

src/StackExchange.Redis/RedisTransaction.cs
```diff
@@ -146,6 +154,33 @@ public QueuedMessage(Message message) : base(message.Db, message.Flags | Command
...
+            /// function has returned with the state of the transaction.
+            ///
+            /// Calling the base complete resets the ResultBox, 
+            /// </summary>
+            public override void Complete()
+            {
+                // still need to activate continuations for GetMessages(),
+                // which might be waiting for the last innerOperation to
+                // complete.
+                ResultBox?.ActivateContinuations();
```
dusty9023, line 169: If I understood the logic behind this correct, in order for the last section of the GetMessage() logic to work, where it waits on the last result box of the innerOperations, this needs to be fired.

This ultimately means that in the lifetime of a QueuedMessage, it will fire/pulse twice. Oncee when it's marked Complete() after it completes it's call, and then again when TransactionComplete() is called when the TransactionMessage runs it's Complete(). (which calls TransactionComplete() on all its innerOperations).

NickCraver:
Hey @dusty9023 - I won't have time to get to this until tonight at the earliest for a really in-depth pass (this needs some thinking time), but wanted to say thanks a ton for a _great_ write-up here. We'll evaluate and see if this works as-is or if we need some tweaks but your investigation and notes are a tremendous help and time save here - thank you!

NickCraver:
Alrighty, trying to grok the issue/changes here (thanks a ton for descriptions and example code!). I understand what you're hitting, but would want to take a much simpler approach to solving it if possible - hope to get to it this week but depends what all else eats time (trying to get a few things in).

PR #2064, opened 2022-04-04
Support COPY

Avital-Fine:
https://redis.io/commands/copy/


src/StackExchange.Redis/Interfaces/IDatabase.cs
```diff
@@ -472,6 +472,18 @@ public interface IDatabase : IRedis, IDatabaseAsync
...
+        /// Copy key from the selected key to the specified destination key.
+        /// </summary>
+        /// <param name="source">Copy the value stored at the source.</param>
+        /// <param name="destination">Copy the value to destination in the logical database used by the connection.</param>
+        /// <param name="database">Specifying an alternative logical database index for the destination key.</param>
+        /// <param name="replace">Removes the destination key before copying the value to it.</param>
+        /// <param name="flags">The flags to use for this operation.</param>
+        /// <returns><see langword="true"/> if key was copied. <see langword="false"/> if key was not copied.</returns>
+        /// <remarks>https://redis.io/commands/copy</remarks>
+        bool KeyCopy(RedisKey source, RedisKey destination, int database = 0, bool replace = false, CommandFlags flags = CommandFlags.None);
```
mgravell, line 485: a: I would strongly recommend using name like `destinationDatabase` here, to be clear what it means
b: I don't think zero is a good default - because zero is a valid database number; I would *expect* that if I've used `GetDatabase(5)` and then call `.Copy(...)` *without* specifying a database, that we copy it to the *current* database; I see that zero is special-cased in the implementation, but that then means we *can't* intentionally copy to database zero, which is also problematic.

Suggestion: use `-1` as the default value, and interpret any negative number as  "current database"

src/StackExchange.Redis/RedisDatabase.cs
```diff
@@ -2732,6 +2744,23 @@ public Task<RedisValue> StringSetRangeAsync(RedisKey key, long offset, RedisValu
             _ => throw new ArgumentException("Expiry time must be either Utc or Local", nameof(when)),
         };
 
+        private Message GetCopyMessage(in RedisKey source, RedisKey destination, int database, bool replace, CommandFlags flags)
+        {
+            if (database < 0) throw new ArgumentOutOfRangeException(nameof(database));
+            else if (database == 0)
```
mgravell, line 2750: see previous comment about preferring negative to zero here, due to zero being a valid database number

src/StackExchange.Redis/RedisDatabase.cs
```diff
@@ -2732,6 +2744,23 @@ public Task<RedisValue> StringSetRangeAsync(RedisKey key, long offset, RedisValu
...
+            {
+                return replace?
+                    Message.Create(Database, flags, RedisCommand.COPY, source, destination):
+                    Message.Create(Database, flags, RedisCommand.COPY, source, destination, RedisLiterals.REPLACE);
+            }
+            else
+            {
+                return replace?
+                    Message.Create(Database, flags, RedisCommand.COPY, source, destination, database):
```
mgravell, line 2759: thought: if we're copying to the same database, should we assert that `source` and `destination` are different?

Avital-Fine: In my opinion, as long as we talk about clients, we don't need to check these cases.
Redis server is throwing an error when using the same key for source and destination

src/StackExchange.Redis/Interfaces/IDatabase.cs
```diff
@@ -472,6 +472,18 @@ public interface IDatabase : IRedis, IDatabaseAsync
...
+        /// Copy key from the selected key to the specified destination key.
+        /// </summary>
+        /// <param name="source">Copy the value stored at the source.</param>
+        /// <param name="destination">Copy the value to destination in the logical database used by the connection.</param>
+        /// <param name="destinationDatabase">Specifying an alternative logical database index for the destination key.</param>
+        /// <param name="replace">Removes the destination key before copying the value to it.</param>
+        /// <param name="flags">The flags to use for this operation.</param>
+        /// <returns><see langword="true"/> if key was copied. <see langword="false"/> if key was not copied.</returns>
+        /// <remarks>https://redis.io/commands/copy</remarks>
+        bool KeyCopy(RedisKey source, RedisKey destination, int destinationDatabase = -1, bool replace = false, CommandFlags flags = CommandFlags.None);
```
NickCraver, line 485: Recommend we call these `sourceKey` and `destinationKey` to be super clear. The documentation above could be between 2 servers for example (confused with migrate). Let's please explicitly say `key` there too (happy to tweaks docs and push if you want!)

src/StackExchange.Redis/Interfaces/IDatabaseAsync.cs
```diff
@@ -448,6 +448,18 @@ public interface IDatabaseAsync : IRedisAsync
...
+        /// Copy key from the selected key to the specified destination key.
+        /// </summary>
+        /// <param name="source">Copy the value stored at the source.</param>
+        /// <param name="destination">Copy the value to destination in the logical database used by the connection.</param>
+        /// <param name="destinationDatabase">Specifying an alternative logical database index for the destination key.</param>
+        /// <param name="replace">Removes the destination key before copying the value to it.</param>
+        /// <param name="flags">The flags to use for this operation.</param>
+        /// <returns><see langword="true"/> if key was copied. <see langword="false"/> if key was not copied.</returns>
+        /// <remarks>https://redis.io/commands/copy</remarks>
+        Task<bool> KeyCopyAsync(RedisKey source, RedisKey destination, int destinationDatabase = -1, bool replace = false, CommandFlags flags = CommandFlags.None);
```
NickCraver, line 461: Same as above!

src/StackExchange.Redis/RedisDatabase.cs
```diff
@@ -2732,6 +2744,23 @@ public Task<RedisValue> StringSetRangeAsync(RedisKey key, long offset, RedisValu
...
+                return replace?
+                    Message.Create(Database, flags, RedisCommand.COPY, source, destination, RedisLiterals.REPLACE):
+                    Message.Create(Database, flags, RedisCommand.COPY, source, destination);
+            }
+            else
+            {
+                return replace?
+                    Message.Create(Database, flags, RedisCommand.COPY, source, destination, database, RedisLiterals.REPLACE):
+                    Message.Create(Database, flags, RedisCommand.COPY, source, destination, database);
+            }
```
NickCraver, line 2761: Can use a `switch` expression to tidy here now that we have conditionals - see `GetStringSetMessage` as an example :)

tests/StackExchange.Redis.Tests/BasicOps.cs
```diff
@@ -431,6 +431,22 @@ public void ShouldUseSharedMuxer()
...
+                var db = muxer.GetDatabase();
+                var src = Me();
+                var dest = Me() + "2";
+                _ = db.StringSetAsync(src, "Heyyyyy");
+                var ke1 = db.KeyCopyAsync(src, dest).ForAwait();
+                var ku1 = db.StringGet(dest);
+                Assert.True(await ke1);
+                Assert.True(ku1.Equals("Heyyyyy"));
+            }
+        }
```
NickCraver, line 448: Let's please test copying to another database - see `GetDedicatedDB` to grab a database in the test suite that won't conflict (and be sure to delete the key there first to make sure we're testing).

We should also be testing the replace case and that it doesn't replace, etc.

NickCraver (changes requested):
This is looking good - thanks! There are a few things I'd like to remedy inline and test a bit better but overall great. I'm happy to make the suggested changes and push if preferred soon as time allows, let me know :)

tests/StackExchange.Redis.Tests/BasicOps.cs
```diff
@@ -431,6 +431,22 @@ public void ShouldUseSharedMuxer()
...
 
+        [Fact]
+        public async Task Copy()
+        {
+            using (var muxer = Create())
+            {
+                var db = muxer.GetDatabase();
+                var src = Me();
+                var dest = Me() + "2";
+                _ = db.StringSetAsync(src, "Heyyyyy");
```
mgravell, line 442: we should probably also have:

- a test of the cross-DB functionality
- a test of what scenario would cause it to return `false` - presumably missing source?

but other than that: looking great, thanks! Note there also a releasenotes.md - could you add to that too, following the pattern?

mgravell (changes requested):

tests/StackExchange.Redis.Tests/BasicOps.cs
Avital-Fine: Having troubles with copying to another database - I don't know why but the test skips (at least not failing)

tests/StackExchange.Redis.Tests/BasicOps.cs
Avital-Fine: Maybe someone can **help**: Having trouble with copying to another database - I don't know why but the test skips (at least not failing)

Also there is one test that returns `false` when the destination key exists and we're not using REPLACE (just added now).

Avital-Fine: Where is the releasenotes.md? Can't find it

tests/StackExchange.Redis.Tests/BasicOps.cs
NickCraver: Fixed in latest - was just missing the DB literal in the command message :)

tests/StackExchange.Redis.Tests/BasicOps.cs
NickCraver: Fixed current tests/commands - will try and add a few more tests today!

Avital-Fine: Thank you! @NickCraver

docs/ReleaseNotes.md
```diff
@@ -15,6 +15,7 @@
 - Adds: Most `ConfigurationOptions` changes after `ConnectionMultiplexer` connections will now be respected, e.g. changing a timeout will work and changing a password for auth rotation would be used at the next reconnect ([#2050 by NickCraver](https://github.com/StackExchange/StackExchange.Redis/pull/2050))
   - **Obsolete**: This change also moves `ConnectionMultiplexer.IncludeDetailInExceptions` and `ConnectionMultiplexer.IncludePerformanceCountersInExceptions` to `ConfigurationOptions`. The old properties are `[Obsolete]` proxies that work until 3.0 for compatibility.
 - Adds: Support for `ZRANGESTORE` with `.SortedSetRangeAndStore()`/`.SortedSetRangeAndStoreAsync()` ([#2052 by slorello89](https://github.com/StackExchange/StackExchange.Redis/pull/2052))
+- Adds: Support for `COPY` ([#2064 by Avital-Fine](https://github.com/StackExchange/StackExchange.Redis/pull/2064))
```
NickCraver, line 18: This should go up under the unreleased section that's empty atm, you're first in the next batch!

NickCraver (approved):
Current is looking great - thanks for this!

PR #2072, opened 2022-04-08
Tests: move to skip by version

NickCraver:
This should again make PRs more straightforward by having 1 way to do things, and doesn't leave us appending kind of random surface area to `RedisFeatures`.

Also tidies up a few `RedisFeatures` versions to be what docs reflect.


src/StackExchange.Redis/RedisFeatures.cs
```diff
@@ -54,7 +54,7 @@ public RedisFeatures(Version version)
         /// <summary>
         /// Does BITOP / BITCOUNT exist?
         /// </summary>
-        public bool BitwiseOperations => Version >= v2_5_10;
+        public bool BitwiseOperations => Version >= v2_6_0;
```
mgravell, line 57: I looked it up, and indeed: various late 2.5.something are aka 2.6 release candidates; this simplification suffices, I think

mgravell (approved):

Issue #2055, opened 2022-03-25
Missing Commands in StackExchange.Redis


slorello89:
Hi @NickCraver @mgravell , as we discussed a few weeks ago (with @chayim and @gkorland) I went through the library and compiled a list of what I believe are the missing commands of StackExchange.Redis.

## Methodology

I basically diffed [RedisCommand.cs](https://github.com/StackExchange/StackExchange.Redis/blob/main/src/StackExchange.Redis/Enums/RedisCommand.cs) with the main Redis Project's [commands.json](https://github.com/redis/redis-doc/blob/master/commands.json), which resulted in ~180 commands, I went through and purged most of the commands that felt like they were admin/config (except ACL) to preserve the more interactive commands, this left ~70. Then I went through the library to see if the sub-commands that came up (a lot of stuff from XGROUP XINFO and a few others) were already covered by a combination of the command Enum + a literal to filter those out as they are already supported. After that, and collapsing the many sub-commands of `FUNCTION` and `ACL`, we're left with 47. I think what follows is mostly correct (in that the thing is ACTUALLY missing, and that it's something that you would probably want to add to the library if it were feasible). I marked the blocking commands, and then applied a bit of my own reason to determine if I thought the command would be addable without a major overhaul of the library. I also added the Redis Version the command was added in (about 40% of the missing commands are 7.0, I don't know what your policy is on adding Release candidate commands, I'm guessing you wouldn't want to :) )

As it stands, there are about 21 commands which are in stable versions of Redis, interactive, and shouldn't require any major architectural changes, call them low-hanging fruit. Seems like there are a few commands that were either just missed, or there's some story behind why they never made it in (`BITFIELD`, some `OBJECT` subcommand, `SCRIPT KILL`, `LOLWUT` (ok maybe a bit silly, but can always use a bit of levity)). And then I'm guessing there hasn't been a command refresh since 6.0 

This is meant as a starting point, for my part, I'm happy to start opening PRs for any commands you'd approve of adding. I understand the reviews can be burdensome and do not want to overwhelm you, so just let me know what you want to be added and I'm happy to get started.

## Low hanging fruit

These commands should be relatively easy to add, and with the exception of maybe `SCRIPT KILL` I don't think they'd be that controversial

| Command name    | Redis Version | feasible addition without a major overhaul?                  |
| --------------- | ------------- | ---------------------------------------------------------- |
| BITFIELD        | 3.2           | yes                                                        |
[...]

PR #2075, opened 2022-04-10
Sorted set new commands

Avital-Fine:
Covers: 

ZDIFF 
ZDIFFSTORE 
ZINTER
ZINTERCARD
ZUNION

#2055


tests/StackExchange.Redis.Tests/SortedSets.cs
```diff
@@ -52,6 +61,153 @@ public class SortedSets : TestBase
...
+                Assert.Throws<StackExchange.Redis.RedisServerException>(() => db.SortedSetCombineAndStore(SetOperation.Union, destination, new RedisKey[]{ key1, key2 }, new double[]{ 1, 2, 3 }));
+            }
+        }
+
+        [Fact]
+        public void SortedSetIntersectionLength()
+        {
+            using (var conn = Create())
+            {
+                Skip.IfBelow(conn, RedisFeatures.v6_9_240);
```
NickCraver, line 191: If you merge main in, this was changed to a 7 RC version so we don't have to visit these after adding :)

tests/StackExchange.Redis.Tests/SortedSets.cs
```diff
@@ -52,6 +61,153 @@ public class SortedSets : TestBase
...
+
+                var union = db.SortedSetCombineAndStore(SetOperation.Union, destination, new RedisKey[]{ key1, key2});
+                Assert.Equal(10, union);
+            }
+        }
+
+        [Fact]
+        public void SortedSetCombineErrors()
+        {
+            using (var conn = Create())
```
NickCraver, line 156: On all new tests that are full scope, let's prefer the new `using` format please! (I adjusted other PRs, happy to push here too lemme know)

src/StackExchange.Redis/RedisDatabase.cs
```diff
@@ -3154,6 +3225,23 @@ private Message GetSortedSetLengthMessage(RedisKey key, double min, double max,
             return Message.Create(Database, flags, RedisCommand.ZCOUNT, key, from, to);
         }
 
+        private Message GetSortedSetIntersectionLengthMessage(RedisKey[] keys, long limit, CommandFlags flags)
+        {
+            if (keys == null) throw new ArgumentNullException(nameof(keys));
+
+            List<RedisValue>? values = new List<RedisValue>();
```
NickCraver, line 3232: No need for nullable here, and we can avoid `List<T>` resizes. For example:
```cs
var values = new List<RedisValue>(1 + keys.Length + (limit > 0 ? 2 : 0));
```
This will construct a list of the correct size from the onset and avoid copies with more than a few keys in play.

src/StackExchange.Redis/RedisDatabase.cs
```diff
@@ -3116,32 +3152,67 @@ private Message GetSortedSetCombineAndStoreCommandMessage(SetOperation operation
...
+                SetOperation.Difference => RedisCommand.ZDIFFSTORE,
                 _ => throw new ArgumentOutOfRangeException(nameof(operation)),
             };
             if (keys == null) throw new ArgumentNullException(nameof(keys));
+            // Difference operation can't have weights or aggregation
+            if (operation == SetOperation.Difference && (weights != null || aggregate != Aggregate.Sum))
+                throw new ArgumentException("ZDIFFSTORE cannot be used with weights or with aggregation.");
 
-            List<RedisValue>? values = null;
+            List<RedisValue>? values = new List<RedisValue>();
```
NickCraver, line 3163: We shouldn't new up the list here if we're not going to use it - should be the previous approach. The ref arg should be nullable co compensate so the reference is set coming back, though.

src/StackExchange.Redis/RedisDatabase.cs
```diff
@@ -3116,32 +3152,67 @@ private Message GetSortedSetCombineAndStoreCommandMessage(SetOperation operation
...
+                SetOperation.Union => RedisCommand.ZUNION,
+                SetOperation.Difference => RedisCommand.ZDIFF,
+                _ => throw new ArgumentOutOfRangeException(nameof(operation)),
+            };
+            if (keys == null) throw new ArgumentNullException(nameof(keys));
+            // Difference operation can't have weights or aggregation
+            if (operation == SetOperation.Difference && (weights != null || aggregate != Aggregate.Sum))
+                throw new ArgumentException("ZDIFF cannot be used with weights or with aggregation.");
+
+            List<RedisValue>? values = new List<RedisValue>();
```
NickCraver, line 3182: See other comments - we should initialize this list with a size and it need not be nullable :) van simplify to `var values` here. In fact, instead of a list let's just create the array directly here and copy in `for` loop style. Again, happy to push changes intended but don't want to step on toes and goal is to get on the same page!

src/StackExchange.Redis/RedisDatabase.cs
```diff
@@ -3116,32 +3152,67 @@ private Message GetSortedSetCombineAndStoreCommandMessage(SetOperation operation
...
+            {
+                SetOperation.Intersect => RedisCommand.ZINTER,
+                SetOperation.Union => RedisCommand.ZUNION,
+                SetOperation.Difference => RedisCommand.ZDIFF,
+                _ => throw new ArgumentOutOfRangeException(nameof(operation)),
+            };
+            if (keys == null) throw new ArgumentNullException(nameof(keys));
+            // Difference operation can't have weights or aggregation
+            if (operation == SetOperation.Difference && (weights != null || aggregate != Aggregate.Sum))
+                throw new ArgumentException("ZDIFF cannot be used with weights or with aggregation.");
```
NickCraver, line 3180: FWIW, this message is confusing to me because `Sum` _is_ one of the aggregations - not 100% sure on intent but happy to suggest better messaging once I read up on what this constraint is doing.

src/StackExchange.Redis/RedisDatabase.cs
```diff
@@ -3116,32 +3152,67 @@ private Message GetSortedSetCombineAndStoreCommandMessage(SetOperation operation
...
-                    (values ??= new List<RedisValue>()).Add(RedisLiterals.AGGREGATE);
-                    values.Add(RedisLiterals.MIN);
+                    (args ??= new List<RedisValue>()).Add(RedisLiterals.AGGREGATE);
+                    args.Add(RedisLiterals.MIN);
                     break;
                 case Aggregate.Max:
-                    (values ??= new List<RedisValue>()).Add(RedisLiterals.AGGREGATE);
-                    values.Add(RedisLiterals.MAX);
+                    (args ??= new List<RedisValue>()).Add(RedisLiterals.AGGREGATE);
+                    args.Add(RedisLiterals.MAX);
```
NickCraver, line 3211: Let's leave this as `values` please - less diff and the meaning is "RedisValue" by the word so was correct.

src/StackExchange.Redis/RedisDatabase.cs
NickCraver: Okay I get it, I propose instead of throwing here, we just say in docs that aggregation argument is ignored with the difference case and remove that portion of the check in methods - thoughts? The alternative I see is adding a `Aggregate.None`, which would have to be explicitly passed, seems simpler to just point out we ignore it.

src/StackExchange.Redis/RedisDatabase.cs
```diff
@@ -1580,6 +1580,30 @@ public Task<long> SortedSetAddAsync(RedisKey key, SortedSetEntry[] values, When
             return ExecuteAsync(msg, ResultProcessor.Int64);
         }
 
+        public RedisValue[] SortedSetCombine(SetOperation operation, RedisKey[] keys, double[]? weights = null, Aggregate aggregate = Aggregate.Sum, CommandFlags flags = CommandFlags.None)
+        {
+            var msg = GetSortedSetCombineCommandMessage(operation, keys, weights, aggregate, false, flags);
```
NickCraver, line 1585: On these, let's make the code a bit more readable for consumers by labelling the `false` arg please:
```suggestion
            var msg = GetSortedSetCombineCommandMessage(operation, keys, weights, aggregate, withScores: false, flags);
```

src/StackExchange.Redis/RedisDatabase.cs
NickCraver: One more: if I understand correctly, should we check that weights has the same length as keys?

NickCraver (dismissed):
Great work here! Added some comments inline - as always happy to adjust and push or just keep commenting, whichever is a bigger time/sync save on your end. A few questions inline for thoughts!

tests/StackExchange.Redis.Tests/SortedSets.cs
Avital-Fine: done :) thank you

src/StackExchange.Redis/RedisDatabase.cs
Avital-Fine: In general, I support using redis exceptions. I mean that if redis already check the size of the weights and the keys, I don't think we should check it too.

Also, I removed the current exception, so now we get `StackExchange.Redis.RedisServerException` instead of `ArgumentException` which I think is ok :woman-shrugging::skin-tone-3:

Let me know what you think :)

Avital-Fine: Now I'm thinking that if we do check the lengths, we can easily decide the length of the list so maybe it's better in this case :sweat_smile:

src/StackExchange.Redis/RedisDatabase.cs
Avital-Fine: Yea maybe I need your help with this part :sweat_smile: 
I am confused because I want to use list to reuse the method that added the weights and the aggregation.

src/StackExchange.Redis/RedisDatabase.cs
NickCraver: If the error from the server makes it clear on how to fix, in total agreement - will take a peek when time allows today!

NickCraver: Alrighty I poked at this locally, if Redis was more informative here I'd agree, but the error message users gets is "ERR Syntax error" with no details - so I opted to add some error checking here in latest to make it much easier to figure out what's been done wrong.

src/StackExchange.Redis/RedisDatabase.cs
NickCraver: Took another approach here - no longer relevant! Avoids the duplicate allocations to only the array (if any) we're going to ref anyhow,

NickCraver:
@slorello89 I've touched this a lot, so would appreciate a third set of eyes if you get a chance!

src/StackExchange.Redis/RedisDatabase.cs
```diff
@@ -3172,32 +3208,99 @@ private Message GetSortedSetCombineAndStoreCommandMessage(SetOperation operation
...
-                (values ??= new List<RedisValue>()).Add(RedisLiterals.WEIGHTS);
+                values = new RedisValue[argsLength];
+                AddWeightsAggregationAndScore(values, weights, aggregate);
+            }
+            return new SortedSetCombineAndStoreCommandMessage(Database, flags, command, destination, keys, values);
+        }
+
+        private Message GetSortedSetCombineCommandMessage(SetOperation operation, RedisKey[] keys, double[]? weights, Aggregate aggregate, bool withScores, CommandFlags flags)
+        {
+            var command = operation switch
```
slorello89, line 3240: Probably want to move this logic to the `SetOperation` enum as an extension method.

src/StackExchange.Redis/RedisDatabase.cs
```diff
@@ -3172,32 +3208,99 @@ private Message GetSortedSetCombineAndStoreCommandMessage(SetOperation operation
...
+            {
+                SetOperation.Intersect => RedisCommand.ZINTER,
+                SetOperation.Union => RedisCommand.ZUNION,
+                SetOperation.Difference =>
```
slorello89, line 3244: do this check after you've gotten the set operation.

src/StackExchange.Redis/RedisDatabase.cs
```diff
@@ -3172,32 +3208,99 @@ private Message GetSortedSetCombineAndStoreCommandMessage(SetOperation operation
...
+            {
+                throw new ArgumentNullException(nameof(keys));
+            }
+            if (weights != null && keys.Length != weights.Length)
+            {
+                throw new ArgumentException("Keys and weights should have the same number of elements.", nameof(weights));
+            }
+
+            var i = 0;
+            var values = new RedisValue[1 + keys.Length +
```
slorello89, line 3261: How much more expensive would it be to initialize a List with this capacity and using `Add`/`AddRange` as opposed to an array and then having to track `i`, this is a bit. . . feels like we're reinventing the wheel here? `List<T>(int)` is effectively just doing the same initialization, since it's initialized with the correct capacity it shouldn't over allocate nor have to dynamically reallocate.

slorello89 (commented):
@NickCraver just have some questions about the efficiency vs maintainability bits of the Message Creation.

src/StackExchange.Redis/RedisDatabase.cs
slorello89: Ahh I gotcha. [TIL](https://github.com/dotnet/coreclr/blob/master/src/System.Private.CoreLib/shared/System/Collections/Generic/List.cs#L1028)

src/StackExchange.Redis/RedisDatabase.cs
NickCraver: :+1: This makes sense with an arg - done!

PR #2076, opened 2022-04-10
Support ZRANDMEMBER

Avital-Fine:
https://redis-stack.io/commands/zrandmember/

#2055


NickCraver (changes requested):
I started making suggested changes to plural naming (to match existing methods) for the array cases but after many comments I decided pushing the changes up would be decidedly less annoying :)

This looks great - I tweaked docs and methods, only thing missing I think is testing on the singular methods.

Avital-Fine:
> missing key cases

Thank you for the help with the docs!
I added tests for missing key cases but didn't understand what did you mean by singular methods :sweat_smile:

NickCraver (approved):
Looking good, thanks!!

src/StackExchange.Redis/Interfaces/IDatabase.cs
```diff
@@ -1419,6 +1419,53 @@ public interface IDatabase : IRedis, IDatabaseAsync
...
         long SortedSetLengthByValue(RedisKey key, RedisValue min, RedisValue max, Exclude exclude = Exclude.None, CommandFlags flags = CommandFlags.None);
 
+        /// <summary>
+        /// Returns a random element from the sorted set value stored at <paramref name="key"/>.
+        /// </summary>
+        /// <param name="key">The key of the sorted set.</param>
+        /// <param name="flags">The flags to use for this operation.</param>
+        /// <returns>The randomly selected element, or <see cref="RedisValue.Null"/> when <paramref name="key"/> does not exist.</returns>
+        /// <remarks>https://redis.io/commands/zrandmember</remarks>
+        RedisValue SortedSetRandomMember(RedisKey key, CommandFlags flags = CommandFlags.None);
```
slorello89, line 1429: Noodling on this, probably want to have a `SortedSetRandomMemberWithScore` method? Where you just pass in `COUNT` 1 and the `WITHSCORES` argument to yield a single SortedSetResult.

slorello89 (commented):
1 comment about the API - looks like it's missing 1 pair of methods.

src/StackExchange.Redis/Interfaces/IDatabase.cs
NickCraver: I was thinking: it's rare enough I'm okay leaving it off - if there was ever a complaint, we could add it later.

NickCraver: @slorello89 If good with the above, I'll go ahead and get this in - thoughts?

slorello89: 👍

slorello89 (approved):
LGTM

PR #2077, opened 2022-04-10
Support SMISMEMBER

Avital-Fine:
https://redis-stack.io/commands/smismember/

#2055


src/StackExchange.Redis/Interfaces/IDatabase.cs
```diff
@@ -1102,6 +1102,19 @@ public interface IDatabase : IRedis, IDatabaseAsync
...
+        /// <summary>
+        /// Returns whether each member is a member of the set stored at key.
+        /// </summary>
+        /// <param name="key">The key of the set.</param>
+        /// <param name="values">The members to check for.</param>
+        /// <param name="flags">The flags to use for this operation.</param>
+        /// <returns>
+        /// <see langword="true"/> if the element is a member of the set.
+        /// <see langword="false"/> if the element is not a member of the set, or if key does not exist.
+        /// </returns>
```
NickCraver, line 1114: In this case doc should say an _array_ of these true/false, matching position, e.g. 
```suggestion
        /// <returns>
        /// An array of booleans corresponding to <paramref name="values"/>, for each:
        /// <see langword="true"/> if the element is a member of the set.
        /// <see langword="false"/> if the element is not a member of the set, or if key does not exist.
        /// </returns>
```

What happens when the key isn't present, all falses?

tests/StackExchange.Redis.Tests/Sets.cs
```diff
@@ -38,6 +38,31 @@ public void SScan()
             }
         }
 
+        [Fact]
+        public void SetContains()
+        {
+            using (var conn = Create())
```
NickCraver, line 44: New `using` style on tests please! I'll add this to `.editorconfig` as a stronger suggestion so it's automatic.

tests/StackExchange.Redis.Tests/Sets.cs
```diff
@@ -38,6 +38,31 @@ public void SScan()
...
+                }
+
+                // Single member
+                var isMemeber = db.SetContains(key, 1);
+                Assert.True(isMemeber);
+
+                // Multi members
+                var areMemebers = db.SetContains(key, new RedisValue[] {1, 2, 3});
+                Assert.Equal(3, areMemebers.Length);
+                Assert.True(areMemebers[1]);
```
NickCraver, line 62: We should test that `false` works here too :)

src/StackExchange.Redis/ResultProcessor.cs
```diff
@@ -1262,6 +1265,22 @@ protected override bool SetResultCore(PhysicalConnection connection, Message mes
             }
         }
 
+        private sealed class BooleanArrayProcessor : ResultProcessor<Boolean[]>
```
NickCraver, line 1268: Use C# types please:
```suggestion
        private sealed class BooleanArrayProcessor : ResultProcessor<bool[]>
```

src/StackExchange.Redis/ResultProcessor.cs
```diff
@@ -1262,6 +1265,22 @@ protected override bool SetResultCore(PhysicalConnection connection, Message mes
...
+        {
+            protected override bool SetResultCore(PhysicalConnection connection, Message message, in RawResult result)
+            {
+                switch (result.Type)
+                {
+                    case ResultType.MultiBulk:
+                        var arr = result.GetItemsAsBooleans()!;
```
NickCraver, line 1275: This assertion is blind, so a null result would go boom - be sure to add a check, e.g.:
```suggestion
                    case ResultType.MultiBulk when !result.IsNull:
                        var arr = result.GetItemsAsBooleans()!;
```

But I'm not sure this is correct. We should consider the case the key is missing, and add tests for that in a few of these PRs to ensure we get the expected result.

tests/StackExchange.Redis.Tests/Sets.cs
NickCraver: These also need some v6_2_0 checking love!

NickCraver (changes requested):
Spotted a few things cross-PR in reviewing these, notes inline! Thanks again, these PRs are fantastic and you're cranking 'em out <3

tests/StackExchange.Redis.Tests/Sets.cs
Avital-Fine: Is that the new style that you meant?

Avital-Fine: If it's right I can fix the style in the other tests in this file :)

Avital-Fine: fixed :)

src/StackExchange.Redis/Interfaces/IDatabase.cs
Avital-Fine: yes

tests/StackExchange.Redis.Tests/Sets.cs
```diff
@@ -14,234 +14,244 @@ public class Sets : TestBase
...
+            var key = Me();
 
-                db.KeyDelete(key, CommandFlags.FireAndForget);
+            db.KeyDelete(key, CommandFlags.FireAndForget);
 
-                var arr = db.SetPop(key, 1);
-                Assert.Empty(arr);
-            }
+            var arr = db.SetPop(key, 1);
+            Assert.Empty(arr);
```
NickCraver, line 254: Let's please only use the new syntax in the PRs (doing as little diff as possible) - I'll do a separate cleanup pass that's much cleaner overall git history. If we do these mass cleanups mixed in other PRs, history is very confusing later on.

NickCraver (approved):
Looking good - thank you for this!

NickCraver:
Pushed one last tidy on the tests, we should nuke the key when starting to ensure a consistent test run - after that clears will merge in.

PR #2082, opened 2022-04-12
Support the ZMSCORE command.

ttingen:
[https://redis.io/commands/zmscore/](https://redis.io/commands/zmscore/)

#2055


src/StackExchange.Redis/ResultProcessor.cs
```diff
@@ -1083,6 +1087,24 @@ protected override bool SetResultCore(PhysicalConnection connection, Message mes
...
+        private sealed class NullableDoubleArrayProcessor : ResultProcessor<double?[]>
+        {
+            protected override bool SetResultCore(PhysicalConnection connection, Message message, in RawResult result)
+            {
+                switch (result.Type)
+                {
+                    case ResultType.Integer:
+                    case ResultType.SimpleString:
+                    case ResultType.BulkString:
+                    case ResultType.MultiBulk:
```
NickCraver, line 1099: I believe only `MultiBulk` is allowed here (per https://redis.io/commands/zmscore/), which means we can simplify to `if (result.Type == ResultType.MultiBulk && !result.IsNull)` (though I don't _think_ the `IsNull` case can happen, let's be safe). If we're asserting the result of `GetItemsAsDoubles` is not null, we should ensure that either way (with the check).

tests/StackExchange.Redis.Tests/SortedSets.cs
```diff
@@ -639,5 +639,95 @@ public void SortedSetRangeStoreFailExclude()
...
+        public void SortedSetScoreSingle()
+        {
+            using var conn = Create();
+            Skip.IfBelow(conn, RedisFeatures.v2_1_0);
+
+            var db = conn.GetDatabase();
+            var me = Me();
+
+            var setKey = $"{me}:ZScoreSet";
+            var memberName = $"{me}:ZScoreMember";
```
NickCraver, line 653: Overall requests on these tests:
- Only the set key needs to be unique, no need to get fancier or more expensive on the members.
- We should delete the set key(s) before starting, to ensure a good test
- Should check the "set doesn't exist" (random key) case
- Also need to test the Async versions :)

NickCraver (changes requested):
I pushed up doc tweaks for ease there, just a few requests on the processor and tests (happy to push to branch if easier and you're slammed). Thanks for this!!

tests/StackExchange.Redis.Tests/SortedSets.cs
ttingen: Resolved.

src/StackExchange.Redis/ResultProcessor.cs
ttingen: Updated with the simplified check.

NickCraver (approved):
@ttingen I think current is good - thank you! I'd appreciate @slorello89 eyes here when time allows.

NickCraver:
@slorello89 @Avital-Fine Can one of y'all provide secondary eyes here please?

slorello89 (approved):
Somehow missed the request @NickCraver sorry about that! LGTM 👍

PR #2083, opened 2022-04-12
Support EXPIRETIME and PEXPIRETIME

Avital-Fine:
Since the method returns TimeSpan and not a number like the redis server (similar to OBJECT IDLETIME), PEXPIRETIME has no difference from EXPIRETIME.

https://redis.io/commands/expiretime/
https://redis.io/commands/pexpiretime/

#2055


src/StackExchange.Redis/Interfaces/IDatabase.cs
```diff
@@ -552,11 +553,14 @@ public interface IDatabase : IRedis, IDatabaseAsync
...
         /// </para>
+        /// <para>
+        /// Since Redis 7.0.0, you can choose on which condition the timeout will be set using <see cref="ExpiryOption"/>. Notice that this option cannot be used with PERSIST command.
+        /// </para>
         /// </remarks>
         /// <remarks>https://redis.io/commands/expire</remarks>
         /// <remarks>https://redis.io/commands/pexpire</remarks>
         /// <remarks>https://redis.io/commands/persist</remarks>
-        bool KeyExpire(RedisKey key, TimeSpan? expiry, CommandFlags flags = CommandFlags.None);
+        bool KeyExpire(RedisKey key, TimeSpan? expiry, CommandFlags flags = CommandFlags.None, ExpiryOption? expiryOption = null);
```
NickCraver, line 563: This isn't a change we can make - when adding a parameter like this it'll break all current consumers. We'll need to add a new overload and think about how this API looks. Think of it this way: any consumer has a binary `.KeyExpire(RedisKey, TimeSpan?, CommandFlags)` method it's looking for - if we do this, it removes that method and they get a `MissingMethodException` at runtime.

Avital-Fine: Then I can add 
`bool KeyExpire(RedisKey key, TimeSpan? expiry, ExpiryOption expiryOption, CommandFlags flags = CommandFlags.None);`
?

Avital-Fine: @NickCraver

Avital-Fine: If it helps, you can take a look at Jedis' PR (the Java client): https://github.com/redis/jedis/pull/2852/files

NickCraver: Gotcha - I'll try and  grok it a little further along (have some other TODOs this AM) - may suggest naming tweaks to clarify!

NickCraver: Sounds great! Apologies I'm a bit slow - y'all are rocking it, hard to keep up with :)

NickCraver (changes requested):
I'm trying to grok the TTL bits and intent but tried to add initial thoughts on everything - this being an overload change is a bit more to consider!

src/StackExchange.Redis/Enums/ExpiryOption.cs
```diff
@@ -0,0 +1,39 @@
+﻿using System;
+
+namespace StackExchange.Redis
+{
+    /// <summary>
+    /// Specifies what side of the list to refer to.
+    /// </summary>
+    public enum ExpiryOption
```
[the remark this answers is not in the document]
Avital-Fine: Yea we can do that, it will be equivalent to the older version\signature of this command

src/StackExchange.Redis/RedisDatabase.cs
```diff
@@ -721,6 +721,42 @@ public Task<bool> KeyExpireAsync(RedisKey key, DateTime? expiry, CommandFlags fl
             return ExecuteAsync(msg, ResultProcessor.Boolean, server: server);
         }
 
+        public bool KeyExpire(RedisKey key, TimeSpan expiryTime, ExpiryOption expiryOption, CommandFlags flags = CommandFlags.None)
+        {
+            var msg = GetExpiryMessage(key, flags, expiryTime, out ServerEndPoint? server, expiryOption);
+            return ExecuteSync(msg, ResultProcessor.Boolean, server: server);
+        }
```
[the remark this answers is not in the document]
Avital-Fine: Agree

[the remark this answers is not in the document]
Avital-Fine: @NickCraver One small problem here that I'm not sure how to fix:
I wanted to make the TimeSpan and DateTime not nullable in these new signatures to avoid using PERSIST, but then I can't call the new methods from the existing methods.
In this current implementation, the user can use `null` as the `expiry` parameter and `ExpiryWhen.Always` as `when` and get PERSIST command which is a bit weird.

Do you think it's ok? Otherwise, we can make it not nullable again.

src/StackExchange.Redis/Interfaces/IDatabase.cs
```diff
@@ -582,6 +608,42 @@ public interface IDatabase : IRedis, IDatabaseAsync
...
+
+        /// <summary>
+        /// Returns the absolute Unix timestamp (since January 1, 1970) in seconds at which the given key will expire.
+        /// </summary>
+        /// <param name="key">The key to get the expiration from.</param>
+        /// <param name="flags">The flags to use for this operation.</param>
+        /// <returns>The time at which the given key will expire, or <see langword="Null"/> if the key does not exist or the key exists but has no associated expiration time.</returns>
+        /// <remarks>https://redis.io/commands/expiretime</remarks>
+        /// <remarks>https://redis.io/commands/pexpiretime</remarks>
+        TimeSpan? KeyExpireTime(RedisKey key, CommandFlags flags = CommandFlags.None);
```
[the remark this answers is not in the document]
Avital-Fine: Most of the clients return long (the number of seconds or milliseconds for PEXPIRETIME), but after I saw the implementation of `KeyIdleTime` I thought I will be more consistent to return TimeSpan.
You're right that now it's similar to `KeyTimeToLive` :sweat_smile:

We can change it to long and then get the absolute Unix timestamp as the redis server does.

NickCraver:
@Avital-Fine need a break for staring at this and a few doc tweaks, but think it's close with `DateTime?` as the "expected in .NET"-ism for the client here. I'll try and get back to PRs this afternoon (not the day job!)

src/StackExchange.Redis/RedisDatabase.cs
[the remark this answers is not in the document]
NickCraver: I tweaked the approach here in latest - think current is good to go :+1:

NickCraver (approved):
@slorello89 I think this is good but I've also been staring at and tweaking it a while, would appreciate an external set of eyes when time allows <3

slorello89 (approved):
LGTM 👍

PR #2085, opened 2022-04-13
Fix hashslot calculation for XACK, XCLAIM and XPENDING

nielsderdaele:
The current implementation of XACK, XCLAIM and XPENDING uses a ComandValuesMessage. Due to this the GetHashSlot() method always returns -1. 

Using these commands in a redis cluster environment will result in MOVED responses as the commands are send to incorrect endpoints.


nielsderdaele:
The commands XREADGROUP, XREAD, XGROUP and XINFO have the same problem. I have created issue #2086 for these commands as they couldn't be fixed by using an already existing MessageCommand.

Avital-Fine (approved):

PR #2088, opened 2022-04-13
Support OBJECT ENCODING

Avital-Fine:
https://redis.io/commands/object-encoding/

#2055


src/StackExchange.Redis/Interfaces/IDatabase.cs
```diff
@@ -516,6 +516,15 @@ public interface IDatabase : IRedis, IDatabaseAsync
...
         byte[]? KeyDump(RedisKey key, CommandFlags flags = CommandFlags.None);
 
+        /// <summary>
+        /// Returns the internal encoding for the Redis object stored at <paramref name="key"/>.
+        /// </summary>
+        /// <param name="key">The key to dump.</param>
+        /// <param name="flags">The flags to use for this operation.</param>
+        /// <returns>The serialized value.</returns>
+        /// <remarks>https://redis.io/commands/object-encoding</remarks>
+        RedisValue KeyEncoding(RedisKey key, CommandFlags flags = CommandFlags.None);
```
NickCraver, line 526: I think this should be a `string?` based on the API - it'll only ever be a bulk string or null right?

src/StackExchange.Redis/Interfaces/IDatabaseAsync.cs
```diff
@@ -492,6 +492,15 @@ public interface IDatabaseAsync : IRedisAsync
...
         Task<byte[]?> KeyDumpAsync(RedisKey key, CommandFlags flags = CommandFlags.None);
 
+        /// <summary>
+        /// Returns the internal encoding for the Redis object stored at <paramref name="key"/>.
+        /// </summary>
+        /// <param name="key">The key to dump.</param>
+        /// <param name="flags">The flags to use for this operation.</param>
+        /// <returns>The serialized value.</returns>
+        /// <remarks>https://redis.io/commands/object-encoding</remarks>
+        Task<RedisValue> KeyEncodingAsync(RedisKey key, CommandFlags flags = CommandFlags.None);
```
NickCraver, line 502: Same `string?` throughout

tests/StackExchange.Redis.Tests/Keys.cs
```diff
@@ -264,5 +264,25 @@ public async Task TouchIdleTimeAsync()
...
+            db.StringSet(key, "new value", flags: CommandFlags.FireAndForget);
+
+            Assert.Equal("embstr", db.KeyEncoding(key));
+            Assert.Equal("embstr", await db.KeyEncodingAsync(key));
+
+            db.KeyDelete(key, CommandFlags.FireAndForget);
+            db.ListLeftPush(key, "new value", flags: CommandFlags.FireAndForget);
+
+            Assert.Equal("quicklist", db.KeyEncoding(key));
+            Assert.Equal("quicklist", await db.KeyEncodingAsync(key));
```
NickCraver, line 285: This will vary by server version as to when it chooses to ziplist, recommend using something like
```suggestion
            // Depending on server version, this is going to vary - we're sanity checking here.
            var listTypes = new [] { "ziplist", "quicklist" };
            Assert.Contains(db.KeyEncoding(key), listTypes);
            Assert.Contains(await db.KeyEncodingAsync(key), listTypes);
```
(note: this assumes string changes above)

src/StackExchange.Redis/Interfaces/IDatabase.cs
```diff
@@ -516,6 +516,15 @@ public interface IDatabase : IRedis, IDatabaseAsync
         /// <remarks>https://redis.io/commands/dump</remarks>
         byte[]? KeyDump(RedisKey key, CommandFlags flags = CommandFlags.None);
 
+        /// <summary>
+        /// Returns the internal encoding for the Redis object stored at <paramref name="key"/>.
+        /// </summary>
+        /// <param name="key">The key to dump.</param>
+        /// <param name="flags">The flags to use for this operation.</param>
+        /// <returns>The serialized value.</returns>
```
NickCraver, line 524:
```suggestion
        /// <returns>The Redis encoding for the value or <see langword="null"/> is the key does not exist.</returns>
```

src/StackExchange.Redis/Interfaces/IDatabaseAsync.cs
```diff
@@ -492,6 +492,15 @@ public interface IDatabaseAsync : IRedisAsync
         /// <remarks>https://redis.io/commands/dump</remarks>
         Task<byte[]?> KeyDumpAsync(RedisKey key, CommandFlags flags = CommandFlags.None);
 
+        /// <summary>
+        /// Returns the internal encoding for the Redis object stored at <paramref name="key"/>.
+        /// </summary>
+        /// <param name="key">The key to dump.</param>
+        /// <param name="flags">The flags to use for this operation.</param>
+        /// <returns>The serialized value.</returns>
```
NickCraver, line 500:
```suggestion
        /// <returns>The Redis encoding for the value or <see langword="null"/> is the key does not exist.</returns>
```

NickCraver (changes requested):
This looks pretty good - main thing is return type should be `string?` on these APIs

src/StackExchange.Redis/Interfaces/IDatabase.cs
Avital-Fine: yup

NickCraver (approved):
Looking good! Thanks :)

Issue #2071, opened 2022-04-07
Breaking changes for `StringSet` from 2.5.43 -> 2.5.61


WeihanLi:
The following code breaks when upgrade 2.5.43 -> 2.5.61

``` c#
database.StringSet(_keyName, Base, _expiresIn, When.NotExists);
```

Error message:
```
Argument 4: cannot conv
ert from 'StackExchange.Redis.When' to 'bool'
```

Seemed it's introduced from https://github.com/StackExchange/StackExchange.Redis/pull/2029

mgravell:
Hmmm; indeed, that will be a code-breaking (but thankfully not binary-breaking) change - we must have overlooked that one, sorry. I think we should change `CommandFlags flags` to `CommandFlags flags = CommandFlags.None`, which would allow this previous usage to continue working. Thoughts?

WeihanLi:
Hi @mgravell , thanks for your quick response.

I think maybe it's more simple to adjust the order of the parameters,

For example:

``` diff
- bool StringSet(RedisKey key, RedisValue value, TimeSpan? expiry = null, When when = When.Always, CommandFlags flags = CommandFlags.None)
+ bool StringSet(RedisKey key, RedisValue value, TimeSpan? expiry = null, When when = When.Always, CommandFlags flags = CommandFlags.None, bool keepTtl = false)
```

Another thought is to add a new method such as `StringSetAndKeepTtl`(maybe a better name)

mgravell:
Reordering is a binary break, and we've always tried to maintain a convention of flags as the last param.

slorello89:
Probably needs several smaller overloads, with each ordered sequence of optional parameters. Then in the final version the new parameter should be just prior to the flags.

erick2red:
This caused some pain for me today. While it works perfectly fine in my development setup in Visual Studio, when I deploy to Azure it crashes with the exception below

```
Method not found: 'System.Threading.Tasks.Task`1<Boolean> StackExchange.Redis.IDatabaseAsync.StringSetAsync(StackExchange.Redis.RedisKey, StackExchange.Redis.RedisValue, System.Nullable`1<System.TimeSpan>, Boolean, StackExchange.Redis.When, StackExchange.Redis.CommandFlags)'
```

NickCraver:
[...]

PR #2098, opened 2022-04-17
Fix for #2071: StringSet compatibility

NickCraver:
I missed an overload case being an idiot - adding the missing source break for the case in #2071.

...also fixing KeyTouch ordering while in here.

Note that adding `CommandFlags` back optional seems like a quick fix and I did try that route, but in a full test suite here it became apparent that created other ambiguous overload cases, so went this route.


src/StackExchange.Redis/RedisDatabase.cs
```diff
@@ -2945,6 +2945,9 @@ public Task<long> StringLengthAsync(RedisKey key, CommandFlags flags = CommandFl
             return ExecuteAsync(msg, ResultProcessor.Int64);
         }
 
+        // Backwards compatibility overloads:
+        public bool StringSet(RedisKey key, RedisValue value, TimeSpan? expiry, When when) =>
+            StringSet(key, value, expiry, false, when, CommandFlags.None);
```
mgravell, line 2950: can't we just make `flags` on the overload below option with a default?

NickCraver: Unfortunately I don't think so - the problem is then usage of `RedisKey key, RedisValue value, TimeSpan? expiry, When when` becomes ambiguous between the 2 methods. This was my first play based on your suggestion and my thoughts too, but couldn't make it work.

NickCraver:
Ah crap, reversed the ordering here and merged docs branch _before_ this went to main, dammit.

NickCraver:
@mgravell apologies for noise (I goofed on PR layering), here's the issue with the simple fix which I think we have to assume may be in use:
image
image

If we can go simpler I'm all for it, just didn't see a way to account for current possible uses.

mgravell (approved):

PR #2100, opened 2022-04-18
Docs: Love! Linking, <remarks> fixes, code formatting, etc.

NickCraver:
This does a few things globally to the interfaces:
- De-dupes `<remarks>` since evidently past the first one doesn't count/render
- Links our redis command links (and all others) so they're easily clickable!
- Moves a few types to proper class files
- In places sync/async methods are adjacent, utilizes `<inheritdoc cref="" /> to de-dupe
- ...and some other misc URL cleanup throughout.

In general: docs only change - I think we should merge this as-is to help PRs coming in, then I'll continue to iterate on docs.

### DIFF
--- a/src/StackExchange.Redis/APITypes/LatencyHistoryEntry.cs
+++ b/src/StackExchange.Redis/APITypes/LatencyHistoryEntry.cs
@@ -1,0 +1,47 @@
+﻿using System;
+
+namespace StackExchange.Redis;
+
+/// <summary>
+/// A latency entry as reported by the built-in LATENCY HISTORY command
+/// </summary>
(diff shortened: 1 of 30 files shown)


mgravell (approved):
I was watching you tweet about this over the egg-shaped weekend; very nice

question, though; the LatencyHistory stuff - doesn't seem to be used; is this prep work? side-effect of multiple PRs? I mean... it looks fine, I'm just not sure it is used?

NickCraver:
@mgravell yep tis used, I didn't remember it either but it's a few commands from IServer and was sitting in IServer.cs so cleaning up there :)

PR #2101, opened 2022-04-18
Updating IsPrimaryOnly command map for new commands

slorello89:
@NickCraver, working on `BITFIELD`/`BITFIELD_RO` and I realized that several of the new commands we've been adding are write-commands and therefore won't work on replicas, opening this to fix the command-map in Message.cs (pretty sure this is the right place but LMK if I'm offbase here) will need to push an update to #2095 #2094 to reflect this as well.

### DIFF
--- a/src/StackExchange.Redis/Message.cs
+++ b/src/StackExchange.Redis/Message.cs
@@ -338,8 +338,10 @@ namespace StackExchange.Redis
                 case RedisCommand.DEL:
                 case RedisCommand.EXPIRE:
                 case RedisCommand.EXPIREAT:
+                case RedisCommand.EXPIRETIME:
                 case RedisCommand.FLUSHALL:
                 case RedisCommand.FLUSHDB:
+                case RedisCommand.GEOSEARCHSTORE:
                 case RedisCommand.GETDEL:
                 case RedisCommand.GETEX:
                 case RedisCommand.GETSET:
@@ -366,6 +368,7 @@ namespace StackExchange.Redis
                 case RedisCommand.PERSIST:
                 case RedisCommand.PEXPIRE:
                 case RedisCommand.PEXPIREAT:
+                case RedisCommand.PEXPIRETIME:
                 case RedisCommand.PFADD:
                 case RedisCommand.PFMERGE:
                 case RedisCommand.PSETEX:
@@ -392,10 +395,12 @@ namespace StackExchange.Redis
                 case RedisCommand.TOUCH:
                 case RedisCommand.UNLINK:
                 case RedisCommand.ZADD:
+                case RedisCommand.ZDIFFSTORE:
                 case RedisCommand.ZINTERSTORE:
                 case RedisCommand.ZINCRBY:
                 case RedisCommand.ZPOPMAX:
                 case RedisCommand.ZPOPMIN:
+                case RedisCommand.ZRANGESTORE:
                 case RedisCommand.ZREM:
                 case RedisCommand.ZREMRANGEBYLEX:
                 case RedisCommand.ZREMRANGEBYRANK:


NickCraver:
@slorello89 agreed, but seems like we need a decent way of testing this with .DemandReplica assuring it throws. How to best make sure we test new methods either way though, hmmm...

slorello89:
@NickCraver 

What about something like [this](https://github.com/StackExchange/StackExchange.Redis/pull/2101/commits/378679f67a1e8e988e5d200b06d5510e69e5677c#diff-d1a4ee261df54641928ec8d69e2c6253262753203664c54c6f9f2840fb5d985bR970) for each write-command in question?

I suppose it doesn't demonstrate that the command will automatically target the primary, but that's probably out of scope for what an individual command is expected to demonstrate.

slorello89:
@NickCraver, if you wanted to test this for all commands, implement a general solution, you'd need to add a test that has a default implementation for each command, attempts to demand replica on each, and breaks or doesn't based on it's validity.

NickCraver:
@slorello89 Oh sorry I didn't mean add a new test per (I think one global test works here) - it's more of ensuring a command is added to the primary check itself - like it has to be there or not. I'll try and play tonight on some options like you _have_ to declare true/false or we throw and the test wouldn't pass, etc. 

We could for instance have 1 method with all commands that has a giant switch of true/false returns and defaults to throw - if you don't add it to the test and the test itself doesn't match is IsPrimaryOnly from Message, it fails - this would ensure we account for it. Or...some other approach?

slorello89:
You'd probably want to iterate over all the `RedisCommand` Enums, have some default implementation, try to demand replica, and check to make sure it fails when expected, or otherwise passes?

NickCraver:
@slorello89 I think this can be simpler by baking it in at the root level. We can change `IsPrimaryOnly` to have _all_ commands that should `return false` as a new chunk, and make the default throw. Then you can't add a command without account for it, pretty much nothing would work.

NickCraver:
Oh I see IsPrimaryOnly is static, we can just loop and call it :)

slorello89:
@NickCraver ahah, I gotcha, yeah that makes sense

slorello89:
Ok, I've worked out how to do this, I'll work on that update & test tonight/tomorrow. Shouldn't be terribly difficult.

NickCraver:
@slorello89 Have a spare hour an ideas - poking at this :)

NickCraver:
@slorello89 Alrighty, pushed up for eyes!

slorello89:
Looks about right. Let me take a pass though. It looks like some commands weren't added to the demand-primary loop when they were added. Some commands in there that do writes that shouldn't be passed onto a replica

src/StackExchange.Redis/Enums/RedisCommand.cs
```diff
@@ -1,236 +1,465 @@
...
+            case RedisCommand.BITPOS:
+            case RedisCommand.CLIENT:
+            case RedisCommand.CLUSTER:
+            case RedisCommand.CONFIG:
+            case RedisCommand.DBSIZE:
+            case RedisCommand.DEBUG:
+            case RedisCommand.DISCARD:
+            case RedisCommand.DUMP:
+            case RedisCommand.ECHO:
+            case RedisCommand.EVAL:
```
slorello89, line 350: Eval and EvalSha can be write commands I'm pretty sure they would be rejected by a replica. That's the impetus behind the RO variants

src/StackExchange.Redis/Enums/RedisCommand.cs
```diff
@@ -1,236 +1,465 @@
...
+            case RedisCommand.SUNION:
+            case RedisCommand.SSCAN:
+            case RedisCommand.SYNC:
+            case RedisCommand.TIME:
+            case RedisCommand.TTL:
+            case RedisCommand.TYPE:
+            case RedisCommand.UNSUBSCRIBE:
+            case RedisCommand.UNWATCH:
+            case RedisCommand.WATCH:
+            case RedisCommand.XACK:
```
slorello89, line 427: Probably considered write. Will try it out.

src/StackExchange.Redis/Enums/RedisCommand.cs
```diff
@@ -1,236 +1,465 @@
...
+            case RedisCommand.XADD:
```
slorello89, line 428: Definitely a write command

src/StackExchange.Redis/Enums/RedisCommand.cs
```diff
@@ -1,236 +1,465 @@
...
+            case RedisCommand.XCLAIM:
```
slorello89, line 429: Same with the next 3

src/StackExchange.Redis/Enums/RedisCommand.cs
```diff
@@ -1,236 +1,465 @@
...
+            case RedisCommand.XDEL:
+            case RedisCommand.XGROUP:
+            case RedisCommand.XINFO:
+            case RedisCommand.XLEN:
+            case RedisCommand.XPENDING:
+            case RedisCommand.XRANGE:
+            case RedisCommand.XREAD:
+            case RedisCommand.XREADGROUP:
```
slorello89, line 437: Dread group id guess is considered a write command

src/StackExchange.Redis/Enums/RedisCommand.cs
```diff
@@ -1,236 +1,465 @@
...
+            case RedisCommand.XREVRANGE:
+            case RedisCommand.XTRIM:
```
slorello89, line 439: XTRIM probably considered write

src/StackExchange.Redis/Enums/RedisCommand.cs
```diff
@@ -1,236 +1,465 @@
...
+            case RedisCommand.QUIT:
+            case RedisCommand.RANDOMKEY:
+            case RedisCommand.READONLY:
+            case RedisCommand.READWRITE:
+            case RedisCommand.REPLICAOF:
+            case RedisCommand.ROLE:
+            case RedisCommand.SAVE:
+            case RedisCommand.SCAN:
+            case RedisCommand.SCARD:
+            case RedisCommand.SCRIPT:
```
slorello89, line 402: Hmmm I wonder if this would be considered a write command?

slorello89 (commented):
Yeah definitely a few in here that are write commands, likely weren't explicitly added to this list when they were added to the library - these changes ought to take care of that

src/StackExchange.Redis/Enums/RedisCommand.cs
NickCraver: For this purpose, we only want to enforce if it _will_ fail, so want to leave the eval bits which we can't tell. For example if we changed those, it may change someone's production load from being spread over 20 replicas to slamming a primary.

src/StackExchange.Redis/Enums/RedisCommand.cs
NickCraver: I'm not sure on how this is handled either - the semantics of stream claiming are odd to me. For example, if you defined the consumer group on the secondary, is it valid to claim there? I have no idea if either is valid.

src/StackExchange.Redis/Enums/RedisCommand.cs
NickCraver: Could be either, so we should leave it in this case...should clarify the comment on these are things that would 100% fail on a read-only sever for any use case, anything else should be `false` - thoughts?

NickCraver:
@slorello89 oh yeah I noticed some stream for sure that looked like it - agree the list needs tweaking to be correct in the catch-up here.

NickCraver:
Hmmm, @slorello89 this sucks :) Not your code, but the reality of "fixing" this means we give a crap message. Compare, today when this gets through:

> RedisServerException : READONLY You can't write against a read only replica.
If we actively block it:
> RedisConnectionException: No connection is active/available to service this operation: XADD auto_id_stream_1650328241728, mc: 1/1/0, mgr: 10 of 10 available, clientName: StreamOpsFailOnReplica, IOCP: (Busy=0,Free=1000,Min=16,Max=1000), WORKER: (Busy=2,Free=2045,Min=16,Max=2047), v: 2.6.26.52977

I think we need to change the messaging when we have a connection, but not a writable one - that's not intuitive IMO. So I'm for fixing the issue here, but we need to improve the exception along with it. Maybe as simple as "No writable connection is..."? Thoughts/suggestions?

NickCraver:
okay this is an oddity in the flow for a non-specific server otherwise we throw a better message here:
```cs
if (message.IsPrimaryOnly() && server.IsReplica)
{
    throw ExceptionFactory.PrimaryOnly(RawConfig.IncludeDetailInExceptions, message.Command, message, server);
}
```
...will noodle on how we re-work that to cover the non-specified case.

slorello89:
I suppose the other wrinkle we need to consider, what about writeable replicas? Would this be BC for the stream commands on writeable replicas? I'd think so right?

NickCraver:
> I suppose the other wrinkle we need to consider, what about writeable replicas? Would this be BC for the stream commands on writeable replicas? I'd think so right?

Sooooooo I'm okay punting on that, because it doesn't work now. We've inadvertently blocked that in the `IsPrimaryOnly()` regardless of settings due to it being command-based for years. It's an issue in #354. I don't think anything is being made _worse_ here - we're just making `IsPrimaryOnly()` complete here. The issue needs a more global solution but I don't think it's made more difficult by anything in here.

NickCraver:
@slorello89 I pushed some goodies here on the message and started a Streams test - can go through the remaining commands and test each and add to that method before moving. Thoughts on hopefully-improved error messaging for users here?

src/StackExchange.Redis/Enums/RedisCommand.cs
slorello89: meant XREADGROUP, sorry was on mobile 😆

slorello89: I've set up a sentinel, I'll take a look at the ones with open questions.

src/StackExchange.Redis/Enums/RedisCommand.cs
slorello89: SCRIPT is definitely not a writeonly command, which is interesting because for things like `SCRIPT LOAD` you'd think it have to be propagated to the replica's?

src/StackExchange.Redis/Enums/RedisCommand.cs
NickCraver: @slorello89 I'm handling X* stuff right now - big test case incoming, working through each command!

src/StackExchange.Redis/Enums/RedisCommand.cs
```diff
@@ -1,236 +1,469 @@
...
+            case RedisCommand.TIME:
+            case RedisCommand.TTL:
+            case RedisCommand.TYPE:
+            case RedisCommand.UNSUBSCRIBE:
+            case RedisCommand.UNWATCH:
+            case RedisCommand.WATCH:
+            case RedisCommand.XACK:
+            case RedisCommand.XCLAIM:
+            case RedisCommand.XDEL:
+            case RedisCommand.XGROUP:
```
slorello89, line 435: can confirm XGROUP and it's sub-commands are all write commands (even `XGROUP HELP`!)

src/StackExchange.Redis/Enums/RedisCommand.cs
slorello89: XTRIM is a write.

src/StackExchange.Redis/Enums/RedisCommand.cs
slorello89: Yep XCLAIM is write.

src/StackExchange.Redis/Enums/RedisCommand.cs
slorello89: Can confirm this is also a write operation

src/StackExchange.Redis/Enums/RedisCommand.cs
NickCraver: Concur - change and test added!

src/StackExchange.Redis/Enums/RedisCommand.cs
NickCraver: Concur - change and test added!

src/StackExchange.Redis/Enums/RedisCommand.cs
NickCraver: Concur - change and test added!

src/StackExchange.Redis/Enums/RedisCommand.cs
NickCraver: Yep tis a write - added in test suite!

src/StackExchange.Redis/Enums/RedisCommand.cs
NickCraver: Yep - found that local too - test suite in play now!

NickCraver:
@slorello89 Alrighty pushed changes and verification up (was testing each before adding to the list) - ready for another pass!

src/StackExchange.Redis/Enums/RedisCommand.cs
slorello89: yep, you're right, reading [this](https://redis.io/docs/manual/replication/#read-only-replica) I thought that eval/evalsha would be considered write commands in and of themselves. [Evidently not](https://github.com/redis/redis/blob/unstable/src/commands.c#L3290) - you'd see `CMD_WRITE` in the flags.

src/StackExchange.Redis/Enums/RedisCommand.cs
```diff
@@ -1,236 +1,470 @@
...
+internal static class RedisCommandExtensions
 {
-    internal enum RedisCommand
+    /// <summary>
+    /// Gets whether a given command can be issued only to a primary, or if any server is eligible.
+    /// </summary>
+    /// <param name="command">The <see cref="RedisCommand"/> to check.</param>
+    /// <returns><see langword="true"/> if the command is primary-only, <see langword="false"/> otherwise.</returns>
+    [System.Diagnostics.CodeAnalysis.SuppressMessage("Style", "IDE0066:Convert switch statement to expression", Justification = "No, it'd be ridiculous.")]
+    public static bool IsPrimaryOnly(this RedisCommand command)
```
slorello89, line 247: This probably shouldn't be public.

NickCraver: The class is internal ;) (whole enum is really), but yeah no harm in being clearer!

src/StackExchange.Redis/ExceptionFactory.cs
```diff
@@ -124,6 +124,11 @@ internal static string GetInnerMostExceptionMessage(Exception? e)
                 // This can happen in cloud environments often, where user disables abort and has the wrong config
                 initialMessage = $"Connection to Redis never succeeded (attempts: {attempts} - check your config), unable to service operation: ";
             }
+            else if (message is not null && message.IsPrimaryOnly())
```
slorello89, line 127: I might be missing some context here, isn't this a bit inferential?

NickCraver: I'm not sure I follow - how so? If we're passed a message and we know it's demanding a primary, we're simply adding that info to the message.

slorello89: yeah, but this could happen in any number of failure scenarios right?

This is kind of implying that you might have been able to connect to a replica, but connecting to a primary was the issue.

You could theoretically see this in any topography, couldn't you?

src/StackExchange.Redis/Enums/RedisCommand.cs
NickCraver: To make this clearer for anyone down the road, gave the same love to all other extensions in #2102.

src/StackExchange.Redis/ExceptionFactory.cs
NickCraver: Correct, the message isn't meant to imply there is a replica, alternative phrasing suggestion?

slorello89: Maybe add to the conditional that you need to be on a topography with replication? It looks like the multiplexers configuration is being interrogated above, so it doesn't seems like that would be totally out-of-band behavior here. And then you can use something along the lines of `"No writeable connection active/available to service this operation: "`?

NickCraver: I tried that at first but felt phrasing was too little to indicate what's happening. The check we can do though - I can add a conditional to being connected to anything. From that, we can infer we only have replica connections. Pushed up a tweak here, and indeed the async test _won't_ get this message now, because it's only connecting to a primary and it's not available, so gets the old message.

slorello89: Sounds good 👍

NickCraver (approved):
Looks good, but I've heavily tainted my review - @mgravell can you lay eyes please?

mgravell (approved):
seems legit

PR #2102, opened 2022-04-19
Extensions: make internals...internal

NickCraver:
This isn't a visibility change, just making public members of internal classes internal themselves to be clearer.

Note: no public API changes in this...so loving that analyzer right now :)


slorello89 (approved):
LGTM 👍

PR #2106, opened 2022-04-19
Builds: GitHub Actions logger upgrade

NickCraver:
Upgrading this for some goodies - see https://twitter.com/Tyrrrz/status/1516371034999504897

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -27,7 +27,7 @@ jobs:
       working-directory: ./tests/RedisConfigs
       run: docker-compose -f docker-compose.yml up -d
     - name: StackExchange.Redis.Tests
-      run: dotnet test tests/StackExchange.Redis.Tests/StackExchange.Redis.Tests.csproj -c Release --logger trx --logger "GitHubActions;report-warnings=false" --results-directory ./test-results/ /p:CI=true
+      run: dotnet test tests/StackExchange.Redis.Tests/StackExchange.Redis.Tests.csproj -c Release --logger trx --logger GitHubActions --results-directory ./test-results/ /p:CI=true
     - uses: dorny/test-reporter@v1
       continue-on-error: true
       if: success() || failure()
@@ -74,7 +74,7 @@ jobs:
         .\redis-server.exe --service-install --service-name "redis-26381" "..\Sentinel\sentinel-26381.conf" --sentinel
         Start-Service redis-*
     - name: StackExchange.Redis.Tests
-      run: dotnet test tests/StackExchange.Redis.Tests/StackExchange.Redis.Tests.csproj -c Release --logger trx --logger "GitHubActions;report-warnings=false" --results-directory ./test-results/ /p:CI=true
+      run: dotnet test tests/StackExchange.Redis.Tests/StackExchange.Redis.Tests.csproj -c Release --logger trx --logger GitHubActions --results-directory ./test-results/ /p:CI=true
     - uses: dorny/test-reporter@v1
       continue-on-error: true
       if: success() || failure()
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -27,7 +27,7 @@ jobs:
       working-directory: ./tests/RedisConfigs
       run: docker-compose -f docker-compose.yml up -d
     - name: StackExchange.Redis.Tests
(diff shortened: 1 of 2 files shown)


Tyrrrz:
Note: I think you need to also add it here:

https://github.com/StackExchange/StackExchange.Redis/blob/ecea7f4ff003934498f843d06634848bf70a6544/.github/workflows/CI.yml#L30

You can have multiple loggers at the same time IIRC, just put multiple `--logger <value>`.

The format of the summary is different from some of the other ones I've seen, so it will probably need more rework to handle projects with massive test suites. There is also a limit of 128kb on the summary to take into account. So essentially I'm looking forward to feedback, nothing is final. :)

NickCraver:
@Tyrrrz yup, latest has it ;) I think for our size, I'd like to see an option for detail on errors/warnings only in the summary, rather than everything. Counts for success is great, with details for failure, at least in our case - thoughts?

Tyrrrz:
> @Tyrrrz yup, latest has it ;) I think for our size, I'd like to see an option for detail on errors/warnings only in the summary, rather than everything. Counts for success is great, with details for failure, at least in our case - thoughts?

That sounds like a reasonable solution. There's probably also some other stuff I could remove to reduce the size further.
I'll try to get a fix in soon-ish (might be towards the end of the week as I'm a bit sick), I'll comment on this PR when I release a new version. Thanks for trying. :)

NickCraver:
@Tyrrrz that sounds awesome - will leave this open to experiment. Hope you feel better!

Directory.Packages.props
```diff
@@ -10,7 +10,7 @@
     
     <!-- Packages only used in the solution, upgrade at will -->
     <PackageVersion Include="BenchmarkDotNet" Version="0.13.1" />
-    <PackageVersion Include="GitHubActionsTestLogger" Version="1.3.0" />
+    <PackageVersion Include="GitHubActionsTestLogger" Version="1.4.1" />
```
Tyrrrz, line 13: Hey @NickCraver, can you test out the new alpha I made?
Sorry it took so long, dotnet test platform is hellish to work with in all honesty (and the war doesn't help either).

```suggestion
    <PackageVersion Include="GitHubActionsTestLogger" Version="2.0.0-alpha" />
```

.github/workflows/CI.yml
```diff
@@ -74,7 +74,7 @@ jobs:
         .\redis-server.exe --service-install --service-name "redis-26381" "..\Sentinel\sentinel-26381.conf" --sentinel
         Start-Service redis-*
     - name: StackExchange.Redis.Tests
-      run: dotnet test tests/StackExchange.Redis.Tests/StackExchange.Redis.Tests.csproj -c Release --logger trx --results-directory ./test-results/ /p:CI=true
+      run: dotnet test tests/StackExchange.Redis.Tests/StackExchange.Redis.Tests.csproj -c Release --logger trx --logger "GitHubActions;report-warnings=false" --results-directory ./test-results/ /p:CI=true
```
Tyrrrz, line 77: This is the default behavior with v2.0

```suggestion
      run: dotnet test tests/StackExchange.Redis.Tests/StackExchange.Redis.Tests.csproj -c Release --logger trx --logger GitHubActions --results-directory ./test-results/ /p:CI=true
```

.github/workflows/CI.yml
```diff
@@ -27,7 +27,7 @@ jobs:
       working-directory: ./tests/RedisConfigs
       run: docker-compose -f docker-compose.yml up -d
     - name: StackExchange.Redis.Tests
-      run: dotnet test tests/StackExchange.Redis.Tests/StackExchange.Redis.Tests.csproj -c Release --logger trx --results-directory ./test-results/ /p:CI=true
+      run: dotnet test tests/StackExchange.Redis.Tests/StackExchange.Redis.Tests.csproj -c Release --logger trx --logger "GitHubActions;report-warnings=false" --results-directory ./test-results/ /p:CI=true
```
Tyrrrz, line 30: This is the default behavior with v2.0

```suggestion
      run: dotnet test tests/StackExchange.Redis.Tests/StackExchange.Redis.Tests.csproj -c Release --logger trx --logger GitHubActions --results-directory ./test-results/ /p:CI=true
```

Directory.Packages.props
NickCraver: Absolutely - thanks for this! Giving it a spin now

Tyrrrz:
Looks quite nice IMO, and the link even takes you to the right place (really proud of this one 😅)

image

NickCraver:
@Tyrrrz agreed - thank you for this! Will merge on in :)

PR #2110, opened 2022-04-19
avoid byte[] allocations when calculating cluster slot

mgravell:
There is a subtle `byte[]` alloc in the cluster slot usage, which will be allocating a *lot* of `byte[]` for cluster; let's... not do that

- [ ] open question: happy for me to add `[SkipLocalsInit]`? improves a few scenarios, but in particular `stackalloc`
- [ ] refactor existing `byte[]` operator to use `CopyTo`
- [ ] check for other related scenarios - equality maybe?


src/StackExchange.Redis/RedisKey.cs
```diff
@@ -297,5 +298,59 @@ internal static RedisKey WithPrefix(byte[]? prefix, RedisKey value)
...
+                destination = destination.Slice(KeyPrefix.Length);
+            }
+            switch (KeyValue)
+            {
+                case null:
+                    break; // nothing to do
+                case string s:
+                    if (s.Length != 0)
+                    {
+#if NETCOREAPP3_1_OR_GREATER || NETSTANDARD2_1_OR_GREATER
```
NickCraver, line 332: Since we only target `netstandard2.0`, simplify to `#if NETCOREAPP`?

NickCraver (approved):
Looks great! One question bout #if defs I wanted to check on intent - thoughts on simplifying? Works as-is, but can simplify I think given one's never hit.

src/StackExchange.Redis/RedisKey.cs
mgravell: Noted; I assume `NETCOREAPP` includes 5.0+?

NickCraver: Aye, it's _not_ netstandard but all .NET Core onwards including 5/6/7, etc.

NickCraver:
@mgravell housekeeping question: we now have `AssemblyInfoHack.cs` (which I can move to .csproj), `.Hacks.cs`, `NullableHacks.cs`, and `SkipLocalsInit.cs` - how about I make a follow-up minimizing/consolidating some of that info a folder or some such after this is in?

mgravell:
> `VectorSafeIndexOfCRLF`

Yeah, I actually started hacking that, but: I'm not 100% sure what it compiles to on netfx, and sharplab can't help me. I'll dig out a desktop IL decompiler tomorrow, and check that it is close enough to what we expect.

PR #2111, opened 2022-04-19
SORT_RO

slorello89:
@NickCraver - I modified the path for message creation for the Sort command to point to using `SORT_RO` when possible, this again had to do a version-check against the multiplexer, which I'm not certain is the correct way - thoughts?

Also, SORT is considered a write command and will be rejected out of hand by a replica (so I'm moving that as well)

```text
127.0.0.1:6378> SORT test
(error) READONLY You can't write against a read only replica.
```

### DIFF
--- a/src/StackExchange.Redis/RedisDatabase.cs
+++ b/src/StackExchange.Redis/RedisDatabase.cs
@@ -1762,25 +1762,25 @@ namespace StackExchange.Redis
 
         public RedisValue[] Sort(RedisKey key, long skip = 0, long take = -1, Order order = Order.Ascending, SortType sortType = SortType.Numeric, RedisValue by = default, RedisValue[]? get = null, CommandFlags flags = CommandFlags.None)
         {
-            var msg = GetSortedSetAddMessage(default(RedisKey), key, skip, take, order, sortType, by, get, flags);
+            var msg = GetSortMessage(RedisKey.Null, key, skip, take, order, sortType, by, get, flags);
             return ExecuteSync(msg, ResultProcessor.RedisValueArray, defaultValue: Array.Empty<RedisValue>());
         }
 
         public long SortAndStore(RedisKey destination, RedisKey key, long skip = 0, long take = -1, Order order = Order.Ascending, SortType sortType = SortType.Numeric, RedisValue by = default, RedisValue[]? get = null, CommandFlags flags = CommandFlags.None)
         {
-            var msg = GetSortedSetAddMessage(destination, key, skip, take, order, sortType, by, get, flags);
+            var msg = GetSortMessage(destination, key, skip, take, order, sortType, by, get, flags);
             return ExecuteSync(msg, ResultProcessor.Int64);
         }
 
         public Task<long> SortAndStoreAsync(RedisKey destination, RedisKey key, long skip = 0, long take = -1, Order order = Order.Ascending, SortType sortType = SortType.Numeric, RedisValue by = default, RedisValue[]? get = null, CommandFlags flags = CommandFlags.None)
         {
-            var msg = GetSortedSetAddMessage(destination, key, skip, take, order, sortType, by, get, flags);
+            var msg = GetSortMessage(destination, key, skip, take, order, sortType, by, get, flags);
             return ExecuteAsync(msg, ResultProcessor.Int64);
         }
 
         public Task<RedisValue[]> SortAsync(RedisKey key, long skip = 0, long take = -1, Order order = Order.Ascending, SortType sortType = SortType.Numeric, RedisValue by = default, RedisValue[]? get = null, CommandFlags flags = CommandFlags.None)
         {
-            var msg = GetSortedSetAddMessage(default(RedisKey), key, skip, take, order, sortType, by, get, flags);
+            var msg = GetSortMessage(RedisKey.Null, key, skip, take, order, sortType, by, get, flags);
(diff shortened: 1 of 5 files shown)


src/StackExchange.Redis/RedisDatabase.cs
```diff
@@ -3500,8 +3500,10 @@ private Message GetSortedSetAddMessage(RedisKey key, RedisValue member, double s
             }
         }
 
-        private Message GetSortedSetAddMessage(RedisKey destination, RedisKey key, long skip, long take, Order order, SortType sortType, RedisValue by, RedisValue[]? get, CommandFlags flags)
+        private Message GetSortMessage(RedisKey destination, RedisKey key, long skip, long take, Order order, SortType sortType, RedisValue by, RedisValue[]? get, CommandFlags flags)
         {
+            var command = destination.IsNull && multiplexer.GetServer(multiplexer.GetEndPoints()[0]).Version >= RedisFeatures.v7_0_0_rc1 ? RedisCommand.SORT_RO : RedisCommand.SORT;
```
NickCraver, line 3505: Here we'd want to add a new _internal_ member to `RedisFeatures`, e.g. `internal bool SortedSetPop => Version >= v7_0_0_rc1;` and use `var features = GetFeatures(key, flags, out ServerEndPoint? server);` like other places - note this means we need to pass server outward and use it upstream (to ensure we went to the server we said had the feature).

These types of switches are the _only_ things we want to be added to `RedisFeatures` overall but let's keep internal for now.

tests/StackExchange.Redis.Tests/Sets.cs
```diff
@@ -343,4 +343,42 @@ public void SetPopMulti_Nil()
...
+        await db.KeyDeleteAsync(key);
+
+        var random = new Random();
+        var items = Enumerable.Repeat(0, 200).Select(x => random.Next()).ToList();
+        await db.SetAddAsync(key, items.Select(x=>(RedisValue)x).ToArray());
+        items.Sort();
+
+        var result = (await db.SortAsync(key)).Select(x=>(int)x);
+        Assert.Equal(items.Count,result.Count() );
+        Assert.Equivalent(items, result);
```
NickCraver, line 363: There's an `IEnumerable<T>` overload to `Assert.Equal()` we'd want to use here:
```suggestion
        Assert.Equal(items, result);
```

tests/StackExchange.Redis.Tests/Sets.cs
```diff
@@ -343,4 +343,42 @@ public void SetPopMulti_Nil()
...
```
NickCraver, line 382:
```suggestion
        Assert.Equal(items, result);
```

src/StackExchange.Redis/RedisDatabase.cs
slorello89: Thanks, @NickCraver, that's much cleaner, I'll clean up my other PR that does that hocky junk as well.

src/StackExchange.Redis/RedisDatabase.cs
```diff
@@ -1762,26 +1762,30 @@ private CursorEnumerable<RedisValue> SetScanAsync(RedisKey key, RedisValue patte
 
         public RedisValue[] Sort(RedisKey key, long skip = 0, long take = -1, Order order = Order.Ascending, SortType sortType = SortType.Numeric, RedisValue by = default, RedisValue[]? get = null, CommandFlags flags = CommandFlags.None)
         {
-            var msg = GetSortedSetAddMessage(default(RedisKey), key, skip, take, order, sortType, by, get, flags);
-            return ExecuteSync(msg, ResultProcessor.RedisValueArray, defaultValue: Array.Empty<RedisValue>());
+            var features = GetFeatures(key, flags, out ServerEndPoint? server);
```
NickCraver, line 1765: This can be inside `GetSortMessage` just add an out param there ;) See `GetStringGetWithExpiryMessage` for an example!

slorello89: fair.

slorello89:
Btw @NickCraver, in the `GetSortMessage` bit, there's a check where it checks your flags to see if you are demanding a replica for a store command -AFAIK this would never have worked against a replica and is sort of strictly in the WRITE camp, that's why I moved the command to the Write section in the GetPrimary extension method - worth nixing that bit?

NickCraver:
@slorello89 I was wondering about that - are we saying SORT in general doesn't work against a replica ever?

NickCraver:
Aside, but it'd be really nice if the docs pages had an indicator of what is/isn't allowed on a read-only server somewhere...

slorello89:
@NickCraver in theory the WRITE tag in the [commands.json](https://github.com/redis/redis-doc/blob/master/commands.json) file is supposed to tell you that, you can see it under the ACLs in [Redis.io](https://redis.io/commands/sort/)

NickCraver:
Gotcha, I couldn't tell if that's "just with STORE" though - need to poke at this and test. If it can operate on a replica without STORE, then we shouldn't move/enforce that as it could break an existing use.

slorello89:
@banker @itamarhaber - I like @NickCraver 's suggestion above - maybe we could add a table to redis.io to make it a bit clearer for our library developers what commands are meant to be write commands?

NickCraver (approved):
Current looks good :+1: Thanks for this!

PR #2116, opened 2022-04-24
Support BYTE|BIT in BITCOUNT and BITPOS commands

Avital-Fine:
The new features were added to Redis 7: now the indexes of both commands ([BITCOUNT](https://redis.io/commands/bitcount/) and [BITPOS](https://redis.io/commands/bitpos/)) can be specified in BITs instead of BYTEs.

(Also added tests for BitPosition since it was missing)


NickCraver:
In this case, I'd guess we want an overload, e.g. with `StringIndexType.Bit`/`Byte` (defaulting to `Byte`) before `CommandFlags` - thoughts?

Avital-Fine:
> In this case, I'd guess we want an overload, e.g. with `StringIndexType.Bit`/`Byte` (defaulting to `Byte`) before `CommandFlags` - thoughts?

Yes, I was considering that but we can't make a default if we overload. 

@NickCraver I can do overload without a default if you think it's better

NickCraver:
We can do with a default, we just need to make sure not to break the old method - see #2098 where I'm doing a setup to ensure it's not broken by testing all overload options someone could be using - we'd want to do the same here. You could merge that PR branch in here and point this PR at that branch instead of `main`, since it'll merge first and has the test clas in place and such.

Avital-Fine:
@NickCraver Ready :)

src/StackExchange.Redis/Interfaces/IDatabaseAsync.cs
```diff
@@ -2477,6 +2478,10 @@ public interface IDatabaseAsync : IRedisAsync
         /// <remarks><seealso href="https://redis.io/commands/append"/></remarks>
         Task<long> StringAppendAsync(RedisKey key, RedisValue value, CommandFlags flags = CommandFlags.None);
 
+        /// <inheritdoc cref="StringBitCountAsync(RedisKey, long, long, StringIndexUnit, CommandFlags)" />
+        [Browsable(false), EditorBrowsable(EditorBrowsableState.Never)]
+        Task<long> StringBitCountAsync(RedisKey key, long start, long end, CommandFlags flags);
```
NickCraver, line 2483: I _think_ these are okay, but we should add a test suite to `OverloadCompat`	to ensure we're all good and haven't missed anything. I've merging in the `StringGet()` bits there as an example!

src/StackExchange.Redis/Enums/StringIndexUnit.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System;
+
+namespace StackExchange.Redis;
+
+/// <summary>
+/// Indicates if we index with bit unit of byte unit.
+/// </summary>
+public enum StringIndexUnit
```
NickCraver, line 8:
```suggestion
public enum StringIndexType
```

src/StackExchange.Redis/Enums/StringIndexUnit.cs
```diff
@@ -0,0 +1,28 @@
...
+{
+    /// <summary>
+    /// Use bit
+    /// </summary>
+    Bit,
+    /// <summary>
+    /// Use byte
+    /// </summary>
```
NickCraver, line 16:
```suggestion
    /// <summary>
    /// Indicates the index is the number of bits into a string.
    /// </summary>
    Bit,
    /// <summary>
    /// Indicates the index is the number of bytes into a string.
    /// </summary>
```

src/StackExchange.Redis/Enums/StringIndexUnit.cs
```diff
@@ -0,0 +1,28 @@
...
+    Byte,
+}
+
+internal static class StringIndexUnitExtensions
```
NickCraver, line 20:
```suggestion
internal static class StringIndexTypeExtensions
```

src/StackExchange.Redis/Interfaces/IDatabaseAsync.cs
Avital-Fine: I had these tests in `Strings.cs` and now moved it to `OverloadCompat.cs`.
@NickCraver Is it ok now?

src/StackExchange.Redis/RedisDatabase.cs
```diff
@@ -2826,13 +2844,31 @@ public Task<long> StringBitOperationAsync(Bitwise operation, RedisKey destinatio
...
         public long StringBitPosition(RedisKey key, bool bit, long start = 0, long end = -1, CommandFlags flags = CommandFlags.None)
         {
-            var msg = Message.Create(Database, flags, RedisCommand.BITPOS, key, bit, start, end);
+            return StringBitPosition(key, bit, start, end, StringIndexType.Byte, flags);
+        }
+
+        public Task<long> StringBitPositionAsync(RedisKey key, bool bit, long start = 0, long end = -1, CommandFlags flags = CommandFlags.None)
+        {
+            return StringBitPositionAsync(key, bit, start, end, StringIndexType.Byte, flags);
+        }
```
NickCraver, line 2853:
```suggestion
        public Task<long> StringBitPositionAsync(RedisKey key, bool bit, long start = 0, long end = -1, CommandFlags flags = CommandFlags.None) =>
            StringBitPositionAsync(key, bit, start, end, StringIndexType.Byte, flags);
```

NickCraver (changes requested):
Alrighty tried to give this a good pass. Sorry time isn't on my side this week, but this is looking great - just about good to go :)

docs/ReleaseNotes.md
```diff
@@ -30,6 +30,7 @@
 - Adds: Support for `OBJECT FREQ` with `.KeyFrequency()`/`.KeyFrequencyAsync()` ([#2105 by Avital-Fine](https://github.com/StackExchange/StackExchange.Redis/pull/2105))
 - Performance: Avoids allocations when computing cluster hash slots or testing key equality ([#2110 by Marc Gravell](https://github.com/StackExchange/StackExchange.Redis/pull/2110))
 - Adds: Support for `SORT_RO` with `.Sort()`/`.SortAsync()` ([#2111 by slorello89](https://github.com/StackExchange/StackExchange.Redis/pull/2111))
+- Adds: Support for `BIT | BYTE` to `BITCOUNT` and `BITPOS` with `.StringBitCount()`/`.StringBitCountAsync()` and `.StringBitPosition()`/`.StringBitPositionAsync()` using overload ([#2116 by Avital-Fine](https://github.com/StackExchange/StackExchange.Redis/pull/2116))
```
NickCraver, line 33:
```suggestion
- Adds: Support for `BIT | BYTE` to `BITCOUNT` and `BITPOS` with `.StringBitCount()`/`.StringBitCountAsync()` and `.StringBitPosition()`/`.StringBitPositionAsync()` [#2116 by Avital-Fine](https://github.com/StackExchange/StackExchange.Redis/pull/2116))
```

NickCraver (approved):
This is looking awesome - thank you for the tweaks and compat checks!

PR #2118, opened 2022-04-26
for pub/sub, treat length-one arrays comparably to simple values

### LINKED ISSUE #2117: Bug: Subscriber client crash when listen to '__redis__:invalidate' messages queue.
Found in :
[https://github.com/StackExchange/StackExchange.Redis/issues/1461](url)
Redis ver 6.2.6  (on docker's container)
Stackexchange ver : 2.2.79 / 2.5.61 ( on dotnet core 3.1 )

Lets say we have a subscriber for invalidate keys queue : (to implement near cache)
  
var subscriber = invalidator.GetSubscriber();
_subscriber.Subscribe(" _ _ redis _ _ :invalidate").OnMessage(message =>
{
    Console.WriteLine($"__redis__:invalidate: {message}");
});_

We never get messages came from values changes or expired (we do get messages came from explicit publish call.)

When looking deeper we can see that the client id of the subscriber change after setting new values of tracking keys.. meaning it's was crash..

I got this : System.InvalidCastException: 'Cannot convert to RedisValue: MultiBulk' .
in this method : 
Internal RedisValue AsRedisValue()

mgravell:
fix #2117

it appears that some of the RESP2 fallback support for client-side caching uses multi-bulk payloads; this currently causes a connection break, so we definitely need to stop *the failure*, but we can also interpret unary arrays similarly to simple values, which should at least allow simple cases to work

### DIFF
--- a/src/StackExchange.Redis/PhysicalConnection.cs
+++ b/src/StackExchange.Redis/PhysicalConnection.cs
@@ -1566,7 +1566,7 @@ namespace StackExchange.Redis
                     case ResultType.BulkString:
                         parsed = value.AsRedisValue();
                         return true;
-                    case ResultType.MultiBulk when value.ItemsCount == 1:
+                    case ResultType.MultiBulk when allowArraySingleton && value.ItemsCount == 1:
                         return TryGetPubSubPayload(in value[0], out parsed, allowArraySingleton: false);
                 }
                 parsed = default;
--- a/src/StackExchange.Redis/PhysicalConnection.cs
+++ b/src/StackExchange.Redis/PhysicalConnection.cs
@@ -1498,10 +1498,10 @@ namespace StackExchange.Redis
                     // invoke the handlers
                     var channel = items[1].AsRedisChannel(ChannelPrefix, RedisChannel.PatternMode.Literal);
                     Trace("MESSAGE: " + channel);
-                    if (!channel.IsNull)
+                    if (!channel.IsNull && TryGetPubSubPayload(items[2], out var payload))
                     {
                         _readStatus = ReadStatus.InvokePubSub;
-                        muxer.OnMessage(channel, channel, items[2].AsRedisValue());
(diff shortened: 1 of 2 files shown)


src/StackExchange.Redis/PhysicalConnection.cs
```diff
@@ -1551,6 +1551,27 @@ private void MatchResult(in RawResult result)
...
+                }
+                switch (value.Type)
+                {
+                    case ResultType.Integer:
+                    case ResultType.SimpleString:
+                    case ResultType.BulkString:
+                        parsed = value.AsRedisValue();
+                        return true;
+                    case ResultType.MultiBulk when value.ItemsCount == 1:
+                        return TryGetPubSubPayload(in value[0], out parsed, allowArraySingleton: false);
```
NickCraver, line 1570: FWIW I think this would be simpler to take the first element as a RedisValue instead of recursion just for ease of understanding/maintenance, but there may be a future case of nested arrays I'm unaware of.

NickCraver (approved):
Looks good, but _if_ we can not recurse I think it'd be a bit simpler. Probably a reason I'm not seeing through 👍

src/StackExchange.Redis/PhysicalConnection.cs
mgravell: @NickCraver yah, avoiding edge cases and avoiding having to repeat the "what combinations are allowed" (the `switch`)

GT185076:
Well done , it's work !
Big question left : Is there a case with multiple values (more than one)..?

mgravell:
Great question. I honestly don't know. If there is, the library will now silently ignore them (since it doesn't know what to do with them) rather than outright dying, which is... better.

mgravell:
The myget feed is here: https://www.myget.org/F/stackoverflow/api/v3/index.json - we'll push it to nuget when we're ready

PR #2124, opened 2022-04-28
Overload old versions of KeyExpire and KeyExpireAsync

Avital-Fine:
Continue #2083 
Now when we can overload function - we can do it also to `KeyExpire` and `KeyExpireAsync`


NickCraver (approved):

NickCraver:
@Avital-Fine trying to get the quicker ones in (others need more time/eyes), thanks for this!

PR #2133, opened 2022-05-07
Fix race condition in Sentinel connection multiplexer

ewisuri:
In the Sentinel Connection Multiplexer, since `OnManagedConnectionRestored` can be executed on a different thread, it's possible for it to dispose the timer just before `OnManagedConnectionFailed` calls `.Change()` on the timer, which leads to an `ObjectDisposedException` being thrown in `OnManagedConnectionFailed`. Since the connection was restored, it seems safe to suppress these exceptions, instead of requiring locking around the reference to `connection.sentinelPrimaryReconnectTimer`.


NickCraver (approved):
Looking good - thanks for the fix here!

PR #2136, opened 2022-05-10
Support GT, LT and CH in ZADD command

Avital-Fine:
[SortedSetAdd](https://redis.io/commands/zadd/) currently not supporting GT and LT features.

Also, CH is not supported. @NickCraver Should I add it here in this PR too?


NickCraver:
I think we should do all the things at once here given it's the same overloads, yep - @mgravell and I are unsure about the `When` because the options are more complex/confusing.  But we should tackle the CH and INCR at the same time given it's all the same method ultimately. I'm back in action this week so trying to resume PRs tonight.

NickCraver:
Hmm yeah okay let's ignore for now - plan is if someone needs the combo of these args _and_ INCR, then we can add it to `SortedSetIncrement` down the road :)

Avital-Fine:
> @Avital-Fine talking about this in call - can we please try renaming the `When` to `SortedSetWhen` which has _all_ the members? Then we don't have the existing `When` in the new overload - it's an overall change to this `[Flags]` enum to specify the options.
> 
> /cc @mgravell

Maybe we can add SortedSetAddFlags and then we can include there CH flag? WDYT?

NickCraver:
@Avital-Fine ACK! Sorry I've been swamped and not on this hardly at all - @mgravell is going to give compat eyes here :)

mgravell (approved):

PR #2145, opened 2022-05-24
fix: corrected fuget.org typo

reecerussell:
In the "API Browser" link text, the reference to fuget.org was furget.org. This change is to correct the spelling of the site.


NickCraver (approved):
Thanks!

PR #2161, opened 2022-06-16
Implement IAsyncDisposable from IConnectionMultiplexer

### LINKED ISSUE #2160: Implement IAsyncDisposable in IConnectionMultiplexer
`IConnectionMultiplexer` allows closing all connections synchronously via `Close` and asynchronously via `CloseAsync`. It also implements `IDisposable` and closes any connections synchronously when disposed, but doesn't implement the corresponding asynchronous [`IAsyncDisposable`](https://docs.microsoft.com/en-us/dotnet/api/system.iasyncdisposable) interface and [`DisposeAsync`](https://docs.microsoft.com/en-us/dotnet/api/system.iasyncdisposable.disposeasync) method. Doing so would allow closing connections asynchronously before disposing without manually calling `CloseAsync` first.

kimsey0:
Fixes #2160.

We may want to implement `IAsyncDisposable` in more public `IDisposable` classes which can dispose their resources asynchronously.

### DIFF
--- a/src/StackExchange.Redis/ConnectionMultiplexer.cs
+++ b/src/StackExchange.Redis/ConnectionMultiplexer.cs
@@ -2025,7 +2025,7 @@ namespace StackExchange.Redis
             await CloseAsync(!_isDisposed);
             if (sentinelConnection is ConnectionMultiplexer sentinel)
             {
-                await (sentinel?.DisposeAsync() ?? default);
+                await sentinel.DisposeAsync();
             }
             var oldTimer = Interlocked.Exchange(ref sentinelPrimaryReconnectTimer, null);
             oldTimer?.Dispose();
--- a/src/StackExchange.Redis/ConnectionMultiplexer.cs
+++ b/src/StackExchange.Redis/ConnectionMultiplexer.cs
@@ -2017,6 +2017,21 @@ namespace StackExchange.Redis
         }
 
         /// <summary>
+        /// Release all resources associated with this object.
+        /// </summary>
+        public async ValueTask DisposeAsync()
+        {
+            GC.SuppressFinalize(this);
+            await CloseAsync(!_isDisposed);
+            if (sentinelConnection is not null)
+            {
+                await sentinelConnection.DisposeAsync();
+            }
+            var oldTimer = Interlocked.Exchange(ref sentinelPrimaryReconnectTimer, null);
+            oldTimer?.Dispose();
+        }
+
+        /// <summary>
         /// Close all connections and release all resources associated with this object.
         /// </summary>
         /// <param name="allowCommandsToComplete">Whether to allow all in-queue commands to complete first.</param>
--- a/src/StackExchange.Redis/ConnectionMultiplexer.cs
+++ b/src/StackExchange.Redis/ConnectionMultiplexer.cs
@@ -2023,9 +2023,9 @@ namespace StackExchange.Redis
         {
             GC.SuppressFinalize(this);
             await CloseAsync(!_isDisposed);
-            if (sentinelConnection is not null)
+            if (sentinelConnection is ConnectionMultiplexer sentinel)
             {
-                await sentinelConnection.DisposeAsync();
+                await (sentinel?.DisposeAsync() ?? default);
             }
             var oldTimer = Interlocked.Exchange(ref sentinelPrimaryReconnectTimer, null);
             oldTimer?.Dispose();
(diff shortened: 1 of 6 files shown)


NickCraver:
Hey @kimsey0 - we should add release notes and move the API to the Shipped set in-order but then good to go. Happy to take care of that to get this in if you want :)

kimsey0:
@NickCraver: I moved the API change to shipped and added a new line under Unreleased in the release notes document. I haven't used `Microsoft.CodeAnalysis.PublicApiAnalyzers` before, so feel free to fix it if I've done something incorrectly.

src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -2016,6 +2016,21 @@ public void Dispose()
...
+        /// Release all resources associated with this object.
+        /// </summary>
+        public async ValueTask DisposeAsync()
+        {
+            GC.SuppressFinalize(this);
+            await CloseAsync(!_isDisposed);
+            if (sentinelConnection is not null)
+            {
+                await sentinelConnection.DisposeAsync();
+            }
```
NickCraver, line 2029: This can race, where `sentinelConnection` is nulled out between the `if` and the `await`. Instead, what you want to do is use the `if (sentinelConnection is ConnectionMultiplexer sentinel) { await sentinel.DisposeAsync(); }` pattern so that you're eliminating the potential race and null ref (more common than you'd think in disposing!)

NickCraver (changes requested):
This is looking good! 1 issue introduced with a null ref in dispose though - need to shuffle a bit on how that's handled (comments inline).

src/StackExchange.Redis/ConnectionMultiplexer.cs
kimsey0: Well spotted. I guess it could also be `await (sentinel?.DisposeAsync() ?? default);` to match the synchronous implementation a little closer, if you prefer? I just don't find it quite as readable.

NickCraver: Would stick with the `if`, you don't want to pay state machine overhead awaiting if we know there's nothing to await :)

kimsey0: I don't think it makes any difference. To the best of my knowledge, the state machine isn't initialized if the (Value)Task is completed synchronously (the `if (!awaiter.IsCompleted)` [here](https://sharplab.io/#v2:CYLg1APgAgTAjAWAFBQAwAIpwKwG5lQDMmM6AwugN7Lq2bFQAcmAbOgJIBmAFAJRU06Q9gEEAzgE8AdgGMAIgEsxABwD2YgIYAjADYBTTI0Ur12/egC86KQFcdO/EiFCFndNybG1m3QaUdxaXklbzMDTxDTXzh+aidnBKgATkMvKP04ADo0sT1A2T5HBNoAX0FS8vpDVnQ5PU4NOwAXPgF4hNFJWRyw1MifcytbeyLi5PcIkwG9AH5s/ryumVaZmfRgesadJt5RiqQSoA===)), but I'm not too knowledgeable on the async/await internals.

In any case, I've pushed the version you suggested. 😄

src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -2016,6 +2016,21 @@ public void Dispose()
...
+        /// <summary>
+        /// Release all resources associated with this object.
+        /// </summary>
+        public async ValueTask DisposeAsync()
+        {
+            GC.SuppressFinalize(this);
+            await CloseAsync(!_isDisposed);
+            if (sentinelConnection is ConnectionMultiplexer sentinel)
+            {
+                await (sentinel?.DisposeAsync() ?? default);
```
NickCraver, line 2028:
```suggestion
                await sentinel.DisposeAsync();
```

NickCraver (changes requested):

src/StackExchange.Redis/ConnectionMultiplexer.cs
NickCraver: Tried to be clear in comment - the `is ConnectionMultiplexer` is doing the null check here - there's zero chance it's null here so let's keep it simple please!

NickCraver: To add: I would fix this and commit but since the checkbox isn't enabled for us to push changes or PR, can only suggest here.

kimsey0: You were perfectly clear. I just messed up the commit after testing the two solutions. I also thought you'd be able to push commits, but maybe it wasn't enabled because I created this first as a draft pull request. Sorry for the inconvenience.

kimsey0: _Allow edits by maintainers_ is enabled. I don't know why it doesn't work.

NickCraver (approved):
Looking good, thank you!

PR #2166, opened 2022-06-21
PrepareScript should work for parameterless scripts; fix #2164

### LINKED ISSUE #2164: Lua Script Cannot be Parsed after upgrading to 2.6.45
I've upgraded `StackExchange.Redis`  nuget package to the latest version 2.6.45 and I've noticed that my previously working Lua script is no longer accepted at startup by Stack Exchange library. The exception thrown is 

```
System.ArgumentException: Count not parse script: [...]
-------------------------------------------------------------------------------
-- API definitions
-------------------------------------------------------------------------------
local MessageStoreAPI = {}

MessageStoreAPI.confirmPendingDelivery = function(smscMessageId, smscDeliveredAt, smscMessageState)
    local messageId = redis.call('hget', "smId:" .. smscMessageId, 'mId')
    if not messageId then
        return nil
    end
    -- delete pending delivery
    redis.call('del', "smId:" .. smscMessageId)

    local mIdK = 'm:'..messageId

    local result = redis.call('hsetnx', mIdK, 'sState', smscMessageState)
    if result == 1 then
        redis.call('hset', mIdK, 'sDlvAt', smscDeliveredAt)
        redis.call('zrem', "msg.validUntil", messageId)
        return redis.call('hget', mIdK, 'payload')
    else
        return nil
    end
end


-------------------------------------------------------------------------------
-- Function lookup
-------------------------------------------------------------------------------

-- None of the function calls accept keys
if #KEYS > 0 then error('No Keys should be provided') end

-- The first argument must be the function that we intend to call, and it must
-- exist
local command_name = assert(table.remove(ARGV, 1), 'Must provide a command as first argument')
local command      = assert(MessageStoreAPI[command_name], 'Unknown command ' .. command_name)

return command(unpack(ARGV))

   at StackExchange.Redis.ScriptParameterMapper.PrepareScript(String script) in /_/src/StackExchange.Redis/ScriptParameterMapper.cs:line 145
   at StackExchange.Redis.LuaScript.Prepare(String script) in /_/src/StackExchange.Redis/LuaScript.cs:line 92
```

The very same script is working when using a previous version of StackExchange.Redis. Any hint on what's going on? The exception's message is not very helpful at pointing the error.

mgravell:
fix breaking change introduced [here](https://github.com/StackExchange/StackExchange.Redis/commit/75da236160c244f1ea0246f20d849519a6875acd#diff-fae8ea62c94ed93bcabb924f204ecff8a3e66a2822d3629f951ed5f22172b69dR138) - fixes #2164

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -2,7 +2,7 @@
 
 ## Pending
 
-- Fix: [#2164](https://github.com/StackExchange/StackExchange.Redis/issues/2164): fix `LuaScript.Prepare` for scripts that don't have parameters ([#2066 by MarcGravell](https://github.com/StackExchange/StackExchange.Redis/pull/2066))
+- Fix: [#2164](https://github.com/StackExchange/StackExchange.Redis/issues/2164): fix `LuaScript.Prepare` for scripts that don't have parameters ([#2166 by MarcGravell](https://github.com/StackExchange/StackExchange.Redis/pull/2166))
 
 ## 2.6.45
 


NickCraver (approved):
👍 Looking good

svalbuena:
When are you planning to release this change @mgravell ? The CI build didn't publish it as a test failed

svalbuena:
it worked! thanks!!

svalbuena:
Do you know when this will get published on nuget.org? I can only fetch it from there for production

mgravell:
we discussed this in a meeting on Tuesday - I think we're planning to give it a day or so to see if anything else comes in from the recent release that needs urgent attention, and then get this out; how about if we haven't seen anything by Tuesday (simply because we have a recurring coordination meeting on Tuesdays), we get the release sorted out?

Issue #1968, opened 2022-01-26
Automatically detect and reupload flushed LUA script


ORuban:
Though executed scripts are guaranteed to be in the script cache of a given execution of a Redis instance forever, things like Redis instance restart do flush the script cache. This causing RedisServerException with `"NOSCRIPT No matching script. Please use EVAL."` errors on client side. Right now the recovery step is to explicitly reapload the script.

It would be awesome if LUA script flushing could be handled by the library internally. It would also be a good fit to what [EVAL command documentation](https://redis.io/commands/eval) describes:

> Practically speaking, for the client it is much better to simply assume that in the context of a given connection, cached scripts are guaranteed to be there unless an administrator explicitly called the SCRIPT FLUSH command.

mgravell:
For the ScriptEvaluate use-case, this should already be the case. Do you have an example of the API you're using when this is failing?

ORuban:
we are using `IDatabaseAsync.ScriptEvaluateAsync`, in particular this overload
```
Task<RedisResult> ScriptEvaluateAsync(byte[] hash, RedisKey[] keys = null, RedisValue[] values = null, CommandFlags flags = CommandFlags.None);
```

Just in case, we deal with `IDatabase` instance that is received via `IConnectionMultiplexer.GetDatabase()`
```
IConnectionMultiplexer multiplexer = ...
var database = multiplexer.GetDatabase();
database.ScriptEvaluateAsync(hash, keys, values, CommandFlags.None);
```

NickCraver:
@ORuban which version of the library are you on? I'm wondering if you have a version from before this was handled.

ORuban:
@NickCraver we are having a fork that last time was merged with StackExchange.Redis master at Nov, 13.
Please, as you say some work has been done, may you point to related threads for reference so we could learn what was changed? Just asking as I had tried to search accross any recent commits, issues using "Lua" /"script" as keyworld, but failed to find anything related.

mgravell:
it looks like you're only supplying the hash; in that scenario, it can't possibly reconstruct the script - all it has is the hash; it isn't going to try and brute-force a script that has the same hash :)

if you use the overload that takes the `string` version, it still *uses* `EVALSHA` - it performs the hashing internally (IIRC, the first time it sees a new script, it performs an `EVAL` and `SCRIPT LOAD` - only switching to `EVALSHA` when it has actually seen a success response from the `SCRIPT LOAD`)

[...]

PR #2170, opened 2022-06-23
Allow LoadedLuaScript fail-over in case the Redis instance is restarted

martintmk:
**Motivation**

We want to take advantage of sending just hashed `LoadedLuaScript` to a server, but still, be able automatically failover and reconstruct the hash in case the Redis instance restarts.

The goal was not to change the public API surface.

Adresses #1968

Unit tests are still missing, first I want to check with you guys whether this is the right approach.

### DIFF
--- a/src/StackExchange.Redis/Interfaces/IDatabase.cs
+++ b/src/StackExchange.Redis/Interfaces/IDatabase.cs
@@ -1272,7 +1272,8 @@ namespace StackExchange.Redis
         /// <param name="values">The values to execute against.</param>
         /// <param name="flags">The flags to use for this operation.</param>
         /// <returns>A dynamic representation of the script's result.</returns>
-        /// <remarks><seealso href="https://redis.io/commands/evalsha"/></remarks>
+        /// <remarks>Please do not use this overload as it's not resilient to Redis server restarts. Use <see cref="ScriptEvaluate(string, RedisKey[], RedisValue[], CommandFlags)"/> instead.</remarks>
+        [EditorBrowsable(EditorBrowsableState.Never)]
         RedisResult ScriptEvaluate(byte[] hash, RedisKey[]? keys = null, RedisValue[]? values = null, CommandFlags flags = CommandFlags.None);
 
         /// <summary>
--- a/src/StackExchange.Redis/Interfaces/IDatabase.cs
+++ b/src/StackExchange.Redis/Interfaces/IDatabase.cs
@@ -1300,17 +1300,6 @@ namespace StackExchange.Redis
         RedisResult ScriptEvaluate(LoadedLuaScript script, object? parameters = null, CommandFlags flags = CommandFlags.None);
 
         /// <summary>
-        /// Execute a Lua script against the server using just the SHA1 hash.
-        /// </summary>
-        /// <param name="script">The loaded script to execute.</param>
-        /// <param name="keys">The keys to execute against.</param>
-        /// <param name="values">The values to execute against.</param>
-        /// <param name="flags">The flags to use for this operation.</param>
-        /// <returns>A dynamic representation of the script's result.</returns>
-        [EditorBrowsable(EditorBrowsableState.Never)]
-        RedisResult LoadedScriptEvaluate(LoadedLuaScript script, RedisKey[]? keys = null, RedisValue[]? values = null, CommandFlags flags = CommandFlags.None);
-
-        /// <summary>
         /// Add the specified member to the set stored at key.
         /// Specified members that are already a member of this set are ignored.
         /// If key does not exist, a new set is created before adding the specified members.
--- a/src/StackExchange.Redis/Interfaces/IDatabase.cs
+++ b/src/StackExchange.Redis/Interfaces/IDatabase.cs
@@ -1272,7 +1272,7 @@ namespace StackExchange.Redis
         /// <param name="values">The values to execute against.</param>
         /// <param name="flags">The flags to use for this operation.</param>
         /// <returns>A dynamic representation of the script's result.</returns>
-        /// <remarks>Please do not use this overload as it's not resilient to Redis server restarts. Use <see cref="ScriptEvaluate(string, RedisKey[], RedisValue[], CommandFlags)"/> instead.</remarks>
+        /// <remarks>Be aware that this method is not resilient to Redis server restarts. Use <see cref="ScriptEvaluate(string, RedisKey[], RedisValue[], CommandFlags)"/> instead.</remarks>
         [EditorBrowsable(EditorBrowsableState.Never)]
         RedisResult ScriptEvaluate(byte[] hash, RedisKey[]? keys = null, RedisValue[]? values = null, CommandFlags flags = CommandFlags.None);
 
--- a/src/StackExchange.Redis/Interfaces/IDatabase.cs
+++ b/src/StackExchange.Redis/Interfaces/IDatabase.cs
@@ -1300,6 +1300,17 @@ namespace StackExchange.Redis
         RedisResult ScriptEvaluate(LoadedLuaScript script, object? parameters = null, CommandFlags flags = CommandFlags.None);
 
         /// <summary>
+        /// Execute a Lua script against the server using just the SHA1 hash.
+        /// </summary>
+        /// <param name="script">The loaded script to execute.</param>
+        /// <param name="keys">The keys to execute against.</param>
+        /// <param name="values">The values to execute against.</param>
+        /// <param name="flags">The flags to use for this operation.</param>
+        /// <returns>A dynamic representation of the script's result.</returns>
+        [EditorBrowsable(EditorBrowsableState.Never)]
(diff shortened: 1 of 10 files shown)


mgravell:
I've had a look at this, and: I think this PR is doing a great job of illustrating the problem, but I'm not sure it is the best direction for the solution. From what I can see, I *think* we can fix everything here much more simply and consistently, simply by changing `LoadedLuaScript` to pass `ExecutableScript` instead of `Hash` when calling `db.ScriptEvaluate[Async](...)`; we would run the lookup against the expected cache automatically, and I'm fine with paying a dictionary trygetvalue for simpler code.

There *is* a pre-existing inconsistency between the sync and async versions of `ScriptEvaluate[Async](string, ...)`, and honestly: I think we should just eat the `async`/`await` here, and make the code consistent - i.e. reinstate the `try`/`catch` (and detect missing script) from the sync version into the async version.

Thoughts?

martintmk:
Thanks Mark, I am all for simplicity :). If your approach is still able to internally reconstruct and reuse the hash, then it's a better, simpler approach. 

I will give this a shot after the weekend.

martintmk:
@mgravell , I did take a look at the implementation and I think the approach you proposed should work as we will be hitting this code-path (the hashed value is sent):

https://github.com/StackExchange/StackExchange.Redis/blob/d6e05f656a7194a041fd369da779b922ab5c01f0/src/StackExchange.Redis/RedisDatabase.cs#L4804

The only remaining problem is the `withKeyPrefix` parameter, which is not currently accepted by the `IDatabase` API:
https://github.com/StackExchange/StackExchange.Redis/blob/d6e05f656a7194a041fd369da779b922ab5c01f0/src/StackExchange.Redis/LuaScript.cs#L265

Are we OK with expanding the public API or should I stick with the `ParameterAndKeyPrefixHolder` trick?

martintmk:
> Re key prefix; the existing implementation also didn't use any privileged APIs - it just expanded them as part of ExtractParameters - so IMO we should just keep doing that exactly the same

Yeah, silly me didn't notice that we can just reuse the existing helper. Fixed.

> Re LoadedLuaScript at all; we can't remove it because of API signature breaking; it does still at least usually offer some benefit re pre-load of the script; not a massive benefit, perhaps, but: ... I'm not sure it is worth doing much in either direction here, i.e. I'm not convinced we should mark it [Obsolete] either.

Fine by me, are we OK with not using the SHA hash at all then in this code path?

src/StackExchange.Redis/RedisDatabase.cs
```diff
@@ -1538,13 +1538,22 @@ public RedisResult ScriptEvaluate(LuaScript script, object? parameters = null, C
...
 
-        public Task<RedisResult> ScriptEvaluateAsync(string script, RedisKey[]? keys = null, RedisValue[]? values = null, CommandFlags flags = CommandFlags.None)
+        public async Task<RedisResult> ScriptEvaluateAsync(string script, RedisKey[]? keys = null, RedisValue[]? values = null, CommandFlags flags = CommandFlags.None)
         {
             var msg = new ScriptEvalMessage(Database, flags, script, keys, values);
-            return ExecuteAsync(msg, ResultProcessor.ScriptResult, defaultValue: RedisResult.NullSingle);
+
+            try
+            {
+                return await ExecuteAsync(msg, ResultProcessor.ScriptResult, defaultValue: RedisResult.NullSingle);
```
martintmk: I just copied the `ScriptEvaluate`, however not sure if the fallback is even needed here. 

When debugging and restarting the Redis the catch branch was never reached.

src/StackExchange.Redis/Interfaces/IDatabase.cs
```diff
@@ -1272,7 +1272,8 @@ public interface IDatabase : IRedis, IDatabaseAsync
         /// <param name="values">The values to execute against.</param>
         /// <param name="flags">The flags to use for this operation.</param>
         /// <returns>A dynamic representation of the script's result.</returns>
-        /// <remarks><seealso href="https://redis.io/commands/evalsha"/></remarks>
+        /// <remarks>Please do not use this overload as it's not resilient to Redis server restarts. Use <see cref="ScriptEvaluate(string, RedisKey[], RedisValue[], CommandFlags)"/> instead.</remarks>
```
mgravell, line 1275: I think it is fine to offer this as an *advisory comment*, but: I'd hesitate to say "please do not" - more just something that gives relevant warning; ditto on the EditorBrowsableState.Never

mgravell:
I'd much rather *not* make the API change here - meaning: the new API added that passes `LoadedLuaScript` down to avoid the regex; instead, I've tested locally, and if we tweak `IsSHA1` trivially: we can just say "yup, that's fine"; I did a test locally of how expensive this regex is:

1000000 tests: 81ms
1000000 tests: 53ms
1000000 tests: 54ms
1000000 tests: 52ms
1000000 tests: 55ms
1000000 tests: 45ms
1000000 tests: 43ms
1000000 tests: 43ms
1000000 tests: 43ms
1000000 tests: 46ms

If we assume that almost all non-SHA scripts *aren't* exactly 40 characters, we can add a length test:

``` c#
internal static bool IsSHA1(string? script) => script is not null && script.Length == 40 && sha1.IsMatch(script);
```

and we get:

1000000 tests: 0ms
1000000 tests: 0ms
1000000 tests: 0ms
1000000 tests: 0ms
1000000 tests: 0ms
1000000 tests: 0ms
1000000 tests: 0ms
1000000 tests: 0ms
1000000 tests: 0ms
1000000 tests: 0ms

I'd much rather do that, and keep the API simple; thoughts?

mgravell:
(FWIW: I experimented with a manually rolled test of the actual character test - no significant improvement, so: no point rewriting that IMO - just leave it as regex; it turns out that regex has been optimized pretty well for this type of scenario)

src/StackExchange.Redis/Interfaces/IDatabase.cs
```diff
@@ -1298,6 +1299,17 @@ public interface IDatabase : IRedis, IDatabaseAsync
...
+        /// <summary>
+        /// Execute a Lua script against the server using just the SHA1 hash.
+        /// </summary>
+        /// <param name="script">The loaded script to execute.</param>
+        /// <param name="keys">The keys to execute against.</param>
+        /// <param name="values">The values to execute against.</param>
+        /// <param name="flags">The flags to use for this operation.</param>
+        /// <returns>A dynamic representation of the script's result.</returns>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        RedisResult LoadedScriptEvaluate(LoadedLuaScript script, RedisKey[]? keys = null, RedisValue[]? values = null, CommandFlags flags = CommandFlags.None);
```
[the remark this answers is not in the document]
martintmk: That's correct. Not directly exposed.

src/StackExchange.Redis/Interfaces/IDatabase.cs
martintmk: Done.

martintmk:
> Re key prefix; the existing implementation also didn't use any privileged APIs - it just expanded them as part of `ExtractParameters` - so IMO we should just keep doing that exactly the same
> 
> Re `LoadedLuaScript` at all; we can't _remove_ it because of API signature breaking; it _does_ still at least usually offer some benefit re pre-load of the script; not a _massive_ benefit, perhaps, but: ... I'm not sure it is worth doing much in either direction here, i.e. I'm not convinced we should mark it `[Obsolete]` either.

How about we do these changes:

* Given that there is no real perf benefit based on your measurements the `LoadedLuaScript.Evaluate` just passes the string to evaluate the method. The hash is just ignored. This makes it resilient to Redis restarts. 
* Can we hide it from the API? At least it will not attract some unnecessary attention.

mgravell:
If you're asking about hiding the hash property: yes, we can safely IMO hide that in every way - [Browsable], [EditorBrowsable] and [Obsolete] (with the latter as a warning only, not an error). Sound good?

martintmk:
> 



> If you're asking about hiding the hash property: yes, we can safely IMO hide that in every way - [Browsable], [EditorBrowsable] and [Obsolete] (with the latter as a warning only, not an error). Sound good?

And what do you think about hiding everything around `LoadedLuaScript`? Or is that too much?

The only real added value is that script pre-load which can be done by:

``` csharp
server.ScriptLoad(LuaScript.ExecutableScript, flags)
```

mgravell:
It's a great question, but I think I'd rather defer on that for now

martintmk:
> It's a great question, but I think I'd rather defer on that for now

Sounds good,  pushed the changes and reverted any API additions. Let me know if the approach is OK, then I'll focus on tests.

mgravell (commented):
approach looks good 👍

martintmk:
> Can you please add the `.Length == 40` check as per the comments above? thanks

Done.

martintmk:
@mgravell Pushed the changes. Anything left to address?

Unfortunately, some random tests failed on Ubuntu should I just retry?

NickCraver (approved):
Current looks good - thanks for the iterations on this to make it non-breaking and beneficial at the same time :)

Issue #2176, opened 2022-06-26
Batch issue with Sorted Sets in version 2.6.45


mikechoinoski:
The latest 2.6.45 version seems to have an issue with batches that previous versions did not.

In the test below a sorted set is created. A batch is then created where another sorted set is created, then an intersect is done between the two sorted sets, then the data should be returned in with SortedSetRangeByRankAsync command. In prior versions, this worked well but in this version no data is returned in the SortedSetRangeByRankAsync command.

It does seem like the first two commands are being executed in the correct order as the intersect key does exist after the test below completes with the correct value. I'm wondering if the SortedSetRangeByRankAsync isn't being executed last though.

In the newest version, I also tried the SortedSetCombineAsync instead of the separate intersect and return data commands and it didn't return any data either.

        using var conn = Create();
        var db = conn.GetDatabase();

        var key = "TestKeyForRedis1";
        var key2 = "TestKeyForRedis2";
        var keyIntersect = "TestKeyForRedisIntersect";

        db.KeyDelete(key);
        db.KeyDelete(key2);
        db.KeyDelete(keyIntersect);
        db.SortedSetAdd(key, "a", 1345);

        var tasks = new List<Task>();
        var batch = db.CreateBatch();
        tasks.Add(batch.SortedSetAddAsync(key2, "a", 4567));
        tasks.Add(batch.SortedSetCombineAndStoreAsync(SetOperation.Intersect,
            keyIntersect, new RedisKey[] { key, key2 }));
        var rangeByRankTask = batch.SortedSetRangeByRankAsync(keyIntersect);
        tasks.Add(rangeByRankTask);
        batch.Execute();

        Task.WhenAll(tasks.ToArray());

        var rangeByRankSortedSetValues = rangeByRankTask.Result;

        int size = rangeByRankSortedSetValues.Length;
        Assert.Equal(1, size);
        string firstRedisValue = rangeByRankSortedSetValues.FirstOrDefault().ToString();
        Assert.Equal("a", firstRedisValue);

mgravell:
Definitely sounds weird. Will look.

mgravell:
oof; confirmed - the batch is somehow out of order; I see:

``` txt
1656409461.024216 [0 127.0.0.1:53158] "UNLINK" "TestKeyForRedis1"
1656409461.024681 [0 127.0.0.1:53158] "UNLINK" "TestKeyForRedis2"
[...]

Issue #2167, opened 2022-06-21
Streams Transaction Issue


askoutaris:
All methods inside RedisDatabase.cs which call ExecuteAsync<T> with parameter 'T defaultValue' (eg StreamReadGroupAsync) does not work with transactions since there is no override ExecuteAsync<T> with parameter 'T defaultValue' in RedisTransaction.cs and as a result ExecuteAsync<T> of RedisBase.cs is used instead of  RedisTransaction.cs

FYI: I downloaded the code and managed it to work by adding the overriden methods shown in the attached screenshot.

![2022-06-21 17_49_48-RedisStreams - Microsoft Visual Studio Preview](https://user-images.githubusercontent.com/13051434/174829740-9db4227d-6754-4750-93c3-c903fa215720.png)

mgravell:
agreed, this looks like a problem; do you have any concrete examples of failure cases here? honestly, the best thing is probably to submit this as a PR including tests

askoutaris:
Thank you for your feedback!
I'll try to find some time in the next days to study and follow all Github's procedures about branching and PR policies in order to help you with that.

PR #2177, opened 2022-06-28
URGENT Fix error in batch/transaction handling

mgravell:
A missing API override in `RedisBatch`/`RedisTransaction` can lead to some operations **NOT ENTERING THE BATCH/TRANSACTION**. This is very bad.

Some undefined set of operations (including at least `ZRANGE`) end up using what appears to be a new overload/path; this API was not overridden in the batch/transaction, with the impact that affected operations use the default behaviour and go directly to the multiplexer/connection. This means a: they are issued even if the transaction/batch *is not ever executed*, and b: if everything is issued, they are issued out-of-order (i.e. before the main batch/transaction)

See: #2167, #2176


NickCraver (approved):
Current looks good, but I've also committed - needs cross-check eyes.

mgravell:
Random thought; do both methods need to be virtual? Should we devirtualize the method that doesn't take the default value? One fewer thing to remember.

NickCraver:
hmm not sure I follow (in text-only atm) - aren't we overriding it in Batch/Transaction both?

mgravell:
> hmm not sure I follow (in text-only atm) - aren't we overriding it in Batch/Transaction both?

I'm saying "what if there was only one method to override"

PR #2183, opened 2022-07-08
moving pre 2.6.45 write commands that were not previously considered primary-only out of primary-only

### LINKED ISSUE #2182: CRITICAL: SORT is now rejected in all replicas
I see this commit started to refuse SORT command on replicas:
https://github.com/StackExchange/StackExchange.Redis/commit/989528f498d5861e8af11cdea7f6e8d841525532

This doesn't seem right. We can use SORT in replicas, the replicas just need to be set as writable. This is a bug that was fixed in Redis 7.

In summary, SORT is a valid command on replicas, in any redis version it that the command exists, it's just not valid on READONLY Replicas

slorello89:
Fixes #2182 

@NickCraver for background in #2101, we changed some of the logic around primary-only to require all commands to be explicitly stated as primary-only or not. During this effort, I went through the command list and checked to see which were considered write commands (which would be rejected by a replica) so that there would be consistency in behavior when stumbling on commands that would ordinarily be rejected by a replica. This made a slightly more restrictive list of commands. 

Enter #2182, where this inconsistency had evidently become load-bearing. The user has evidently designated their replicas as writeable (I think I was subconsciously aware of this capability but have never actually seen anyone use it). As it turns out By resolving this inconsistency in #2101 and the follow-on issue when I introduced `SORT_RO` in #2111 I apparently introduced a break.

This PR reverts all the pre-2.6.45 commands' primary vs replica disposition back to their previous state which will remove the break. Any command that was introduced in 2.6.45 is correctly dispositioned.


NickCraver:
@slorello89 recommend continuing with a revert here to a known good. If we want to get smarter about it later we can, but let's switch back to a state that caused fewer issues overall :)

NickCraver (approved):
Looking good :+1:

PR #2191, opened 2022-07-18
adding some smarts to GetFeatures

### LINKED ISSUE #2016: RedisCommandException for "Command cannot be issued to a replica" is difficult to handle
Hey all,

I work on a large service using Azure Redis Premium. Our normal pattern for using Redis is something like:

try{
  db.StringGet()
}
catch (RedisException)
...
catch (TimeoutException)
...

This works pretty well, but we have some low level of failures that look like this:

RedisCommandException: Exception: 'Command cannot be issued to a slave: UNLINK <key>'

(In newer versions of the SDK, it says "Command cannot be issued to a replica" instead.)

This is a reasonable error, but I am not intentionally targeting a replica. I assume Azure has failed over from under me, changing the recipient to a replica, while it was primary when the command was sent.

I would like to just treat this as a transient failure, log it, and move on with life as though it were a timeout. This is a little tricky though because the exception is RedisCommandException (which doesn't inherit from RedisException). I am hesitant to catch and handle RedisCommandException, because I understand that it usually means the command is bad, like maybe a bad LUA script.

Basically, I'd like some way to differentiate this exception so I can treat it as transient.

My suggestion is to make this case a RedisException or RedisServerException, but I'm open to other resolutions, including being told that I should be catching some different set of exceptions as transient.

Thanks,
Adam

slorello89:
@NickCraver - This will fix #2016 

The issue is that when `GetFeatures` does the server selection, it uses `PING` as the command it's asking about, so when the multiplexer responds with a server, it responds with any server. Of course, then when it goes to execute the command, it's been explicitly handed a server, so it honors that choice but checks it to make sure it's a valid server to send the command to, so if you're executing a write command and `GetFeatuers` just so happened to output a read-only replica, when the multiplexer does the 'is this a valid server?' check, it determines it is not and propagates the error. that's what causes the blowup. By passing in the RedisCommand to `GetFeatures`, we allow it to make an informed choice of server for that Redis Command.

The other option is to simply not pass the server on down the line when we we've done a feature check, and allow the muxer to make it's own decision, this would cause an extra run of `Select` which the current pattern e.g. see [sort](https://github.com/StackExchange/StackExchange.Redis/blob/main/src/StackExchange.Redis/RedisDatabase.cs#L1816) tries to avoid

One weird case, with SORT/SORT_RO, if the RO command is what we prefer to run (because we don't have a destination key to output it to), and `SORT_RO` is not available (because they aren't on Redis 7+), we cannot trust the output of GetFeatures, so I just set the selected server back to null and let the muxer figure it out down the line.

The rub with this PR, though is that I'm not 100% sure about how to go about testing this - do you have any thoughts?

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -4,6 +4,7 @@
 
 - Fix [#2182](https://github.com/StackExchange/StackExchange.Redis/issues/2182): Be more flexible in which commands are "primary only" in order to support users with replicas that are explicitly configured to allow writes ([#2183 by slorello89](https://github.com/StackExchange/StackExchange.Redis/pull/2183))
 - Adds: `IConnectionMultiplexer` now implements `IAsyncDisposable` ([#2161 by kimsey0](https://github.com/StackExchange/StackExchange.Redis/pull/2161))
+- Fix [#2016](https://github.com/StackExchange/StackExchange.Redis/issues/2016): Align server selection with supported commands (e.g. with writable servers) to reduce `Command cannot be issued to a replica` errors ([#2191 by slorello89](https://github.com/StackExchange/StackExchange.Redis/pull/2191))
 
 ## 2.6.48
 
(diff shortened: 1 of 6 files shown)


NickCraver:
@slorello89 had to find time to wrap my head around it - I don't think there's a good existing way to ensure this, will keep thinking on it, but I see the issue/fix and don't think we should delay getting better in place while we figure out regression testing on it :)

slorello89:
Sounds good @NickCraver

mgravell (approved):

PR #2197, opened 2022-07-26
Use 3 parts as assembly version

Kielek:
Found while implementing OpenTelemetry AutoIntrumention https://github.com/open-telemetry/opentelemetry-dotnet-instrumentation/pull/979

It will be useful to have a possibility to detect assembly version on runtime. As you are using 3 parts as a release version, the precision should be set to `build`. If you prefer have all 4 parts, it can be changed to `revision`.

Before changes

```csharp
[assembly: System.Reflection.AssemblyVersionAttribute("2.0.0.0")]
[assembly: System.Reflection.AssemblyFileVersionAttribute("2.6.53.65123")]
[assembly: System.Reflection.AssemblyInformationalVersionAttribute("2.6.53+63fe68c7de")]
```

After changes
```csharp
[assembly: System.Reflection.AssemblyVersionAttribute("2.6.53.0")]
[assembly: System.Reflection.AssemblyFileVersionAttribute("2.6.53.65123")]
[assembly: System.Reflection.AssemblyInformationalVersionAttribute("2.6.53+63fe68c7de")]
```

### DIFF
--- a/version.json
+++ b/version.json
@@ -1,7 +1,9 @@
 {
   "version": "2.6",
   "versionHeightOffset": -1,
-  "assemblyVersion": "2.0",
+  "assemblyVersion": {
+    "precision": "build"
+  },
   "publicReleaseRefSpec": [
     "^refs/heads/main$",
     "^refs/tags/v\\d+\\.\\d+"


mgravell:
This is not accidental, and is a pattern pretty common among a wide range of libraries. The reality is: having a granular assembly-version places a very large burden on consumers, in particular (but not quite exclusively) on .NET Framework - mostly due to assembly binding redirect problems. Having this break constantly creates unnecessary friction, which is often avoided by pinning the assembly version to majors. Or at a push: major+minor.

My suggestion here would be for OpenTelemetry AutoIntrumention should be tweaked to be a little sympathetic to this intent, perhaps considering other attributes.

Kielek:
Thanks for the explanation. I will be just easier for us in the future to maintain this functionality. For now, there is a way how we can handle it differently.

Issue #2020, opened 2022-03-04
Reading from streams with command flag "PreferReplica", does not recover after rebooting replica node


stedmanm:
We have an Azure Redis cluster (2 masters, 2 replicas), and are using SE version `2.2.88` in a `.NET Core 3.1` service. We are reading streams from the replica nodes (using `PreferReplica` command flag). After we reboot the replica node where the stream is located, we get continuous errors when trying to read from the stream, it never seems to recover. The error that we get is `Key has MOVED to Endpoint xx.xx.xx.xx:port and hashslot 0 but CommandFlags.NoRedirect was specified - redirect not followed for XREAD`. We are able to consistently reproduce this issue with the code snippet below. We are assuming this issue is related to streams, because if we replace `StreamReadAsync` with `StringGetAsync` we can't reproduce the issue. Also, we can't reproduce the issue if we read the stream without specifying the command flag, and reboot the master node.
``` C#
var options = ConfigurationOptions.Parse(
    "host123:1234,password=123,ssl=True,abortConnect=False");

options.CommandMap = CommandMap.Create(new HashSet<string> { "SUBSCRIBE" }, false);
options.CertificateValidation += (sender, certificate, chain, sslPolicyErrors) => true;

var connectionMultiplexer = await ConnectionMultiplexer.ConnectAsync(options);

const string key = "{06S}TestPreferReplicaStream";

await connectionMultiplexer.GetDatabase().StreamAddAsync(key, "Field", "hello world from shard 0");

await connectionMultiplexer.GetDatabase().KeyExpireAsync(key, TimeSpan.FromHours(1));

while (true)
{
    try
    {
        await connectionMultiplexer.GetDatabase().StreamReadAsync(key, "0-0",
            1000, CommandFlags.PreferReplica);
    }
    catch (Exception e)
    {
        await Task.Delay(100);
        Console.WriteLine("------------------------------------------------------------------------------------------------------------");
        Console.WriteLine($"{DateTime.Now}: " + e);
    }
}
```
Below is the output from the code snippet above.
```
------------------------------------------------------------------------------------------------------------
[...]

PR #2201, opened 2022-07-28
Add more error information when the cast fails

rkarthick:
When redis/envoy  proxy servers return error, currently we do not propagate the error string to the client. Client gets an errors

`Cannot convert to RedisValue: Error`

It will be nice to propagate the error that is sent by the upstream servers here (redis or proxy) to the client for better observability.

After this PR we should see the errors similar to the following:

`Cannot convert to RedisValue: Error: upstream link failure`

### DIFF
--- a/src/StackExchange.Redis/RawResult.cs
+++ b/src/StackExchange.Redis/RawResult.cs
@@ -134,14 +134,14 @@ namespace StackExchange.Redis
                     }
                     return default;
                 default:
-                    throw new InvalidCastException("Cannot convert to RedisChannel: " + Type);
+                    throw new InvalidCastException("Cannot convert to RedisChannel: " + ToString());
             }
         }
 
         internal RedisKey AsRedisKey() => Type switch
         {
             ResultType.SimpleString or ResultType.BulkString => (RedisKey)GetBlob(),
-            _ => throw new InvalidCastException("Cannot convert to RedisKey: " + Type),
+            _ => throw new InvalidCastException("Cannot convert to RedisKey: " + ToString()),
         };
 
         internal RedisValue AsRedisValue()
@@ -157,7 +157,7 @@ namespace StackExchange.Redis
                 case ResultType.BulkString:
                     return (RedisValue)GetBlob();
             }
-            throw new InvalidCastException("Cannot convert to RedisValue: " + Type);
+            throw new InvalidCastException("Cannot convert to RedisValue: " + ToString());
         }
 
         internal Lease<byte>? AsLease()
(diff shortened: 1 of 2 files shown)


mgravell:
I think this is probably OK for error scenarios, but: if I understand the PR, this could also potentially include the value in the case of non-error types. That is undesirable, as it could leak data into exception logs - or at the very least, it can only be done if the relevant "include detail in exceptions" flag is set. IMO we should explicitly limit this change to error types. If I've misunderstood: let me know!

mgravell:
Thanks for tweaking that.

Secondary question: how can we observe this behaviour? Normally, I would have expected the library to have already converted this to an exception, so I'm a little surprised that we're seeing this happen at all. Is there any specific scenario where you're seeing this? And: is this testable?

rkarthick:
Thank you for feedback and request for more details.

I think the error message from server is getting converted rightly as error here: https://github.com/StackExchange/StackExchange.Redis/blob/main/src/StackExchange.Redis/Message.cs#L539 

May be the library throws an error only when it is not able to RESP decode? https://github.com/StackExchange/StackExchange.Redis/blob/main/src/StackExchange.Redis/PhysicalConnection.cs#L1883

Repro steps below:

1. Connect to redis envoy proxy which fronts 3 redis masters and 3 redis replicas
2. Bring down one of the masters and its replica
3. Try to get a key that belongs to that master
4. Envoy proxy returns an error `upstream link failure`

https://www.envoyproxy.io/docs/envoy/latest/intro/arch_overview/other_protocols/redis#failure-modes
```
$ redis-cli MGET a b c d e
1) "alpha"
2) "bravo"
3) (error) upstream failure
4) (error) upstream failure
5) "echo"
```

Testing would be not straightforward using the current CI setup (technically we could modify supervisord to disable restarts and kill -9 one of the redis servers to test it). However I am not sure if we wanted to do that for the exception scenario.

I am pushing in a testcase for the error too. Do let me know if that makes sense

NickCraver:
@rkarthick We still have a fundamental issue here in that this isn't checking if detailed exceptions are enabled - it's unconditionally logging the information which could be sensitive. For instance if `ConfigurationOptions.IncludeDetailInExceptions` is false, this shouldn't log. But...it's not easy to get that here, without adding overhead (e.g. a ref which gets very tricky/expensive/dangerous here). Without this, we can't add the information to the exception since that'll cause problems with users in the multi-tenant/security-sensitive environments.

src/StackExchange.Redis/RawResult.cs
```diff
@@ -133,6 +133,8 @@ internal RedisChannel AsRedisChannel(byte[]? channelPrefix, RedisChannel.Pattern
                         return new RedisChannel(copy, mode);
                     }
                     return default;
+                case ResultType.Error:
+                    throw new InvalidCastException("Cannot convert to RedisChannel: " + ToString());
```
mgravell, line 137: I think we should reduce this down to the raw redis error; my suggestion:

``` c#
throw new RedisServerException(GetString());
```

a few differences here:

- we're emphasizing that this is an error coming from the server
- we're keeping the error #2020 
- the use of `GetString` avoids additional words

Can we use this in all the places? with that, I think we're good

rkarthick:
Closing this out as per @NickCraver's comment

PR #2203, opened 2022-07-31
Add ConnectionMultiplexer.GetServers() API

NickCraver:
Making things a bit easier for when you need to iterate over servers for a multiplexer

### DIFF
--- a/src/StackExchange.Redis/ConnectionMultiplexer.cs
+++ b/src/StackExchange.Redis/ConnectionMultiplexer.cs
@@ -1030,6 +1030,20 @@ namespace StackExchange.Redis
         }
 
         /// <summary>
+        /// Obtain configuration APIs for all servers this multiplexer.
+        /// </summary>
+        public IServer[] GetServers()
+        {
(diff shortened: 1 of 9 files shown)


src/StackExchange.Redis/ConnectionMultiplexer.cs
```diff
@@ -1029,6 +1029,20 @@ public IServer GetServer(EndPoint? endpoint, object? asyncState = null)
             return new RedisServer(this, server, asyncState);
         }
 
+        /// <summary>
+        /// Obtain configuration APIs for all servers this multiplexer.
```
philon-msft, line 1033: "for all servers IN this multiplexer"? or "for all servers this multiplexer IS CONNECTED TO"? 

(this string also appears in IConnectionMultiplexer.cs)

philon-msft (approved):

src/StackExchange.Redis/ConnectionMultiplexer.cs
NickCraver: Good catch - fixed!

PR #2221, opened 2022-08-18
Initial POC commit for adding ReadOnlyMemory overloads to StringSet o…

Stabzs:
…perations.


src/StackExchange.Redis/Interfaces/IDatabaseAsync.cs
```diff
@@ -2817,6 +2817,20 @@ public interface IDatabaseAsync : IRedisAsync
...
+        /// If <see cref="When.NotExists"/> is specified, this will not perform any operation at all even if just a single key already exists.
+        /// </summary>
+        /// <param name="values">The keys and values to set.</param>
+        /// <param name="when">Which condition to set the value under (defaults to always).</param>
+        /// <param name="flags">The flags to use for this operation.</param>
+        /// <returns><see langword="true"/> if the keys were set, <see langword="false"/> otherwise.</returns>
+        /// <remarks>
+        /// <seealso href="https://redis.io/commands/mset"/>,
+        /// <seealso href="https://redis.io/commands/msetnx"/>
+        /// </remarks>
```
mgravell, line 2831: possibly worth a warning about FireAndForget being dangerous here if the ReadOnlyMemory is going to be recycled/reused?

Stabzs: Definitely seems wise, thank you. I'll add warnings for all of the new methods taking ReadOnlyMemory. If the approach looks good overall, I'll continue expanding the surface area.

src/StackExchange.Redis/KeyspaceIsolation/WrapperBase.cs
```diff
@@ -804,6 +807,25 @@ internal WrapperBase(TInner inner, byte[] keyPrefix)
...
+            {
+                return outer;
+            }
+            else
+            {
+                KeyValuePair<RedisKey, RedisValue>[] inner = new KeyValuePair<RedisKey, RedisValue>[outer.Length];
+
+                for (int i = 0; i < outer.Length; ++i)
+                {
+                    inner[i] = ToInner(outer.Span[i]);
```
mgravell, line 822: Accessing .Span is relatively expensive; this should be done outside the loop and resused.

mgravell: The ++i is very unusual for C#, note

Stabzs: I'll make the changes to cache all call sites for .Span.

The ++i was copied from line 801 of the diff. I assumed that was intentional and copied it here. Is that an incorrect usage?

mgravell:
Not wrong, just atypical. Consistency/copy-paste is a valid reason here!

Issue #2219, opened 2022-08-18
TLS cipher issues on linux servers in .NET 5+


slorello89:
Hey 👋  @NickCraver @mgravell - this is something of a follow on to #1998,

We've had some customers who run Redis Enterprise on some older OS's (e.g. RHEL 7) which support older tls ciphers run into some meaningful difficulties when trying to connect to Redis when upgrading to .NET 5+, the particular customer in question I'm working with is trying to upgrade to .net 6 but this is backwards to .NET 5 as that's where the [BC](https://docs.microsoft.com/en-us/dotnet/core/compatibility/cryptography/5.0/default-cipher-suites-for-tls-on-linux) was introduced.

The issue only occurs on linux servers, and of course is due to OpenSSL's configuraiton. But some folks have pre-existing images they need use in their entirety, and making even minor configuration changes is extremely difficult. Thus I wanted to see if we could perhaps add a mechanism to remediate this in the client via some means of configuration.

## Reproduction

I wrote a minimal reproduction and put it [here](https://github.com/slorello89/rhel7-issue-minimal-repro) - if you want to see the issue, follow the steps laid out in the README, basically, all it's doing is creating an instance of Redis-enterprise with a very narrow set of ciphers (AES128-SHA256:AES256-SHA256), and then running your .NET app in a separate ubuntu container who's OpenSSL config is not set to handle these ciphers.

## The Rub

The real rub with whatever form this fix would take is that the class & enum used to configure SslStream's configuration settings [SslClientAuthenticationOptions](https://docs.microsoft.com/en-us/dotnet/api/system.net.security.sslclientauthenticationoptions?view=net-6.0) and [TlsCipherSuite](https://docs.microsoft.com/en-us/dotnet/api/system.net.security.tlsciphersuite?view=net-6.0) are simply not available in the .NET standard, so any leverage of them would have to be guarded from the compiler by pre-processor directives.

Now, I have to imagine adding non-.NET standard data structures to the configuration is just not going to fly but I had a couple of ideas as to how we could remediate this

1. Follow the pattern laid out for handling certs (see the `SERedis_ClientCertPfxPath`), I actually wrote a fix that does remediate this issue using an environment variable to pull out the list of acceptable ciphers `SERedis_EnabledCiphers` but this seemed pretty hacky so I don't really want to open a PR
[...]

PR #2224, opened 2022-08-21
SSL/TLS: Add SslClientAuthenticationOptions configurability

NickCraver:
Proposal for #2219, allowing explicit configuration of the SSL connection for advanced use cases. This needs some thought, but pitching the general idea here as an option available for the frameworks that support it.


slorello89 (commented):
@NickCraver - this LGTM in principle, my initial thought was to avoid any public API surface area to make it a more palatable addition. But I think this is way more usable and enduring and will give users maximal flexibility moving forward.

Directory.Packages.props
```diff
@@ -11,7 +11,7 @@
     <!-- Packages only used in the solution, upgrade at will -->
     <PackageVersion Include="BenchmarkDotNet" Version="0.13.1" />
     <PackageVersion Include="GitHubActionsTestLogger" Version="2.0.0-alpha" />
-    <PackageVersion Include="Microsoft.CodeAnalysis.PublicApiAnalyzers" Version="3.3.3" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.PublicApiAnalyzers" Version="3.3.4-beta1.22362.3" />
```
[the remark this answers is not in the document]
NickCraver: Yep, support for multiple API lists (to differentiate based on TFM) wasn't added until very recently. Turns out we didn't actually have any conditional APIs for newer versions yet.

btchoum:
@NickCraver 

I like this idea a lot. It solves the immediate issue while giving us more flexibility. I will give it a try in my environment and let you know how it goes.

Thank you so much for considering this issue.

NickCraver:
@btchoum minor but important fix pushed if trying this...otherwise it'll double auth because I'm an idiot who didn't hit save.

src/StackExchange.Redis/PhysicalConnection.cs
```diff
@@ -1432,6 +1432,17 @@ internal async ValueTask<bool> ConnectedAsync(Socket socket, LogProxy? log, Sock
...
+                            if (configOptions is not null)
+                            {
+                                await ssl.AuthenticateAsClientAsync(configOptions);
+                            }
+                            else
+                            {
+                                ssl.AuthenticateAsClient(host, config.SslProtocols, config.CheckCertificateRevocation);
+                            }
+#endif
                             ssl.AuthenticateAsClient(host, config.SslProtocols, config.CheckCertificateRevocation);
```
btchoum, line 1446: @NickCraver 
This pre-processor directive needs a **else** block, otherwise runs ssl.AuthenticateAsClient **twice** in NETCOREAPP3_1_OR_GREATER

Something like this:

``` csharp
#if NETCOREAPP3_1_OR_GREATER
                            var configOptions = config.SslClientAuthenticationOptions?.Invoke(host);
                            if (configOptions is not null)
                            {
                                await ssl.AuthenticateAsClientAsync(configOptions);
                            }
                            else
                            {
                                ssl.AuthenticateAsClient(host, config.SslProtocols, config.CheckCertificateRevocation);
                            }
#else 
                   ssl.AuthenticateAsClient(host, config.SslProtocols, config.CheckCertificateRevocation);
#endif
```

NickCraver: See comment above (and latest code) :)

btchoum: @NickCraver 

So far things are looking good in my environment. I was able to connect to a Redis instance with restricted ciphers from a .NET 6.0 client, both from Windows and Linux.

I needed to wrap the call to the CipherSuitesPolicy constructor in a try/catch because that will fail on Windows. 

Please let me know if the below snippet is how you indent this new API to be used.

``` csharp
var redisConfig = ConfigurationOptions.Parse(_configuration.GetConnectionString("Redis"));

redisConfig.SslClientAuthenticationOptions = (host) =>
{
	var sslOptions = new SslClientAuthenticationOptions
	{
		TargetHost = host
	};

	try
	{
		sslOptions.CipherSuitesPolicy = new CipherSuitesPolicy(new[]
		{
			TlsCipherSuite.TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256,
			// add more allowed ciphers here ...
		});
	}
	catch (PlatformNotSupportedException)
	{
		// Ignore: The CipherSuitesPolicy constructor is expected to fail on Windows. It only works on Linux with OpenSSL 1.1.1 or higher or MacOS
		// https://docs.microsoft.com/en-us/dotnet/api/system.net.security.ciphersuitespolicy?view=net-6.0
		// if (Environment.OSVersion.Platform == PlatformID.Unix) could also work here
	}

	return sslOptions;
};

var redis = ConnectionMultiplexer.Connect(redisConfig);
```

NickCraver: Maybe! Alternatively, you could only set the property at all on Linux - thoughts?

slorello89: As this is providing a lot of flexibility, I don't think there's any need for extra customization here other than wrapping up and re-throwing any errors that emanate from the SslClientOptions invocation. There are probably other permutations where that invocation could throw an exception, they should be pretty transparent though, and this is flagged as a somewhat advanced use-case. All the ConfigurationOptions really needs to worry about is if the ssl-config delegate responded with valid options or null, it's really none of its business how those were derived as those are the responsibility of the consumer.

btchoum: Agreed. My use case only needs to set the SslClientAuthenticationOptions when running on Linux.

So I'm going with something like below.


``` csharp
var redisConfig = ConfigurationOptions.Parse(_configuration.GetConnectionString("Redis"));

if (Environment.OSVersion.Platform == PlatformID.Unix)
{
	redisConfig.SslClientAuthenticationOptions = (host) =>
		new SslClientAuthenticationOptions
		{
			TargetHost = host,
			CipherSuitesPolicy = new CipherSuitesPolicy(new[]
				{
					TlsCipherSuite.TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256,
					// add more allowed ciphers here ...
				})
		};
}

var redis = ConnectionMultiplexer.Connect(redisConfig);
```

NickCraver: Oh yeah @slorello89 makes a good point, returning null is equivalent so however you want to structure the code there works. As long as this is working, it has legs and I very much appreciate the testing! :)

src/StackExchange.Redis/ConfigurationOptions.cs
```diff
@@ -490,6 +490,14 @@ public int ResponseTimeout
         /// </remarks>
         public SocketManager? SocketManager { get; set; }
 
+#if NETCOREAPP3_1_OR_GREATER
+        /// <summary>
+        /// A <see cref="SslClientAuthenticationOptions"/> provider for a given host, for custom TLS connection options.
+        /// Note: this overrides *all* other TLS and certificate settings, only for advanced use cases.
+        /// </summary>
+        public Func<string, SslClientAuthenticationOptions>? SslClientAuthenticationOptions { get; set; }
```
mgravell, line 498: add to `Clone()` ?

slorello89:
@NickCraver - I've confirmed this fix works for my purposes :) - one thing I will note that may or may not be worth addressing. The way the tls authentication works now, it actually swallows the authentication failures in tls.

When I was initially testing this I was trying to run something like:

```cs
var options = new ConfigurationOptions();
options.EndPoints.Add($"{host}:12000");
options.Ssl = true;
options.SslClientAuthenticationOptions = delegate(string s)
{
    var opts = new SslClientAuthenticationOptions();
    opts.TargetHost = host;
    opts.EnabledSslProtocols = SslProtocols.Tls12;
   
    opts.CipherSuitesPolicy = new CipherSuitesPolicy(new TlsCipherSuite[]
    {
        TlsCipherSuite.TLS_RSA_WITH_AES_128_CBC_SHA256,
        TlsCipherSuite.TLS_RSA_WITH_AES_256_CBC_SHA256
    });
    
    opts.RemoteCertificateValidationCallback += (sender, certificate, chain, errors) =>
    {
        Console.WriteLine(errors);
        return true;
    };
    return opts;
};

options.SslProtocols = SslProtocols.Tls12;
options.CertificateValidation += (sender, certificate, chain, errors) => true;
```

This is erroneous because `options.CertificateValidation` and `opts.RemoteCertificateValidationCheck` both assign the delegate, but the only error I was getting was:

```
It was not possible to connect to the redis server(s). There was an authentication failure; check that passwords (or client certificates) are configured correctly.
```

If you really dig though, you see that you get:

```
The 'RemoteCertificateValidationCallback' option was already set in the SslStream constructor.
```

Now if I dropped a text-writer into the constructor of the multiplexer, this of course lights up the issue, and maybe this is a broader discussion, but some of these failures - might it be worth adding them as inner-exceptions to the exceptions the muxer is already throwing?

NickCraver:
@slorello89 hmmm I see the issue, but changing the flow is a bit more complicated because we don't throw from that method. I agree we should improve this for sure - let's get tests on this, commit as-is as an enhancement, then separately improve the error logging. I have general ideas of per-endpoint having the last exception and we can sum those up, etc. - lots we can improve here for sure in surfacing this.

NickCraver:
Started to add tests here but remembered we have a secure (password) image in the docker-compose but not a SSL/certificate one...which we can now do for Redis 6+ and light up these tests. No expertise there so I'll need to investigate how to light up a new port that's SSL-enabled (self-signed/test cert - doesn't have to be valid).

slorello89:
Hey @NickCraver - yeah so when it comes to building an environment for this, it gets a bit sticky, as you're aware TLS was added in Redis 6, what you might not know (depending how deep you are into the weeds to redis tls), is that it's not shipped in the standard binaries/distros of redis. Rather you need to build redis from source with the tls flag set. I'm not sure how you want to handle this - you're obviously a lot more familiar with how the StackExchange.Redis test-suite/environment is setup - I couldn't gather much beyond it looking like you run everything on a windows-server image?  

I put together [this repo](https://github.com/slorello89/cipher-restricted-redis) which has an image that you can use to validate the fix (it uses the cipher `AES256-SHA256` - that's `TLS_RSA_WITH_AES_256_CBC_SHA256` in the IANA parlance that .NET uses), and ran this bit of code against it to validate the cipher-suite fix (as well as just generally insuring that the options worked :))

```csharp
var options = new ConfigurationOptions();
options.EndPoints.Add("localhost:6379");
options.Ssl = true;
options.SslClientAuthenticationOptions = delegate(string s)
{
    var opts = new SslClientAuthenticationOptions();
    opts.TargetHost = s;
    opts.EnabledSslProtocols =  SslProtocols.Tls12 | SslProtocols.Tls13 ;
    // opts.CipherSuitesPolicy = new CipherSuitesPolicy(new[] {TlsCipherSuite.TLS_RSA_WITH_AES_128_CBC_SHA256}); // change to this and it will break
    opts.CipherSuitesPolicy = new CipherSuitesPolicy(new [] {TlsCipherSuite.TLS_RSA_WITH_AES_256_CBC_SHA256});
    opts.CertificateRevocationCheckMode = X509RevocationMode.NoCheck;
    opts.RemoteCertificateValidationCallback += (_, _, _, _) => true;
    return opts;
};

var muxer = await ConnectionMultiplexer.ConnectAsync(options);
var db = muxer.GetDatabase();

db.StringSet("foo", "bar");
Console.WriteLine("Completed. . .");

```

NickCraver:
@slorello89 overall we're using the redis-latest docker image for most builds, so only looking to solve it there. I believe this is already built with support (https://github.com/docker-library/redis/blob/5c0bbb5dfce3d4999649cbc3ba8bf7c123bcadff/7.0/Dockerfile) so I _think_ it's a matter of figuring out the docker bits to get _any_ certificate on there and loaded (I don't care about validity). Am I reading that right?

slorello89:
@NickCraver  - yep looks like they are building the docker image with it (I was a bit lost in the forest for the trees) - I just pushed an update to that repo slimming it down so it uses the redis image and the bare minimum of what it would need to test things.

As far as validity is concerned, you can work around that with the configuration options - notice I just return true when it's trying to validate above. Or you could create a very long living cert.

slorello89:
@NickCraver - I think I understand how to get it in - do you mind if I push to this branch?

NickCraver:
@slorello89 Not at all - if you have the bandwidth I'd very much appreciate it. Don't worry about tests failing if you add a new instance and default config for SSL server/port for the new instance (I'd expect Windows to start failing the non-skipped tests) - can figure that out once it works at all.

slorello89:
@NickCraver - that worked on my machine (granted I did not spin up the StackExchange.Redis build enviornment) heads up, GH might complain to the two of us about leaked cert files but those certs aren't real certs :)

slorello89:
yeah no not quite :) ok I might see where I've gone wrong.

slorello89:
@NickCraver - sorry for spamming this PR with commits looks like I needed to disaggregate the tls versions - probably a cipher-suite negotiation thing with the server it kept failing when given the choice of TLS 1.3 since I was passing it bogus pre-tls ciphers - lol - looks like the tests are passing (at least for now) - Is this something that should be skipped on windows since it's using Redis 3?

NickCraver:
@slorello89 Thanks a ton for the docker setup work here, much appreciated! I'm taking a poke locally at getting tests cross-platform friendly, simplifying a bit, lighting up the existing SSL tests...and maybe improving those error messages a bit while in here. I think I need it to have test sanity anyway so seeing how hard that is to tackle.

.github/workflows/CI.yml
```diff
@@ -64,6 +64,7 @@ jobs:
         .\redis-server.exe --service-install --service-name "redis-6379" "..\Basic\primary-6379-3.0.conf"
         .\redis-server.exe --service-install --service-name "redis-6380" "..\Basic\replica-6380.conf"
         .\redis-server.exe --service-install --service-name "redis-6381" "..\Basic\secure-6381.conf"
+        .\redis-server.exe --service-install --service-name "redis-6384" "..\Basic\tls-ciphers-6384.conf"
```
slorello89, line 67: This might not be necessary - I pushed it up but I'm assuming this is pointing at redis 3?

slorello89 (changes requested):

docs/Configuration.md
```diff
@@ -76,26 +76,26 @@ The `ConfigurationOptions` object has a wide range of properties, all of which a
...
 | password={string}      | `Password`             | `null`                       | Password for the redis server                                                                             |
 | user={string}          | `User`                 | `null`                       | User for the redis server (for use with ACLs on redis 6 and above)                                        |
 | proxy={proxy type}     | `Proxy`                | `Proxy.None`                 | Type of proxy in use (if any); for example "twemproxy/envoyproxy"                                         |
 | resolveDns={bool}      | `ResolveDns`           | `false`                      | Specifies that DNS resolution should be explicit and eager, rather than implicit                          |
-| serviceName={string}   | `ServiceName`          | `null`                       | Used for connecting to a sentinel primary service                                                          |
+| serviceName={string}   | `ServiceName`          | `null`                       | Used for connecting to a sentinel primary service                                                         |
```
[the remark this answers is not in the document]
NickCraver: Yep - aligned all the things :)

.github/workflows/CI.yml
NickCraver: oh yeah good call - let me roll this back when I get a sec, it won't work so let's remove.

src/StackExchange.Redis/ExceptionFactory.cs
```diff
@@ -383,17 +385,29 @@ private static string GetLabel(bool includeDetail, RedisCommand command, Message
         internal static Exception UnableToConnect(ConnectionMultiplexer muxer, string? failureMessage = null)
         {
             var sb = new StringBuilder("It was not possible to connect to the redis server(s).");
-            if (muxer != null)
+            Exception? inner = null;
+            if (muxer is not null)
             {
-                if (muxer.AuthSuspect) sb.Append(" There was an authentication failure; check that passwords (or client certificates) are configured correctly.");
-                else if (muxer.RawConfig.AbortOnConnectFail) sb.Append(" Error connecting right now. To allow this multiplexer to continue retrying until it's able to connect, use abortConnect=false in your connection string or AbortOnConnectFail=false; in your code.");
+                if (muxer.AuthException is Exception aex)
```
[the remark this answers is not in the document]
NickCraver: Not sure I follow - what's the ask on this one?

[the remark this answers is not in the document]
slorello89: `muxer.AuthException` is a `Nullable<Exception>`, really what you're looking to do is just check if `muxer.AuthException != null` right? this looks like an extra-unneeded cast.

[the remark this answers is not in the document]
NickCraver: So not exactly - the key here is I'm memoizing it in the method to a local before use. If instead you checked for null and then tried to use it, `muxer.AuthException` could be cleared between those 2 and null ref.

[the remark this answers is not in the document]
NickCraver: Ooh hey the same race can apply to out inner set too - not catastrophic since it'd be fine, but fixing!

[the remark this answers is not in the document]
slorello89: Gotcha, that makes sense.

slorello89 (approved):
LGTM 👍

NickCraver:
@slorello89 thanks for the eyes and help here <3

btchoum:
@NickCraver @slorello89 
Thank you so much for solving this issue. We’ve been using the latest version of the package and so far works perfectly. Much appreciated

Issue #2146, opened 2022-05-25
Small timeouts: Time elapsed is much greater than timeout


tyzoid:
When a timeout of 100ms is set, it appears that StackExchange.Redis takes upwards of 1000ms to actually time out. I would expect that this would time out much closer to the 100ms limit that's set in the configuration.
```
StackExchange.Redis.RedisTimeoutException :: StackExchange.Redis.RedisTimeoutException: Timeout awaiting response (outbound=0KiB, inbound=0KiB, 922ms elapsed, timeout is 100ms), command=GET, next: GET XXXXXXXXXX, inst: 0, qu: 20, qs: 1362, aw: True, bw: WritingMessage, rs: ReadAsync, ws: Writing, in: 0, in-pipe: 0, out-pipe: 0, serverEndpoint: XXX.XXX.XXX.XXX:6379, mc: 1/1/0, mgr: 9 of 10 available, clientName: XXXXX(SE.Redis-v2.5.61.22961), IOCP: (Busy=2,Free=998,Min=8,Max=1000), WORKER: (Busy=12,Free=8179,Min=8,Max=8191), v: 2.5.61.22961 (Please take a look at this article for some common client-side issues that can cause timeouts: https://stackexchange.github.io/StackExchange.Redis/Timeouts)
       at System.Runtime.CompilerServices.TaskAwaiter.ThrowForNonSuccess(Task task)
       at System.Runtime.CompilerServices.TaskAwaiter.HandleNonSuccessAndDebuggerNotification(Task task)
       at XXXX<Application Code>
```

Here's the configuration we're supplying:
```csharp
ConfigurationOptions options = new ConfigurationOptions
{
	Password = password,
	Ssl = false,
	AbortOnConnectFail = false,
	SyncTimeout = 100,
	AsyncTimeout = 100,
	ReconnectRetryPolicy = new LinearRetry(200),
	BacklogPolicy = BacklogPolicy.FailFast
};
```

NickCraver:
The timeout is only evaluated on the heartbeat interval (every 1000ms), so if you set it far lower it may be anywhere from 100ms -> ~1000ms (depending on system timers) when it triggers. Basically: once a second, we crawl the queue and check for violators. This is a global tradeoff between fidelity and paying a high overhead for timeout evaluation, and currently intentional.

tyzoid:
[...]

PR #2243, opened 2022-09-04
Heartbeat interval: Make it configurable

NickCraver:
When I was looking at a separate timer for async timeout evaluation < 1000ms fidelity it became apparent _most_ of the cost in a heartbeat is the timeout evaluation anyway, but also: if you are chasing a lower timeout fidelity you likely want to observe a server outage ASAP as well so I think it makes more sense to actually open up the heartbeat to make it configurable.

I added some docs about how this is risky but that portion and comments  could use scrutiny (if we even agree with doing this) on being strong enough warnings around this. I would prefer to leave this as code-only due to potential downsides people may not appreciate otherwise.

Thoughts?


NickCraver:
cc @deanward81

src/StackExchange.Redis/PhysicalBridge.cs
```diff
@@ -20,7 +20,7 @@ internal sealed class PhysicalBridge : IDisposable
 
         private const int ProfileLogSamples = 10;
 
-        private const double ProfileLogSeconds = (ConnectionMultiplexer.MillisecondsPerHeartbeat * ProfileLogSamples) / 1000.0;
+        private const double ProfileLogSeconds = (1000 /* ms */ * ProfileLogSamples) / 1000.0;
```
tyzoid, line 23: Why multiply by a thousand just to divide again?

tyzoid (commented):
Thanks for this. I'll re-run some of the load testing that got me to #2146 in the first place this week. This seems reasonable to me as a way of resolving our original issue.

Given the cautionary notes I've seen from developers here regarding shortening this interval, is this a measurable/meaningful impact? Decreasing the interval from 1s to 100ms will 10x the total amount of work the library will incur overall, but how much of an impact is this? Would this just be burning extra cycles in a separate thread, or does this impact request processing and throughput?

One of the workarounds we implemented in our codebase is await on a `Task.Delay` with a short timeout, and just abandon the request if the timeout is exceeded. Not a great solution, since this still leaves tasks floating around (since there doesn't seem to be a cancellation interface), and also retries seem to still select the same failing connection.

If decreasing this heartbeat interval is seen as too drastic a measure, perhaps an alternative implementation to achieve a similar result is to present a cancellation interface to the client allowing the requester to abort the request early? I have no data to support either conclusion, hence the need for testing, but I don't have a great way of fixturing this library directly without digging into the code more in depth, only in the black-box method of testing we've performed so far.

src/StackExchange.Redis/PhysicalBridge.cs
NickCraver: This was really just showing intent of the net change without doing a bigger rename, etc. - can be simplified for sure.

NickCraver:
@tyzoid We can't abort it early really, because the command _has been sent_, so we have to handle/account for it no matter what - any per-command tracking (_especially out of order_) would be much more costly.

On the overall cost of this...it'd be awesome to experiment and see. The work of a heartbeat isn't crazy _but_ there's a brief lock around the work queue to check timeouts. If people lower it a ton though, say to 10ms, it's probably not awesome and going to have some measurable jitter in spikes. I don't have exact numbers, by nature it's going to depend on the workload a lot as to if it'll have any impact or how much.

slorello89 (commented):
@NickCraver - So I understand, the issue here is that the event that wakes up to check the heartbeat is the same that checks all of the timeouts?

If the goal here is to allow for tighter propagation of aggressively set timeouts, this seems like a somewhat roundabout way to get there. Wouldn't it make more sense to decouple the process that checks the timeouts from the hearbeat and make that interval configurable? It strikes me that the heartbeat needs a network call and a full round trip to Redis, which, even if it's piggybacking along with other packets being multiplexed to Redis, sounds like an expense you shouldn't need to pay to check some timeouts and wakeup the hung tasks/threads associated with those timeouts?

NickCraver:
@slorello89 We only connect to Redis [if we haven't written in a long time](https://github.com/StackExchange/StackExchange.Redis/blob/main/src/StackExchange.Redis/PhysicalBridge.cs#L561), this shouldn't actually increase the network activity any...and most of the work left _is_ the timeout and messages. There's no additional thread here - only effectively the timeout checks and a few if statements ahead of them.

I agree it sounds roundabout and my first work was the split path, but 2 things drove me back to simpler:
1. Most of the work is message timeouts anyway, or all timeout related (e.g. connections timing out themselves, from Redis)
2. People who want crazy low timeouts are almost 1:1 with wanting to know Redis is offline and handle it ASAP, so a quicker window in those cases to evaluate the backup checks (if we didn't actively detect a socket failure) is usually desired for the same crowd.

philon-msft (approved):

PR #2246, opened 2022-09-06
Auth: Better exception for bad passwords

### LINKED ISSUE #1879: Unable to connect to Redis due to authentication - ConnectionFailureType
Hi,

In case the connection with Redis failed due to authentication failure (e.g. password is configured incorrectly due to key rotation),
the ConnectionFailureType that is returned is ConnectionFailureType.UnableToConnect.

My suggestion is in this case to return ConnectionFailureType.AuthenticationFailure
Thus, me as a client, can decide to disconnect Redis and retry with a new primary key that I have fetched.

The automatically retry mechanism of IMultiplexer would never success to connect with its own retry policy, but I don't want to catch any RedisConnectionException.

Is there any reason that the function returns ConnectionFailureType.UnableToConnect instead of ConnectionFailureType.AuthenticationFailure in this case?

https://github.com/StackExchange/StackExchange.Redis/blob/51d2946b7f1499605d90b95a00bcb8f776ff0b49/src/StackExchange.Redis/ExceptionFactory.cs#L392-L403

Thanks,
Sagi

NickCraver:
Fixes #1879. In the case where a password is wrong, now indicates "WRONGPASS invalid username-password pair or user is disabled" rather than "NOAUTH Returned - connection has not authenticated" to help users diagnose their problem a bit easier.

### DIFF
--- a/src/StackExchange.Redis/ResultProcessor.cs
+++ b/src/StackExchange.Redis/ResultProcessor.cs
@@ -229,7 +229,7 @@ namespace StackExchange.Redis
             {
                 if (result.StartsWith(CommonReplies.NOAUTH))
                 {
-                    bridge?.Multiplexer?.SetAuthSuspect(new RedisServerException("NOAUTH Returned - connection has not authenticated"));
+                    bridge?.Multiplexer?.SetAuthSuspect(new RedisServerException("NOAUTH Returned - connection has not yet authenticated"));
                 }
                 else if (result.StartsWith(CommonReplies.WRONGPASS))
                 {
--- a/src/StackExchange.Redis/ResultProcessor.cs
+++ b/src/StackExchange.Redis/ResultProcessor.cs
@@ -227,7 +227,14 @@ namespace StackExchange.Redis
             }
             if (result.IsError)
             {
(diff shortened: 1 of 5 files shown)


slorello89 (approved):
LGTM 👍 - FWIW the change in message seems to have come in 6.0 with the dawn of ACLs

src/StackExchange.Redis/ResultProcessor.cs
```diff
@@ -227,7 +227,14 @@ public virtual bool SetResult(PhysicalConnection connection, Message message, in
             }
             if (result.IsError)
             {
-                if (result.StartsWith(CommonReplies.NOAUTH)) bridge?.Multiplexer?.SetAuthSuspect(new RedisServerException("NOAUTH Returned - connection has not authenticated"));
+                if (result.StartsWith(CommonReplies.NOAUTH))
+                {
+                    bridge?.Multiplexer?.SetAuthSuspect(new RedisServerException("NOAUTH Returned - connection has not authenticated"));
```
philon-msft, line 232: "has not BEEN authenticated"?

philon-msft (approved):

src/StackExchange.Redis/ResultProcessor.cs
NickCraver: Maybe "not yet authenticated"? I worry about confusion with "we can't verify the validity of what you connected to" on phrasing - thoughts?

mgravell (approved):

PR #2250, opened 2022-09-09
Remove SCRIPT LOAD from ScriptEvalMessage executions

rkarthick:
The current implementation of `ScriptEvalMessage` optimistically calls `SCRIPT LOAD <script>` before `EVAL <script>` for every new script that is seen by the client to load the script in redis cache. However, calling `SCRIPT LOAD <script>` in our implementation, is redundant. as the following `EVAL <script>` also loads the script in REDIS cache [1].

`SCRIPT LOAD` returns SHA1 of the script. In this PR,  we are calculating the SHA1 value within the client and passing on to `EVALSHA`. This change also has a nice side effect that `ScriptEvalMessage` no longer is a IMultiMessage interface.

A reference implementation in go-redis can be found here: https://github.com/go-redis/redis/blob/master/script.go#L59 Please note that, unlike the stackexchange client, it does not use a client side cache to track the loaded scripts.

[1] https://redis.io/docs/manual/programmability/eval-intro/ (scroll down to Script cache section)

Redis monitor before the change while calling `ScriptEvaluate` twice

```
1662073157.821955 [0 172.17.0.1:45914] "SCRIPT" "LOAD" "redis.call('set', ARGV[1], ARGV[2])"
1662073157.821967 [0 172.17.0.1:45914] "EVAL" "redis.call('set', ARGV[1], ARGV[2])" "1" "foo" "foo" "123"
1662073163.983800 [0 lua] "set" "foo" "123"
1662073165.117343 [0 172.17.0.1:45914] "EVALSHA" "738ee76b1389bf5e4bb46e6b3b23e24382d54905" "1" "foo" "foo" "123"
1662073165.117366 [0 lua] "set" "foo" "123"
```

Redis monitor after the change while calling `ScriptEvaluate` twice

```
1662682698.470518 [0 172.17.0.1:41268] "EVAL" "redis.call('set', ARGV[1], ARGV[2])" "1" "foo" "foo" "123"
1662682698.470551 [0 lua] "set" "foo" "123"
1662682698.476800 [0 172.17.0.1:41268] "EVALSHA" "738ee76b1389bf5e4bb46e6b3b23e24382d54905" "1" "foo" "foo" "123"
1662682698.476819 [0 lua] "set" "foo" "123"
```

`738ee76b1389bf5e4bb46e6b3b23e24382d54905` SHA1 of the script is calculated by C# in the second instance

### DIFF
--- a/src/StackExchange.Redis/ServerEndPoint.cs
+++ b/src/StackExchange.Redis/ServerEndPoint.cs
@@ -546,11 +546,11 @@ namespace StackExchange.Redis
                 return;
             }
 
-            var hash = Sha1(script);
+            var hash = Encoding.ASCII.GetBytes(Sha1(script));
 
             lock(knownScripts)
             {
-                knownScripts[script] = Encoding.ASCII.GetBytes(hash);
+                knownScripts[script] = hash;
             }
         }
 
--- a/src/StackExchange.Redis/ServerEndPoint.cs
+++ b/src/StackExchange.Redis/ServerEndPoint.cs
@@ -547,7 +547,6 @@ namespace StackExchange.Redis
             }
 
             var hash = GetSHA1(script);
-
             lock(knownScripts)
             {
                 knownScripts[script] = hash;
@@ -901,7 +900,6 @@ namespace StackExchange.Redis
             byte[] outputBytes = sha1.ComputeHash(inputBytes);
             return Encoding.ASCII.GetBytes(BitConverter.ToString(outputBytes).Replace("-", "").ToLower());
         }
-
 
         private PhysicalBridge? CreateBridge(ConnectionType type, LogProxy? log)
         {
--- a/src/StackExchange.Redis/ServerEndPoint.cs
+++ b/src/StackExchange.Redis/ServerEndPoint.cs
@@ -4,6 +4,7 @@ using System.Collections.Generic;
 using System.Linq;
 using System.Net;
 using System.Runtime.CompilerServices;
+using System.Security.Cryptography;
 using System.Text;
 using System.Text.RegularExpressions;
 using System.Threading;
@@ -536,6 +537,29 @@ namespace StackExchange.Redis
                 }
             }
             return found;
+        }
+
+        internal void SetScriptHash(string script, RedisCommand command)
+        {
+            if (command != RedisCommand.EVAL)
+            {
+                return;
+            }
+
+            var hash = Sha1(script);
+
+            lock(knownScripts)
+            {
+                knownScripts[script] = Encoding.ASCII.GetBytes(hash);
+            }
+        }
+
+        internal string Sha1(string input)
+        {
+            var sha1 = SHA1.Create();
+            byte[] inputBytes = Encoding.ASCII.GetBytes(input);
+            byte[] outputBytes = sha1.ComputeHash(inputBytes);
+            return BitConverter.ToString(outputBytes).Replace("-", "").ToLower();
         }
 
         internal string? GetStormLog(Message message) => GetBridge(message)?.GetStormLog();
--- a/src/StackExchange.Redis/ServerEndPoint.cs
+++ b/src/StackExchange.Redis/ServerEndPoint.cs
@@ -546,20 +546,12 @@ namespace StackExchange.Redis
                 return;
             }
 
-            var hash = Encoding.ASCII.GetBytes(Sha1(script));
+            var hash = GetSHA1(script);
 
             lock(knownScripts)
             {
                 knownScripts[script] = hash;
             }
-        }
-
-        internal string Sha1(string input)
-        {
-            var sha1 = SHA1.Create();
-            byte[] inputBytes = Encoding.ASCII.GetBytes(input);
-            byte[] outputBytes = sha1.ComputeHash(inputBytes);
-            return BitConverter.ToString(outputBytes).Replace("-", "").ToLower();
         }
 
         internal string? GetStormLog(Message message) => GetBridge(message)?.GetStormLog();
@@ -901,6 +893,15 @@ namespace StackExchange.Redis
             }
             return default;
         }
+
+        private static byte[] GetSHA1(string input)
+        {
+            var sha1 = SHA1.Create();
+            byte[] inputBytes = Encoding.ASCII.GetBytes(input);
+            byte[] outputBytes = sha1.ComputeHash(inputBytes);
+            return Encoding.ASCII.GetBytes(BitConverter.ToString(outputBytes).Replace("-", "").ToLower());
+        }
+
 
         private PhysicalBridge? CreateBridge(ConnectionType type, LogProxy? log)
         {
--- a/src/StackExchange.Redis/RedisDatabase.cs
+++ b/src/StackExchange.Redis/RedisDatabase.cs
@@ -4742,7 +4742,7 @@ namespace StackExchange.Redis
             public override int ArgCount => _args.Count;
         }
 
-        private sealed class ScriptEvalMessage : Message, IMultiMessage
+        private sealed class ScriptEvalMessage : Message
         {
             private readonly RedisKey[] keys;
             private readonly string? script;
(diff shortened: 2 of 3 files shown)


mgravell:
Note: the same docs you linked to caution about the problems when pipelining. SE.Redis *always pipelines*, so some amount of care is recommended here - it is reasonable to blip if someone explicitly flushes the cache, but getting odd results every first run (on a cold server): very undesirable. If EVAL doubles as a silent SCRIPT LOAD, then that is genuinely news to me - I'll have to check, but you may also want to verify for what server versions that has been true.

rkarthick:
>  it is reasonable to blip if someone explicitly flushes the cache, but getting odd results every first run (on a cold server): very undesirable.
I take your point on extra care with pipelining. i will spend some more time on the failure cases there. However not sure if I follow how changing the implementation from

t0: script load
t1: eval
t2: evalsha

to 

t0: eval
t1: evalsha

will result in odd results on a cold server. On a cold server in either case, if evalsha fails we revert to the eval path, no?

> If EVAL doubles as a silent SCRIPT LOAD, then that is genuinely news to me - I'll have to check, but you may also want to verify for what server versions that has been true.

The best I could find is this excerpt below.
```Every script you execute with [EVAL](https://redis.io/commands/eval) is stored in a dedicated cache that the server keeps. The cache's contents are organized by the scripts' SHA1 digest sums, so the SHA1 digest sum of a script uniquely identifies it in the cache.``` 
in the document [linked](https://redis.io/docs/manual/programmability/eval-intro/).  Let me see if I can find  early implementations of EVAL and if they do the same.

**Comment by mgravell** [comment]
Ah, that makes more sense - I was reviewing on mobile, and thought it was switching to just EVALSHA with the existing fallback if the specific script error was found. If this works as described, then: seems like a good win, indeed. I'll need to validate for myself that it *does* work as described, because I am paranoid (I've been burned many times), but: I'm assuming you're probably entirely correct here (those burned times are still a small fraction of all the total things :p - just: a painful small fraction)

There's probably some things we'd want to do around the sha compute implementation, but: I am intrigued.

File: src/StackExchange.Redis/ServerEndPoint.cs [hunk 1]
```diff
@@ -878,6 +893,14 @@ internal ValueTask WriteDirectOrQueueFireAndForgetAsync<T>(PhysicalConnection? c
             return default;
         }
 
+        private static byte[] GetSHA1(string input)
+        {
+            var sha1 = SHA1.Create();
+            byte[] inputBytes = Encoding.ASCII.GetBytes(input);
+            byte[] outputBytes = sha1.ComputeHash(inputBytes);
+            return Encoding.ASCII.GetBytes(BitConverter.ToString(outputBytes).Replace("-", "").ToLower());
```
rkarthick, line 901: Not sure if there is an optimal way to convert the bytes to hexBytes directly.

rkarthick:
> Ah, that makes more sense - I was reviewing on mobile, and thought it was switching to just EVALSHA with the existing fallback if the specific script error was found. If this works as described, then: seems like a good win, indeed. I'll need to validate for myself that it _does_ work as described, because I am paranoid (I've been burned many times), but: I'm assuming you're probably entirely correct here (those burned times are still a small fraction of all the total things :p - just: a painful small fraction)
> 
> There's probably some things we'd want to do around the sha compute implementation, but: I am intrigued.

Thank you for the feedback @mgravell! I changed the implementation of Sha1 a little, and added a comment on the current limitation of us using the bytes directly. Would love to get your thoughts on it. Does this require a test case?  Would be a little tricky as we need to check the monitor command from redis to ensure the expected commands are flowing through. Thoughts?

rkarthick:
@mgravell / @NickCraver Would love to get your thoughts :)

mgravell:
random question for @slorello89, I think; has `EVAL` *always* functioned as `SCRIPT LOAD`? this isn't documented in the docs AFAIK, but I'm also quite favourable towards this PR (currently the library uses `SCRIPT LOAD` and `EVAL` until it has seen a result from `SCRIPT LOAD`, then it switches to `EVALSHA`; this PR removes the explicit `SCRIPT LOAD`, and just relies on the `EVAL` side-effect)

slorello89:
@mgravell - Admittedly this is something I'd never heard of. I built 2.6.0 from source and yeah this works all the way back at that version. And the docs on scripting does make an allusion to this behavior so I think this is probably pretty safe. I think if it were me I'd err on the side of doing the LOAD if for no other reason than to be explicit, even if it means sacrificing that precious extra round trip.

rkarthick:
Thank you for your thoughts @slorello89 and @mgravell. The main motivation behind the change is to make script evaluation execution compatible with the proxy route to redis (EVAL + EVALSHA vs LOAD + EVALSHA). Envoy proxy does not support SCRIPT load command (for good reasons to hide the underlying cluster details from the client using envoy proxy route to redis). It does support `EVALSHA` and `EVAL`.

mgravell:
Relevant https://github.com/envoyproxy/envoy/issues/17926

rkarthick:
Envoy proxy forwards the command as-is to the corresponding redis node. It does _not_ track or backfill based on known nodes.  EVAL from the library will be routed to the right redis node by envoy, and the script will be loaded in the redis node as expected. The library keeps track of all the loaded scripts per envoy node. If envoy node changes, the library needs to send EVAL to refresh library's internal cache that is maintained per envoy node. If a redis node changes, it does force a refresh of  all envoy endpoints maintained by the library. That is the tradeoff of this approach. 

We did consider doing "SCRIPT LOAD" at envoy that will broadcast the command to every redis node. But that would require introducing a new paradigm of broadcast commands in envoy. The approach here, IMO does not break any fundamental paradigms of how script is handled within the redis library. In the normal non-proxy route, it does offer some minor optimization.

As of today, in envoy proxy mode the library does what you are suggesting (without no-cache option). It always uses EVAL, as SCRIPT LOAD is not a part of envoy supported commands with the library. The optimization here is added to reduce the network bandwidth, avoiding sending huge scripts for every execution.

mgravell:
Is "the library" in the above SE.Redis? Because SE.Redis has no knowledge or interest in envoy nodes, and won't know to reissue anything. If it sees a script error due to a missing script, it will reissue if it can, but... that's a bit late and can't be used in all cases (MULTI/EXEC for example). If we don't already, I guess the most we could do here would be to disable the SCRIPT command via the map. Maybe that's what works today.

So: perhaps the sweet spot here might be to use SCRIPT LOAD by default, unless SCRIPT is disabled, in which case use EVAL switching to EVALSHA once observed successfully?

rkarthick:
Yeah, Sorry the library here is SE.Redis. SE.Redis keeps an [internal cache per endpoint](https://github.com/StackExchange/StackExchange.Redis/blob/main/src/StackExchange.Redis/ServerEndPoint.cs#L30) which is flushed when it sees a script error. The endpoint could be either a redis node or an envoy node. It flushes the [entire cache though](https://github.com/StackExchange/StackExchange.Redis/blob/5ceb775ca05a99226a0d88ec00f72013c08f856c/src/StackExchange.Redis/ResultProcessor.cs#L1772) when a script error is received.  Yeah, SCRIPT is disabled via the map, and I think thats how it works today.

> So: perhaps the sweet spot here might be to use SCRIPT LOAD by default, unless SCRIPT is disabled, in which case use EVAL switching to EVALSHA once observed successfully?

This is definitely reasonable and I can look into this :).

slorello89:
Yeah so my thoughts are in line with @mgravell - I poked some of our core engineers internally to gather their thoughts, and the consensus opinion is that `SCRIPT LOAD` is the correct way to load a script. From Envoy's perspective, it should probably fan the `SCRIPT LOAD` out to all it's nodes, that's actually what the Redis Enterprise proxy does which gets to Marc's point earlier:

> Hmmm. That just exacerbates the problem and raises more questions, but now instead of worrying about the defined behaviour of one product, we need to consider the defined behaviour of two products interacting with eachother.

Where one implementation of a proxy made one assumption whereas another made a different assumption.

So I think I would agree, guarding any such behavior behind either `SCRIPT LOAD` being disabled, or some other configuration option to explicitly skip the loads and trust `EVAL` to do the job (I think the latter makes more sense to me so the default behavior remains without an explicit configuration/code change in an application using SE.R).

Another design question though, shouldn't you also keep track at the slot level as to whether or not a particular slot has seen the script yet? The Envoy proxy demultiplexes, from one proxy to each shard, so if you have shards 1, 2, and 3 and the keys of your first eval are only slotted to shard 1, if the multiplexer saves the known script and then tries to execute the script again against slot 2, slot 2 is now going to respond with an error, forcing you to reload the script again, and incurring an extra round trip? So this will result in error/reloads presumably until the all of the nodes are filled in behind the proxy? 

I wonder if perhaps it wouldn't make sense to keep watch over the slots have seen the script? Perhaps this doesn't make sense because it would be a fair bit of memory to track each slot. But it strikes me that this implementation as it currently stands leans pretty heavily into errors communicating the current disposition of the scripts in a shard. Which is a legitimate way to handle it, I suppose the tradeoff would be extra bandwidth/memory on the one hand (because your going to be making excess `EVAL` calls unless there's an intelligent way to for the proxy to communicate the slot mapping), and extra round trips on the other.

mgravell:
> The endpoint could be either a redis node or an envoy node.

Sorry, what I meant is: if a single endpoint (which happens to be envoy) happens to wrap multiple *server* nodes: the library will only know about what it sees as the endpoint, and the library's script caching will be based on the assumption that the state is consistent for that endpoint; I *think* for hybrid SCRIPT/EVAL+EVALSHA to work reliably through an envoy node that talks to multiple server nodes, we'd need more info than that - info that we simply don't have from the library (it would require insight into the envoy sharding details). This is the same scenario that is covered by the discussion starting "The Envoy proxy demultiplexes, from one proxy to each shard, so..."; so again, without a fan-out `SCRIPT LOAD`, I'd be hella cautious of whether scripts that we expect to be loaded: are actually loaded - i.e. whether the proposal to use EVAL first time would actually help any. It *seems* like we can reasonably expect this to *cause errors* - in at least two scenarios:

1. pipelined usage where we don't pay latency between related operations (where things could end up out of the intended order)
2. the same inside a transaction, which ... yeah, awkwardness

Happy to keep discussing options, but I really don't think this (in the envoy case) is as simple as just switching to EVAL

rkarthick:
Thank you for all the extra context. The weirdness around using envoy with the current logic makes a lot of sense.  I think we can close out this PR. We are exploring a way to optimistically load the script, use `EVALSHA` all the time and when it fail s use `EVAL` (without using a cache) in our private repo. The implementation will be similar to [go-redis](https://github.com/go-redis/redis/blob/master/script.go#L70).  Will post that PR when it is ready.

PR #2254, opened 2022-09-20
Correction on documented behavior of 'httpRuntime targetFramework="4.5"'

johnaceous:
The old wording equated `<httpRuntime targetFramework="4.5" />` to `<add key="aspnet:UseTaskFriendlySynchronizationContext" value="false" />`, when the linked citation makes it clear that the effect is the opposite: `<httpRuntime targetFramework="4.5" />` is inferred as adding the setting `<add key="aspnet:UseTaskFriendlySynchronizationContext" value="true" />`.

> Second, `<httpRuntime targetFramework="4.5" />` is a shortcut that allows the ASP.NET runtime to infer a wide array of configuration settings. If the runtime sees this setting, it will expand it out just as if you had written the following:
> ```
>     ...
>     <add key="aspnet:UseTaskFriendlySynchronizationContext" value="true" />
> ```
https://devblogs.microsoft.com/dotnet/all-about-httpruntime-targetframework/

### DIFF
--- a/docs/ThreadTheft.md
+++ b/docs/ThreadTheft.md
@@ -44,7 +44,7 @@ configure ASP.NET with:
 <add key="aspnet:UseTaskFriendlySynchronizationContext" value="false" />
 ```
 
-or if you do NOT add the following line:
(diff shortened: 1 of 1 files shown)


NickCraver (approved):
Agree this needed some love - thanks for this!

PR #2267, opened 2022-10-12
Respect IncludeDetailInExceptions in MOVED etc scenarios from ResultProcessor

mgravell:
ResultProcessor should be more reticent if IncludeDetailInExceptions is disabled


NickCraver (approved):
👍 Looks like bridge can't be null down there and all good, thanks for this!

PR #2274, opened 2022-10-17
Configuration/Connections: Allow HTTP tunneling

mgravell:
1. add a tunnel config option, which is used to influence connection creation
2. implement http-tunnel as a well-known tunnel, integrated via the `http:` prefix on the tunnel option

tunnels are implemented as concrete subclasses of the `Tunnel` type; an HTTP proxy "connect" implementation is provided as a well-known version that is supported inside `Parse` - but custom 3rd-party tunnel implementations can also be provided via the object-model (not `Parse`)

A `Tunnel` allows:

- overriding the `EndPoint` used to create `Socket` connections, or to suppress `Socket` creation entirely (by default, the same logical endpoint requested is provided back out)
- provide a "before socket connect" twin (mirrors delegate approach, but: async; by default, do nothing)
- provide a "before authenticate" injection point, which can a: perform additional handshake operations, and b: subvert the entire `Stream` (by default, nothing is done and no custom stream is returned)

### DIFF
--- a/src/StackExchange.Redis/Configuration/Tunnel.cs
+++ b/src/StackExchange.Redis/Configuration/Tunnel.cs
@@ -48,10 +48,6 @@ namespace StackExchange.Redis.Configuration
             {
                 if (socket is not null)
                 {
-
-                    // TODO: make write+read async
-                    // TODO: compare the response more appropriately?
-
                     var encoding = Encoding.ASCII;
                     var ep = Format.ToString(endpoint);
                     const string Prefix = "CONNECT ", Suffix = " HTTP/1.1\r\n\r\n", ExpectedResponse = "HTTP/1.1 200 OK\r\n\r\n";
--- a/src/StackExchange.Redis/Configuration/Tunnel.cs
+++ b/src/StackExchange.Redis/Configuration/Tunnel.cs
@@ -1,0 +1,91 @@
+﻿using System;
+using System.Buffers;
+using System.IO;
+using System.Net;
+using System.Net.Sockets;
+using System.Text;
+using System.Threading;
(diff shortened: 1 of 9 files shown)


NickCraver (approved):
Tidied docs/APIs a bit - looking good!

src/StackExchange.Redis/Configuration/Tunnel.cs
```diff
@@ -0,0 +1,114 @@
...
+
+            public override ValueTask<EndPoint?> GetSocketConnectEndpointAsync(EndPoint endpoint, CancellationToken cancellationToken) => new(Proxy);
+
+            public override async ValueTask<Stream?> BeforeAuthenticateAsync(EndPoint endpoint, ConnectionType connectionType, Socket? socket, CancellationToken cancellationToken)
+            {
+                if (socket is not null)
+                {
+                    var encoding = Encoding.ASCII;
+                    var ep = Format.ToString(endpoint);
+                    const string Prefix = "CONNECT ", Suffix = " HTTP/1.1\r\n\r\n", ExpectedResponse = "HTTP/1.1 200 OK\r\n\r\n";
```
flobernd, line 52: @mgravell According to https://datatracker.ietf.org/doc/html/draft-luotonen-web-proxy-tunneling-01#section-3.2, the expected response should be `HTTP/1.1 200 Connection established` instead of `HTTP/1.1 200 OK`. Currently it's not possible to use a HTTP proxy that correctly implements this behavior (e.g. Squid).

flobernd: Created a PR: #2448

PR #2276, opened 2022-10-19
Timeouts: add "last-in" for the last payload's length to exception messages

NickCraver:
In a multiplexed setup we often see a timeout _behind_ a large payload in the process of parsing (or not). This is meant to help better diagnose that case, by recording what the last payload size was that came through this connection.

Needs some @mgravell eyes on the record point here and if we want to adjust (or if it's wrong due to buffers in a way I'm not understanding). I realize this won't work on a payload so large enough it blows all buffers completely, but I don't see how to solve that either the way we properly stream parse. I think it'll still improve most cases we see this in.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -8,7 +8,7 @@ Current package versions:
 
 ## Unreleased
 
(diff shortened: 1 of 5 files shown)


andrebekema:
This will be a very useful addition! I can imagine users asking the next logical question "Ok, so which command is clogging the pipe?". Do you think there are possibilites (as a next feature enhancement) of providing more information about the command other than its payload size?

mgravell (approved):

PR #2282, opened 2022-10-25
Internals Rename: *Wrapper -> "KeyPrefixed"

NickCraver:
This renames the various wrappers to IMO more intuitive "KeyPrefixedX" - just couldn't get used to the previous model when touching N files for an API addition, hoping this makes it a little clearer overall, also has the benefit of batching them in UI due to common prefix.


NickCraver:
@mgravell this was the rename I talked about forever ago that kept tripping me up - thoughts on it equally/more intuitive to you? If not can leave as-is, just keeps locking my brain up

mgravell (approved):

PR #2284, opened 2022-10-26
Update README to reflect heartbeatInterval which applies only to async

rkarthick:
From my reading of code, the sync execute path uses `syncTimeout` and applies it [instantaneously through](https://github.com/StackExchange/StackExchange.Redis/blob/main/src/StackExchange.Redis/ConnectionMultiplexer.cs#L1879) `Monitor.Wait`. 

Looks like the heartbeatInterval only applies to async commands?
https://github.com/StackExchange/StackExchange.Redis/blob/main/src/StackExchange.Redis/PhysicalConnection.cs#L690


rkarthick:
QQ on this comment. Do you know what returning "source" to the pool here means in the context of the timeout? 
https://github.com/StackExchange/StackExchange.Redis/blob/main/src/StackExchange.Redis/ConnectionMultiplexer.cs#L1888

NickCraver:
@rkarthick the `SimpleResultBox` is in an unknown/tainted state, we don't want it back in the object pool for re-use :)

Issue #2074, opened 2022-04-08
"-READONLY You can't write against a read only replica." on ScriptEvaluate(...)


dusty9023:
Background:
- AWS MemoryDB (REDIS Engine 6.2)
- Version: 2.5.43.

During some of testing, we've noticed that the error "-READONLY You can't write against a read only replica.", is showing up in our logs when we attempt to resize our shards in MemoryDB (change the size of the nodes, which causes it to fail over to a replica). 

Attempting to retry the operation (we do 3 retries with exponential backoff) it fails all 3 times before finally being thrown a final time. Therefore, it looks like it's not reassigning the primary on these errors.

Looking at the exception, this only seems to affect ScriptEvaluate (LUA Scripts). We don't see any of our other various calls throwing this error:
`at StackExchange.Redis.RedisDatabase.ScriptEvaluate(String script, RedisKey[] keys, RedisValue[] values, CommandFlags flags) in /_/src/StackExchange.Redis/RedisDatabase.cs:line 1212`

Full Stacktrace:
```
message=ERR Error running script (call to f_f3...0): @user_script:50: @user_script: 50: -READONLY You can't write against a read only replica.
Stack trace
   at StackExchange.Redis.ConnectionMultiplexer.ExecuteSyncImpl[T](Message message, ResultProcessor`1 processor, ServerEndPoint server) in /_/src/StackExchange.Redis/ConnectionMultiplexer.cs:line 2836
   at StackExchange.Redis.RedisBase.ExecuteSync[T](Message message, ResultProcessor`1 processor, ServerEndPoint server) in /_/src/StackExchange.Redis/RedisBase.cs:line 53
   at StackExchange.Redis.RedisDatabase.ScriptEvaluate(String script, RedisKey[] keys, RedisValue[] values, CommandFlags flags) in /_/src/StackExchange.Redis/RedisDatabase.cs:line 1212
```

**We don't see this behaviour when we do other cluster operations:**
- Primary Node Fail-overs to Replicants
- Adding/removing shards, which cause a hashslot migrations

I will attempt to dig deeper next week, but I wanted to report it.

slorello89:
Ahh, interesting, well if the Lua script is performing writes, it is illegal to execute it on a replica. To Remove this error you should be able to use the `CommandFlag` `DemandMaster` in your `ScriptEvaluate` commands that are performing writes.

[...]

Issue #1660, opened 2021-01-05
When the master in a cluster is failed over, the writes fail with InternalError until the next auto discovery cycle


dreijer:
I've been testing the StackExchange.Redis library against a Redis cluster with three shards. My topology consists of three physical nodes each running three Redis instances, i.e. one master and two replicas (for the other two shards) on each physical node, resulting in a total of nine Redis instances. Each of the physical nodes has its own public IP address and each Redis instance.

When the master is failed over to one of its replicas, the library doesn't seem to handle the MOVED response properly on set commands. Specifically, when the library processes the MOVED response as it tries to set a value on the old master (which is now a replica), it correctly updates `ServerSelectionStrategy.map` for the given hash slot (i.e. it changes the slot in the array to the `ServerEndpoint` of the new master), but when it tries to re-send the set command, [the logic in ServerSelectionStrategy.Select()](https://github.com/StackExchange/StackExchange.Redis/blob/6409b9d1f2afd01a7eab9498a525360d1ca2b752/src/StackExchange.Redis/ServerSelectionStrategy.cs#L299) causes the old master to be chosen again because the `ServerEndpoint` isn't marked as a master yet:
```
private ServerEndPoint FindMaster(ServerEndPoint endpoint, RedisCommand command)
{
    int max = 5;
    do
    {
        if (!endpoint.IsReplica && endpoint.IsSelectable(command)) return endpoint;

        endpoint = endpoint.Master;
    } while (endpoint != null && --max != 0);
    return null;
}
```
Interestingly, `ServerSelectionStrategy.Select()` has a comment that states that all the entries in the 'map' are masters, which is correct, so why do we need to even call `FindMaster()` on the node if we could just use it directly?
```
ServerEndPoint endpoint = arr[slot], testing;
// but: ^^^ is the MASTER slots; if we want a replica, we need to do some thinking
```
The end result of the current logic is that the set operation ultimately fails with an **InternalServer** error since the same endpoint is tried twice and it's no longer the master.

dreijer:
Several potential solutions come to mind, but I'm not an expert in the library and would love some feedback.

One solution could be to add an override flag that will allow the library to write to the `ServerEndpoint` because we got the MOVED hint, even though it’s still set as a replica, which will make sets work until the next auto discovery will fix everything up. In other words, one could extend the logic in `PhysicalBridge.WriteMessageToServerInsideWriteLock` to check an additional flag on the following line:
[...]

PR #2286, opened 2022-10-27
Cluster: Proactively reconfigure when we hit a MOVED response

NickCraver:
Meant to help address #1520, #1660, #2074, and #2020.

I'm not 100% sure about this because if there is a MOVED happening (e.g. bad proxy somewhere) this would just continually re-run...but only once every 5 seconds. Overall though, we linger in a bad state retrying moves until a discovery happens today and this could be resolved much faster.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -8,9 +8,15 @@ Current package versions:
 
 ## Unreleased
 
+No pending changes for the next release yet.
+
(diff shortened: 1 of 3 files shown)


mgravell (approved):
I'm fine with the hard-coded 5s as an initial; we can revisit based on feedback and knowledge in the future, if appropriate

PR #2320, opened 2022-12-09
Drop unsupported versions of .NET, Update LangVersion to 11, Update .NET SDK to 7.0.x

Romfos:
What was changed:
1) Now minimal supported .NET framework is 4.6.2
2) .NET 5 was replaced by .NET 6
3) .NET SDK was updated to 7,0.x
4) LangVersion now is 11

Motivation:

- NET Framework 4.6.1 is out of support, .NET Framework 4.6.2 is the older version that is still supported
https://devblogs.microsoft.com/dotnet/net-framework-4-5-2-4-6-4-6-1-will-reach-end-of-support-on-april-26-2022/

- .NET 5 is out of support
https://devblogs.microsoft.com/dotnet/dotnet-5-end-of-support-update

- C# 11 has some perf improvements for delegates
https://devblogs.microsoft.com/dotnet/welcome-to-csharp-11/

### DIFF
--- a/.github/workflows/CI.yml
+++ b/.github/workflows/CI.yml
@@ -4,10 +4,10 @@ on:
   pull_request:
   push:
     branches:
-    - main
+      - main
     paths:
-    - '*'
-    - '!/docs/*' # Don't run workflow when files are only in the /docs directory
+      - "*"
+      - "!/docs/*" # Don't run workflow when files are only in the /docs directory
 
 jobs:
   main:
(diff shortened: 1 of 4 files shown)


mgravell:
build langver and SDK: no problem whatsoever, as long as the CI works; it gives us better platform tools to work with - 👍 

dropping TFMs is a much more complex topic, as it impacts users - I don't need to be the arbiter of someone else's business challenges; now, `net461` - I don't think I can raise much of a concern about this one; `net5.0` however ... unless we actively *gain* something by dropping it, I'm not sure I see the point in changing this. We're not even *reducing* the TFM count (and thus nupkg size) - we're just exchanging `net5.0` for `net6.0`. That'll definitely happen in time, but; is there an actual purpose in doing that right now? what do we *achieve*, in trade for cutting some people off?

Romfos:
@mgravell 

Some people that are still on .NET 5 will be able to use this package because of .NET Core 3.1 is still there(+ .NET Standard 2.0)
the main reason: reduce cognitive difficulty by supporting\testing\bugfix on unsupported versions of .NET

Maybe you are right, let's review it later

PR #2322, opened 2022-12-09
Honor select disposition in transactions

### LINKED ISSUE #2321: SELECT command mapping not honored in transactions
Hey @NickCraver @mgravell  👋 
Follow on to #1988 - originally fixed by #2023 looks like there's a case where if you remove select from the command map and try to run a transaction with ad-hoc commands, it will have a write failure. As it turns out when the RedisTransaction enqueues a message it checks to see if it's unknown/eval/evalsha and creates a select message and it enqueues that afterwards. I'll open up a PR in a couple minutest to address - figured I'd open an issue for tracking.

slorello89:
Fixes #2321 

For transactions, if you remove `SELECT` from the command map, it will still try to send one if you issue an unknown/eval/evalsha command - this adds a bit of logic when enqueuing messages to interrogate the multiplexer to see if SELECT is in the command map.

### DIFF
--- a/src/StackExchange.Redis/RedisTransaction.cs
+++ b/src/StackExchange.Redis/RedisTransaction.cs
@@ -121,7 +121,8 @@ namespace StackExchange.Redis
                     case RedisCommand.UNKNOWN:
                     case RedisCommand.EVAL:
                     case RedisCommand.EVALSHA:
-                        if (multiplexer.CommandMap.IsAvailable(RedisCommand.SELECT))
+                        var server = multiplexer.SelectServer(message);
+                        if (server != null && server.SupportsDatabases)
                         {
                             // people can do very naughty things in an EVAL
                             // including change the DB; change it back to what we
--- a/src/StackExchange.Redis/RedisTransaction.cs
+++ b/src/StackExchange.Redis/RedisTransaction.cs
@@ -116,14 +116,13 @@ namespace StackExchange.Redis
             lock (SyncLock)
             {
                 (_pending ??= new List<QueuedMessage>()).Add(queued);
-
-                if(multiplexer.CommandMap.IsAvailable(RedisCommand.SELECT))
+                switch (message.Command)
                 {
-                    switch (message.Command)
-                    {
-                        case RedisCommand.UNKNOWN:
-                        case RedisCommand.EVAL:
-                        case RedisCommand.EVALSHA:
+                    case RedisCommand.UNKNOWN:
+                    case RedisCommand.EVAL:
+                    case RedisCommand.EVALSHA:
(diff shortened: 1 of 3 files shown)


src/StackExchange.Redis/RedisTransaction.cs
```diff
@@ -117,20 +117,23 @@ private void QueueMessage(Message message)
             {
                 (_pending ??= new List<QueuedMessage>()).Add(queued);
 
-                switch (message.Command)
```
mgravell, line 120: can we move the test *inside* the `switch`? I'd rather not execute the `IsAvailable` check etc unless we think there's a good need; other than that: makes sense

mgravell: not related to this change, but: unless I'm misreading the code, it *looks* like we'd issue this even on "cluster"; open question: should we also add a single-DB check? we usually know the DB count, either from server metadata (when available) or from the server type

NickCraver:
Dropping a reminder: Release notes please! Happy to add when ready at the end too, just making sure we remember :)

src/StackExchange.Redis/RedisTransaction.cs
slorello89: sounds good wrt to rearranging things.

WRT preventing an issue with cluster - what do you think of this new approach?

slorello89:
> Dropping a reminder: Release notes please! Happy to add when ready at the end too, just making sure we remember :)

sorry @NickCraver - added :)

mgravell:
The more I look at this, the more I wonder whether this is actually needed, or the right place; the comparison is to here: https://github.com/StackExchange/StackExchange.Redis/blob/02c29ef8e8ddf8de9556c113a8a6fb1563f46a83/src/StackExchange.Redis/PhysicalBridge.cs#L1449 - IMO RediaTransaction might not need to do anything here, if we can refactor that later a bit to handle sub-messages. I can look on Monday

slorello89:
> The more I look at this, the more I wonder whether this `RedisTransaction` code is actually needed, or the right place; the comparison is to here:
> 
> https://github.com/StackExchange/StackExchange.Redis/blob/02c29ef8e8ddf8de9556c113a8a6fb1563f46a83/src/StackExchange.Redis/PhysicalBridge.cs#L1449
> 
> - IMO `RediaTransaction` might not need to do anything here, if we can refactor that linked code above a bit to handle sub-messages. I can look on Monday

I see - it does seem a tad bit out of place. The code does intersect there to set the database to dirty I think - so it might just work if we ripped this bit out.

mgravell (approved):

PR #2337, opened 2023-01-03
better API support on ServerSnapshot

mgravell:
so we aren't limited *just* to `ReadOnlySpan<ServerEndpoint>`, and don't need the allocatey `ToArray()`

implemented as custom iterator which allows `async` and LINQ to work directly' existing code still uses span for efficiency, with the new API used in limited scenarios only


NickCraver (approved):
This looks awesome - no notes, much cleaner and dig the implementation. Good call going this route!

PR #2338, opened 2023-01-03
PhysicalBridge: Fix orphaning cases of PhysicalConnection where it never finishes

NickCraver:
What happens here is that `PhysicalConnection` attempts to connect but never hears back, so it's stuck in `State.ConnectedEstablishing` state, which is handled in the heartbeat code. In the situation where the "last write seconds ago" passes in the heartbeat and we haven't heard back, we fire an `PhysicalBridge.OnDisconnected()` which clears out `physical` and orphans the socket listening forever. This now properly disposes of that `PhysicalConnection` mirroring like we do in `State.Connecting` which will properly fire `OnDisconnected()` and clean up the orphan socket.

The situation manifests where we establish a TCP connection, but not a Redis connection. All of the cases in the memory dump we're analyzing are some bytes sent and 0 received. Likely a Redis server issue, but the client is then handling it incorrectly and leaking.

I nuked the unused `RemovePhysical` method just to prevent further oops here.

### DIFF
--- a/src/StackExchange.Redis/PhysicalBridge.cs
+++ b/src/StackExchange.Redis/PhysicalBridge.cs
@@ -576,6 +576,7 @@ namespace StackExchange.Redis
                                 else
                                 {
                                     OnDisconnected(ConnectionFailureType.SocketFailure, tmp, out bool ignore, out State oldState);
+                                    using (tmp) { } // dispose etc
                                 }
                             }
                             else if (writeEverySeconds <= 0 && tmp.IsIdle()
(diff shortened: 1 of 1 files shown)


philon-msft (approved):

src/StackExchange.Redis/PhysicalBridge.cs
```diff
@@ -576,6 +576,7 @@ internal void OnHeartbeat(bool ifConnectedOnly)
                                 else
                                 {
                                     OnDisconnected(ConnectionFailureType.SocketFailure, tmp, out bool ignore, out State oldState);
+                                    using (tmp) { } // dispose etc
```
mgravell, line 579: Well, there's no real "etc" here (if we ignore the `try`/`finally` since the `try` is empty) - `tmp.Dispose();` (or `?.`) would seem more direct and would avoid the comment? But I don't care as such

mgravell (approved):

src/StackExchange.Redis/PhysicalBridge.cs
NickCraver: Was mirror above there only - agree on simplifying (did both). Also added some more comments to hopefully help later - take another peek pretty please?

mgravell (approved):
I can't see any way it could be null, so 👍 (note `using` is effectively `?.Dispose()` for comparison)

PR #2346, opened 2023-01-16
API suggestion: add ROM for Lua+exec

mgravell:
only Lua implemented for now, to discuss API shape

- [x] add `ReadOnlyMemory<T>` API for Lua
- [x] add single-key API for Lua
- [x] fix incorrect use of non-RO onward call for RO Lua in key-prefix path
- [ ] add `ReadOnlyMemory<object>` for exec (modules)? or try to be more clever? maybe a new `readonly struct RedisKeyOrValue` which is a known union of the two, and a `ReadOnlyMemory<RedisKeyOrValue>` ? (at the moment we take `object` because the API can't know the positions of key vs non-key args, and we need to know which are keys for routing - but using `object` forces a lot of boxing)
- [ ] add / fixup tests
- [ ] (open question) any other APIs where we should consider `ReadOnlyMemory<T>` APIs? `mget`? `hmget`? `del`? `touch`? `mset`? `hset`?
- [ ] (open question) add on `IServer` path? not sure `IServer` needs super optimize
- [ ] (open question) change `[Loaded]LuaScript` to exploit this with leased buffers? (I am not a fan of that API)

in particular note the lease approach used in the key-prefix path

discussion topic: any usage passing the literals `null` or `default` as the current `RedisKey[] keys` arg is now ambiguous due to the single-key overload (`RedisKey` has some operators); this is source-breaking, but is trivially fixable. How much do we care about that? in reality, the number of Lua scripts that don't take any keys can be approximated as zero, so I *do not* anticipate this impacting much code.


NickCraver:
Poking at this locally, I think the main problem I see is the current `null`/`default` case, I can see that likely breaking someone. Specifically the set of:
```cs
        db.ScriptEvaluate(script, default); // BOOM
        db.ScriptEvaluate(script, null); // BOOM
```
I was thinking if we removed the `default` defaults for the `ReadOnlyMemory` overloads and forced the user to pass `default` there it'd improve things but the most basic conflict is early on: that `RedisKey` (from `null->string->RedisKey` implicit) and `RedisKeys[]?` parameter set which I'm not sure how we'd get around - that's always going by be ambiguous with a single argument and I'd imagine that is used by someone today.

PR #2351, opened 2023-01-22
LuaScript: use invariant regex on parameter parsing (fixes #2350)

### LINKED ISSUE #2350: ParameterExtractor of cultureInfo issue
There is a CultureInfo problem in the process of finding the "@blabla" parameters defined in the script text with regex. 
For example
A parameter named "@testIId" cannot be found in script text when CulturerInfo is "tr-TR". The character "I" has always been a problem for this Culture :)  RegexOptions.CultureInvariant can be added as a solution.

Old

```csharp 
private static readonly Regex ParameterExtractor = new Regex(@"@(?<paramName> ([a-z]|_) ([a-z]|_|\d)*)", RegexOptions.Compiled | RegexOptions.IgnoreCase | RegexOptions.IgnorePatternWhitespace);
```

New 

```csharp 
private static readonly Regex ParameterExtractor = new Regex(@"@(?<paramName> ([a-z]|_) ([a-z]|_|\d)*)", RegexOptions.Compiled | RegexOptions.IgnoreCase | RegexOptions.IgnorePatternWhitespace | RegexOptions.CultureInvariant);
```

NickCraver:
This changes the regex to use `RegexOptions.CultureInvariant` and adds an easy way to do tests like this in the future with `[Fact, TestCulture("tr-TR")]` for example, which will set and restore the culture for a specific test.

Before/after break/fix test included for the Turkish case.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -8,7 +8,7 @@ Current package versions:
 
 ## Unreleased
 
-No pending changes for the next release yet.
+- Fix [#2350](https://github.com/StackExchange/StackExchange.Redis/issues/2350): Properly parse lua script paramters in all cultures ([#2351 by NickCraver](https://github.com/StackExchange/StackExchange.Redis/pull/2351))
 
 ## 2.6.90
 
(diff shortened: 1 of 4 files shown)


mgravell (approved):

mgravell:
Nice attribute approach

mgravell:
(I've eyeballed the other `Regex` usages, and they're very specific to the point where culture sensitivity doesn't really apply and isn't a factor as far as I can tell)

PR #2370, opened 2023-02-09
Implement RedisValue.Length for all underlying storage kinds

### LINKED ISSUE #2368: RedisValue's Length() method
Hi , 
I'm trying to collect statistics while using the Redis client ( in order to spot hot keys , big keys which are frequently used etc. ) on our PROD env ,  and wanted to use the RedisValue's Length() method , however seems it's only partially implemented - 

 StorageType.Null => 0,
            StorageType.Raw => _memory.Length,
            StorageType.String => Encoding.UTF8.GetByteCount((string)_objectOrSentinel!),
            _ => throw new InvalidOperationException("Unable to compute length of type: " + Type),
            
  Since the "Type" property is internal, I can't query the value in order to check whether it would be safe to call this method or not.
  
Is there any workaround  / another suggestion ?

mgravell:
fix #2368

- implement Length() for other encodings (using format layout)
- unify format code
- switch to C# 11 for u8 strings (needed a few "scoped" modifiers adding)
- tests for format and Length

### DIFF
--- a/src/StackExchange.Redis/Format.cs
+++ b/src/StackExchange.Redis/Format.cs
@@ -174,7 +174,7 @@ namespace StackExchange.Redis
 
         internal static bool CouldBeInteger(string s)
         {
-            if (string.IsNullOrEmpty(s) || s.Length > PhysicalConnection.MaxInt64TextLen) return false;
+            if (string.IsNullOrEmpty(s) || s.Length > Format.MaxInt64TextLen) return false;
             bool isSigned = s[0] == '-';
             for (int i = isSigned ? 1 : 0; i < s.Length; i++)
             {
@@ -185,7 +185,7 @@ namespace StackExchange.Redis
         }
         internal static bool CouldBeInteger(ReadOnlySpan<byte> s)
         {
-            if (s.IsEmpty | s.Length > PhysicalConnection.MaxInt64TextLen) return false;
+            if (s.IsEmpty | s.Length > Format.MaxInt64TextLen) return false;
             bool isSigned = s[0] == '-';
             for (int i = isSigned ? 1 : 0; i < s.Length; i++)
             {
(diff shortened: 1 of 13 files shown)


src/StackExchange.Redis/Format.cs
```diff
@@ -355,5 +355,88 @@ internal static unsafe string GetString(ReadOnlySpan<byte> span)
...
+            {
+                if (double.IsPositiveInfinity(value))
+                {
+                    if (!"+inf"u8.TryCopyTo(destination))
+                    {
+                        ThrowFormatFailed();
+                    }
+                    return 4;
+                }
+                if (double.IsNegativeInfinity(value))
```
NickCraver, line 385: Can this just be a cheaper else, given the context?

mgravell: possibly; I was mirroring existing code, but yes: I can't think of a third kind of infinity :)

mgravell: tweaked; please recheck

NickCraver: Looking good 👍

NickCraver:
If we want to go .NET 7, will need to bump Ubuntu too it looks like, not sure what's on their images (have to run in a few but can investigate this afternoon and tidy to get unblocked if you're on other stuff)

NickCraver:
Hmm, image manifest says 7.0.102 is on there: https://github.com/actions/runner-images/blob/main/images/linux/Ubuntu2204-Readme.md, but GitHub/Actions are loading hella slow this morning, makes it hard to dig much more atm.

src/StackExchange.Redis/Format.cs
mgravell: two infinities, and beyond! 🚀

NickCraver (approved):
Looking good 👍

PR #2371, opened 2023-02-09
RESP 3 Support

shacharPash:


NickCraver:
Closing this to eliminate build noise and we're not yet agreed on how things will work - the next step is for Marc to take a pass at this and figure out protocol parsing. We need to be aligned on how this should work (issue) before having the discussion move to PR details.

PR #2383, opened 2023-02-27
API: ConfigureEndpoints for Proxies

usjpin:
Hi team, this PR is a proposal to add new interface APIs `ConfigureEndpoints` and `ConfigureEndpointsAsync` to ConnectionMultiplexer. 

In a distributed environment, we frequently have updates in endpoints from service discovery - due to node refreshes, expansion / shrinking of instances, reschedules from hardware issues, etc. 

For Redis Clusters, this is not an issue since the library is able to rebalance slots using `CLUSTER NODES`. However for proxies, this command is not supported. The current solution we use is to delete and recreate a new ConnectionMultiplexer, which is not ideal due to the incurred downtime. 

The new changes allow for obsolete proxy connections to be replaced, without losing availability of the ConnectionMultiplexer as a whole.

Would like to hear thoughts on this.

### DIFF
--- a/src/StackExchange.Redis/ConnectionMultiplexer.cs
+++ b/src/StackExchange.Redis/ConnectionMultiplexer.cs
@@ -43,7 +43,7 @@ namespace StackExchange.Redis
         internal bool IsDisposed => _isDisposed;
 
         internal CommandMap CommandMap { get; }
-        internal EndPointCollection EndPoints { get; }
+        internal EndPointCollection EndPoints { get; set; }
         internal ConfigurationOptions RawConfig { get; }
         internal ServerSelectionStrategy ServerSelectionStrategy { get; }
         internal Exception? LastException { get; set; }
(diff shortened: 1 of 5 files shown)


NickCraver:
@usjpin The main issue is changing endpoints during runtime and outside of our connections and reconnections has a lot of issues, that's why it's explicitly left out of being reconfigurable or respecting the latest `ConnectionOptions` instance (there are a _lot_ of race conditions when swapping the lists of servers, the endpoints, bridges, connections, and socket we have).

I'm not sure this is a case we want to support, at least in this way. The more appropriate API would be to respect ConnectionOptions's endpoint list in a reconfigure instead of having a specific API for it...but to do that, we have to solve all the edge cases and I'm not sure it's worth it. I think I understand the use case, but we'd want to approach this in a more generic way and not as a one-off API :)

usjpin:
Hi @NickCraver , thanks for the response. I'm wondering if you could give me some example of the issues? I modeled my logic after what happens during Redis Cluster Rebalancing. The old connections are marked unselectable and `ReconfigureAsync()`on top of the newly added ones. 

I understand that we might not want to support this as an API and instead in `ConfigurationOptions` like you said. But I'm wondering what kind of runtime issues this may run into in the case it *is* used as an API? Maybe I'm misunderstanding some part of the behavior here

NickCraver:
@usjpin Anything that relies on endpoints overall: subscriptions and restorations, server selections that are in-flight, etc. - in general this adds a lot of complexity to those scenarios to even think about, to the point I'd say "maybe this is where you'd want a fork of the library", if not using a stable proxy or routable endpoint like most common scenarios (e.g. a load balancer).

NickCraver:
Unfortunately we found another area this will cause problems: async completions. Though we're fixing this in #2413 for current paths, this would create another orphan scenario where a heartbeat would neither root nor complete any pending messages or async continuations on orphaned bridges from the prior endpoints. Given everything this causes (largely why it doesn't exist today) my vote would be: we don't accept this API addition.

mgravell:
Just to be explicit, @NickCraver - your thoughts here are around reassigning Endpoints in a way that removes some, yes?

NickCraver:
@mgravell yeah that happening in several places is not awesome - we should either find a way to do this in ConfigurationOptions that allows you to change them, or not do this IMO. The fact it's not mutable there _but then changes here_ via another API is odd as a starting point, but complicating factors like the heartbeat pathing, continuations, whether DNS is resolved and how we improve that for re-connects...this locks us in and reduces options.

PR #2408, opened 2023-03-21
Timeouts: Improve Backlog handling and errors for users + GC rooting fixes for outstanding scenarios

NickCraver:
## Changes
This combination PR is both fixing a GC issue (see below, and #2413 for details) and improves timeout exception. Basically if a timeout happens for a message that was in the backlog but was never sent, the user now gets a much more informative message like this:

> Exception: The message timed out in the backlog attempting to send because no connection became available - Last Connection Exception: InternalFailure on 127.0.0.1:6379/Interactive, Initializing/NotStarted, last: GET, origin: ConnectedAsync, outstanding: 0, last-read: 0s ago, last-write: 0s ago, keep-alive: 500s, state: Connecting, mgr: 10 of 10 available, last-heartbeat: never, last-mbeat: 0s ago, global: 0s ago, v: 2.6.99.22667, command=PING, inst: 0, qu: 0, qs: 0, aw: False, bw: CheckingForTimeout, last-in: 0, cur-in: 0, sync-ops: 1, async-ops: 1, serverEndpoint: 127.0.0.1:6379, conn-sec: n/a, aoc: 0, mc: 1/1/0, mgr: 10 of 10 available, clientName: NAMISTOU-3(SE.Redis-v2.6.99.22667), IOCP: (Busy=0,Free=1000,Min=32,Max=1000), WORKER: (Busy=2,Free=32765,Min=32,Max=32767), POOL: (Threads=18,QueuedItems=0,CompletedItems=65), v: 2.6.99.22667 (Please take a look at this article for some common client-side issues that can cause timeouts: https://stackexchange.github.io/StackExchange.Redis/Timeouts)

Today, this isn't intuitive especially for connections with `AbortOnConnectFail` set to `false`. What happens is a multiplexer _never_ connects successfully, but the user just gets generic timeouts. This makes the error more specific and includes the inner exception (also as `.InnerException`) for more details, informing the user of a config/auth/whatever error underneath as to why things are never successfully sending.

Also adds `aoc: (0|1)` to the exception message for easier advice in issues (reflecting what `AbortOnConnectFail` is set to).

## Performance
Before:
|             Method |        Job |              Runtime |       Mean |    Error |   StdDev |    Op/s |   Gen 0 |  Gen 1 | Allocated |
|------------------- |----------- |--------------------- |-----------:|---------:|---------:|--------:|--------:|-------:|----------:|
[...]

PR #2415, opened 2023-03-29
Add test for multiplexer GC rooting.

kornelpal:
Add test for multiplexer GC rooting in #2413.

### DIFF
--- a/tests/StackExchange.Redis.Tests/GarbageCollectionTests.cs
+++ b/tests/StackExchange.Redis.Tests/GarbageCollectionTests.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Threading;
 using System.Threading.Tasks;
 using Xunit;
 using Xunit.Abstractions;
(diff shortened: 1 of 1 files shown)


tests/StackExchange.Redis.Tests/GarbageCollectionTests.cs
```diff
@@ -52,4 +53,43 @@ public async Task MuxerIsCollected()
...
+            Assert.False(conn.IsConnected);
+
+            var pingTask = Assert.ThrowsAsync<RedisConnectionException>(() => db.PingAsync());
+            startGC.SetResult(true);
+            await pingTask;
+        }).WithTimeout(5000);
+
+        // Use sync wait and sleep to ensure a more timely GC
+        Task.WaitAny(startGC.Task, testTask);
+        while (!testTask.IsCompleted)
```
mgravell, line 88: May be worth a timeout exit failure condition here, perhaps after N iterations for some N

Also, to check that it hasn't been collected: really, you need to ask a WeakReference? The test is fine, just: we're not *directly* testing whether it has been collected - we're testing whether the task completed. These are causally related, sure; but if the test name says it is checking collection: we should probably actually check for collection. Or rename the test.

kornelpal:
Sure, I'll reopen it after pushing some improvements. Note that this needs to be merged after #2408 so that the test is not failing.

tests/StackExchange.Redis.Tests/GarbageCollectionTests.cs
kornelpal: Note that testTask actually has a timeout via WithTimeout, and without the GC rooting fix this test times out.

Unless you have some good idea, I'll likely just rename the test because a WeakReference could be used to confirm that the multiplexer is not collected before the ping completes, but could not be used to confirm that it was not collected after that as the GC was free to collect it, and this would require a lot of extra code to avoid strong references.

On the other hand because of the strong reference in using in the body of testTask, the task completing implies that the connection was still alive. I'll add a comment about this at the end of testTask.

mgravell:
Renaming it "something something async task completes something": sure;
(naming is hard)

PR #2430, opened 2023-04-05
Fix concurrency issue between AbandonPendingBacklog() and CheckBacklogForTimeouts(), and remove backlog locking

kornelpal:
There is only one ProcessBacklogAsync() thread running at a time and all current backlog locks are within that thread, so there is no need for these locks. On the other hand AbandonPendingBacklog() can run concurrently with the ProcessBacklogAsync() thread, that runs CheckBacklogForTimeouts(), but AbandonPendingBacklog() is not locking the backlog that can result in concurrency issues. This can result in CheckBacklogForTimeouts() leaving the dequeued message abandoned in an uncompleted (hung) state. This fix the resolves the concurrency issue by introducing an _abandonPendingBacklogException field that also enables removing the lock. The "failed" message is completed with the thrown exception to make any potential concurrency issues more visible.

### DIFF
--- a/src/StackExchange.Redis/PhysicalBridge.cs
+++ b/src/StackExchange.Redis/PhysicalBridge.cs
@@ -41,6 +41,7 @@ namespace StackExchange.Redis
         private int _backlogProcessorIsRunning = 0;
         private int _backlogCurrentEnqueued = 0;
         private long _backlogTotalEnqueued = 0;
+        private Exception? _abandonPendingBacklogException;
 
         private int activeWriters = 0;
         private int beating;
@@ -483,11 +484,18 @@ namespace StackExchange.Redis
 
         private void AbandonPendingBacklog(Exception ex)
         {
+            // Peeking at the backlog, checking message and then dequeuing is not thread-safe.
+            // CheckBacklogForTimeouts() depends on this being set to properly complete dequeued messages.
+            Volatile.Write(ref _abandonPendingBacklogException, ex);
+
             while (BacklogTryDequeue(out Message? next))
             {
                 Multiplexer.OnMessageFaulted(next, ex);
                 next.SetExceptionAndComplete(ex, this);
             }
+
+            // Best effort cleanup to avoid false positive thread safey check failures in CheckBacklogForTimeouts().
+            if (_backlogStatus != BacklogStatus.CheckingForTimeout) Interlocked.CompareExchange(ref _abandonPendingBacklogException, null, ex);
         }
 
         internal void OnFullyEstablished(PhysicalConnection connection, string source)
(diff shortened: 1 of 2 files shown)


mgravell:
This feels like we're solving the wrong problem; IMO we should be fixing
whatever gap it is falling in currently. I need to look carefully at what
is going on here, but I don't see that we should need the new exception
bits.

mgravell:
>  that can result in concurrency issues

Please can you be *very explicit* about what concurrency issue we're discussing? what is the actual symptom/issue that we're looking at resolving here? To understand whether this resolves them, first I need to have a clear vision of what that "them" are.

So: talk me through it; what scenario are we discussing?

kornelpal:
Although PhysicalBridge._backlog is a ConcurrentQueue, PhysicalBridge.CheckBacklogForTimeouts() is using it in a non-thread-safe way. The existing comment from that method describes it best:
> Because peeking at the backlog, checking message and then dequeuing, is not thread-safe, we do have to use
> a lock here, for mutual exclusion of backlog DEQUEUERS. Unfortunately.

When not all dequeuers are locking the backlog then CheckBacklogForTimeouts() can dequeue a message then abandon it without ever being completed.

Code from inside the lock in CheckBacklogForTimeouts() annotated by me for the problematic scenario:
```csharp
// There is a message in the backlog, so no break.
if (!_backlog.TryPeek(out message)) break;
// The message peeked at has timed out, so no break.
if (!message.HasTimedOut(now, timeout, out var _)) break;
// Another thread without locking the backlog already dequeued the previous message
// between the TryPeek() and BacklogTryDequeue() calls.
// Scenario 1; there were no messages left: This is not really an issue.
// Scenario 2; another message (message2) was dequeued: It may or may not be timed out,
// but the current logic does not care, just abandons the message and it will not be completed
// as it is not stored anywhere else. This is a problem for async messages only,
// not for sync (wait timeout), or F+F (not completed otherwise either).
if (!BacklogTryDequeue(out var message2) || (message != message2))
{
    // In both Scenario 1 and 2 the backlog processing thread fails,
    // but a new one will be started by the heartbeat or by adding a message to the backlog.
    throw new RedisException("Thread safety bug detected! A queue message disappeared while we had the backlog lock");
}
```

Methods dequeuing from the backlog:
- CheckBacklogForTimeouts(): Properly locks the backlog, and only runs on the backlog processing thread.
- ProcessBacklogAsync(): Properly locks the backlog, and only runs on the backlog processing thread.
- AbandonPendingBacklog(): Does not lock the backlog and can run concurrently with the backlog processing thread. Called from PhysicalBridge.Dispose(), ~PhysicalBridge() and PhysicalBridge.OnConnectionFailed(). 

Since the two methods that actually lock the backlog cannot run concurrently, the current lock is just an overhead.

On the other hand not locking the backlog in AbandonPendingBacklog() can cause the concurrency issue described in the annotated code above that can cause one task per occurrence to be left in a hung state.

kornelpal:
I've added a test for the Dispose() case. It fails without the fix and succeeds with the fix. Should be possible to cause the issue for BacklogPolicy.AbortPendingOnConnectionFailure = true too, but I don't know how to simulate a connection failure with a large backlog.

kornelpal:
I just realized that clearing _abandonPendingBacklogException at the end of AbandonPendingBacklog() can result in CheckBacklogForTimeouts() failing when AbandonPendingBacklog() is running on multiple threads in parallel, so more complexity (like a wrapper for the backlog) would be needed reliable bug detection in CheckBacklogForTimeouts().

kornelpal:
I have one more idea, inspired by PhysicalBridge.HasPendingCallerFacingItems(); Instead of removing items, CheckBacklogForTimeouts() could be changed to enumerate the items, and ProcessBridgeBacklogAsync() could be changed to ignore completed items. This way the concurrency issue was eliminated and there was no need for a lock or the exception field. Although adds some more compute overhead, checking for timeout in ProcessBridgeBacklogAsync() again might be simpler than adding tweaks at other places to complete timed out sync messages and identify timed out F+F messages (that never have a result box).

PR #2444, opened 2023-04-25
Address misleading documentation

### LINKED ISSUE #2443: Misleading documentation about the evaluation of server loaded Lua scripts
Recently, I had to investigate an issue with server failovers which cleared the `SHA-1` hash cache and prevented me from evaluating back-end cached lua scripts because the hashes themselves did not exist on the server. The `StackExchange.Redis` version was quite outdated and did not feature the improvements in `v2.6.66` which change how loaded scripts were evaluated.

Instead of sending the `SHA-1` hash value, now the script itself is evaluated which works for our use case. However, I was trying to replicate the issue we experienced in an isolated environment using a docker container. I was not able to reproduce the initial issue after I updated the library, whenever I rebooted the container. The `XML` documentation for the `LoadedLuaScript::Evaluate()` & `LoadedLuaScript::EvaluateAsync()` do not reference the new behavior and mislead me. Finally, I had to take a look at the source code in order to understand why after a `SCRIPT FLUSH [SYNC|ASYNC]` command, the loaded lua scripts were able to be evaluated even though the hashes were not present on the server.

SonnyRR:
Remove references that mention SHA-1 hashes being passed when evaluating server loaded Lua scripts. The current implementation passes the script itself instead of the hash, generated when the script was initially loaded, due to resiliency concerns.

Resolves #2443


SonnyRR:
This whole abstraction: `LoadedLuaScript` abstraction seems kind of weird to me in it's current state. I guess it's not deprecated due to compatibility concerns. But I was wondering is there another reason that I might be unaware of ? Running `SCRIPT EXIST sha1` before evaluating the script by it's hash does not sound ideal to me, but let's say we're always checking for a single script the complexity should be `O(1)`. Are there any gains to be made if a check was introduced to evaluate the script by it's hash - if it exists, and if not - evaluate it with `EVAL`.

docs/Scripting.md
```diff
@@ -35,10 +35,10 @@ Any object that exposes field or property members with the same name as @-prefix
  - RedisKey
  - RedisValue
 
-StackExchange.Redis handles Lua script caching internally. It automatically transmits the Lua script to redis on the first call to 'ScriptEvaluate'. For further calls of the same script only the hash with [`EVALSHA`](https://redis.io/commands/evalsha) is used.
+StackExchange.Redis handles Lua script caching internally. It automatically transmits the Lua script to redis on the first call to 'ScriptEvaluate'. For further calls of the same script [`EVAL`](https://redis.io/commands/eval) is used instead of [`EVALSHA`](https://redis.io/commands/evalsha), due to resiliency concerns.
 
 For more control of the Lua script transmission to redis, `LuaScript` objects can be converted into `LoadedLuaScript`s via `LuaScript.Load(IServer)`.
```
SonnyRR, line 40: I was wondering if we should get rid of this paragraph at all, since the loaded script abstraction doesn't evaluate the script by its hash anymore. Any thoughts on this ?

mgravell:
Thing is... I'm not sure it is wrong. It should still hit the local script cache against our endpoint instance, so once we've seen it load: it should switch to the hash version. I'd want to investigate this via redis-cli monitor before updating any docs (noting that it will still use the full script initially)

SonnyRR:
@mgravell Thanks for the heads up, I saw that `EVAL` was being used after a `SCRIPT LOAD` command was executed to load a script that previously was cached. When I saw that the changes to the loaded script abstraction, that do not invoke the `ScriptEvaluate(byte[] hash, ...)` overload, I presumed it always used the `EVAL` redis command. I'm still new to this codebase, so I've missed that local cache instance. Can you also verify that this is the behavior you're seeing in the monitor logs? Unfortunately, I cannot share examples, but I can probably setup a sandbox project that replicates the problem.

I'll re-word my changes to make it clear what the current behavior is.

docs/Scripting.md
SonnyRR: This comment is irrelevant.

PR #2445, opened 2023-04-25
DefaultOptionsProvider: allow providing User/Password

NickCraver:
For wrappers which intend to provide this (e.g. managed service accounts and such), the intent was for them to override these and return their "current" values (e.g. as a token rotates), but I missed them in the initial pass...even though this was the original extensibility reason, because I suck! Fixing.


src/StackExchange.Redis/Configuration/DefaultOptionsProvider.cs
```diff
@@ -172,6 +172,16 @@ public static void AddProvider(DefaultOptionsProvider provider)
         /// </summary>
         public virtual TimeSpan ConfigCheckInterval => TimeSpan.FromMinutes(1);
 
+        /// <summary>
+        /// The user to use to authenticate with the server.
```
philon-msft, line 176: "The user NAME to use"?

philon-msft (approved):

src/StackExchange.Redis/Configuration/DefaultOptionsProvider.cs
NickCraver: Done!

mgravell (approved):

PR #2448, opened 2023-04-26
Fix HTTP tunnel

flobernd:
According to https://datatracker.ietf.org/doc/html/draft-luotonen-web-proxy-tunneling-01#section-3.2, the expected response to a `CONNECT` should be `HTTP/1.1 200 Connection established` instead of `HTTP/1.1 200 OK`. Allow both responses for now (as still a lot of proxy implementations fail to follow the standard).

A HTTP(S), SOCKS, etc. proxy is traditionally specified by an URI of form `scheme://[user:pass@]host[:port]`. Accept both formats for backwards compatibility.


src/StackExchange.Redis/ConfigurationOptions.cs
```diff
@@ -904,19 +904,25 @@ private ConfigurationOptions DoParse(string configuration, bool ignoreUnknown)
                             {
                                 Tunnel = null;
                             }
-                            else if (value.StartsWith("http:"))
+                            else
                             {
-                                value = value.Substring(5);
-                                if (!Format.TryParseEndPoint(value, out var ep))
+                                // For backwards compatibility with `http:address_with_port`.
```
mgravell, line 909: IMO we should add tests for this change, i.e. that we get the expected values back out - even if that means poking at internal state after the parse

mgravell (changes requested):
functional change looks good; parsing change: IMO needs tests

src/StackExchange.Redis/ConfigurationOptions.cs
flobernd: I added the new format to the roundtrip test. This should be sufficient to confirm that the resulting `Tunnel` is the same for both input formats (`HttpProxyTunnel.ToString()` internally uses `this.EndPoint.ToString()` which means `address` and `port` must be the same to end up with the same string).

mgravell (approved):

flobernd:
Seems like the CI is broken 👀

NickCraver:
@flobernd updated release notes and landed on - thank you!!

PR #2453, opened 2023-05-02
DefaultOptionsProvider: allow overriding LibName

NickCraver:
If a wrapper package is generally in use in a deployment, it may want to override what we set as the library name in `CLIENT SETINFO lib-name <name>`. This allows doing so via the `DefaultOptionsProvider` (intentionally not on `ConfigurationOptions` directly as version isn't either).

Note that this does NOT upgrade the test suite to 7.2.0 RC1. I did test and this works, but there are other breaks we need to evaluate - I'll open another PR separately to demonstrate.

### DIFF
--- a/src/StackExchange.Redis/Configuration/DefaultOptionsProvider.cs
+++ b/src/StackExchange.Redis/Configuration/DefaultOptionsProvider.cs
@@ -198,6 +198,12 @@ namespace StackExchange.Redis.Configuration
              ?? "StackExchange.Redis") + "(SE.Redis-v" + LibraryVersion + ")";
 
         /// <summary>
+        /// Gets the library name to use for CLIENT SETINFO lib-name calls to Redis during handshake.
+        /// Defaults to "SE.Redis".
+        /// </summary>
+        public virtual string LibraryName => "SE.Redis";
+
+        /// <summary>
         /// String version of the StackExchange.Redis library, for use in any options.
         /// </summary>
         protected static string LibraryVersion => Utils.GetLibVersion();
--- a/src/StackExchange.Redis/ServerEndPoint.cs
+++ b/src/StackExchange.Redis/ServerEndPoint.cs
@@ -936,7 +936,7 @@ namespace StackExchange.Redis
                     log?.WriteLine($"{Format.ToString(this)}: Setting client lib/ver");
 
                     var libName = Multiplexer.RawConfig.Defaults.LibraryName;
-                    if (!string.IsNullOrEmpty(libName))
+                    if (!string.IsNullOrWhiteSpace(libName))
                     {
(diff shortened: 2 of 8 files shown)


src/StackExchange.Redis/Configuration/DefaultOptionsProvider.cs
```diff
@@ -197,6 +197,12 @@ public static void AddProvider(DefaultOptionsProvider provider)
              ?? ComputerName
              ?? "StackExchange.Redis") + "(SE.Redis-v" + LibraryVersion + ")";
 
+        /// <summary>
+        /// Gets the library name to use for CLIENT SETINFO lib-name calls to Redis during handshake.
+        /// Defaults to "SE.Redis".
+        /// </summary>
+        public virtual string LibraryName => "SE.Redis";
```
philon-msft, line 204: Should use `RedisLiterals.SE_Redis`?

src/StackExchange.Redis/ServerEndPoint.cs
```diff
@@ -935,10 +935,14 @@ private async Task HandshakeAsync(PhysicalConnection connection, LogProxy? log)
                     // server version, so we will use this speculatively and hope for the best
                     log?.WriteLine($"{Format.ToString(this)}: Setting client lib/ver");
 
-                    msg = Message.Create(-1, CommandFlags.FireAndForget, RedisCommand.CLIENT,
-                        RedisLiterals.SETINFO, RedisLiterals.lib_name, RedisLiterals.SE_Redis);
-                    msg.SetInternalCall();
-                    await WriteDirectOrQueueFireAndForgetAsync(connection, msg, ResultProcessor.DemandOK).ForAwait();
+                    var libName = Multiplexer.RawConfig.Defaults.LibraryName;
+                    if (!string.IsNullOrEmpty(libName))
```
philon-msft, line 939: `IsNullOrWhiteSpace`?

NickCraver: YES, meant this - good catch!

src/StackExchange.Redis/ServerEndPoint.cs
```diff
@@ -935,10 +935,14 @@ private async Task HandshakeAsync(PhysicalConnection connection, LogProxy? log)
                     // server version, so we will use this speculatively and hope for the best
                     log?.WriteLine($"{Format.ToString(this)}: Setting client lib/ver");
```
mgravell, line 936: a bit late now, but: maybe this should be inside the null/empty check

NickCraver: Agree it's weird, but there are 2 distinct checks and 1 log message atm - we could log each, change the method, or something else - thoughts? Happy to tidy up!

PR #2454, opened 2023-05-02
Redis 7.2.0 RC1: Prep for upgrades

NickCraver:
I'm seeing a few issues here that we need to resolve for the upcoming Redis 7.2 release:
- [x] The internal encoding has changed 1 spot to listpack, which is correct from https://github.com/redis/redis/pull/11303, but is missing in the docs at https://redis.io/commands/object-encoding/ (fixed in tests themselves)
- [x] The `HackyGetPerf` reliably returns 0 now, regardless of how long has passed (e.g. upping iterations tremendously)...this may legit be bugged.
- [x] `StreamAutoClaim_IncludesDeletedMessageId` expectations are broken, not sure what to make of this yet but it's an odd change to hit between 7.0 and 7.2 versions.

Note: no release notes because these are all test tweaks.

### DIFF
--- a/tests/StackExchange.Redis.Tests/ScriptingTests.cs
+++ b/tests/StackExchange.Redis.Tests/ScriptingTests.cs
@@ -98,27 +98,6 @@ public class ScriptingTests : TestBase
     }
 
     [Fact]
-    public void HackyGetPerf()
-    {
-        using var conn = GetScriptConn();
-
-        var key = Me();
-        var db = conn.GetDatabase();
-        db.StringSet(key + "foo", "bar", flags: CommandFlags.FireAndForget);
-        var result = (long)db.ScriptEvaluate(@"
-redis.call('psetex', KEYS[1], 60000, 'timing')
-for i = 1,500000 do
-    redis.call('set', 'ignore','abc')
-end
-local timeTaken = 60000 - redis.call('pttl', KEYS[1])
-redis.call('del', KEYS[1])
-return timeTaken
-", new RedisKey[] { key }, null);
-        Log(result.ToString());
-        Assert.True(result > 0);
-    }
-
-    [Fact]
     public async Task MultiIncrWithoutReplies()
     {
         using var conn = GetScriptConn();


NickCraver:
cc @slorello89 for heads up on these, may affect others!

tests/StackExchange.Redis.Tests/ScriptingTests.cs
```diff
@@ -107,7 +107,7 @@ public void HackyGetPerf()
         db.StringSet(key + "foo", "bar", flags: CommandFlags.FireAndForget);
         var result = (long)db.ScriptEvaluate(@"
 redis.call('psetex', KEYS[1], 60000, 'timing')
-for i = 1,5000 do
+for i = 1,500000 do
```
uglide, line 110: @NickCraver What is the reason for changing the number of iterations for the loop?

slorello89:
Hey Nick - the error your seeing here is caused by slightly different behavior between the Release Candidate and the actual release of Redis 7:

```
While iterating the PEL, if XAUTOCLAIM stumbles upon a message which doesn't exist in the stream anymore (either trimmed or deleted by [XDEL](https://redis.io/commands/xdel)) it does not claim it, and deletes it from the PEL in which it was found. This feature was introduced in Redis 7.0. These message IDs are returned to the caller as a part of XAUTOCLAIMs reply.
```

Might seem innocuous but when you take it in context with:

```
The optional <count> argument, which defaults to 100, is the upper limit of the number of entries that the command attempts to claim.
```

Tells you what's going on, the test is sending the XAUTOCLAIM with a COUNT of 1 - it attempts to claim the deleted message. THAT COUNTS as its one message it will attempt to claim, and it breaks off. It returns the next message as the place to start the next autoclaim, returns a null entry as the actual messages claimed, and returns the deleted message in the third spot.

The fix is simple enough if you want me to push it (just change the counts to 2 :) )

NickCraver:
@slorello89 If the test is wrong and server is right, absolutely - would be much appreciated. I'd love for tests to be wrong here :)

tests/StackExchange.Redis.Tests/ScriptingTests.cs
slorello89: to be clear - only during lua script execution (catches up when the script exits)

slorello89:
@NickCraver

Insofar as `XAUTOCLAIM` the server is behaving correctly, so the update is correct

WRT the other issue with the perf test - this is a BC coming to Redis 7.2:

https://github.com/redis/redis/issues/10182
https://github.com/redis/redis/pull/10300

See the Breaking section of the [release notes](https://github.com/redis/redis/releases/tag/7.2-rc1) - basically what's happening is that Redis 7.2 is freezing the clock during script execution. Consequently, that test is no longer correct and will always fail. Obviously there's a couple ways to remediate, but I'm not entirely sure what that test is really testing?

Anyway - hopefully this is helpful

NickCraver:
@slorello89 _very_ helpful - thank you!!

@mgravell any thoughts about the perf test/intent? For quick context, this is the entire content of the test: 
```cs
    [Fact]
    public void HackyGetPerf()
    {
        using var conn = GetScriptConn();

        var key = Me();
        var db = conn.GetDatabase();
        db.StringSet(key + "foo", "bar", flags: CommandFlags.FireAndForget);
        var result = (long)db.ScriptEvaluate(@"
redis.call('psetex', KEYS[1], 60000, 'timing')
for i = 1,5000 do
    redis.call('set', 'ignore','abc')
end
local timeTaken = 60000 - redis.call('pttl', KEYS[1])
redis.call('del', KEYS[1])
return timeTaken
", new RedisKey[] { key }, null);
        Log(result.ToString());
        Assert.True(result > 0);
    }
```

The test originally comes from a Booksleeve migration 9 years ago...I'm leaning towards "delete it": https://github.com/StackExchange/StackExchange.Redis/commit/c002a00e19ad22652c230213939244de3e76b551

mgravell (approved):
Lgtm; fine with deleting the hack perf thing

PR #2483, opened 2023-06-14
add DIM for Publish[Async]

mgravell:
The recent change to `RedisChannel` doesn't impact `Publish`/`PublishAsync`, as publish is always single-channel; to minimize the code-changes required, this PR adds Default Interface Methods (DIMs) to shim `Publish[Async](channel, ...)` via `Publish[Async](RedisChannel.Literal(channel), ...)`

This means that - at least on modern runtimes - code that uses `Publish(string)` today will not need to be adjusted.

It was investigated whether the same could be achieved via an extension method, and: it could not (the non-extension overload with implicit conversion ranks as preferable). Adding regular interface methods seems unnecessary when the implementation is always the same.

It is not seen as necessary to add `byte[]` overloads, as `byte[]` channels are so niche as to be irrelevant.

Question: is this overkill?

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -8,6 +8,8 @@ Current package versions:
 
 ## Unreleased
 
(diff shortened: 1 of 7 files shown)


mgravell:
Fair enough; shelving

PR #2494, opened 2023-06-28
Fix nullability annotation of IConnectionMultiplexer.RegisterProfiler

eerhardt:
RegisterProfiler allows for a null ProfilingSession to be returned - it is even documented with "or returning null to not profile".

Fixing the nullable annotation to indicate that null is an acceptable result of the profilingSessionProvider function.


NickCraver (approved):
Perfect - thanks for this, and apologies for the miss when doing NRTs here!

PR #2497, opened 2023-06-29
Enable net6.0 build

eerhardt:
This allows for using new APIs introduced in net6.0.

In order to enable building for net6.0, we also need to revert the thread pool changes in #1939 and #1950. This was already effectively reverted in #1992 by not building for net6.0. Now that we are building for net6.0 again, these if-defs need to be removed.

Questions:

1. Does this need a ReleaseNotes entry? I don't think there is any user-visible affect to this change.
1. Should `net5.0` be removed? It has been end-of-life for over a year. I left it in, for now. It can be removed in the future, if we think it should be removed.

cc @NickCraver @mgravell

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -8,6 +8,7 @@ Current package versions:
 
 ## Unreleased
 
+- Change: Target net6.0 instead of net5.0, since net5.0 is end of life. ([#2497 by eerhardt](https://github.com/StackExchange/StackExchange.Redis/pull/2497))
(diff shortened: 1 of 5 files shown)


NickCraver:
FWIW I'm good with dropping `net5.0` here given it's non-LTS and EOL over a year ago - @mgravell thoughts?

NickCraver (approved):
Looking good - thanks for the tweaks!

0xfeeddeadbeef:
Why not drop .NET Core 3.1 as well while at it?

NickCraver:
@0xfeeddeadbeef because .NET Core 3.1 is newer and was an LTS which many more people are on for the long-term. If we're dropping an LTS, I'd agree with a minor version bump. For a non-LTS over a year past EOL I'm a lot less concerned :)

PR #2498, opened 2023-06-29
Support sharded pubsub commands

vandyvilla:
Support SPUBLISH and SSUBSCRIBE, SUNSUBSCRIBE commands. 

We tested this PR internally and will add more tests to this PR. 

Would like to gather some early feedback. Thanks! 

cc @NickCraver @mgravell


vandyvilla:
Not sure how to fix below build error: 

D:\a\StackExchange.Redis\StackExchange.Redis\Build.csproj : error :   MSB4276: The default SDK resolver failed to resolve SDK "Microsoft.Build.Traversal" because directory "C:\Program Files\dotnet\sdk\7.0.304\Sdks\Microsoft.Build.Traversal\Sdk" did not exist.
D:\a\StackExchange.Redis\StackExchange.Redis\Build.csproj : error MSB4236: The SDK 'Microsoft.Build.Traversal/3.0.2' specified could not be found.

mgravell:
`RedisChannel` seems to have a new `bool` field, increasing the size; to simplify design and avoid additional storage, since `SSUBSCRIBE` does not support wildcards, I wonder if we should implement this as a new `PatternMode` value instead of a separate `bool` - perhaps with a `public static RedisChannel Sharded(...)` pair to match the `Literal` and `Pattern` methods. At the moment there is a `private readonly bool _isPatternBased`, but ultimately this is padded to 4 bytes, so would be identical to `private readonly PatternMode _mode;`. As an alternative, we could use bit-bashing to combine the two bools as bits, to avoid increasing the size - but honestly I think the `PatternMode` approach is simpler.


My main concern, however, is around shard rebalancing; I don't see anything that handles this currently; presumably we'd need to detect shard movement, and validate that subscriptions are still against the correct server - right now I don't see that happening; we may or may not also need to do this during resubscribe against an existing connection?

mgravell:
Does this actually work without error? I'm a little surprised because I would have expected `CheckReadOnlyOperations` to fail, because I don't think you've modified `IsPrimaryOnly`?

src/StackExchange.Redis/PhysicalConnection.cs
```diff
@@ -1579,8 +1579,14 @@ private void MatchResult(in RawResult result)
                     }
 
                     // invoke the handlers
-                    var channel = items[1].AsRedisChannel(ChannelPrefix, RedisChannel.PatternMode.Literal);
-                    Trace("MESSAGE: " + channel);
+                    RedisChannel channel;
+                    if (items[0].IsEqual(message)) {
```
mgravell, line 1583: feels like we're testing this a lot; can we hoist this to a `bool` somewhere? maybe:

``` c#
if (items.Length >= 3 && IsSimpleMessage(items[0], out bool isSharded))
```

then just use `isSharded` in all the places? for example `channel = items[1].AsRedisChannel(ChannelPrefix, RedisChannel.PatternMode.Literal, isSharded);`

src/StackExchange.Redis/RedisSubscriber.cs
```diff
@@ -485,6 +489,7 @@ private bool UnregisterSubscription(in RedisChannel channel, Action<RedisChannel
             return false;
         }
 
+        // TODO: We need a new api to support SUNSUBSCRIBE all. Calling this now would unsubscribe both sharded and unsharded channels.
```
mgravell, line 492: do we actually need such an API?

src/StackExchange.Redis/ResultProcessor.cs
```diff
@@ -1325,7 +1325,7 @@ protected override bool SetResultCore(PhysicalConnection connection, Message mes
                 {
                     case ResultType.MultiBulk:
                         var final = result.ToArray(
-                                (in RawResult item, in ChannelState state) => item.AsRedisChannel(state.Prefix, state.Mode),
+                                (in RawResult item, in ChannelState state) => item.AsRedisChannel(state.Prefix, state.Mode, isSharded: false),
```
mgravell, line 1328: do we need to support `PUBSUB SHARDCHANNELS` (see `SubscriptionChannels[Async]`) ?

mgravell: ditto `SHARDNUMSUB` - see `SubscriptionPatternCount` / `SubscriptionSubscriberCount`

src/StackExchange.Redis/RedisSubscriber.cs
vandyvilla: currently UnsubscribeAll will unsubscribe both unsharded and sharded channels. Alternatively, we need to pass some flag to indicate whether we want to unsubscribe all sharded channels to match the SUNSUBSCRIBE semantics.

src/StackExchange.Redis/ResultProcessor.cs
vandyvilla: yes, plan to add it in a separate PR.

src/StackExchange.Redis/RedisSubscriber.cs
mgravell: I understand that. I just don't know that it is obvious whether that represents a problem. What genuinely likely scenario are we thinking of where this nuance is needed?

vandyvilla:
> 
good idea for leveraging the pattern mode to save one additional bool variable. 

Thanks for bringing up the shard movement concern, which is critical. Based on my testing, the subscriber gets a SUNSUBSCRIBE message reply from Redis when the shard is moved, which triggers a re-subscription against the correct server. Does this make sense to you?

dmitrig89:
@mgravell No pressure, but any idea when we might expect this feature to be merged? 
We're currently experiencing performance bottlenecks when using the pub\sub pattern on a Redis cluster.
I'm hopeful that sharding the pub\sub will help with the issues we're having

Thanks

mgravell:
Sorry, took my eyes off this one; let's get this done! if you're not still eager, I can probably take it over; but!

1: I really don't like these constructors:

``` diff
+ StackExchange.Redis.RedisChannel.RedisChannel(string! value, bool isSharded) -> void
+ StackExchange.Redis.RedisChannel.RedisChannel(byte[]? value, bool isSharded) -> void
```

can we instead use (to mirror `.Literal` and `.Pattern`:

``` c#
public static RedisChannel Sharded(string value) => ...
public static RedisChannel Sharded(byte[] value) => ....
```

? To me that feels more consistent.

2: I still think we should squash the 2 `bool`, but: let's not worry about those *today*

mgravell:
I've taken the liberty of pushing the merge fixes; I've also proposed the single-field refactor here: https://github.com/vandyvilla/StackExchange.Redis/pull/1 - would love to get your input / thoughts here @vandyvilla , although I guess I can push it through either way; sorry this has been delayed

mgravell:
working on adding some tests; very much needed (this could be from my changes, note):

image

mgravell:
found and fixed; that was a genuine bug in RedisDatabase

image

mgravell:
right, I think that's everything - @vandyvilla  if you want to merge https://github.com/vandyvilla/StackExchange.Redis/pull/1, then I should be able to pull in the combined effort (preserving your credit, etc)

mgravell:
important TODOs before final merge:

- [ ] investigate break/reconnect logic
- [ ] investigate cluster shard migration logic

src/StackExchange.Redis/ResultProcessor.cs
mgravell: ssub handled in my extensions to this PR

caoyang1024:
hi @NickCraver Could you please review this merge request at your earliest convenience?

nevermore-LQ:
@mgravell is this mr still active?

locus84:
bump.. I need this too..

uglide:
@mgravell @NickCraver Are there any blockers to get this merged besides the conflicts?

mgravell:
@uglide  resurrecting over at https://github.com/StackExchange/StackExchange.Redis/pull/2887 so I can clean up properly and re-evaluate what needs doing

@vandyvilla PR has been merged with credit - not to `main` yet, but: working on it

PR #2502, opened 2023-07-07
add LibraryName at ConfigurationOptions level

mgravell:
context: desire to set library name more granularly than provider - for example output cache provider, distributed cache provider, etc

the plan is also to, where possible, respect the pre-existing lib name - i.e. if the provider says "foostore", we'll use the new `GetProvider` API, ask for that info, and specifiy "foostore ocache", "foostore dcache", etc

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -11,7 +11,7 @@ Current package versions:
 - Change: Target net6.0 instead of net5.0, since net5.0 is end of life. ([#2497 by eerhardt](https://github.com/StackExchange/StackExchange.Redis/pull/2497))
 - Fix: Fix nullability annotation of IConnectionMultiplexer.RegisterProfiler ([#2494 by eerhardt](https://github.com/StackExchange/StackExchange.Redis/pull/2494))
 - Add: `Timer.ActiveCount` under `POOL` in timeout messages on .NET 6+ to help diagnose timer overload affecting timeout evaluations ([#2500 by NickCraver](https://github.com/StackExchange/StackExchange.Redis/pull/2500))
-- Add: `LibraryName` / `libname` configuration option; allows the library name to be controlled at the individual options level (in addition to the existing controls in `DefaultOptionsProvider`) ([#2502 by mgravell](https://github.com/StackExchange/StackExchange.Redis/pull/2502))
+- Add: `LibraryName` configuration option; allows the library name to be controlled at the individual options level (in addition to the existing controls in `DefaultOptionsProvider`) ([#2502 by mgravell](https://github.com/StackExchange/StackExchange.Redis/pull/2502))
 - Add: `DefaultOptionsProvider.GetProvider` allows lookup of provider by endpoint ([#2502 by mgravell](https://github.com/StackExchange/StackExchange.Redis/pull/2502))
 
 ## 2.6.116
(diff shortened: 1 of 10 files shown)


NickCraver:
Question here: how important is the connection string version? IMO we can always add it later, and I'd like to avoid the commas problem, which I can easily see someone hitting here - thoughts?

mgravell:
@NickCraver I'd be fine with dropping it and making it a code-only feature

NickCraver:
If cool with that, IMO better play and maybe it never comes up...don't want to have to deal with parsing if we can just play coy :)

src/StackExchange.Redis/Configuration/DefaultOptionsProvider.cs
```diff
@@ -49,10 +49,12 @@ public static void AddProvider(DefaultOptionsProvider provider)
         /// </summary>
         public virtual bool IsMatch(EndPoint endpoint) => false;
 
+        private static readonly DefaultOptionsProvider s_DefaultProvider = new DefaultOptionsProvider();
```
NickCraver, line 52: I'm a little uncomfortable here because there's a chance this becomes mutable at some point down the road (and we're not returning a new one each time). Thoughts?

mgravell: @NickCraver the `AddProvider` API already kinda assumes that it isn't, since that is instance based; so if it comes up as a mutability issue: we already have a bigger problem

mgravell: @NickCraver  for example, *today* you could just do `DefaultOptionsProvider.AddProvider(new DefaultOptionsProvider());` - and that needs to work

mgravell:
pushed changes to revert `libname` config string support, and move to "shipped"

src/StackExchange.Redis/Configuration/DefaultOptionsProvider.cs
NickCraver: I agree that'd happen _if you did that_ due to matching, but today every caller would be getting their own instance, no? I'm not sure how shifting this changes the example, but does return the same instance for all.

I'm saying this because `clientName` is stateful, and has the potential to change on subsequent caller/connections as the environment does (e.g. restarting your Azure app after an environment change isn't uncommon).

mgravell: Then any subclass already has this problem?

NickCraver: hmm, yeah it could indeed, good point - this just makes it worse for the default Azure case.

mgravell: I'm happy to revert the default instance for now and take that point separately; this isn't a hill for today

NickCraver: If we can please - I know this will break some scenarios (silently reporting old IDs, few would notice)

mgravell: @NickCraver done

docs/ReleaseNotes.md
```diff
@@ -11,6 +11,8 @@ Current package versions:
 - Change: Target net6.0 instead of net5.0, since net5.0 is end of life. ([#2497 by eerhardt](https://github.com/StackExchange/StackExchange.Redis/pull/2497))
 - Fix: Fix nullability annotation of IConnectionMultiplexer.RegisterProfiler ([#2494 by eerhardt](https://github.com/StackExchange/StackExchange.Redis/pull/2494))
 - Add: `Timer.ActiveCount` under `POOL` in timeout messages on .NET 6+ to help diagnose timer overload affecting timeout evaluations ([#2500 by NickCraver](https://github.com/StackExchange/StackExchange.Redis/pull/2500))
+- Add: `LibraryName` / `libname` configuration option; allows the library name to be controlled at the individual options level (in addition to the existing controls in `DefaultOptionsProvider`) ([#2502 by mgravell](https://github.com/StackExchange/StackExchange.Redis/pull/2502))
```
NickCraver, line 14:
```suggestion
- Add: `LibraryName` configuration option; allows the library name to be controlled at the individual options level (in addition to the existing controls in `DefaultOptionsProvider`) ([#2502 by mgravell](https://github.com/StackExchange/StackExchange.Redis/pull/2502))
```

NickCraver (approved):
One release notes tweak to match delta, looking good 👍

PR #2522, opened 2023-08-09
[Tests] Remove Moq as a dependency

NickCraver:
This is a test-only change. I don't want to risk an upgrade and harvesting PII from anyone who works on our project, so I'm removing Moq immediately. See https://github.com/moq/moq/issues/1372 for details/discussion.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -15,10 +15,9 @@
     <PackageVersion Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.2" />
     <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.1.0" />
     <PackageVersion Include="Microsoft.SourceLink.GitHub" Version="1.1.1" />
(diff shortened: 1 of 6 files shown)


philon-msft (approved):

mgravell (approved):

kzu:
Heya @NickCraver. I'd like to gather thoughts from folks on how maintenance/sponsorships of oss projects can me encouraged more broadly without resorting to things like this. Would love your thoughts on it at https://github.com/moq/moq/issues/1374 if you might?

Thanks!

(BTW, I'm not discarding the possibility of reverting the change for now until folks get more comfortable or improvements are implemented, so if you want to hold up, just skip updating Moq for now while we figure this out?)

thetredev:
@kzu you reap what you sow I guess.

mgravell:
I responded: https://github.com/moq/moq/issues/1374#issuecomment-1671166436

Issue #2520, opened 2023-08-06
[Improvement] Race condition for being selectable when some nodes in the cluster are down


Matiszak:
Context:
We have a cluster consisting of 3 masters + 3 replicas (1 replica for each master). We are simulating situation when some of our cluster is down (in this example we shut down 3 replicas). Our .NET clients have all of the 6 nodes put into configuration. Also option "resolveDns=true" is set. (resolveDns is not crucial here, it's just because nodes know themselves through their ips so that I set it to have not duplicated connections by hostname and by ip after discovery).

When some of the nodes in the cluster are down they are not responsive. So they exhaust default connection timeout of 5000 milliseconds. They exhaust even much higher connection timeouts (tens of seconds). It was noticed in previous versions of StackExchange.Redis (<2.2.xx) that this behavior caused problems because after .Connect() randomly some of the nodes were unselectable by reason DidNotRespond. Upgrading to 2.6.xx mostly solved this issue (for which im glad) due to queueing feature. But I noticed that the underlying issue is still there and in some scenarios might cause problems so that's the reason I'm posting this. I found one use case in which it manifests but you might be aware of more.

Issue:
1. The first iteration of connecting to nodes takes 5 seconds and connects to 3 healthy nodes but fails to connect to remaining 3 wrong ones.
2. Second iteration issues PINGs to healthy nodes because of `sendTracerIfConnected: true` in  `server.OnConnectedAsync(log, sendTracerIfConnected: true, autoConfigureIfConnected: reconfigureAll);`
3. Since first iteration exhaused timeout the second iteration exits immediately without waiting
4. For a brief moment after .Connect() some of the healthy nodes are undelectable even though connection succeeded in the first iteration
5. It causes issues in some scenarios (e.g SET with BacklogPolicy = FailFast

Reproduction code:
```
var options = ConfigurationOptions.Parse("server1:6116,server2:6116,server3:6116,server4:6116,server5:6116,server6:6116,password=some_password,resolveDns=true");
options.BacklogPolicy = BacklogPolicy.FailFast;
var connection = ConnectionMultiplexer.Connect(options, new LoggerWriter());
var status = connection.GetStatus();
connection.GetDatabase(0).StringSet("test2", "test2");
```

Exception:
```
[...]

PR #2525, opened 2023-08-16
Do not send tracers when running second iteration of cluster nodes di…

Matiszak:
…scovery (#2520)

Sending tracers is not necessary because we just connected to the nodes a few seconds ago. It causes problems because sent tracers do not have enough time to respond.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -10,6 +10,7 @@ Current package versions:
 
 - Change: Target net6.0 instead of net5.0, since net5.0 is end of life. ([#2497 by eerhardt](https://github.com/StackExchange/StackExchange.Redis/pull/2497))
 - Fix: Fix nullability annotation of IConnectionMultiplexer.RegisterProfiler ([#2494 by eerhardt](https://github.com/StackExchange/StackExchange.Redis/pull/2494))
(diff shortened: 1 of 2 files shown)


Matiszak:
I'm currently testing the change below on your CI but it seems I somwhow broke 1/2 tests, I'm gonna check why:
```
// This awaits either the endpoint's initial connection, or a tracer if we're already connected
 // (which is the reconfigure case**, except second iteration which is only for newly discovered cluster members**).
var isFirstIteration = iter == 0;
available[i] = server.OnConnectedAsync(log, sendTracerIfConnected: isFirstIteration, autoConfigureIfConnected: reconfigureAll);
```

Matiszak:
It's good that all checks have passed but I noticed this changed version as well as unchanged version (based on tag 2.6.122) is quite flaky and sometimes failes, sometimes not.
Examples:
https://github.com/StackExchange/StackExchange.Redis/actions/runs/5881471149/job/15950011765?pr=2525
https://github.com/StackExchange/StackExchange.Redis/actions/runs/5881109242/job/15948901359?pr=2525
https://github.com/StackExchange/StackExchange.Redis/actions/runs/5879015803/job/15942221751?pr=2525
https://github.com/StackExchange/StackExchange.Redis/actions/runs/5881109242?pr=2525

NickCraver (approved):
Wanted to land RESP3 ahead to prevent conflicts here, but agree with this change. Thanks for the investigation and the improvement here!

PR #2527, opened 2023-08-23
first stab at raw API for redis notifications

mgravell:
proposed API:

``` c#
muxer.EnableServerAssistedClientSideTracking(keyCallback, ... options ...);
// ...
```

to get notification when a tracked key is invalidated, and `CommandFlags.ClientCaching` on any operation to enable tracking per-call.


mgravell:
This is obviated by v4 client-side-caching work

PR #2547, opened 2023-09-10
Options: support changing lib-ver

### LINKED ISSUE #2533: LibraryVersion in ConfigurationOptions
I saw that now we can set the `LibraryName` in the ConfigurationOptions, what about `LibraryVersion`?
can we add `LibraryVersion` as well like [here](https://github.com/StackExchange/StackExchange.Redis/blob/2e9e8e969552b0a0ba7c1baff615d363847e2769/src/StackExchange.Redis/ConfigurationOptions.cs#L254)?

NickCraver:
This adds similar support from LibraryName to LibraryVersion so that consumers can override both if they so choose - this works in all the same places, DefaultOptionsProvider, etc.

Resolves #2533.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -10,17 +10,18 @@ Current package versions:
 
 - Adds: RESP3 support ([#2396 by mgravell](https://github.com/StackExchange/StackExchange.Redis/pull/2396)) - see https://stackexchange.github.io/StackExchange.Redis/Resp3
 - Fix [#2507](https://github.com/StackExchange/StackExchange.Redis/issues/2507): Pub/sub with multi-item payloads should be usable ([#2508 by mgravell](https://github.com/StackExchange/StackExchange.Redis/pull/2508))
-- Add: connection-id tracking (internal only, no public API) ([#2508 by mgravell](https://github.com/StackExchange/StackExchange.Redis/pull/2508))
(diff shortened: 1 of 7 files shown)


shacharPash:
@NickCraver Is there an estimate of when it is going to be merged?

mgravell:
@shacharPash  I believe the snag on this one is the static -> virtual, which is a big change to make in a minor; in voice call, we discussed options here, including breaking the change into two parts, so that there's no single hard break - for example, in version X we could do:

``` diff
- protected static string LibraryVersion => Utils.GetLibVersion();
+ [Obsolete("words, basically: stop using this")]
+ protected static string LibraryVersion => DefaultLibraryVersion;
+ protected static string DefaultLibraryVersion => Utils.GetLibVersion();
```

then in change Y

``` diff
- [Obsolete("words, basically: stop using this")]
- protected static string LibraryVersion => DefaultLibraryVersion; 
+ public virtual string LibraryVersion => DefaultLibraryVersion;
```

If we *do* go that route, it'll take at least 2 deploys to get it released. However, I'm open to other options. Just; the hard break here is not ideal, even if it is a very niche API that most folks won't be using.

NickCraver:
@shacharPash After talking about this and looking through it - I believe it will be a 3.x addition due to the need for a break here. We have many calls to `Utils` directly to get this (that can't necessarily use options), so having an intermediate non-breaking solution which we'd still not want to break for 3.x either is pretty messy.

shacharPash:
@NickCraver Thanks for updating 👍

PR #2553, opened 2023-09-14
Fix message balancing between replicas with Redis Sentinel

Grumov:
The current distribution of messages between replicas is uneven. 
The replica following the master in the list of servers is selected more often than others. 
In the configuration there is a master and two replicas: 
- the first replica receives 66% of messages 
- the second replica receives 34%.


NickCraver:
I see we do have an issue #2554, awesome - let's discuss the issue there and figure out how to proceed.

PR #2575, opened 2023-10-17
Update Transactions.md

DavidKlempfner:
Using await for an async method.

### DIFF
--- a/docs/Transactions.md
+++ b/docs/Transactions.md
@@ -70,7 +70,7 @@ mixed together with other callers. So our example becomes:
 var newId = CreateNewId();
 var tran = db.CreateTransaction();
 tran.AddCondition(Condition.HashNotExists(custKey, "UniqueID"));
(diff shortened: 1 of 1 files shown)


mgravell:
No. That's not correct here, and will effectively jam your execution - I would say block your thread, but the async, threadless version of blocking a thread)

If you want to await it to get a result, you must do that **after** the execute, by storing the pending result in a local.

DavidKlempfner:
> No. That's not correct here, and will effectively jam your execution (I would say block your thread, but the async, threadless version of blocking a thread)
> 
> If you want to await it to get a result, you must do that **after** the execute, by storing the pending result in a local:
> 
> ```
> bar pending = tran.HashSetAsync(custKey, "UniqueID", newId);
> bool committed = tran.Execute();
> bar result = await pending;
> ```

I understand now, thanks!

PR #2578, opened 2023-10-19
Add (ReadOnly)Memory<byte> implicit conversions to RedisKey

kevin-montrose:
At a high level, this just changes RedisKey.KeyPrefix to a ReadOnlyMemory<byte> (from a byte[]).

The change is slightly more complicated to deal with Prepend/Append.

Also noticed [this line](https://github.com/kevin-montrose/StackExchange.Redis/blob/5504ed9a93f729f204e6cb53280cb4efa1ccd390/src/StackExchange.Redis/PhysicalConnection.cs#L1153C31-L1153C31) ; which looks like a place where prefix wouldn't be written?  "Fixed" it and all tests still seem to pass.

### DIFF
--- a/src/StackExchange.Redis/KeyspaceIsolation/KeyPrefixed.cs
+++ b/src/StackExchange.Redis/KeyspaceIsolation/KeyPrefixed.cs
@@ -3,6 +3,7 @@ using System.Collections.Generic;
 using System.Diagnostics.CodeAnalysis;
 using System.Linq;
 using System.Net;
+using System.Runtime.InteropServices;
 using System.Threading.Tasks;
 
 namespace StackExchange.Redis.KeyspaceIsolation
@@ -843,7 +844,8 @@ namespace StackExchange.Redis.KeyspaceIsolation
 
         protected RedisChannel ToInner(RedisChannel outer)
         {
-            var combined = RedisKey.ConcatenateBytes(Prefix, null, (byte[]?)outer);
+            var combined = RedisKey.ConcatenateBytesArray(Prefix, null, (byte[]?)outer);
+
             return new RedisChannel(combined, outer.IsPattern ? RedisChannel.PatternMode.Pattern : RedisChannel.PatternMode.Literal);
         }
 
--- a/src/StackExchange.Redis/PhysicalConnection.cs
+++ b/src/StackExchange.Redis/PhysicalConnection.cs
@@ -1148,19 +1148,30 @@ namespace StackExchange.Redis
             return value < 10 ? (byte)('0' + value) : (byte)('a' - 10 + value);
         }
 
-        internal static void WriteUnifiedPrefixedString(PipeWriter writer, byte[]? prefix, string? value)
+        internal static void WriteUnifiedPrefixedString(PipeWriter writer, ReadOnlyMemory<byte> prefix, string? value)
         {
-            if (value == null)
+            if (prefix.IsEmpty && value == null)
             {
                 // special case
                 writer.Write(NullBulkString.Span);
+            }
+            else if (value == null)
+            {
+                int totalLength = prefix.Length;
+
+                var span = writer.GetSpan(3 + Format.MaxInt32TextLen);
+                span[0] = (byte)'$';
+                int bytes = WriteRaw(span, totalLength, offset: 1);
+                writer.Advance(bytes);
+
+                writer.Write(prefix.Span);
             }
             else
             {
                 // ${total-len}\r\n         3 + MaxInt32TextLen
                 // {prefix}{value}\r\n
                 int encodedLength = Encoding.UTF8.GetByteCount(value),
-                    prefixLength = prefix?.Length ?? 0,
+                    prefixLength = prefix.Length,
                     totalLength = prefixLength + encodedLength;
 
                 if (totalLength == 0)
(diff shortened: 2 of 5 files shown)


mgravell:
> which looks like a place where prefix wouldn't be written?

well, historically I believe it was "value only" or "prefix+value" - it is your change that makes "prefix only" a thing; it bothers me a little that if we *are* using key-prefix (which is non-trivial usage - aspnet exposes it, for example), then this gets kicked into non-working; I had a separate design somewhere that refactored the *value* part of the key instead, using object+int+int fields, with type testing to support:

- `string`, `char[]`, `MemoryManager<char>` (all handled as `ReadOnlyMemory<char>`)
- `byte[]`, `MemoryManager<byte>` (all handled as `ReadOnlyMemory<byte>`)

I wonder if that's the way we should go here

mgravell:
here was the core part from last time I looked at this:

``` c#
using System.Buffers;
using System.Runtime.InteropServices;
 
Describe(default);
 
string s = "abcdefg";
Describe(s);
Describe(""); // should look like default
Describe(s.AsMemory());
Describe(s.AsMemory().Slice(2,3));
 
var bytes = new byte[42];
Describe(bytes);
Describe(Array.Empty<byte>()); // should look like default
Describe(bytes.AsMemory());
Describe(bytes.AsMemory().Slice(10, 3));
 
var chars = new char[42];
Describe(chars);
Describe(chars.AsMemory());
Describe(chars.AsMemory().Slice(10, 3));
 
// not shown: custom memory managers
 

static void Describe(Prototype value)
{
    if (value.TryGetValue(out ReadOnlyMemory<byte> bytes))
    {
        Console.WriteLine($"Value is {bytes.Length} bytes");
    }
    if (value.TryGetValue(out ReadOnlyMemory<char> text))
    {
        Console.WriteLine($"Value is {text.Length} chars");
    }
    Console.WriteLine("---");
    Console.WriteLine();
}
readonly struct Prototype
{
    private readonly object? _obj;
    private readonly int _start, _length;
 
    private Prototype(object? obj, int start, int length)
    {
        if (length is 0)
        {
            // don't track the obj, then
            this = default;
        }
        else
        {
            _obj = obj;
            _start = start;
            _length = length;
        }
    }
    public static implicit operator Prototype(string value) => new(value, 0, value is null ? 0 : value.Length);
 
    public static implicit operator Prototype(byte[] value) => new(value, 0, value is null ? 0 : value.Length);
    public static implicit operator Prototype(char[] value) => new(value, 0, value is null ? 0 : value.Length);
 
    public static implicit operator Prototype(Memory<char> value) => (Prototype)(ReadOnlyMemory<char>)value;
    public static implicit operator Prototype(ReadOnlyMemory<char> value)
    {
        // string and array are pretty common
        if (MemoryMarshal.TryGetString(value, out var text, out int start, out int length))
        {
            return new Prototype(text, start, length);
        }
        if (MemoryMarshal.TryGetArray(value, out var segment))
        {
            return new Prototype(segment.Array, segment.Offset, segment.Count);
        }
        // this is pretty rare
        if (MemoryMarshal.TryGetMemoryManager(value, out MemoryManager<char>? mgr, out start, out length))
        {
            return new Prototype(mgr, start, length);
        }
        // we don't expect this
        return Fail();
    }
 
    public static implicit operator Prototype(Memory<byte> value) => (Prototype)(ReadOnlyMemory<byte>)value;
    public static implicit operator Prototype(ReadOnlyMemory<byte> value)
    {
        // array is pretty common
        if (MemoryMarshal.TryGetArray(value, out var segment))
        {
            return new Prototype(segment.Array, segment.Offset, segment.Count);
        }
        // this is pretty rare
        if (MemoryMarshal.TryGetMemoryManager(value, out MemoryManager<byte>? mgr, out var start, out var length))
        {
            return new Prototype(mgr, start, length);
        }
        // we don't expect this
        return Fail();
    }
    private static Prototype Fail() => throw new InvalidOperationException("Unexpected ReadOnlyMemory<> scenario; please report this.");
 
    public bool TryGetValue(out ReadOnlyMemory<byte> value)
    {
        if (_length == 0)
        {
            value = default;
            return true;
        }
        switch (_obj)
        {
            case byte[] arr:
                value = new ReadOnlyMemory<byte>(arr, _start, _length);
                return true;
            case MemoryManager<byte> mgr:
                value = mgr.Memory.Slice(_start, _length);
                return true;
            default:
                value = default;
                return false;
        }
    }
    public bool TryGetValue(out ReadOnlyMemory<char> value)
    {
        if (_length == 0)
        {
            value = default;
            return true;
        }
        switch (_obj)
        {
            case string text:
                value = text.AsMemory().Slice(_start, _length);
                return true;
            case char[] arr:
                value = new ReadOnlyMemory<char>(arr, _start, _length);
                return true;
            case MemoryManager<char> mgr:
                value = mgr.Memory.Slice(_start, _length);
                return true;
            default:
                value = default;
                return false;
        }
    }
}
```

mgravell:
btw, I'm not saying "you should go do this" - this is the what I'm more inclined towards, but that doesn't make it your problem

kevin-montrose:
> > which looks like a place where prefix wouldn't be written?
> 
> well, historically I believe it was "value only" or "prefix+value" - it is your change that makes "prefix only" a thing

Ah, I think that's true - I was just looking at the signatures.

mgravell:
@kevin-montrose there is a "get lease" API for redis strings, which might be what you mean? Or do you mean the keys?

kevin-montrose:
> @kevin-montrose there is a "get lease" API for redis strings, which might be what you mean? Or do you mean the keys?

@mgravell YUP, that'll do - couldn't find it (was looking for "pool" or something), and seems basically undocumented?

Ideally we'd even be able to reuse the Lease<byte>... but that's maybe overkill.

PR #2610, opened 2023-12-05
Connections: detect and handle the Linux dead socket case

NickCraver:
In Linux, we see 15 minute socket stalls due to OS-level TCP retries. What this means is the PhysicalConnection detects no issues on the pipe that's retrying, but is also not receiving data at all leading to long stalls in client applications. The goal here is to detect that we're timing out commands people have issued to the connection but we're getting _NOTHING_ back on the socket at all. In this case, we should assume the socket is dead and issue a reconnect so that we get out of the hung situation much faster.

For an initial go at this, we've chosen 4x the timeout interval as a threshold, but could make this configurable if needed.

### DIFF
--- a/src/StackExchange.Redis/PhysicalBridge.cs
+++ b/src/StackExchange.Redis/PhysicalBridge.cs
@@ -624,13 +624,13 @@ namespace StackExchange.Redis
                                 KeepAlive();
                             }
                             else if (timedOutThisHeartbeat > 0
-                                && tmp.LastReadSecondsAgo > (tmp.BridgeCouldBeNull?.Multiplexer.AsyncTimeoutMilliseconds * 4000))
+                                && tmp.LastReadSecondsAgo * 1_000 > (tmp.BridgeCouldBeNull?.Multiplexer.AsyncTimeoutMilliseconds * 4))
                             {
                                 // If we've received *NOTHING* on the pipe in 4 timeouts worth of time and we're timing out commands, issue a connection failure so that we reconnect
                                 // This is meant to address the scenario we see often in Linux configs where TCP retries will happen for 15 minutes.
                                 // To us as a client, we'll see the socket as green/open/fine when writing but we'll bet getting nothing back.
                                 // Since we can't depend on the pipe to fail in that case, we want to error here based on the criteria above so we reconnect broken clients much faster.
-                                tmp.BridgeCouldBeNull?.Multiplexer.Logger?.LogWarning("Dead socket detected, no reads in " + tmp.LastReadSecondsAgo + " seconds, issuing disconnect");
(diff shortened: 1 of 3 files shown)


src/StackExchange.Redis/PhysicalBridge.cs
```diff
@@ -623,6 +623,17 @@ internal void OnHeartbeat(bool ifConnectedOnly)
                                 // queue, test the socket
                                 KeepAlive();
                             }
+                            else if (timedOutThisHeartbeat > 0
+                                && tmp.LastReadSecondsAgo > (tmp.BridgeCouldBeNull?.Multiplexer.AsyncTimeoutMilliseconds * 4000))
```
philon-msft, line 627:
```suggestion
                                && (tmp.LastReadSecondsAgo * 1000) > (tmp.BridgeCouldBeNull?.Multiplexer.AsyncTimeoutMilliseconds * 4)
```

src/StackExchange.Redis/PhysicalConnection.cs
```diff
@@ -248,6 +248,7 @@ private enum ReadMode : byte
         private readonly WeakReference _bridge;
         public PhysicalBridge? BridgeCouldBeNull => (PhysicalBridge?)_bridge.Target;
 
+        public long LastReadSecondsAgo => unchecked(Environment.TickCount - Thread.VolatileRead(ref lastReadTickCount)) / 1000;
```
philon-msft, line 251: nit: changing this from seconds to LastReadMillisecondsAgo would eliminate some math

src/StackExchange.Redis/PhysicalConnection.cs
```diff
@@ -720,8 +721,13 @@ internal void GetStormLog(StringBuilder sb)
             }
         }
 
-        internal void OnBridgeHeartbeat()
+        /// <summary>
+        /// Runs on every heartbeat for a bridge, timing out any commands that are overdue and returning an integer of how many we timed out.
+        /// </summary>
+        /// <returns>How many commands were overdue and threw timeout exceptions.</returns>
+        internal int OnBridgeHeartbeat()
```
philon-msft, line 728: We only really _need_ this to return a `bool` indicating if _any_ commands were timed out. Or will the `int` be useful in some scenarios?

philon-msft (changes requested):

src/StackExchange.Redis/PhysicalConnection.cs
NickCraver: I meant to include this in the exception...woops. Fixed!

src/StackExchange.Redis/PhysicalConnection.cs
NickCraver: Agreed, but since we have a sibling property for the Write side here, wanted this to match. Cheers on the math sanity checks and catching epic fail <3

philon-msft (approved):

bkaracivi:
Hello,
Have you ever faced this issue for 2.6.122 client versions running on Windows Server environment?
Is this change, corrects similar cases for clients on Windows Server?
Thanks in advance

PR #2621, opened 2023-12-20
add type-forward on IsExternalInit

### LINKED ISSUE #2619: System.MissingMethodException method not found: 'Void StackExchange.Redis.ConfigurationOptions.set_EndPoints(StackExchange.Redis.EndPointCollection)'
### Description

I have netstandard2.0 project which is wrapper around StackExchange.Redis nuget. I have net8.0 console project which is using netstandard2.0 one.
I'm getting the following runtime error:
```
"Method not found: 'Void StackExchange.Redis.ConfigurationOptions.set_EndPoints(StackExchange.Redis.EndPointCollection)'."}
    ClassName: null
    Data: {System.Collections.ListDictionaryInternal}
    HResult: -2146233069
    HasBeenThrown: true
    HelpLink: null
    InnerException: null
    MemberName: null
    Message: "Method not found: 'Void StackExchange.Redis.ConfigurationOptions.set_EndPoints(StackExchange.Redis.EndPointCollection)'."
    SerializationStackTraceString: "   at CacheWrapper.CacheWrapper.BuildConfigurationOptions()\r\n   at CacheWrapper.CacheWrapper..ctor() in CacheWrapper.cs:line 11\r\n   at Program.<Main>$(String[] args) in Program.cs:line 3"
    SerializationWatsonBuckets: null
    Signature: null
    Source: "CacheWrapper"
```

As I understand that the problem is in StackExchange.Redis nuget package which uses **init only setters** in netstandard version of the library.

### Reproduction Steps

CacheWrapper.csproj
```
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <TargetFramework>netstandard2.0</TargetFramework>
    <LangVersion>latest</LangVersion>
  </PropertyGroup>

  <ItemGroup>
    <PackageReference Include="StackExchange.Redis" Version="2.7.10" />
  </ItemGroup>

</Project>
```
CacheWrapper.cs
```
using StackExchange.Redis;

namespace CacheWrapper
{
    public class CacheWrapper
    {
        private readonly ConfigurationOptions _options;

        public CacheWrapper()
        {
            _options = BuildConfigurationOptions();
        }

        public IDatabase Connect()
        {
            var connectionMultiplexer = ConnectionMultiplexer.Connect(_options);
            return connectionMultiplexer.GetDatabase();
        }

        internal ConfigurationOptions BuildConfigurationOptions()
        {
            var endPoints = new EndPointCollection { { "127.0.0.1", 11111 } };

            return
                new ConfigurationOptions
                {
                    AbortOnConnectFail = false,
                    EndPoints = endPoints,
                    SyncTimeout = 30000
                };
        }
    }
}
```
CacheUser.csproj
```
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>net8.0</TargetFramework>
    <ImplicitUsings>enable</ImplicitUsings>
    <Nullable>enable</Nullable>
  </PropertyGroup>

  <ItemGroup>
    <ProjectReference Include="..\CacheWrapper\CacheWrapper.csproj" />
  </ItemGroup>

</Project>
```
Program.cs
```
var cache = new CacheWrapper.CacheWrapper();
```
This code works if my console application is targeting net48.

mgravell:
fix #2619

as per guidance from the runtime folks, we should declare a type-forward for `IsExternalInit`, so that it gets correctly bound (in particular when netstandard is involved, as the build and execution environments are more likely to differ, leading to modreq confusion)

### DIFF
--- a/src/StackExchange.Redis/Hacks.cs
+++ b/src/StackExchange.Redis/Hacks.cs
@@ -1,5 +1,7 @@
-﻿#if !NET5_0_OR_GREATER
-
+﻿#if NET5_0_OR_GREATER
+// context: https://github.com/StackExchange/StackExchange.Redis/issues/2619
+[assembly: System.Runtime.CompilerServices.TypeForwardedTo(typeof(System.Runtime.CompilerServices.IsExternalInit))]
+#else
 // To support { get; init; } properties
 using System.ComponentModel;
 
@@ -8,5 +10,4 @@ namespace System.Runtime.CompilerServices
     [EditorBrowsable(EditorBrowsableState.Never)]
     internal static class IsExternalInit { }
 }
-
 #endif
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -11,7 +11,7 @@
     <!-- Packages only used in the solution, upgrade at will -->
     <PackageVersion Include="BenchmarkDotNet" Version="0.13.1" />
     <PackageVersion Include="GitHubActionsTestLogger" Version="2.0.0-alpha" />
-    <PackageVersion Include="Microsoft.CodeAnalysis.PublicApiAnalyzers" Version="3.3.4-beta1.22362.3" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.PublicApiAnalyzers" Version="3.3.4" />
     <PackageVersion Include="Microsoft.Extensions.Logging.Abstractions" Version="6.0.0" />
     <PackageVersion Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.2" />
     <PackageVersion Include="Microsoft.NET.Test.Sdk" Version="17.1.0" />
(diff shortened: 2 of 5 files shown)


src/StackExchange.Redis/Hacks.cs
```diff
@@ -1,5 +1,7 @@
-﻿#if !NET5_0_OR_GREATER
-
+﻿#if NET5_0_OR_GREATER
```
WeihanLi, line 1: Should it be `NET6_0_OR_GREATER` since we currently support `net6.0` other than `net5.0`

mgravell: The "currently" is the key word there. Ok, in reality we're not going to add a 5.0 at this point, but by sticking with the relevant framework version *for the feature*, if we change the target frameworks: any such `#if` will sort themselves out automatically, without anything needing to be revisited.

In short: I'm happier with the `NET5_OR_GREATER` here

NickCraver: Agreed - updating all the definitions each time is precisely what these defines we're meant to avoid. It tells us which version this code started to be needed in and that's not a terrible thing in itself. I'd vote not to chase these overall - it's just a lot of yearly diff that never ends :)

NickCraver (approved):

PR #2622, opened 2023-12-21
Use GlobalPackageReference for Microsoft.NETFramework.ReferenceAssemblies

WeihanLi:
use `GlobalPackageReference` since it's for all projects and `PrivateAssets="All"`

https://learn.microsoft.com/en-us/nuget/consume-packages/central-package-management#global-package-references

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -35,7 +35,4 @@
   <ItemGroup Condition="'$(Configuration)' == 'Release' and '$(SourceRoot)'==''">
     <SourceRoot Include="$(MSBuildThisFileDirectory)/"/>
   </ItemGroup>
-  <ItemGroup>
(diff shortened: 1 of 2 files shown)


mgravell:
Not sure I see the benefit of this - it doesn't reduce any duplication, and the existing code is perfectly clear. It seems to be a change for change's sake?

WeihanLi:
We can have version and reference defined in one place, without splitting into two places with this

mgravell:
> We can have version and reference defined in one place, without splitting into two places

Which is how literally every other reference works. I'm not sure I see the benefit of changing it, honestly, other than "just because"

WeihanLi:
I'm fine with it either way, thanks for the opinion

PR #2646, opened 2024-02-11
Support Alibaba pseudo-cluster configurations

### LINKED ISSUE #2642: Feature Request: DB selection support for proxy-bridged cluster
**Description:**
Some of the Proxy-bridged cluster can support multiple db selection. 
In order to give full support of advanced features, the proxy-bridged cluster worked as a single node cluster that:

- support "Cluster" command, if user emit "cluster nodes" it will return its IP with full slot range [0, 16383]
- "cluster_enabled" is set to 1 in info, with corresponding "redis_mode" a cluster
- it supports mget/mset series and multi-key commands (non-atomic) 

in that mode, it can give the compatibility at best effort and user could write same code for standalone mode or cluster mode, especially when user scales standalone mode to cluster, with db selection data inside.

An unhandled timeout exception was encountered when doing db selection in cluster mode in some early version of the library, and we tried the latest package 2.7.17 but also failed.

```
Unhandled exception. StackExchange.Redis.RedisTimeoutException: Timeout performing HSET (10000ms), inst: 0, qu: 0, qs: 0, aw: False, rs: ReadAsync, ws: Idle, in: 0, in-pipe: 0, out-pipe: 0, serverEndpoint: 172.123.35.71:6379, mc: 1/1/0, mgr: 10 of 10 available, clientName: 47e4280ce570, PerfCounterHelperkeyHashSlot: 11933, IOCP: (Busy=0,Free=1000,Min=32,Max=1000), WORKER: (Busy=0,Free=32767,Min=32,Max=32767), v: 2.2.79.4591
```

the case can be reproduced by using code below,

```
        // singleton connection
        ConnectionMultiplexer cm = RedisConnSingleton.getRedisConn();
        var db = cm.GetDatabase();
        

        for (int i = 1; i <= 100; i++)
        {
            string key = i.ToString();
            string value = i.ToString();

            // after switch db selection, the GetDatabase returned a unexpect handle so failed to send
            db.GetDatabase(i); 

            db.StringSet(key, value);
            String ret = db.StringGet(key);
            Console.WriteLine("get key: " + ret);

            Thread.Sleep(1000);
        }    
```

- Hope to support db-selection in cluster mode
- Is there a way of getting correct cluster link to send raw "select" command? Please advise.

Thanks,
Baron

mgravell:
1: don't treat trivial clusters as clusters - Alibaba uses this config
2: report synchronous failures immidiately and accurately

fix #2642

For context, Alibaba reports from `cluster nodes` as 2 nodes:

``` txt
aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa {proxy endpoint} myself,master - 0 0 0 connected 0-16383
bbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbbb {proxy endpoint} slave aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa 0 0 0 connected
```

which confuses things somewhat; this isn't a valid/normal cluster configuration, so: ignore it!

Other relevant metadata:

from `config`:

```
databases=256
```

from `info`:

```
# Cluster
cluster_enabled:1
databases:256
nodecount:2
```

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -7,7 +7,8 @@ Current package versions:
 | [![StackExchange.Redis](https://img.shields.io/nuget/v/StackExchange.Redis.svg)](https://www.nuget.org/packages/StackExchange.Redis/) | [![StackExchange.Redis](https://img.shields.io/nuget/vpre/StackExchange.Redis.svg)](https://www.nuget.org/packages/StackExchange.Redis/) | [![StackExchange.Redis MyGet](https://img.shields.io/myget/stackoverflow/vpre/StackExchange.Redis.svg)](https://www.myget.org/feed/stackoverflow/package/nuget/StackExchange.Redis) |
 
 ## Unreleased
-No unreleased changes
+
+- Fix [#2642](https://github.com/StackExchange/StackExchange.Redis/issues/2642): Detect and support multi-DB pseudo-cluster/proxy scenarios ([#2646](https://github.com/StackExchange/StackExchange.Redis/pull/2646) by mgravell)
 
 ## 2.7.17
 
--- a/src/StackExchange.Redis/ConnectionMultiplexer.cs
+++ b/src/StackExchange.Redis/ConnectionMultiplexer.cs
@@ -1,4 +1,7 @@
-﻿using System;
+﻿using Microsoft.Extensions.Logging;
+using Pipelines.Sockets.Unofficial;
+using StackExchange.Redis.Profiling;
+using System;
 using System.Collections;
 using System.Collections.Generic;
 using System.ComponentModel;
@@ -11,9 +14,6 @@ using System.Runtime.InteropServices;
 using System.Text;
 using System.Threading;
 using System.Threading.Tasks;
-using Microsoft.Extensions.Logging;
-using Pipelines.Sockets.Unofficial;
-using StackExchange.Redis.Profiling;
 
 namespace StackExchange.Redis
 {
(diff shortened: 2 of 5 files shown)


NickCraver:
How does this affect single node clusters? For example, even in Azure hosting people start with a single cluster node in a lot of cases which makes scaling to `n` easier later...but they don't support `SELECT` because they are cluster. I worry about the assumption of 2+ here being needed.

mgravell:
Untested and it is a good question, but: if the worst side effect here is that we get a server-side "nope" rather than a client-side "nope" in genuine single-server clusters: I'm comfortable with that...?

NickCraver:
I used SELECT as an example, but I think this had many more side-effects such as not detecting when the cluster changes and such. If for example we were viewing this in Opserver…it'd be wrong. Any code basing off cluster or not on top of us has the same downstream fun. Lots of potential there.

I'd vote we should not make such an assumption based on node count - likely better for the server-side to not represent itself as a cluster here.

mgravell:
@NickCraver  thanks for keeping me honest; refactored to use the `"databases"` entry as reported from `config`; this data is also available under `info cluster`, but I don't propose to add an extra `info` call unless this doesn't work reliably

NickCraver (approved):
ADD SOME RELEASE NOTES YOU HEATHEN

yangbodong22011:
@mgravell Thank you for your quick fix. We have verified the 2.7.20 version of MyGet and it works very well. When do you expect to release it to NuGet? After that, we will push users to upgrade.

mgravell:
Thanks for confirming. I'll discuss release in our weekly sync tomorrow (20th) - assuming there's no known blockers, I expect we'll release to NuGet either end-of-day 20th, or very shortly after

PR #2659, opened 2024-02-28
add new AddLibraryNameSuffix API for annotating connections with usage

mgravell:
This usage is intended to help people identify their connections by usage; for example, by using:

``` c#
conn.AddLibraryNameSuffix("cache");
```

the connection will be advertised as (by default) `SE.Redis-cache`. In particular, this targets scenarios where the `ConfigurationOptions` may be shared between uses, or otherwise unavailable, so it is not possible to simply add this to `ConfigurationOptions`


NickCraver (approved):
Chatted on call - +1 to approach, recommend we put an example in the method doc

Issue #2650, opened 2024-02-22
Intermittent connection errors after MemoryDB cluster version upgrade


schickosky:
I have an application using AWS MemoryDB for storage which recently underwent a version upgrade from 6.2 to 7.1, and ever since have been seeing fairly frequent connection drops with the following message:

```
StackExchange.Redis.RedisConnectionException: InternalFailure on insights-silo-memorydb-cluster-0001-002.insights-silo-memorydb-cluster.g748ly.memorydb.us-west-2.amazonaws.com:6379/Interactive, Idle/Faulted, last: SSCAN, origin: ReadFromPipe, outstanding: 17, last-read: 0s ago, last-write: 0s ago, unanswered-write: 0s ago, keep-alive: 60s, state: ConnectedEstablished, mgr: 9 of 10 available, in: 0, last-heartbeat: 0s ago, last-mbeat: 0s ago, global: 0s ago, v: 2.7.17.27058
 ---> System.InvalidOperationException: Unexpected response prefix: {
    at StackExchange.Redis.PhysicalConnection.TryParseResult(ResultFlags flags, Arena`1 arena, ReadOnlySequence`1& buffer, BufferReader& reader, Boolean includeDetilInExceptions, ServerEndPoint server, Boolean allowInlineProtocol) in /_/src/StackExchange.Redis/PhysicalConnection.cs:line 2118
   at StackExchange.Redis.PhysicalConnection.ProcessBuffer(ReadOnlySequence`1& buffer) in /_/src/StackExchange.Redis/PhysicalConnection.cs:line 1859
   at StackExchange.Redis.PhysicalConnection.ReadFromPipe() in /_/src/StackExchange.Redis/PhysicalConnection.cs:line 1803
   --- End of inner exception stack trace ---
```

The error only seems to occur during set scans, and appears to cause the connection to drop briefly for the client. We do also have sorted set scans being read from/written to even more frequently, but I do not see the error from that command. The data being retrieved is large JSON blobs stored as strings, and our workload is very write-heavy (we are continually popping elements off the sets to work in memory to then re-persist, even when there is nothing to be done). Suspected the data may have become corrupted due to being written/deleted during the upgrade (which was eventful and took 2 days), but I can successfully scan the entire set from `redis-cli`. I have not been able to reproduce the error outside of our production environments, however.

[...]

PR #2660, opened 2024-03-01
new tunnel-based RESP logging and validation

mgravell:
context: https://github.com/StackExchange/StackExchange.Redis/issues/2650

adds:

- opt-in full RESP logging to file (supports TLS, logged data is not encrypted)
- RESP validation of captured files
- RESP local replay of captured files (for display purposes; this is not replay *against a server*)

Usage md: https://github.com/StackExchange/StackExchange.Redis/blob/4b450ce557d52a05af42fbecf3fcc169f0e048ce/docs/RespLogging.md

note that this API is considered (and is annotated) experimental, and as such does not yet feature in the API file list; this is intentional


docs/RespLogging.md
```diff
@@ -0,0 +1,150 @@
...
+```
+
+If the RESP streams are *not* valid, an exception will provide further details.
+
+**An exception here is strong evidence that there is a fault either in the redis server, or an intermediate proxy**.
+
+Conversely, if the library reported a protocol failure but the validation step here *does not* report an error, then
+that is strong evidence of a library error; [**please report this**](https://github.com/StackExchange/StackExchange.Redis/issues/new) (with details).
+
+You can also *replay* the conversation locally, seeing the individual requests and responses:
```
philon-msft, line 111: Could we "replay" files into the library itself for some tests (as opposed to running an actual Redis server locally)?

src/StackExchange.Redis/Configuration/LoggingTunnel.cs
```diff
@@ -0,0 +1,624 @@
...
+    public static async Task<long> ReplayAsync(Stream @out, Stream @in, Action<RedisResult, RedisResult> pair)
+    {
+        using Arena<RawResult> arena = new();
+        var outPipe = StreamConnection.GetReader(@out);
+        var inPipe = StreamConnection.GetReader(@in);
+
+        long count = 0;
+        while (true)
+        {
+            var sent = await ReadOneAsync(outPipe, arena, isInbound: false);
```
philon-msft, line 42: Append `.ForAwait();` on async calls?

docs/RespLogging.md
mgravell: there's a few problems there; we'd need to isolate away anything non-deterministic, which means:

1. anything involving multiple threads, including anything that is a side-effect of heartbeat and (due to timing) anything pub/sub on RESP2
2. anything random, such as the unique muxer identifier (we'd have to make it deterministic in for this scenario, or make it forgive the value received)
3. anything involving multiple connections (cluster)

If we had a concrete scenario where all of the above would be met, and replay would be useful: fine, but I don't think we should do anything preemptively.

src/StackExchange.Redis/Configuration/LoggingTunnel.cs
mgravell: good spot; fixed in 64f31748b70998614a0ea06f5c0d07743540254c, although oddly that isn't showing on GitHub yet (despite being pushed)

src/StackExchange.Redis/Configuration/LoggingTunnel.cs
```diff
@@ -0,0 +1,624 @@
...
+                break; // no more data, or trailing incomplete messages
+            }
+        }
+        return default;
+    }
+
+    /// <summary>
+    /// Validate a RESP stream and return the number of top-level RESP fragments.
+    /// </summary>
+    public static async Task<long> ValidateAsync(string path)
```
philon-msft, line 117: Add <param> doc comment about how you can pass a path to a file OR directory?

mgravell: fair; 5ac13be3a5ad1f06ed725ebd5bbd8fcf18324476

philon-msft (approved):

PR #2665, opened 2024-03-07
CheckTrustedIssuer: Fixes for invalid chains

NickCraver:
This issue was brought to my attention last night (thanks to Badrish Chandramouli): https://github.com/dotnet/dotnet-api-docs/pull/6660

This changeset ensures that we do not honor self-signed certs or partial/broken chains as a result of `X509VerificationFlags.AllowUnknownCertificateAuthority` downstream and adds a few tests and utilities to generate test certificates (currently valid for ~9000 days). Instead we are checking that the certificate we're being told to trust is explicitly in the chain, given that the result of `.Build()` cannot be trusted for this case.

This also resolves an issue where `TrustIssuer` could be called but we'd error when _no errors_ were detected (due to requiring chain errors in our validator), this means users couldn't temporarily trust a cert while getting it installed on the machine for instance and migrating between the 2 setups was difficult.

This needs careful eyes, please scrutinize heavily. It's possible this breaks an existing user, but...it should be broken if so unless there's a case I'm not seeing.


src/StackExchange.Redis/ConfigurationOptions.cs
```diff
@@ -296,15 +296,25 @@ private static RemoteCertificateValidationCallback TrustIssuerCallback(X509Certi
...
-                    && CheckTrustedIssuer(v2, issuer);
+            return (object _, X509Certificate? certificate, X509Chain? __, SslPolicyErrors sslPolicyError) =>
+            {
+                // If we're already valid, there's nothing further to check
+                if (sslPolicyError == SslPolicyErrors.None)
+                {
+                    return true;
+                }
+                // If we're not valid due to chain errors - check against the trusted issuer
+                return sslPolicyError == SslPolicyErrors.RemoteCertificateChainErrors
```
mgravell, line 307: this is a `[Flags]` enum; that's fine *if* we mean "if the *only* error is the SSL chain, then we'll check downstream" - which is indeed "as written" - just want to check our intent (maybe worth a comment?)

src/StackExchange.Redis/ConfigurationOptions.cs
```diff
@@ -313,7 +323,23 @@ private static bool CheckTrustedIssuer(X509Certificate2 certificateToValidate, X
             chain.ChainPolicy.UrlRetrievalTimeout = new TimeSpan(0, 0, 0);
 
             chain.ChainPolicy.ExtraStore.Add(authority);
-            return chain.Build(certificateToValidate);
+            // This only verifies that the chain is valid, but with AllowUnknownCertificateAuthority could trust
+            // self-signed or partial chained vertificates
+            var chainIsVerfiied = chain.Build(certificateToValidate);
```
mgravell, line 328: `chainIsVerified`

tests/StackExchange.Redis.Tests/Certificates/create_certificates.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+set -eu
+# Adopted from https://github.com/stewartadam/dotnet-x509-certificate-verification/blob/main/create_certificates.sh
```
mgravell, line 3: Adapted from?

docs/ReleaseNotes.md
```diff
@@ -9,6 +9,7 @@ Current package versions:
 ## Unreleased
 
 - Add new `LoggingTunnel` API; see https://stackexchange.github.io/StackExchange.Redis/Logging ([#2660 by mgravell](https://github.com/StackExchange/StackExchange.Redis/pull/2660))
+- Fix `CheckTrustedIssuer` validation for broken chain scenarios ([#2665 by NickCraver](https://github.com/StackExchange/StackExchange.Redis/pull/2665))
```
mgravell, line 12: (no change needed here) When we do the release notes, we should find the biggest font available to emphasize a possible change in connectivity

mgravell:
Conceptually I'm :+1: here, however: I *do not* consider myself a certificates expert. Locally, have we tried connecting to (say) azure redis after this change?

docs/ReleaseNotes.md
NickCraver: Check latest - agree and tried to make it a bit more obvious/descriptive.

src/StackExchange.Redis/ConfigurationOptions.cs
NickCraver: Good call - added comment and some tests against to make it a bit more obvious there.

NickCraver:
Same here - reaching out for more test scenarios to be careful here.

src/StackExchange.Redis/ConfigurationOptions.cs
```diff
@@ -313,7 +324,23 @@ private static bool CheckTrustedIssuer(X509Certificate2 certificateToValidate, X
             chain.ChainPolicy.UrlRetrievalTimeout = new TimeSpan(0, 0, 0);
 
             chain.ChainPolicy.ExtraStore.Add(authority);
-            return chain.Build(certificateToValidate);
+            // This only verifies that the chain is valid, but with AllowUnknownCertificateAuthority could trust
+            // self-signed or partial chained vertificates
+            var chainIsVerified = chain.Build(certificateToValidate);
```
vcsjones, line 329: The "right" way to do this in .NET 5+ is to use [CustomTrustStore](https://learn.microsoft.com/en-us/dotnet/api/system.security.cryptography.x509certificates.x509chainpolicy.customtruststore?view=net-8.0#system-security-cryptography-x509certificates-x509chainpolicy-customtruststore). But unfortunately it does not work for your needs since it is based on a trusted root, not a trusted issuer.

The reason I point this out is that AllowUnknownCertificateAuthority can be unreliable. Android, for example, does not work with AllowUnknownCertificateAuthority (reference: https://github.com/dotnet/runtime/blob/80b1e143c686c9548ee5930f2466805a76f095bd/src/libraries/System.Security.Cryptography/tests/X509Certificates/DynamicChainTests.cs#L198). The Android chain builder just doesn't give back partial or untrusted chains. This is "fine" since it doesn't cause you to fail open (that is, not a security issue) but someone could run in to a situation where `TrustIssuer` doesn't work.

_Who is calling Redis from Android, Kevin?_

Well, you never know, plus there are other platform quirks about returning partial chains.

_So, should I use CustomTrustStore?_

Initially I was going to recommend it for .NET 5+, but seeing how your implementation is based on trusting an arbitrary anchor, not a root, it makes it difficult to recommend, or would otherwise have different behavior.

_You still aren't looking at the pull request Kevin_

Oh, right. Let's take a look. Well, it builds a chain and permits exactly one error, AUCA, which is fine. If it's valid, then see if any of the certificates in the chain match the "trusted" certificate by thumbprint. If nothing matches, then "false".

That all seems fine.

Suggestions

1. It is possible for X509Chain.Build to throw CryptographicException in very rare scenarios. You may want to catch it and return false instead of letting the exception bubble up.
2. X509Chain is disposable. You may want to dispose of it instead of leaving it up to the GC.

src/StackExchange.Redis/ConfigurationOptions.cs
```diff
@@ -313,7 +324,23 @@ private static bool CheckTrustedIssuer(X509Certificate2 certificateToValidate, X
...
+            if (chainIsVerified)
+            {
+                // Our method is "TrustIssuer", which means any intermediate cert we're being told to trust
+                // is a valid thing to trust, up until it's a root CA
+                foreach (var chainElement in chain.ChainElements)
+                {
+                    if (chainElement.Certificate.Thumbprint == authority.Thumbprint)
```
bartonjs, line 336: `cert.Thumbprint` is based on SHA-1 (and can't ever be upgraded, because compat).  If you're using .NET Framework 4.8+ or .NET 5+ (or various old versions of .NET Core) you could use cert.GetCertHashString(HashAlgorithmName.SomethingBetterThanSha1).

Or, save off authority.RawData(Memory) and just SequenceEqual the element cert.RawData against the actual cert you want to compare.  Hashing isn't free, and reading Thumbprint (or calling GetCertHashString) just causes the underlying libraries to hash RawData (though with memoization).

src/StackExchange.Redis/ConfigurationOptions.cs
```diff
@@ -296,15 +296,26 @@ private static RemoteCertificateValidationCallback TrustIssuerCallback(X509Certi
...
 
         private static bool CheckTrustedIssuer(X509Certificate2 certificateToValidate, X509Certificate2 authority)
         {
-            // reference: https://stackoverflow.com/questions/6497040/how-do-i-validate-that-a-certificate-was-created-by-a-particular-certification-a
+            // Reference:
+            // https://stackoverflow.com/questions/6497040/how-do-i-validate-that-a-certificate-was-created-by-a-particular-certification-a
+            // https://github.com/stewartadam/dotnet-x509-certificate-verification
             X509Chain chain = new X509Chain();
             chain.ChainPolicy.RevocationMode = X509RevocationMode.NoCheck;
             chain.ChainPolicy.RevocationFlag = X509RevocationFlag.ExcludeRoot;
```
bartonjs, line 321: Nit: Since RevocationMode is NoCheck, RevocationFlag has no effect.

(Was long dissertation on how revocation + AllowUnknownCertificateAuthority is bad; then eventually clocked that you set NoCheck the line above)

src/StackExchange.Redis/ConfigurationOptions.cs
```diff
@@ -313,7 +324,23 @@ private static bool CheckTrustedIssuer(X509Certificate2 certificateToValidate, X
             chain.ChainPolicy.UrlRetrievalTimeout = new TimeSpan(0, 0, 0);
```
bartonjs, line 324: VerificationTime is being set to DateTime.Now.  This is either redundant to the constructor (which did the same thing), or is a few ticks later.

This is creating a skew problem from when the original request happened (the cert could have exited validity), so you probably want to copy the chain.ChainPolicy.VerificationTime value from the chain you were given in the callback.

tests/StackExchange.Redis.Tests/Certificates/CertValidationTests.cs
```diff
@@ -0,0 +1,64 @@
...
+
+public class CertValidationTests : TestBase
+{
+    public CertValidationTests(ITestOutputHelper output) : base(output) { }
+
+    [Fact]
+    public void CheckIssuerValidity()
+    {
+        // The endpoint cert is the same here
+        var endpointCert = LoadCert(Path.Combine("Certificates", "device01.foo.com.pem"));
```
bartonjs, line 21: Instead of checking in certificates that will eventually expire; you could use System.Security.Cryptography.X509Certificates.CertificateRequest to just create them on the fly.  (net472+/net5+)

src/StackExchange.Redis/ConfigurationOptions.cs
bartonjs: > It is possible for X509Chain.Build to throw CryptographicException in very rare scenarios.

FWIW, I believe all of those scenarios are deterministic, based on the EE certificate.  And since this is from an SslStream callback, X509Chain has already run once (and not kerploded), so it's _probably_ fine.

But adding the try/catch doesn't hurt.

src/StackExchange.Redis/ConfigurationOptions.cs
```diff
@@ -296,15 +296,26 @@ private static RemoteCertificateValidationCallback TrustIssuerCallback(X509Certi
...
+                }
+                // If we're not valid due to chain errors - check against the trusted issuer
+                // Note that we're only proceeding at all here if the *only* issue is chain errors (not more flags in SslPolicyErrors)
+                return sslPolicyError == SslPolicyErrors.RemoteCertificateChainErrors
+                       && certificate is X509Certificate2 v2
+                       && CheckTrustedIssuer(v2, issuer);
+            };
         }
 
         private static bool CheckTrustedIssuer(X509Certificate2 certificateToValidate, X509Certificate2 authority)
```
bartonjs, line 314: OK.  Last comment (I think :wink:).

Aside from Internet connectivity flakiness, the only difference between this chain build and the one you were given in the callback is that you've definitely added the specified pin/anchor to ExtraStore.  That means the chain will realistically only be different if the incoming chain terminated in PartialChain and the specified pin/anchor is the next hop.

_That_ means that you could run this same logic on the existing chain, with two minor changes:
1) You'd need to check each chainElement.ChainElementStatus for the presence of an error other than `PartialChain` or `UnknownRootAuthority`.
2) If it did have PartialChain, and you didn't find your anchor, _then_ you'd run it again with this logic.

Both the flow I suggested here and what the code is trying to accomplish, you'll potentially have issues regarding cross-certification.  Since you can't ask .NET to ask the OS to consider this intermediate as a trust anchor, the OS might prefer a variant of the certificate.  For example, the Let's Encrypt R3 intermediate was signed by both "ISRG Root X1" and "DST Root CA X3" (https://crt.sh/?caid=183267).  If the anchor in your flow was the R3 signed by DST-X3; but the underlying OS trusts ISRG-X1 and not DST-X3, then the OS will completely ignore the ExtraStore input as irrelevant because it always prefers to build trusted chains over untrusted chains.  If it trusts both, or neither, then "stuff happens" (I don't know if the ExtraStore nudge will act as a tiebreaker, or if it comes down to load-order, phase of the moon, etc).  Really this is just a concern for the inputs: the certificates should always be from nice single-path segments of a chain.  Once you've made that constraint/assumption, then you don't also have a reason to rerun the chain build just because you don't see the intended anchor.

And, finally, regarding disposing things... for maximum reduction of finalization, you should dispose all of the certificates from chain.ChainElements in the chain the callback gave you.  .NET Framework didn't dispose them (so they went off to the finalizer).  And I thought the .NET Core line only did it when you didn't have a callback; but looking at the code just now it seems to do it in a finally and has no suppression for if the chain was passed to a callback.  Huh.

src/StackExchange.Redis/ConfigurationOptions.cs
```diff
@@ -296,15 +296,26 @@ private static RemoteCertificateValidationCallback TrustIssuerCallback(X509Certi
...
-            return (object _, X509Certificate? certificate, X509Chain? __, SslPolicyErrors sslPolicyError)
-                => sslPolicyError == SslPolicyErrors.RemoteCertificateChainErrors
-                    && certificate is X509Certificate2 v2
-                    && CheckTrustedIssuer(v2, issuer);
+            return (object _, X509Certificate? certificate, X509Chain? __, SslPolicyErrors sslPolicyError) =>
+            {
+                // If we're already valid, there's nothing further to check
+                if (sslPolicyError == SslPolicyErrors.None)
+                {
+                    return true;
```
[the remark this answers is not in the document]
vcsjones: Hm. Based on the documentation I did not _think_ the intention was pinning

https://github.com/StackExchange/StackExchange.Redis/blob/441f89afeff6d9a55ecf6e413ed5bd29b5aec19d/src/StackExchange.Redis/ConfigurationOptions.cs#L282

But a good question to ask.

src/StackExchange.Redis/ConfigurationOptions.cs
bartonjs: I'm somewhere between amused and horrified that it was previously the more correct thing, and then switched to using thumbprint with a subsequent commit.  In my best Mr. Garrison voice: "Thumbprint's bad, mm'kay?"

src/StackExchange.Redis/ConfigurationOptions.cs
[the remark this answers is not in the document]
bartonjs: Since @vcsjones and I have had a discussion offline regarding what that sentence means... if it is pinning, the code's wrong.  If it isn't pinning, the docs would probably be more clear by replacing "even if" with "when".

* I think "even if" means "when normal, and when also", which would make it pinning.
* @vcsjones agrees, but thinks "even if not" is somehow special. Which I disagree with.

/me and @vcsjones now watch to see what Nick says the right answer should be.

src/StackExchange.Redis/ConfigurationOptions.cs
NickCraver: This is indeed why I wanted expert eyes - switched to RawData!

tests/StackExchange.Redis.Tests/Certificates/CertValidationTests.cs
```diff
@@ -0,0 +1,64 @@
...
+        Assert.False(callback(this, endpointCert, null, SslPolicyErrors.RemoteCertificateNameMismatch));
+        Assert.False(callback(this, endpointCert, null, SslPolicyErrors.RemoteCertificateNotAvailable));
+        Assert.False(callback(this, endpointCert, null, SslPolicyErrors.RemoteCertificateChainErrors | SslPolicyErrors.RemoteCertificateNameMismatch));
+        Assert.False(callback(this, endpointCert, null, SslPolicyErrors.RemoteCertificateChainErrors | SslPolicyErrors.RemoteCertificateNotAvailable));
+    }
+
+    private static X509Certificate2 LoadCert(string certificatePath) =>
+        new X509Certificate2(StripPEM(File.ReadAllBytes(certificatePath)));
+    private static byte[] StripPEM(byte[] pemData) =>
+        Convert.FromBase64String(Regex.Replace(Encoding.UTF8.GetString(pemData), "(-+BEGIN CERTIFICATE-+)|(-+END CERTIFICATE-+)", "").Trim());
```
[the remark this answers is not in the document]
NickCraver: ACK - making this tweak then looking at generated-on-the-fly anyhow.

src/StackExchange.Redis/ConfigurationOptions.cs
NickCraver: Done!

src/StackExchange.Redis/ConfigurationOptions.cs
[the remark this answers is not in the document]
NickCraver: Playing favorites only gets me in trouble, tried going with making everyone angry instead - thoughts?

src/StackExchange.Redis/ConfigurationOptions.cs
```diff
@@ -313,7 +324,23 @@ private static bool CheckTrustedIssuer(X509Certificate2 certificateToValidate, X
...
             chain.ChainPolicy.ExtraStore.Add(authority);
-            return chain.Build(certificateToValidate);
+            // This only verifies that the chain is valid, but with AllowUnknownCertificateAuthority could trust
+            // self-signed or partial chained vertificates
+            var chainIsVerified = chain.Build(certificateToValidate);
+            if (chainIsVerified)
+            {
+                // Our method is "TrustIssuer", which means any intermediate cert we're being told to trust
+                // is a valid thing to trust, up until it's a root CA
+                foreach (var chainElement in chain.ChainElements)
```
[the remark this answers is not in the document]
vcsjones: The early return would mean a few may still be undisposed. You could do something like:


```C#
bool found = false;

foreach (var chainElement in chain.ChainElements)
{
    if (!found && chainElement.RawData.SequenceEqual(authority.RawData))
    {
        found = true;
    }

    chainElement.Dispose();
}

return found;
```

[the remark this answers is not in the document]
NickCraver: Tweaked!

src/StackExchange.Redis/ConfigurationOptions.cs
```diff
@@ -296,24 +297,61 @@ private static RemoteCertificateValidationCallback TrustIssuerCallback(X509Certi
...
+                    bool found = false;
+                    foreach (var chainElement in chain.ChainElements)
+                    {
+                        using var chainCert = chainElement.Certificate;
+                        if (!found && chainCert.RawData.SequenceEqual(authority.RawData))
+                        {
+                            found = true;
+                        }
+                    }
+                    return found;
```
vcsjones, line 345: The `RawData` property has the behavior of allocating by returning defensive copies. That can't really be avoided for each intermediate, but you could stash the RawData from the authority so it doesn't allocate in each loop.

.NET 8 has RawDataMemory which does not allocate if and when you ever start to target .NET 8, you could consider conditionally using `RawDataMemory`.

```suggestion
                    bool found = false;
                    byte[] authorityBytes = authority.RawData;
                    foreach (var chainElement in chain.ChainElements)
                    {
                        using var chainCert = chainElement.Certificate;
                        if (!found && chainCert.RawData.SequenceEqual(authorityBytes))
                        {
                            found = true;
                        }
                    }
                    return found;
```

NickCraver: ACK - no target at the moment but will keep in mind!

mgravell: @NickCraver  maybe we should add a pre-emptive:

``` c#
#if NET8_0_OR_GREATER
#error TODO: use RawDataMemory (needs testing)
#endif
```

??

mgravell (approved):

philon-msft (approved):

src/StackExchange.Redis/ConfigurationOptions.cs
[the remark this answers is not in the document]
bartonjs: So is the intention that anything that's trusted by default is acceptable, and if it's not trusted then (and only then) this anchor is checked as a fallback?

If that's the intention, my concern in the docs is with the word "even";

> Create a certificate validation check that checks against the supplied issuer even when not known by the machine.

Sounds like pinning to me.

> Create a certificate validation check that checks against the supplied issuer when not known by the machine.

Sounds like a fallback.

src/StackExchange.Redis/ConfigurationOptions.cs
```diff
@@ -296,24 +297,64 @@ private static RemoteCertificateValidationCallback TrustIssuerCallback(X509Certi
...
+#if NET8_0_OR_GREATER
+#error TODO: use RawDataMemory (needs testing)
+#endif
+                        using var chainCert = chainElement.Certificate;
+                        if (!found && chainCert.RawData.SequenceEqual(authority.RawData))
+                        {
+                            found = true;
+                        }
+                    }
+                    return found;
```
bartonjs, line 348: I think I saw Kevin mention it already; but you could lift `authority.RawData` out of the loop

```suggestion
                    bool found = false;
                    byte[] authorityData = authority.RawData;
                    foreach (var chainElement in chain.ChainElements)
                    {
#if NET8_0_OR_GREATER
#error TODO: use RawDataMemory (needs testing)
#endif
                        using var chainCert = chainElement.Certificate;
                        if (!found && chainCert.RawData.SequenceEqual(authorityData))
                        {
                            found = true;
                        }
                    }
                    return found;
```

to reduce the number of defensive copies of that array

src/StackExchange.Redis/ConfigurationOptions.cs
```diff
@@ -296,24 +297,64 @@ private static RemoteCertificateValidationCallback TrustIssuerCallback(X509Certi
...
+            chain.ChainPolicy.VerificationTime = chainToValidate?.ChainPolicy?.VerificationTime ?? DateTime.Now;
             chain.ChainPolicy.UrlRetrievalTimeout = new TimeSpan(0, 0, 0);
 
             chain.ChainPolicy.ExtraStore.Add(authority);
-            return chain.Build(certificateToValidate);
+            // This only verifies that the chain is valid, but with AllowUnknownCertificateAuthority could trust
+            // self-signed or partial chained vertificates
+            try
+            {
+                var chainIsVerified = chain.Build(certificateToValidate);
```
bartonjs, line 331: The comment feels misplaced to me.  I believe it's explaining that `chainIsVerified == true` is necessary, but not sufficient.  Which means (to me) it should be inside the try, attached to the call to chain.Build.  I think it got separated when the try/catch was introduced

NickCraver:
@bartonjs @vcsjones Latest comments appreciated - if you have time for final eyes would appreciate it, want to do a release with this and other fixes today if we're good.

PR #2670, opened 2024-03-12
TLS: Check EKU in X509 chain checks

NickCraver:
Further hardening following #2665. This is an additional check to match the .NET implementation for TLS cert checks so that we don't treat a cert flagged as non-TLS-server effectively. This ensures that a certificate either doesn't have OIDs here (valid, backwards compatible) or has the server-certificate OID indicating it's valid for consumption over TLS for us.

Cheers @bartonjs for the report and info here.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -7,7 +7,7 @@ Current package versions:
 | [![StackExchange.Redis](https://img.shields.io/nuget/v/StackExchange.Redis.svg)](https://www.nuget.org/packages/StackExchange.Redis/) | [![StackExchange.Redis](https://img.shields.io/nuget/vpre/StackExchange.Redis.svg)](https://www.nuget.org/packages/StackExchange.Redis/) | [![StackExchange.Redis MyGet](https://img.shields.io/myget/stackoverflow/vpre/StackExchange.Redis.svg)](https://www.myget.org/feed/stackoverflow/package/nuget/StackExchange.Redis) |
 
 ## Unreleased
-No pending/unreleased changes.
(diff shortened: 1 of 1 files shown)


bartonjs (approved):

src/StackExchange.Redis/ConfigurationOptions.cs
```diff
@@ -312,6 +312,8 @@ private static RemoteCertificateValidationCallback TrustIssuerCallback(X509Certi
             };
         }
 
+        private static readonly Oid _serverAuthOid = new Oid("1.3.6.1.5.5.7.3.1", "1.3.6.1.5.5.7.3.1");
```
TimLovellSmith, line 315: Just wondering why not the friendly name parameter is something like serverAuth..? Or else using the single parameter constructor?

mgravell: Good question; in my local testing, this gives the `FriendlyName` of `"TLS Web Server Authentication"` which is perhaps a little confusing since we're not really connection to a "web server", so perhaps `new Oid("1.3.6.1.5.5.7.3.1", "TLS Server Authentication");` might be more sensible. Thoughts? @NickCraver also?

NickCraver: I...didn't know that was a thing... Either sounds good here - cheers for the TIL!

bartonjs: Mainly "because no one will look at the friendly name".  The single param ctor causes the friendly name to be looked up, and Windows XP through Win 8.1 sometimes caused that to kick off a lookup over to the domain controller... so in the libraries we always use the two parameter version when the Oid instance won't be seen outside the library.

PR #2708, opened 2024-04-30
Added SetUnion, SetIntersect and SetDiff to IDatabase. All of them are used SetCombine method

mohsenbgi:
Added SetUnion, SetIntersect and SetDiff to IDatabase.
All of them are used SetCombine method

### DIFF
--- a/src/StackExchange.Redis/Interfaces/IDatabase.cs
+++ b/src/StackExchange.Redis/Interfaces/IDatabase.cs
@@ -1352,6 +1352,343 @@ namespace StackExchange.Redis
         long SetAdd(RedisKey key, RedisValue[] values, CommandFlags flags = CommandFlags.None);
 
         /// <summary>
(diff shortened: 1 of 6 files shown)


NickCraver:
Hey there! I'm curious what functionality is missing in the current APIs here - can you please explain what's not current covered by the `SetCombine(AndStore)(Async)` methods today?

mohsenbgi:
Hi there, I think that if Redis commands are named intuitively, they can be easier to use and enhance the readability of the code.

NickCraver:
I hear you, and that's certainly a good argument I agree with for any new APIs being added to be intuitive as always, but adding duplicate commands creating both confusion and a lot of overhead does not further that goal.

If we have thoughts on how to make the existing APIs more discoverable, certainly open to options there but we shouldn't add dozens of duplicates to the surface area to achieve that :)

mohsenbgi:
You're right. Thanks for taking the time to review 🙌.

PR #2716, opened 2024-05-20
Feature: Support for hash field expiration

### LINKED ISSUE #2715: Support for Hash Field Expiration commands
There is an upcoming set of commands which will provide a new feature of Redis called Hash Field Expiration.
Related development is in progress and could be checked within the [Redis branch](https://github.com/redis/redis/tree/hash-field-expiry-integ) and initial draft of it [here](https://github.com/redis/redis/pull/13172).

To put it simply, hash field expiration will enable expiration for individual fields in a hash, in a way similar to the key expiration functions in Redis.

### list of related Redis commands
 
    HEXPIRE - sets the remaining time to live in seconds for individual members of hash
    HPEXPIRE - sets the remaining time to live in milliseconds for individual members of hash
    HEXPIREAT - sets the expiration time to a UNIX timestamp in seconds for individual members of hash
    HPEXPIREAT - sets the expiration time to a UNIX timestamp in milliseconds for individual members of hash
    HPERSIST - removes the expiration time for individual members of hash
    HTTL - get the remaining time to live in seconds for individual members of hash
    HPTTL - get the remaining time to live in milliseconds for individual members of hash
    HEXPIRETIME - get the expiration time as a UNIX timestamp in seconds for individual members of hash
    HPEXPIRETIME - get the expiration time as a UNIX timestamp in milliseconds for individual members of hash
		
		
### structure and arguments of commands
**commands:**

	HEXPIRE `key` `seconds` [NX | XX | GT | LT] <`FIELDS `count` `field` [`field` ...]> - for each specified field: set the remaining time to live in seconds
	HPEXPIRE `key` `milliseconds` [NX | XX | GT | LT] <`FIELDS` `count` `field` [`field` ...]> - for each specified field: set the remaining time to live in milliseconds
	HEXPIREAT `key` `unix-time-seconds` [NX | XX | GT | LT] <`FIELDS` `count` `field` [`field` ...]> - for each specified field: set the expiration time to a UNIX timestamp specified in seconds since the Unix epoch
	HPEXPIREAT `key` `unix-time-milliseconds` [NX | XX | GT | LT] <`FIELDS` `count` `field` [field ...]> - for each specified field: set the expiration time to a UNIX timestamp specified in milliseconds since the Unix epoch
>
**arguments:**

	- `key` must be type of hash 
	- `second`, `milliseconds`, `unix-time-seconds`, and `unix-time-milliseconds` are non-negative integers.
	- `NX` – set only when the field has no expiration
	- `XX` – set only when the field has an existing expiration
	- `GT` – set only when the field's new expiration time is greater than current one. 
	- `LT` – set only when the field's new expiration time is less than current one. 
	- `count` must be a positive integer; number of fields following must match `count`
>
**returns:**

	- If `key` does not exist: nil
	- If `key` exists: an array containing result for each field given in arguments
		2: field deleted because the specified expiration time is due, 
		1: expiration time set/updated, 
		0:  expiration time NOT set/updated (a specified NX | XX | GT | LT condition not met)
		-2: if no such field

		
___		
**commands:**

	HPERSIST `key` <FIELDS count field [field ...]> - for each specified field: remove the expiration time
	
**arguments:**

	- `key` must be type of hash 
	- `count` must be a positive integer; number of fields following must match `count`
	
**returns:**

	- If `key `does not exist: nil
	- If `key` exists: an array containing result for each field given in arguments
		1: if the expiration time was removed
		-1: if the expiration time was NOT removed (field has no associated expiration time), 
		-2: if no such field

___	
**commands:**

	HEXPIRETIME key <FIELDS count field [field ...]> - for each specified field: get the expiration time as a Unix timestamp in seconds since the Unix epoch
	HPEXPIRETIME key <FIELDS count field [field ...]> - for each specified field: get the expiration time as a Unix timestamp in milliseconds since the Unix epoch
	HTTL key <FIELDS count field [field ...]> - for each specified field: get the remaining time to live in seconds
	HPTTL key <FIELDS count field [field ...]> - for each specified field: get the remaining time to live in milliseconds	
	
**arguments:**

	- `key` must be type of hash 
	- `count` must be a positive integer; number of fields following must match `count`
	
**returns:**

	- If `key `does not exist: nil
	- If `key` exists: an array containing result for each field given in arguments
		expire time: depends on command, either time to live (in seconds/milliseconds) or UNIX timestamp (in seconds/milliseconds)
		1 : if field has no associated expiration time
		-2: if no such field

___

atakavci:
Closes/Fixes #2715 

this PR add support for a set of new commands related to expiration of individual members of hash

> 
> 	**_HashFieldExpire_** exposes the functionality of commands HPEXPIRE/HPEXPIREAT, for each specified field, it gets the value and sets the field's remaining time to live or expireation timestamp

> 	**_HashFieldExpireTime_** exposes the functionality of command HPEXPIRETIME, for specified field, it gets the remaining time to live in milliseconds or expiration timestamp

> 	**_HashFieldPersist_** exposes the functionality of command HPERSIST, for each specified field, it removes the expiration time

> 	**_HashFieldTimeToLive_** expoes the functionality of command HPTTL, for specified field, it gets the remaining time to live in milliseconds or expiration timestamp


mgravell:
I don't want to merge this until these commands actually land in redis, but we can certainly take a look at the proposal in advance of that 👍

src/StackExchange.Redis/Enums/HashFieldFlags.cs
```diff
@@ -0,0 +1,73 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace StackExchange.Redis;
+
+/// <summary>
+///  Specifies the options to the HSETF command when to create hash/fields and the return data
+/// </summary>
+[Flags]
```
mgravell, line 10: It feels like this is a duplicate re `When`, perhaps coupled with some API shaping - I'd need to think. Either way: we've tried to lean away from abbreviations

src/StackExchange.Redis/Enums/ExpireResult.cs
```diff
@@ -0,0 +1,27 @@
+﻿using System;
+
+namespace StackExchange.Redis;
+
+/// <summary>
+/// Specifies when to set the expiry for a key.
```
mgravell, line 6: Not sure this description is right

atakavci:
sure!! aim is to be prepared for the time this feature get ready in Redis. 
please take a look when you have time and i d be happy to get your thoughts !!

src/StackExchange.Redis/Enums/HashFieldFlags.cs
```diff
@@ -0,0 +1,73 @@
...
+    GETNEW = 8,
+    /// <summary>
+    /// When GETOLD is specified: returns the old value of given fields  
+    /// </summary>
+    GETOLD = 16,
+}
+
+internal static class HashFieldFlagsExtensions
+{
+    internal static bool isNone(this HashFieldFlags flags) =>
```
mgravell, line 42: .NET naming conventions are Proper case; I also wonder whether a single `.HasAny(this value, flag) => (value & flag) != 0;` achieves more with less

src/StackExchange.Redis/Enums/HashFieldFlags.cs
```diff
@@ -0,0 +1,73 @@
...
+    internal static bool isDOF(this HashFieldFlags flags) =>
+        flags.HasFlag(HashFieldFlags.DOF);
+
+    internal static bool isGETNEW(this HashFieldFlags flags) =>
+        flags.HasFlag(HashFieldFlags.GETNEW);
+
+    internal static bool isGETOLD(this HashFieldFlags flags) =>
+        flags.HasFlag(HashFieldFlags.GETOLD);
+
+    internal static List<RedisValue> ToRedisValueList(this HashFieldFlags flags)
```
mgravell, line 59: I suspect we can be much more efficient about this

src/StackExchange.Redis/Interfaces/IDatabase.cs
```diff
@@ -325,6 +325,153 @@ public interface IDatabase : IRedis, IDatabaseAsync
...
+        /// <param name="expiry">The timeout to set.</param>
+        /// <param name="when">under which condition the expiration will be set using <see cref="ExpireWhen"/>.</param>
+        /// <param name="flags">The flags to use for this operation.</param>
+        /// <returns> null if the key does not exist. Otherwise returns the result of operation for given field;
+        ///		2: field deleted because the specified expiration time is due
+        ///		1: expiration time set/updated
+        /// 	0: expiration time is not set/update (a specified ExpireWhen condition is not met)
+        ///		-2 : no such field
+        /// </returns>
+        ExpireResult? HashFieldExpire(RedisKey key, RedisValue hashField, TimeSpan expiry, ExpireWhen when = ExpireWhen.Always, CommandFlags flags = CommandFlags.None);
```
mgravell, line 344: In what scenario would this return null?

src/StackExchange.Redis/Interfaces/IDatabase.cs
```diff
@@ -325,6 +325,153 @@ public interface IDatabase : IRedis, IDatabaseAsync
...
+        /// <param name="expiry">The timeout to set.</param>
+        /// <param name="when">under which condition the expiration will be set using <see cref="ExpireWhen"/>.</param>
+        /// <param name="flags">The flags to use for this operation.</param>
+        /// <returns> null if the key does not exist. Otherwise returns an array where each item is the result of operation for given fields;
+        ///		2: field deleted because the specified expiration time is due
+        ///		1: expiration time set/updated
+        /// 	0: expiration time is not set/update (a specified ExpireWhen condition is not met)
+        ///		-2 : no such field
+        /// </returns>
+        ExpireResult[]? HashFieldExpire(RedisKey key, RedisValue[] hashFields, TimeSpan expiry, ExpireWhen when = ExpireWhen.Always, CommandFlags flags = CommandFlags.None);
```
mgravell, line 361: In what scenario would this return a null array?

src/StackExchange.Redis/Interfaces/IDatabase.cs
```diff
@@ -325,6 +325,153 @@ public interface IDatabase : IRedis, IDatabaseAsync
...
+        /// </summary>
+        /// <param name="key">The key of the hash.</param>
+        /// <param name="hashField">The field in the hash to get expire time.</param>
+        /// <param name="flags">The flags to use for this operation.</param>
+        /// <returns>null if the key does not exist. Otherwise returns the result of operation for given field
+        ///     expiration time: as a UNIX timestamp in milliseconds
+        ///     -1: if field has no associated expiration time 
+        ///     -2: no such field
+        /// </returns>
+        long? HashFieldExpireTime(RedisKey key, RedisValue hashField, CommandFlags flags = CommandFlags.None);
```
mgravell, line 408: This should be DateTime? presumably? Namewise, this sounds like an action rather than a query - something "get expiry" perhaps? Also comparable to KeyTimeToLive (although that's the "TimeSpan" rather than "DateTime" approach)

src/StackExchange.Redis/Interfaces/IDatabase.cs
```diff
@@ -325,6 +325,153 @@ public interface IDatabase : IRedis, IDatabaseAsync
...
+        /// </summary>
+        /// <param name="key">The key of the hash.</param>
+        /// <param name="hashField">The field in the hash to remove expire time.</param>
+        /// <param name="flags">The flags to use for this operation.</param>
+        /// <returns>null if the key does not exist. Otherwise returns the result of operation for given field
+        ///     1: if the expiration time was removed
+        ///     -1: if field has no associated expiration time 
+        ///     -2: no such field
+        /// </returns>
+        long? HashFieldPersist(RedisKey key, RedisValue hashField, CommandFlags flags = CommandFlags.None);
```
mgravell, line 434: These naked results aren't ideal IMO

src/StackExchange.Redis/Enums/HashFieldFlags.cs
atakavci: abbreviations fixed

src/StackExchange.Redis/Enums/HashFieldFlags.cs
atakavci: fixed

src/StackExchange.Redis/Interfaces/IDatabase.cs
atakavci: fixed

src/StackExchange.Redis/Interfaces/IDatabase.cs
atakavci: when the key does not exist, it returns in compliance with response from server

uglide:
> I don't want to merge this until these commands actually land in redis, but we can certainly take a look at the proposal in advance of that 👍

Redis 7.4-RC1 was released. See https://github.com/StackExchange/StackExchange.Redis/pull/2739

src/StackExchange.Redis/Interfaces/IDatabaseAsync.cs
```diff
@@ -312,6 +312,79 @@ public interface IDatabaseAsync : IRedisAsync
...
+        /// </summary>
+        /// <param name="key">The key of the hash.</param>
+        /// <param name="hashFields">The fields in the hash to get expire time.</param>
+        /// <param name="flags">The flags to use for this operation.</param>
+        /// <returns>Empty array if the key does not exist. Otherwise returns the result of operation for given fields
+        ///     expiration time: as a UNIX timestamp in milliseconds
+        ///     -1: if field has no associated expiration time 
+        ///     -2: no such field
+        /// </returns>
+        Task<long[]> HashFieldGetExpireDateTimeAsync(RedisKey key, RedisValue[] hashFields, CommandFlags flags = CommandFlags.None);
```
mgravell, line 360: Normally we'd want to parse this out to `TimeSpan` here, but I guess we can't do that conveniently here because of the alternative (`-1`/`-2`) results; I think it's probably fine, all things considered - any concerns @NickCraver  ?

src/StackExchange.Redis/RedisDatabase.cs
```diff
@@ -387,6 +388,98 @@ public Task<bool> HashExistsAsync(RedisKey key, RedisValue hashField, CommandFla
...
+            {
+                ExpireWhen.Always => new List<RedisValue> { expiry, RedisLiterals.FIELDS, hashFields.Length },
+                _ => new List<RedisValue> { expiry, when.ToLiteral(), RedisLiterals.FIELDS, hashFields.Length },
+            };
+            values.AddRange(hashFields);
+            var msg = Message.Create(Database, flags, cmd, key, values.ToArray());
+            return executor(msg, processor);
+        }
+
+        private RedisCommand PickExpireCommandByPrecision(bool useSeconds) => useSeconds ? RedisCommand.HEXPIRE : RedisCommand.HPEXPIRE;
```
mgravell, line 433: this should be static; this is important to avoid additional delegate creation when using this; to see the difference, see [sharplab example](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0ATEBqAPgAQCYBGAWACh8BmAAiJoGEKBvCm9u2/FGgVQGcYASQB2/AC4BDEWBgBZGOIAWEDAAoAlDQC8APhoARCABUlASxEBzNaInTZC5ao0BuNh2p0eAmAGUp4mZgjirqWnqGJuZWav6SgcGKoa7u7Kk0AA5QZgBu8TBekaYW1vgArAA8wBAQADb6AGYArjLh+spQEADuNCIwPQByEOJCALYZtTCjMCLiMBgAogiyGYEQIppu5Bw06dV1NLZSMvJJqpo67UqdPX2Dw2MTUzNzi8swq2brm+n4xEg0fa1GhxBIhc5tGgdbq9fo0IYjcaTaazeZLFZrDYpcgAXyAA===) - see how the static usage gets memoized (into `<>O.<0>__StaticMethod`)

mgravell (changes requested):
a couple of thoughts and nits; also aware that the whitespace deltas are going to compete with 2722 🤷

src/StackExchange.Redis/RedisDatabase.cs
atakavci: done

atakavci:
> a couple of thoughts and nits; also aware that the whitespace deltas are going to compete with 2722 🤷

@mgravell, lets merge this one first if you are good with it. And i ll handle whitespaces with #2722.

src/StackExchange.Redis/Interfaces/IDatabaseAsync.cs
NickCraver: Yeah I think we're in magical territory there with TimeSpan, unfortunately - best option is as-is.

NickCraver:
@atakavci I'm about to merge the StyleCop bits in now (prepped locally), but I'll immediately fix this up and eat the merge on this end - not asking committers to do it :)

NickCraver:
Updates pushed here - docs tweaked and release notes added as well - if builds are all good 👍on merging this in a bit.

uglide:
@NickCraver Could you merge this PR?

PR #2722, opened 2024-05-22
Add support for HSCAN NOVALUES

### LINKED ISSUE #2721: Support for HSCAN NOVALUES option
add support for NOVALUES option to HSCAN command where Redis return only field names/keys of hash

check [here](https://redis.io/docs/latest/commands/scan/#the-novalues-option) for more details

atakavci:
Closes/Fixes #2721 

brings new functions to the API ;

- IDatabase.HashScanNoValues
- IDatabase.HashScanNoValues
- IDatabaseAsync.HashScanNoValuesAsync

to enable the return type consisting of keys in the hash.
added some unit and integration tests in paralled to what is there for `HashScan` and `HashScanAsnyc`.


src/StackExchange.Redis/RedisDatabase.cs
```diff
@@ -4629,21 +4660,30 @@ internal class ScanEnumerable<T> : CursorEnumerable<T>
...
 
             private protected override ResultProcessor<CursorEnumerable<T>.ScanResult> Processor { get; }
 
             private protected override Message CreateMessage(in RedisValue cursor)
             {
+                if (noValues)
+                {
+                    if (CursorUtils.IsNil(pattern) && pageSize == CursorUtils.DefaultRedisPageSize) return Message.Create(db, flags, command, key, cursor, RedisLiterals.NOVALUES);
+                    if (CursorUtils.IsNil(pattern)) return Message.Create(db, flags, command, key, cursor, RedisLiterals.COUNT, pageSize, RedisLiterals.NOVALUES);
+                    return Message.Create(db, flags, command, key, new RedisValue[] { cursor, RedisLiterals.MATCH, pattern, RedisLiterals.COUNT, pageSize, RedisLiterals.NOVALUES });
```
NickCraver, line 4684: Let's use the non-array for `RedisValue` `Message.Create` here please

src/StackExchange.Redis/RedisDatabase.cs
```diff
@@ -4504,17 +4535,17 @@ private Message GetStringGetWithExpiryMessage(RedisKey key, CommandFlags flags,
...
                     _ => throw new ArgumentOutOfRangeException(nameof(when)),
                 };
             }
 
             return when switch
             {
-                When.Always    => Message.Create(Database, flags, RedisCommand.SET, key, value, RedisLiterals.PX, milliseconds, RedisLiterals.GET),
-                When.Exists    => Message.Create(Database, flags, RedisCommand.SET, key, value, RedisLiterals.PX, milliseconds, RedisLiterals.XX, RedisLiterals.GET),
+                When.Always => Message.Create(Database, flags, RedisCommand.SET, key, value, RedisLiterals.PX, milliseconds, RedisLiterals.GET),
+                When.Exists => Message.Create(Database, flags, RedisCommand.SET, key, value, RedisLiterals.PX, milliseconds, RedisLiterals.XX, RedisLiterals.GET),
```
NickCraver, line 4548: There's a lot of whitespace changes in here which affect git history - can we please remove the non-diff so we only have relevant changes in the PR?

tests/StackExchange.Redis.Tests/HashTests.cs
```diff
@@ -126,6 +126,93 @@ public void Scan()
...
+        Array.Sort(v3);
+        Array.Sort(v4);
+
+        Assert.Equal(new RedisValue[] { "abc", "ghi", "mno" }, v1);
+        Assert.Equal(new RedisValue[] { "ghi" }, v2);
+        Assert.Equal(new RedisValue[] { "abc", "ghi", "mno" }, v3);
+        Assert.Equal(new RedisValue[] { "ghi" }, v4);
+    }
+
+
```
NickCraver, line 215: Nit: extra line here

tests/StackExchange.Redis.Tests/ScanTests.cs
```diff
@@ -334,6 +334,57 @@ public void HashScanLarge(int pageSize)
...
+    public void HashScanNoValuesLarge(int pageSize)
+    {
+        using var conn = Create();
+
+        RedisKey key = Me() + pageSize;
+        var db = conn.GetDatabase();
+        db.KeyDelete(key, CommandFlags.FireAndForget);
+
+        for (int i = 0; i < 2000; i++)
+            db.HashSet(key, "k" + i, "v" + i, flags: CommandFlags.FireAndForget);
```
NickCraver, line 382: Braces here please!

NickCraver (changes requested):
A few changes requested here but looking awesome - with tidy we should be good, thanks for this!

WeihanLi:
thanks very much for drafting the PR, since this is a new feature, we may need to consider adding Redis server version checking before executing the command with `NOVALUES`,  if `NOVALUES` is not supported by the Redis server, maybe we should use `HSCAN`?

tests/StackExchange.Redis.Tests/ScanTests.cs
atakavci: fixed

tests/StackExchange.Redis.Tests/HashTests.cs
atakavci: fixed

src/StackExchange.Redis/RedisDatabase.cs
atakavci: fixed

src/StackExchange.Redis/RedisDatabase.cs
atakavci: removed those changes. 
here i see tendency to keep it more readable with manual formatting/aligning and i appreciate the effort . Most of the whitespace diff comes from that. But here it seems we kinda leave the auto-format features in IDE, which is a great thing,  and there are other formatting styles we see like braces with/without new line, improper white spaces before/after keywords etc... which causes codebase has non-standard formatting and usages here and there. 
do you think it is a fair trade off ? 

to be able to use auto-fomat, i can send another dedicated PR just to fix these whitespace diff show up here,  how that sounds ??

atakavci:
hey @WeihanLi , checking where the similar approach is applied and will try that. thanks !

WeihanLi:
> where is this quote from?
> 
> > This feature will be released in Redis 8.0 and is available since 7.2.5 in Valkey from my research

From my research:
redis: https://github.com/redis/redis/pull/12765

valkey: 
- https://github.com/valkey-io/valkey/blob/unstable/src/commands/hscan.json
- https://github.com/valkey-io/valkey/blob/7.2.5/src/commands/hscan.json

double checked with valkey, not supported so far, 7.2.5 does not support this feature 😢

atakavci:
my 2 cents on this,,
making developers life easier and keeping client compatible in more situations/environments is cool.
on the other hand handling such scenarios and doing more (than actually needed) under the hood, this introduces the possibility of significant increase in resource usage both on client and server side while people probably wont even get noticed about it (probably until something bad).
what do you think about it ?

src/StackExchange.Redis/RedisDatabase.cs
NickCraver: Sorry for the delay here - your notes are totally valid, going to add StyleCop and do one big cleanup pass to enforce consistency with more PRs coming in, but want to get your existing efforts merged first so I eat the diff rather than you on cleanup.

atakavci: then we are good to merge this, right ?

atakavci:
hey @NickCraver, how about taking another look to this one?? Let me know if you have concerns or there are things to fix and i ll deal with it.

mgravell (approved):
LGTM; the unrelated whitespace changes are ... not ideal in a feature PR, but I think we should just eat them - they're fine

TODO (not necessarily as part of this PR, for timing): release notes

NickCraver:
I noticed in updating this that we have a mismatch here on `HashScanNoValues` (2 overloads)` vs `HashScanNoValuesAsync` (1 method).

I'd ask when adding here: that's the purpose of the smaller overload for `HashScanNoValues`? It seems like we should just remove it and normalize on the one method, but maybe I'm missing a use case.

atakavci:
> I'd ask when adding here: that's the purpose of the smaller overload for `HashScanNoValues`? It seems like we should just remove it and normalize on the one method, but maybe I'm missing a use case.

hey @NickCraver,, trying to remember my reasoning.. and after a quick look back to the code, you are most probably right. 
I guess it was just the sake of keeping the convention and provide same method signatures with `HashScan` and `HashScanAsync` already there.
i dont have a strong opinion on it. What do you prefer ??

NickCraver:
@atakavci Let's trim down and remove please - usually the #1 reason for overload is backwards compat, which isn't in play here :)

atakavci:
> @atakavci Let's trim down and remove please - usually the #1 reason for overload is backwards compat, which isn't in play here :)

done :+1:

NickCraver (approved):

PR #2723, opened 2024-05-22
update to handle null messages in FailNoServer method

halilkocaoz:
After this exception:
> System.NullReferenceException: Object reference not set to an instance of an object.
>        at StackExchange.Redis.RedisBatch.FailNoServer(ConnectionMultiplexer muxer, List`1 messages) in /_/src/StackExchange.Redis/RedisBatch.cs:line 120
>               at StackExchange.Redis.RedisBatch.Execute() in /_/src/StackExchange.Redis/RedisBatch.cs:line 30

I was able to have a look at the code and found this:
`(messages == null)` expression is always false because of the nullable annotation.

### DIFF
--- a/src/StackExchange.Redis/RedisBatch.cs
+++ b/src/StackExchange.Redis/RedisBatch.cs
@@ -112,7 +112,7 @@ namespace StackExchange.Redis
         internal override T ExecuteSync<T>(Message? message, ResultProcessor<T>? processor, ServerEndPoint? server = null, T? defaultValue = default) where T : default =>
             throw new NotSupportedException("ExecuteSync cannot be used inside a batch");
 
-        private static void FailNoServer(ConnectionMultiplexer muxer, List<Message> messages)
+        private static void FailNoServer(ConnectionMultiplexer muxer, List<Message>? messages)
         {
             if (messages == null) return;
             foreach(var msg in messages)


halilkocaoz:
Hi @mgravell and @NickCraver 

I submitted a pull request a week ago to handle exception when checking null messages in the FailNoServer method. I would appreciate it if someone could review it when they have a moment.

NickCraver:
Thanks for the report! I don't think what you're seeing is caused or helped by this though, this implies we have a null message _in the list_ or an inlined call is failing with a misleading line number e.g. on the `.Fail` call.

Is this happening while your ConnectionMultiplexer is being disposed by chance?

halilkocaoz:
> Thanks for the report! I don't think what you're seeing is caused or helped by this though, this implies we have a null message _in the list_ or an inlined call is failing with a misleading line number e.g. on the `.Fail` call.
> 
> Is this happening while your ConnectionMultiplexer is being disposed by chance?

I don't think, but I will reproduce it and inform again

halilkocaoz:
> > Thanks for the report! I don't think what you're seeing is caused or helped by this though, this implies we have a null message _in the list_ or an inlined call is failing with a misleading line number e.g. on the `.Fail` call.
> > Is this happening while your ConnectionMultiplexer is being disposed by chance?
> 
> I don't think, but I will reproduce it and inform again

I could not reproduce the situation, if the messages are null, I saw that it got a null exception without entering the iteration. The problem may not be related to this situation. I can re-evaluate and maybe try to explain the situation with the issue, thanks.

PR #2727, opened 2024-05-23
support for CLIENT KILL MAXAGE

### LINKED ISSUE #2726: Support MAXAGE option for CLIENT KILL
Support for the option to specify max age to filter client connections to be killed.

for more details check [here](https://redis.io/docs/latest/commands/client-kill/)

atakavci:
Closes/Fixes #2726 

added two overloads and use a [filter object](https://github.com/StackExchange/StackExchange.Redis/compare/main...atakavci:StackExchange.Redis:ali/maxage?expand=1#diff-6c39ce347d72497c65974f0d78ab9a986be3198e7b2555975ff06410a9831bd8)  to avoid a breaking change.


NickCraver:
Reviewed on the call today - in general, this pattern won't work due to the server-side API supporting `n` filters, we'd need to allow `n` of them ordered on the command. We'd either need to take a `params` here or more likely do a builder pattern for `n` filters. In a builder, returning itself, we could do things like making a method _per_ filter that's very specific and only has the args needed for the 7 doced paths:

- `CLIENT KILL ADDR ip:port`
- `CLIENT KILL LADDR ip:port`
- `CLIENT KILL ID client-id`
- `CLIENT KILL TYPE type`
- `CLIENT KILL USER username`
- `CLIENT KILL SKIPME yes/no`
- `CLIENT KILL MAXAGE maxage`

This wouldn't be very dissimilar to the constructors you have now in terms of one-per, but we can make it intuitive and pit-of-success here for users to do the right thing. And we can make it write the message directly on the backend since it's in the base library. Picture for example:

```cs
IServer.ClientKillAsync(new ClientKillFilter().WithAddr(...).WithSkipMe(true));`
```

Thoughts?

atakavci:
hey @NickCraver , sounds nice and clear ! please take a look with new commit.

NickCraver (changes requested):
Took a look here local and pushed up some fixes, but it looks like we don't implement the `LADDR` or `USER` filters in the actual send, so we need to add those and preferably add tests around the filter calling each command and making sure these are set.

atakavci:
@NickCraver , yeah i missed that two. Adding them..
Now i am looking for convenient way to check the arguments of the `Message` in unit test. Could you spot a sample ?

NickCraver (approved):
Latest is looking good - thank you!

PR #2741, opened 2024-06-13
Implement high integrity mode for commands

### LINKED ISSUE #2706: Feature proposal: high integrity mode
This is a client-library alternative suggestion to https://github.com/redis/redis/issues/13240, to address the issue of protocol desync going undetected for at least "some" commands, and causing mayhem.

Proposal: "high integrity mode"

- new `ConfigurationOptions` option/connection-string/etc
- adds a correlation identifier (cid) onto every top-level message (not inside `multi`/`exec`); this could be random or monotonic incrementing (the latter is probably cheaper?)
- adds a hidden `echo {cid}` or `ping {cid}` (depending on what is available) before ***all*** commands (at least on the interactive connection, probably not needed on pub/sub, but is possible there via `ping`)
- the cid is verified before processing all responses, burning connections if it ever fails and screaming very loudly
- should possibly focus on the overhauled API
- needs performance tuning and impact measurement
- need to document / make enquiries about server cost impact ("do you charge by op? are ping/echo included in that?")

mgravell:
Fixes #2706

This adds additional `ECHO` items after every outbound request, and adds validation of that correlation marker before treating responses as valid.

This is initial draft only; needs conversion to sequence and testing of failure modes.

#### Discussion points
- [x] Entropy vs counter - Going with Counter
- [x] Naming
- [ ] Confidence testing (success and failure)
- [x] Performance impact testing
- [x] Depending on performance impact: should this be defaulted for any scenarios? azure?
- [x] Needs checks of sync vs async, F+F, etc

### DIFF
--- a/src/StackExchange.Redis/Message.cs
+++ b/src/StackExchange.Redis/Message.cs
@@ -1,6 +1,7 @@
 ﻿using Microsoft.Extensions.Logging;
 using StackExchange.Redis.Profiling;
 using System;
+using System.Buffers.Binary;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
@@ -51,6 +52,8 @@ namespace StackExchange.Redis
     internal abstract class Message : ICompletable
     {
         public readonly int Db;
+
+        private int _highIntegrityChecksum;
 
         internal const CommandFlags InternalCallFlag = (CommandFlags)128;
 
@@ -197,6 +200,18 @@ namespace StackExchange.Redis
         }
 
         public bool IsAsking => (Flags & AskingFlag) != 0;
+
+        public bool IsHighIntegrity => _highIntegrityChecksum != 0;
+        public int HighIntegrityChecksum => _highIntegrityChecksum;
+
+        internal void WithHighIntegrity(Random rng)
+        {
+            // create a required value if needed; avoid sentinel zero
+            while (_highIntegrityChecksum is 0)
+            {
+                _highIntegrityChecksum = rng.Next();
+            }
+        }
 
         internal bool IsScriptUnavailable => (Flags & ScriptUnavailableFlag) != 0;
 
@@ -704,6 +719,28 @@ namespace StackExchange.Redis
                 WriteImpl(physical);
             }
             catch (Exception ex) when (ex is not RedisCommandException) // these have specific meaning; don't wrap
+            {
+                physical?.OnInternalError(ex);
+                Fail(ConnectionFailureType.InternalFailure, ex, null, physical?.BridgeCouldBeNull?.Multiplexer);
+            }
+        }
+
+        private static ReadOnlySpan<byte> ChecksumTemplate => "$4\r\nXXXX\r\n"u8;
+
+        internal void WriteHighIntegrityChecksumRequest(PhysicalConnection physical)
+        {
+            Debug.Assert(IsHighIntegrity, "should only be used for high-integrity");
+            try
+            {
+                physical.WriteHeader(RedisCommand.ECHO, 1); // use WriteHeader to allow command-rewrite
+
+                Span<byte> chk = stackalloc byte[10];
+                Debug.Assert(ChecksumTemplate.Length == chk.Length, "checksum template length error");
+                ChecksumTemplate.CopyTo(chk);
+                BinaryPrimitives.WriteInt32LittleEndian(chk.Slice(4, 4), _highIntegrityChecksum);
+                physical.WriteRaw(chk);
+            }
+            catch (Exception ex)
             {
                 physical?.OnInternalError(ex);
                 Fail(ConnectionFailureType.InternalFailure, ex, null, physical?.BridgeCouldBeNull?.Multiplexer);
--- a/src/StackExchange.Redis/Message.cs
+++ b/src/StackExchange.Redis/Message.cs
@@ -53,7 +53,7 @@ namespace StackExchange.Redis
     {
         public readonly int Db;
 
-        private int _highIntegrityChecksum;
+        private uint _highIntegrityToken;
 
         internal const CommandFlags InternalCallFlag = (CommandFlags)128;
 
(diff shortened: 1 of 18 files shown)


src/StackExchange.Redis/Message.cs
```diff
@@ -198,6 +201,18 @@ public bool IsAdmin
...
+        public int HighIntegrityChecksum => _highIntegrityChecksum;
+
+        internal void WithHighIntegrity(Random rng)
+        {
+            // create a required value if needed; avoid sentinel zero
+            while (_highIntegrityChecksum is 0)
+            {
+                _highIntegrityChecksum = rng.Next();
+            }
+        }
```
NickCraver, line 214: Discussed on call: switching to incrementing number for cheapness and we can error indicating how many we dropped.

src/StackExchange.Redis/PhysicalConnection.cs
```diff
@@ -1680,10 +1699,24 @@ private void MatchResult(in RawResult result)
                 // if it didn't look like "[p]message", then we still need to process the pending queue
             }
             Trace("Matching result...");
-            Message? msg;
+
+            Message? msg = null;
+            // check whether we're waiting for a high-integrity mode post-response checksum (using cheap null-check first)
+            if (_awaitingChecksum is not null && (msg = Interlocked.Exchange(ref _awaitingChecksum, null)) is not null)
```
NickCraver, line 1705:
```suggestion
            if (_awaitingChecksum is not null && Interlocked.Exchange(ref _awaitingChecksum, null)) is Message msg)
```

NickCraver: Ah we likely can't due to assignment uncertainty - ignore if so.

src/StackExchange.Redis/PhysicalConnection.cs
```diff
@@ -1696,11 +1729,55 @@ private void MatchResult(in RawResult result)
...
+                bool isValid = false;
+                if (result.Resp2TypeBulkString == ResultType.BulkString)
+                {
+                    var payload = result.Payload;
+                    if (payload.Length == 4)
+                    {
+                        int interpreted;
+                        if (payload.IsSingleSegment)
+                        {
+                            interpreted = BinaryPrimitives.ReadInt32LittleEndian(payload.First.Span);
```
NickCraver, line 1759: Diccussed: might be we can simplify if we store the expected sequence on the `Message` do we can check/mate it when dequeue from _writtenAwaitingResponse and then we don't need to track state anywhere else

src/StackExchange.Redis/PhysicalConnection.cs
```diff
@@ -2032,6 +2109,7 @@ internal enum ReadStatus
             PubSubPMessage,
             Reconfigure,
             InvokePubSub,
+            ResponseChecksum, // high-integrity mode only
```
NickCraver, line 2112: `ResponseSequenceCheck`?

src/StackExchange.Redis/PublicAPI/PublicAPI.Unshipped.txt
```diff
@@ -1 +1,3 @@
-﻿
\ No newline at end of file
+﻿StackExchange.Redis.ConfigurationOptions.HighIntegrity.get -> bool
+StackExchange.Redis.ConfigurationOptions.HighIntegrity.set -> void
+virtual StackExchange.Redis.Configuration.DefaultOptionsProvider.HighIntegrity.get -> bool
```
NickCraver, line 3: Need to move to `PublicAPI.Shipped.txt`

tests/StackExchange.Redis.Tests/TestBase.cs
```diff
@@ -389,7 +394,7 @@ internal void ThrowIfBelowMinVersion(IInternalConnectionMultiplexer conn, Versio
         bool logTransactionData = true,
         int? defaultDatabase = null,
         BacklogPolicy? backlogPolicy = null,
-        RedisProtocol? protocol = null,
+        RedisProtocol? protocol = null, bool highIntegrity = false,
```
NickCraver, line 397:
```suggestion
        RedisProtocol? protocol = null,
        bool highIntegrity = false,
```

src/StackExchange.Redis/Message.cs
```diff
@@ -198,6 +201,18 @@ public bool IsAdmin
 
         public bool IsAsking => (Flags & AskingFlag) != 0;
 
+        public bool IsHighIntegrity => _highIntegrityChecksum != 0;
+        public int HighIntegrityChecksum => _highIntegrityChecksum;
+
+        internal void WithHighIntegrity(Random rng)
+        {
+            // create a required value if needed; avoid sentinel zero
+            while (_highIntegrityChecksum is 0)
```
NickCraver, line 210: If we wrap around, will need to handle 0 in the sequence version

NickCraver (commented):
Went through on call - notes inline

src/StackExchange.Redis/Message.cs
mgravell: switched

src/StackExchange.Redis/PhysicalConnection.cs
mgravell: we literally do exactly that already - `Message._highIntegrityToken` (note renamed slightly) - however, the problem is that *currently* the read loop only handles a single inbound at a time, and would require a significant refactor; the "move all the things" branch has scope to do better here

src/StackExchange.Redis/Message.cs
mgravell: done

src/StackExchange.Redis/PhysicalConnection.cs
mgravell: renamed

tests/BasicTest/BasicTest.csproj
```diff
@@ -6,7 +6,7 @@
     <AssemblyName>BasicTest</AssemblyName>
     <OutputType>Exe</OutputType>
     <PackageId>BasicTest</PackageId>
-    <RuntimeIdentifiers>win7-x64</RuntimeIdentifiers>
+    <!--<RuntimeIdentifiers>win7-x64</RuntimeIdentifiers>-->
```
philon-msft, line 9: Uncomment or remove?

tests/BasicTestBaseline/BasicTestBaseline.csproj
```diff
@@ -6,7 +6,7 @@
     <AssemblyName>BasicTestBaseline</AssemblyName>
     <OutputType>Exe</OutputType>
     <PackageId>BasicTestBaseline</PackageId>
-    <RuntimeIdentifiers>win7-x64</RuntimeIdentifiers>
+    <!--<RuntimeIdentifiers>win7-x64</RuntimeIdentifiers>-->
```
philon-msft, line 9: Uncomment or remove?

philon-msft (approved):

tests/BasicTestBaseline/BasicTestBaseline.csproj
mgravell: it isn't the core - I'm fine with leaving it for context

PR #2757, opened 2024-07-11
Project: Enable StyleCop and fix existing rules to make PRs easier going forward

NickCraver:
This adopts StyleCop and fixes most issues (and disables things we wouldn't want) to make PRs more consistent in an automated way and prevent formatting problems affecting git history, etc. It also does all the documentation enforcement.

Note: since I had to fix up many anyway, I finally did the `<inheritdoc />` minimization on `IDatabase`/`IDatabaseAsync` to remove 10% of duplicate documentation. That should also make PRs and maintenance less monotonous.


philon-msft (approved):

mgravell (approved):
merge hell, but necessary IMO

PR #2765, opened 2024-07-23
Fix Zero length check

### LINKED ISSUE #2764: Invalid Zero Length check expression
I noticed many invalid expression when checking for a zero elements in a collection.

Here is explanation

```
List<string> items = null;

if(items?.Count == 0)
{
    // This will evaluate false since null does not equals 0.
}

if(items?.Any() ==false)
{
    // This will evaluate false since items is null. which is translated to items != null && items.Count == 1
}
if(items == null || items.Count == 0)
{
    // This will evaluate true since items either null OR is not null and it's count is 0.
}

if(items?.Count > 0)
{
    // This check is okay because null items will return false and if the Count is 0 will also false.
}
```

MikeAlhayek:
Fix #2764

### DIFF
--- a/src/StackExchange.Redis/RedisKey.cs
+++ b/src/StackExchange.Redis/RedisKey.cs
@@ -13,7 +13,7 @@ namespace StackExchange.Redis
     {
         internal RedisKey(byte[]? keyPrefix, object? keyValue)
         {
-            KeyPrefix = keyPrefix?.Length == 0 ? null : keyPrefix;
+            KeyPrefix = keyPrefix == null || keyPrefix.Length == 0 ? null : keyPrefix;
             KeyValue = keyValue;
         }
 
@@ -253,7 +253,7 @@ namespace StackExchange.Redis
         /// Obtain the <see cref="RedisKey"/> as a <see cref="T:byte[]"/>.
         /// </summary>
         /// <param name="key">The key to get a byte array for.</param>
-        public static implicit operator byte[]? (RedisKey key)
+        public static implicit operator byte[]?(RedisKey key)
         {
             if (key.IsNull) return null;
             if (key.TryGetSimpleBuffer(out var arr)) return arr;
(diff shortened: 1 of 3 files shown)


mgravell:
Comments added in #2764; there is currently insufficient context as to whether there is an error here. The syntax used is not, by itself, either correct or incorrect. It is simply syntax, and has a current meaning, which this PR changes. Whether the original code, the proposed changed code, neither, or both; are correct - is a great question, but one that neither the issue nor the PR discusses. I suggest we start *there*.

src/StackExchange.Redis/RedisKey.cs
```diff
@@ -13,7 +13,7 @@ namespace StackExchange.Redis
     {
         internal RedisKey(byte[]? keyPrefix, object? keyValue)
         {
-            KeyPrefix = keyPrefix?.Length == 0 ? null : keyPrefix;
```
mgravell, line 16: Let's compare using:

``` c#
using System;

Original(null);
Original("");
Original("abc");

Proposed(null);
Proposed("");
Proposed("abc");

static void Original(string? keyPrefix) => Write(keyPrefix?.Length == 0 ? null : keyPrefix);
static void Proposed(string? keyPrefix) => Write(keyPrefix == null || keyPrefix.Length == 0 ? null : keyPrefix);

static void Write(string? value)
    => Console.WriteLine(value is null ? "(null)" : "\"" + value + "\"");
```

outputs:

keyPrefix input | original output | proposed output
---|---|---
null | null | null
"" | null | null
"abc" | "abc" | "abc"


So... this has no functional change

src/StackExchange.Redis/RedisResult.cs
```diff
@@ -401,7 +401,7 @@ internal override bool AsBoolean()
                     : Array.ConvertAll(_value, x => x.AsByteArray()!);
 
             private bool IsSingleton => _value?.Length == 1;
-            private bool IsEmpty => _value?.Length == 0;
```
mgravell, line 404: this changes the value of `IsEmpty` for `null` values of `_value`; currently a `null` will return `false` for `IsEmpty`; with this change it will return `true`. The question is what *should* `IsEmpty` return for `null` values? IMO we should retain the current behaviour, where `null` reports `false` here (note the separate `IsNull` test). Changing the result of this property is a significant proposition that would require justification.

mgravell:
Closing this; it appears to be part of a bulk creation of similar PRs on the assumption that `?.Length == 0` is automatically incorrect.

PR #2771, opened 2024-08-03
switch to docker compose

atakavci:
docker-compose is not available by default anymore on `ubuntu-latest`
[details here]( https://github.com/actions/runner-images/issues/9692)


NickCraver (approved):
Thanks!

Issue #2778, opened 2024-08-22
Invalid Server Selection after Redis Cluster Failover


leachdaniel-clark:
During maintenance we cycle our Redis Servers and failovers happen and cluster is becomes healthy again. However, even hours after doing an experiment where we rebooted all the servers, the `ConnectionMultiplexer` continues have issues with server section with a stack trace related to [this](https://github.com/StackExchange/StackExchange.Redis/blob/1c6b59e6c7751f03aec74679bb2f0d9123b74e3d/src/StackExchange.Redis/ConnectionMultiplexer.cs#L1967). The only way to resolve the problem is to restart the service or recreate multiplexers. We've tried the using the latest client version 2.8.0 with `HeartbeatConsistencyChecks` enabled.

We are not using an `CommandFlags` to override server selection and are using the method overload:
```csharp

KeyDeleteAsync(RedisKey[] keys, CommandFlags flags = CommandFlags.None);
```

Since Redis Cluster is not actually throwing this error the bug must be an issue with a mismatch in local state of this client triggered by server maintenance.

leachdaniel-clark:
Please disregard, I believe the problem is in the cluster itself

leachdaniel-clark:
related to #1874 "Seamlessly handling Elasticache (AWS)" failovers but not specific to AWS

leachdaniel-clark:
![image](https://github.com/user-attachments/assets/9a173c12-9686-45b7-9e59-aea125eb5db9)

PR #2779, opened 2024-08-24
AutoConfigure detects protocol failure HeartBeats allow ReplicaReconfig

leachdaniel-clark:
When connecting to a Redis Cluster and enabling the `HeartbeatConsistencyChecks` this prevents `CheckInfoReplication()` from ever being called and `ConfigCheckSeconds` is no longer respected.

This change will do the replication check but adds the `RecordConnectionFailed` like the `TracerProcessor` was doing for `HeartbeatConsistencyChecks`.

I haven't been able to run all the tests locally so I'm not sure if this change breaks anything, but was able to build the package and use it successfully.

The main thing I would be concerned about is if there are scenarios that AutoConfigureProcessor shouldn't parse anything, this would be recorded as a protocol failure.

This should fix the issue mentioned in #2778.

### DIFF
--- a/src/StackExchange.Redis/PhysicalBridge.cs
+++ b/src/StackExchange.Redis/PhysicalBridge.cs
@@ -610,19 +610,19 @@ namespace StackExchange.Redis
                                 checkConfigSeconds = ServerEndPoint.ConfigCheckSeconds;
 
                             if (state == (int)State.ConnectedEstablished && ConnectionType == ConnectionType.Interactive
-                                && tmp.BridgeCouldBeNull?.Multiplexer.RawConfig.HeartbeatConsistencyChecks == true)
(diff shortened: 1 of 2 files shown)


NickCraver:
We looked at this in the weekly call - I agree with the issue, but am very hesitant to change the result processor behavior. I've pitched an alternative solution in #2784.

The problem with failing on the result processor is that we get some invalid responses from some servers and proxies that today would fall back to using the configured state, but this would change that situation into looping on connection fails, so I'd rather not shift any on that front :)

NickCraver:
Closing this out as a PR, but will be fixed in a release we're cutting shortly.

Thanks a ton for raising this to our attention, and included you in the release notes on this as well (can remove if you'd rather not, but we're big on credit for things).

Cheers!

PR #2780, opened 2024-08-27
Update KeysScan.md

japjit:
Found through trial and error that not providing the database results in a timeout.

### DIFF
--- a/docs/KeysScan.md
+++ b/docs/KeysScan.md
@@ -43,7 +43,7 @@ Simple: start from a server, not a database.
 var server = conn.GetServer(someServer);
 
 // show all keys in database 0 that include "foo" in their name
-foreach(var key in server.Keys(pattern: "*foo*")) {
+foreach(var key in server.Keys(0, pattern: "*foo*")) { //if you dont provide the database the system can timeout.
     Console.WriteLine(key);
 }
 


mgravell:
Before we jump into changes, can we understand the problem? I cannot reproduce a problem here:

``` c#
using StackExchange.Redis;
using System;
using System.Linq;

var muxer = ConnectionMultiplexer.Connect("127.0.0.1:6379");

var db = muxer.GetDatabase();
db.StringSet("foo", "foo");
db.StringSet("bar", "bar");
var server = muxer.GetServers().Single();


Console.WriteLine("sync...");
foreach (var key in server.Keys(pattern: "*"))
{
    Console.WriteLine(key);
}

Console.WriteLine("async...");
await foreach (var key in server.KeysAsync(pattern: "*"))
{
    Console.WriteLine(key);
}
```

outputs

``` txt
sync...
bar
foo
async...
bar
foo
```

At the code level, [from here](https://github.com/StackExchange/StackExchange.Redis/blob/main/src/StackExchange.Redis/RedisServer.cs#L356), not supplying a database means the default [is applied](https://github.com/StackExchange/StackExchange.Redis/blob/e151cd5c8046bb511f144db1999d32b205fef20f/src/StackExchange.Redis/ConnectionMultiplexer.cs#L1106), which via `GetValueOrDefault()` means it should default to `0` if not already set (see also [here](https://github.com/StackExchange/StackExchange.Redis/blob/e151cd5c8046bb511f144db1999d32b205fef20f/src/StackExchange.Redis/ConfigurationOptions.cs#L447).

So: what is the context here? Can we try to get a repro? If there *is* a problem, I'd rather fix the code than the docs, since the docs aren't wrong. Anything relevant in the server you're connecting to here? or maybe (speculating) the problem is the pattern application itself. Note that redis *does not have* an efficient way of filtering keys by pattern - you're essentially walking the entire keyspace chunk by chunk, which can take time. If you're looking for something that has very few matches in a huge database, it can take a *lot* of round trips to find the first match, due to how `SCAN` works. You can trade reduced round-trips for increased server-time per trip by increasing the `pageSize` parameter (which defaults to `250`, but you can try `1000`, `5000`, `10000` etc to see how it changes things).

mgravell:
The other possible cause of a timeout here is if the server is low-version (or we cannot determine the version), and we therefore use `KEYS` instead of `SCAN`, and the database is large; this is massively disruptive to the server, which is why `KEYS` is largely considered deprecated with `SCAN` replacing it.

As a side note: both `KEYS` and `SCAN` should only really be used for administrative purposes like keyspace analysis - it should *not* be used for routine database operations. For any scenario where a user might use `KEYS`/`SCAN` for *routine* operations: there are usually *much* better options, that might require some minor rework.

japjit:
In this case I tried it with a timeout of 10000 and 30000, the minute I
gave it the default database nos it connected immediately.  Else it times
out. The Stackexchange Redis sdk version  is 2.8 running on windows 2022
server and Redis version is 7.2.5 running on wsl2. The documentation says
it automatically uses SCAN under the hood from what I understand.

Cheers, Japjit

mgravell:
Indeed it should; IMO the first thing we should do here is check what it *is* issuing; any chance you can use a `redis-cli` session to the same instance to issue the `MONITOR` command? this should allow us to see what is being issued when this happens; what I *expect* is to see some (potentially large) number of `SCAN` commands being issued; if it has failed to determine the server version, we might see a single `KEYS` command instead.

Additionally, it would be useful to see the result of `DBSIZE`, so we can see what sort of database we're dealing with here - is it 200 keys, or 200 million keys.

japjit:
My apologies, I withdraw my PR. I cant reproduce it now. After much ado I
went back and omitted the database parameter and ended up with getting the
data back. It must have been a networking glitch on my side as I was
running two parallel server. Sorry again.
Cheers, Japjit

japjit:
Just one final note what was interesting was that I could connect to
database.get but couldnt not to server.key till I gave it the default
database. Not sure why that was the case. After it started working it kept
working. Potential initialization issue? but obviously can't reproduce it
any more.
Cheers, Japjit

mgravell:
If this recurs, I'd be interested in helping investigate - I'm not saying "there was never a problem" - I'm simply saying "it almost certainly wasn't anything to do with the `database` argument". So: if it happens again, please do get in touch in an issue, so we can try to investigate.

NickCraver:
Closing this out for now to tidy up, but happy to follow-up if we see it again :)

PR #2787, opened 2024-09-04
Fix #2785: Support optionally disabling pipelining on AUTH flow

### LINKED ISSUE #2785: Pipelining causes NOAUTH errors with External Authentication on Envoy
Hello,

We have a situation where our Redis client is returning NOAUTH during the connection flow, but if we manually connect to it through _redis-cli_ and send an `AUTH <password>` command, from that exact same machine, we get OK and can send other commands after.

There is no place where we can see the actual result of the AUTH command sent by the SDK since, from what we've seen in code, it's a fire and forget message.
Any ideas how to debug this?

**Important Note**
We have multiple redis clusters that we're connecting to. If we put this cluster first, it works. If we put it last, it doesn't!

For example, this doesn't work:
```C#
ConnectionMultiplexer.Connect(options1);
...
ConnectionMultiplexer.Connect(options2);
...
ConnectionMultiplexer.Connect(options3);
...
ConnectionMultiplexer.Connect(options4);
```

But this does:
```C#
ConnectionMultiplexer.Connect(options4);
...
ConnectionMultiplexer.Connect(options2);
...
ConnectionMultiplexer.Connect(options3);
...
ConnectionMultiplexer.Connect(options1);
```

We're wondering if somehow the multiplexer might be using a wrong connection from a completely different instance. If there is some kind of static or thread-local state.

Some logs:

```
Connecting (sync) on .NET 8.0.8 (StackExchange.Redis: v2.8.0.27420)
<redacted>.internal.office.com:6380,syncTimeout=5000,connectTimeout=17000,abortConnect=False
<redacted>.internal.office.com:6380/Interactive: Connecting...
<redacted>.internal.office.com:6380: BeginConnectAsync
1 unique nodes specified (with tiebreaker)
<redacted>.internal.office.com:6380: OnConnectedAsync init (State=Connecting)
Allowing 1 endpoint(s) 00:00:17 to respond...
Awaiting 1 available task completion(s) for 17000ms, IOCP: (Busy=0,Free=1000,Min=1,Max=1000), WORKER: (Busy=0,Free=32767,Min=4,Max=32767), POOL: (Threads=6,QueuedItems=0,CompletedItems=404,Timers=28)
Configuring TLS
TLS connection established successfully using protocol: Tls13
<redacted>.internal.office.com:6380/Interactive: Connected
<redacted>.internal.office.com:6380: Server handshake
<redacted>.internal.office.com:6380: Authenticating (password)
<redacted>.internal.office.com:6380: Sending critical tracer (handshake): PING
<redacted>.internal.office.com:6380: Flushing outbound buffer
<redacted>.internal.office.com:6380: OnEstablishingAsync complete
<redacted>.internal.office.com:6380: Starting read
Response from <redacted>.internal.office.com:6380/Interactive / PING: Error: NOAUTH Authentication required.
<redacted>.internal.office.com:6380: OnConnectedAsync completed (Disconnected)
All 1 available tasks completed cleanly, IOCP: (Busy=0,Free=1000,Min=1,Max=1000), WORKER: (Busy=2,Free=32765,Min=4,Max=32767), POOL: (Threads=6,QueuedItems=0,CompletedItems=422,Timers=27)
Endpoint summary:
<redacted>.internal.office.com:6380: Endpoint is (Interactive: Connecting, Subscription: Disconnected)
Task summary:
<redacted>.internal.office.com:6380: Returned, but incorrectly
Endpoint Summary:
<redacted>.internal.office.com:6380: Envoyproxy v6.0.0, primary; 1 databases; keep-alive: 00:01:00; int: Connecting; sub: n/a; not in use: DidNotRespond
<redacted>.internal.office.com:6380: int ops=2, qu=0, qs=0, qc=0, wr=0, socks=2; sub ops=0, qu=0, qs=0, qc=0, wr=0
<redacted>.internal.office.com:6380: Circular op-count snapshot; int: 0+2=2 (0.20 ops/s; spans 10s); sub:
Sync timeouts: 0; async timeouts: 0; fire and forget: 0; last heartbeat: -1s ago
Resetting failing connections to retry...
AuthenticationFailure on <redacted>.internal.office.com:6380/Interactive, Flushed/ComputeResult, last: PING, origin: SetResult, outstanding: 0, last-read: 0s ago, last-write: 0s ago, keep-alive: 60s, state: ConnectedEstablishing, mgr: 9 of 10 available, last-heartbeat: never, last-mbeat: 0s ago, global: 0s ago, v: 2.8.0.27420
```

dbarbosapn:
resolves #2785 

This PR is (for now) an initial proposal for fixing the issue mentioned above.
With the new _external authentication_ feature of _Envoy_, pipelining AUTH with further commands can cause unexpected behavior where we see `+OK` for AUTH, followed by `NOAUTH` for the critical trace.

We fix this by adding an **optional** `ConfigurationOptions` flag that disables this behaviour.

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -100,6 +100,7 @@ The `ConfigurationOptions` object has a wide range of properties, all of which a
 | setlib={bool}          | `SetClientLibrary`     | `true`                       | Whether to attempt to use `CLIENT SETINFO` to set the library name/version on the connection              |
 | protocol={string}      | `Protocol`             | `null`                       | Redis protocol to use; see section below                                                                  |
 | highIntegrity={bool}   | `HighIntegrity`        | `false`                      | High integrity (incurs overhead) sequence checking on every command; see section below                    |
+| waitForAuth={bool}     | `WaitForAuth`          | `false`                      | Wait before the result of the `AUTH` command is returned before trying to send any other commands to the server |
 
 Additional code-only options:
 - LoggerFactory (`ILoggerFactory`) - Default: `null`
--- a/src/StackExchange.Redis/Configuration/DefaultOptionsProvider.cs
+++ b/src/StackExchange.Redis/Configuration/DefaultOptionsProvider.cs
@@ -116,6 +116,17 @@ namespace StackExchange.Redis.Configuration
         public virtual bool HighIntegrity => false;
 
         /// <summary>
+        /// A Boolean value that specifies whether the client should wait for the server to return
+        /// response for the initial AUTH command before trying any further commands.
+        /// </summary>
+        /// <remarks>
+        /// This is especially useful when connecting to Envoy proxies with external authentication
+        /// providers.
+        /// The default and recommended value is false.
+        /// </remarks>
+        public virtual bool WaitForAuth => false;
+
+        /// <summary>
         /// The number of times to repeat the initial connect cycle if no servers respond promptly.
         /// </summary>
         public virtual int ConnectRetry => 3;
--- a/src/StackExchange.Redis/ConfigurationOptions.cs
+++ b/src/StackExchange.Redis/ConfigurationOptions.cs
@@ -111,7 +111,8 @@ namespace StackExchange.Redis
                 Tunnel = "tunnel",
                 SetClientLibrary = "setlib",
                 Protocol = "protocol",
-                HighIntegrity = "highIntegrity";
+                HighIntegrity = "highIntegrity",
+                WaitForAuth = "true";
 
             private static readonly Dictionary<string, string> normalizedOptions = new[]
             {
@@ -143,6 +144,7 @@ namespace StackExchange.Redis
                 CheckCertificateRevocation,
                 Protocol,
                 HighIntegrity,
+                WaitForAuth,
             }.ToDictionary(x => x, StringComparer.OrdinalIgnoreCase);
 
             public static string TryNormalize(string value)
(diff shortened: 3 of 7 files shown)


mgravell:
So my random idea of a bug in envoy pipelining support was correct? If so, the "correct" fix here should be squarely inside envoy. I'll have a look at this and see if I can find any concerning side-effects, but: they could be subtle.

dbarbosapn:
> So my random idea of a bug in envoy pipelining support was correct? If so, the "correct" fix here should be squarely inside envoy. I'll have a look at this and see if I can find any concerning side-effects, but: they could be subtle.

Envoy supports pipelining but this new feature doesn't since it calls a gRPC dependency (asynchronous) to validate the AUTH result. Would you say it makes more sense to make such call synchronous on Envoy side? I imagine it would have more performance implications... I would guess it would be best for that to be on client-side.

WDYT?

mgravell:
It doesn't need to be synchronous per-se, but they can't go and process other commands while it is pending. If the problem is that envoy is implemented using thread-per-connection, then yes: it probably needs to be synchronous or effectively synchronous, since they can't meaningfully process anything else on that connection until the external result is known. So in that context: there's nothing to lose by making it synchronous (or in async terms; they need to `await` the gRPC call before processing anything else on that same connection).

dbarbosapn:
> It doesn't need to be synchronous per-se, but they can't go and process other commands while it is pending. If the problem is that envoy is implemented using thread-per-connection, then yes: it probably needs to be synchronous or effectively synchronous, since they can't meaningfully process anything else on that connection until the external result is known. So in that context: there's nothing to lose by making it synchronous (or in async terms; they need to `await` the gRPC call before processing anything else on that same connection).

Thank you! Will investigate on Envoy side and come back with a result on the issue ticket. I'll close the PR for now. Thanks!

PR #2813, opened 2024-10-31
format ipv6 endpoints correctly

### LINKED ISSUE #2811: ConfigurationOptions.ToString does not add brackets to IPv6 address
Hello!

I'm having an issue when calling `ConfigurationOptions.ToString` method, when my connection string has an IPv6 address.

Just for some context, I'm using Redis as a backplane for SignalR and I parse my original connection string to enhance it. Then, I convert it back to string to pass it to `RedisScaleoutConfiguration` (`Microsoft.AspNet.SignalR.StackExchangeRedis@2.4.3` - I'm using .Net Framework 4.8). 

That's why I need to use `ConfigurationOptions.ToString` method, and I'd expect my connection string IPv6 address to have brackets added to it.

The following example reproduces the behavior:

```csharp
 var connectionString = "[::1]:6379,password=mypass";
 var configuration = ConfigurationOptions.Parse(connectionString);
 configuration.AbortOnConnectFail = false;
 configuration.ClientName = "my-client-name";

 var converted = configuration.ToString();
 Console.WriteLine(converted); //expected [::1]:6379, actual: ::1:6379

 var endpoint = configuration.EndPoints.FirstOrDefault().ToString();
 Console.WriteLine(endpoint); //EndPoint.ToString method correctly adds the brackets
```

From my observation, the `ConfigurationOptions.ToString` method writes the endpoints using `Format` class, which [directly writes](https://github.com/StackExchange/StackExchange.Redis/blob/main/src/StackExchange.Redis/Format.cs#L101) the IP address without checking if that's an IPv6 address.

mgravell:
fixes #2811

### DIFF
--- a/src/StackExchange.Redis/Format.cs
+++ b/src/StackExchange.Redis/Format.cs
@@ -97,7 +97,20 @@ namespace StackExchange.Redis
                     if (dns.Port == 0) return dns.Host;
                     return dns.Host + ":" + Format.ToString(dns.Port);
                 case IPEndPoint ip:
-                    if (ip.Port == 0) return ip.Address.ToString();
+                    var addr = ip.Address.ToString();
+
+                    if (ip.Port == 0)
+                    {
+                        // no port specified; use naked IP
+                        return addr;
+                    }
+
+                    if (addr.IndexOf(':') >= 0)
+                    {
+                        // ipv6 with port; use "[IP]:port" notation
+                        return "[" + addr + "]:" + Format.ToString(ip.Port);
+                    }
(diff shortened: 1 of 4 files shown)


mgravell:
Mixed messages from CI:

```
Passed!  - Failed:     0, Passed:  1941, Skipped:  1082, Total:  3023, Duration: 1 m 3 s - StackExchange.Redis.Tests.dll (net6.0)
Error with tests, aborting build.
Command executed with exception
```

Will try to find which cog fell out

PR #2842, opened 2025-01-21
add minId overloads

### LINKED ISSUE #1718: Add support for stream XTRIM MINID command
From the redis command document, I learned about that we could use `XTRIM mystream MINID 649085820` to implement  all entries that have an ID lower than 649085820-0 will be evicted, so I'm wondering if we could add support for `MINID` subcommand for stream `XTRIM`.

Besides, I noticed that it seemed we could use `MINID` subcommand when we try to `XADD` new entry to a stream, maybe we could have it implemented?

- https://redis.io/commands/xadd
- https://redis.io/commands/xtrim

kijanawoodard:
Fixes #1718

Initial commit. Still orienting on how to contribute.
It's building. Tests pass. 

Before I continue, does the diff look sane?

I think I got the "public api" situated correctly.

The strangest thing I ran into was this error about default parameters. I'm not sure if this is pushing me towards adding a new method rather than an overload, e.g. StreamTrimWithMinId. I took out the default CommandFlags to build successfully.

`error RS0026: Symbol 'StreamTrim' violates the backcompat requirement: 'Do not add multiple overloads with optional parameters'`

I'm having trouble running the tests locally. I tried running the tests before making any changes and ~2/3 passed. It might be because I don't have all the versions of .net installed.

### DIFF
--- a/src/StackExchange.Redis/Interfaces/IDatabase.cs
+++ b/src/StackExchange.Redis/Interfaces/IDatabase.cs
@@ -2780,10 +2780,12 @@ namespace StackExchange.Redis
         /// </summary>
         /// <param name="key">The key of the stream.</param>
         /// <param name="minId">All entries with an id (timestamp) earlier minId will be removed.</param>
+        /// <param name="useApproximateMaxLength">If true, the "~" argument is used to allow the stream to exceed minId by a small number. This improves performance when removing messages.</param>
+        /// <param name="limit">The maximum number of entries to remove per call when useApproximateMaxLength = true. If 0, the limiting mechanism is disabled entirely.</param>
         /// <param name="flags">The flags to use for this operation.</param>
         /// <returns>The number of messages removed from the stream.</returns>
         /// <remarks><seealso href="https://redis.io/topics/streams-intro"/></remarks>
-        long StreamTrim(RedisKey key, RedisValue minId, CommandFlags flags);
+        long StreamTrimByMinId(RedisKey key, RedisValue minId, bool useApproximateMaxLength = false, int? limit = null, CommandFlags flags = CommandFlags.None);
 
         /// <summary>
         /// If key already exists and is a string, this command appends the value at the end of the string.
--- a/src/StackExchange.Redis/Interfaces/IDatabase.cs
+++ b/src/StackExchange.Redis/Interfaces/IDatabase.cs
@@ -2776,6 +2776,16 @@ namespace StackExchange.Redis
         long StreamTrim(RedisKey key, int maxLength, bool useApproximateMaxLength = false, CommandFlags flags = CommandFlags.None);
 
         /// <summary>
+        /// Trim the stream to a specified minimum timestamp.
+        /// </summary>
+        /// <param name="key">The key of the stream.</param>
+        /// <param name="minId">All entries less than minId will be removed.</param>
+        /// <param name="flags">The flags to use for this operation.</param>
+        /// <returns>The number of messages removed from the stream.</returns>
+        /// <remarks><seealso href="https://redis.io/topics/streams-intro"/></remarks>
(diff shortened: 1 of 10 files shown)


kijanawoodard:
I can't tell what's wrong in the appveyor build. It says zero tests failed, but still errors.

kijanawoodard:
I was able to get devcontainer working on WSL enough to build, but the tests can't connect to redis in the devcontainer.

I'll post my changes in case anyone else can follow the breadcrumbs.

I needed to enable Ubuntu integration in docker desktop
image

I had to modify docker-compose.yml to point redis back to the dockerfile. I tried running the dockerfile(s) under tests, but that didn't work at all. `dockerfile: ../../.devcontainer/Dockerfile`
image

I added dotnet 8 installation to the Dockerfile under .devcontainer
RUN curl -sSL https://dot.net/v1/dotnet-install.sh | bash /dev/stdin --channel 8.0 --install-dir /usr/share/dotnet/

kijanawoodard:
I wasn't able to get dev containers running. The dockerfile in .devcontainer refers to a Dockerfile in `tests/RedisConfigs` that doesn't exist. I tried using the other ones, but they appear to depend on the `docker-compose.yml` file in the `tests/RedisConfigs` folder.

I was able to run the tests with only one failure: `FailoverTests.SubscriptionsSurviveConnectionFailureAsync(RESP2) [FAIL]`

I got that far by cloning the project in wsl2, running `docker compose up` in `tests/RedisConfigs` and then running `dotnet test StackExchange.Redis.sln`  from the project root.

kijanawoodard:
I was wrong about appveyor having no test failures. There were 3 failures that don't seem to be related to my changes here. If there are known test issues, let me know.

kijanawoodard:
@NickCraver @mgravell sorry to @ you guys, but I want to make sure I haven't missed some step in the PR process. I looked for a contributing doc, but didn't see one. Is there anything else I need to do at the moment, or is it fine to just let this PR sit tight?

WeihanLi:
There's a failed test case, but think it's not related to the changes
image

Maybe add an entry for the release note

https://github.com/StackExchange/StackExchange.Redis/blob/main/docs/ReleaseNotes.md

mgravell:
I very much agree with you about the docs being less than clear on this detail.

kijanawoodard:
The original comment for the MINID PR also brings in LIMIT to work with ~. 
Seems definitive to me as it appears, from reading the entire thread, that the top comment was updated as they made decisions on the PR.

https://github.com/redis/redis/pull/8169

kijanawoodard:
> the API shape gets tricky, because there is an inherent ambiguity since RedisValue is convertable from int; I wonder whether an explicit API name might be preferable, for example StreamTrimByMinId

I agree. An explicit API makes sense.

kijanawoodard:
Side note: more XTRIM options were added a couple days ago:
https://github.com/redis/redis/pull/14130

kijanawoodard:
@mgravell 

I changed the method name to StreamTrimByMinId and implemented useApproximateMaxLength and limit.
From the docs, limit is only in play when useApproximateMaxLength = true. (Noted in method documentation)
I made limit nullable because Zero means "no limits". (Noted in method documentation)

The test with useApproximateMaxLength sets the limit to 100. Less than 100, deleted zero items.
I didn't do a test without a limit since I think that's non-deterministic.

Hmm. Writing this up here, useApproximateMaxLength is tied to the MaxLength version of XTRim. Maybe I should rename?

Additional questions:
0. Rename useApproximateMaxLength for the MINID overload?
1. The original purpose of LIMIT was to action on the MAXLENGTH version as well. Implement that here or do that in another PR?
2. Should we throw if the limit is set without setting useApproximateMaxLength since it will have no effect or should we let the documentation suffice?

mgravell:
Looking at this now; there are some additional things proposed in 8.2 that I'm looking at

Issue #2843, opened 2025-01-22
Memory optimization for ScriptEvaluate


pairbit:
Please add method overloads.
This will allow using ArrayPool<RedisValue>.Rent and reuse buffers

```csharp
Task<RedisResult> ScriptEvaluateMemoryAsync(string script, ReadOnlyMemory<RedisKey> keys, ReadOnlyMemory<RedisValue> values, 
CommandFlags flags = CommandFlags.None)

Task<Lease<byte>?> ScriptEvaluateMemoryLeaseAsync(string script, ReadOnlyMemory<RedisKey> keys, ReadOnlyMemory<RedisValue> values, 
CommandFlags flags = CommandFlags.None)
```

atakavci:
hey @mgravell , @NickCraver , would you like to see a PR for this ?? i can go for it.

PR #2844, opened 2025-01-28
Add Lease and non-boxing versions of Execute(Async)

kevin-montrose:
I've got a scenario where I need to do a _ton_ of `.Execute(Async)` calls, and the lack of ways to avoid allocations are hurting.

So I've added some.

Added methods (generally on both `IServer` and `IDatabase(Async)`, as paired with existing `Execute(Async)` methods):
 - `ExecuteLease(Async)`
   * Returns a `Lease<byte>?` (or `Task<Lease<byte>?>`) - documented as only allowing simple string, bulk string, and integer returns per-`LeaseProcessor` behavior
 - `ExecuteLeaseExplicit(Async)`
   * Identical to the non-`params` `ExcuteLease(Async)` methods except rather than taking an `ICollection<object>` it takes an `ICollection<RedisKey>` and an `ICollection<RedisValue>`
   * I introduced a new `Message` accordingly
   * The `RedisKey` values are not passed, they are only used to determine slots when cluster is enabled
   * Because I cannot identify the key values in args, the key prefixed classes throw `NotSupportException`s if `ExecuteLeaseExplicit(Async)` methods are called
   
I'm not married to these exact methods, but they seem the simplest way to get Execute-ish functionality without boxing and while reusing result allocations.


mgravell:
Hey Monty; what's your timescales? Basically, I have a side branch that completely rewrites the fundamental writer/parser core, and one of the secondary aims is to enable zero-alloc custom commands from external callers.

Assuming you're coming at this with your work hat on, we could perhaps use yourself as a 1st party client to give me an excuse to merge that work on company time.

kevin-montrose:
> Hey Monty; what's your timescales? Basically, I have a side branch that completely rewrites the fundamental writer/parser core, and one of the secondary aims is to enable zero-alloc custom commands from external callers.
> 
> Assuming you're coming at this with your work hat on, we could perhaps use yourself as a 1st party client to give me an excuse to merge that work on company time.

@mgravell It's short but not imminent - second week of February would be ideal.

To expand on the use case, this is a benchmarking tool.  I suspect the functionality that is being Execute'd is already fast enough, but cannot prove it.  Based on profiling, I believe my extra tail latency is allocation / GC related.

kevin-montrose:
{5209E2F6-F36C-4B20-BD83-FCFE5C45D591}
I thought we were friends @NickCraver .

mgravell:
Considerations:

0. this overlaps hugely with some separate work we have planned re the IO core; I propose we hold this PR for now, pending on that, which we can explore directly
1. on the API: the keys/values thing (to avoid boxing) is confusing and does not directly map to the underlying API - and in particular is so similar in shape, but different in operation, to ScriptEvaluate, that it could cause significant problems; this also makes prefixing untenable (as we don't know which args are keys)
2. we would prefer `ReadOnlyMemory<T>` over `ICollection<T>` (for example, like #2843)
3. your scenario seems to be perf critical, making the separate work even more relevant

We're discussing this directly (off-GitHub), and will keep doing so.

mgravell:
Fixed via a separate API (`ExecuteResp`) in 3.2.0, and even more so in v4 (PoC only)

PR #2860, opened 2025-03-14
Subscribe to all primaries (for key-space-notifications)

### LINKED ISSUE #2846: subscribe to redis cluster
I deployed redis cluster with following command

```
helm install my-release oci://registry-1.docker.io/bitnamicharts/redis-cluster
```

helm created 6 nodes: 3 masters and 3 slaves

I activated notifications on EACH pod
```
D:\src\red>kubectl exec -ti my-red-redis-cluster-1 -- /bin/bash
I have no name!@my-red-redis-cluster-1:/$ redis-cli config set notify-keyspace-events AKEnm
OK
```

Then I try to subscribe to key updates. For each i in 0..5 I did

``` C#
    var redis = await ConnectionMultiplexer.ConnectAsync(
        $"my-red-redis-cluster-{i}.my-red-redis-cluster-headless.default.svc.cluster.local:6379,password={password}");
    var sub = redis.GetSubscriber();
    var channel = new RedisChannel("__key*@*__:*", RedisChannel.PatternMode.Pattern);
    await sub.SubscribeAsync(channel, Accept);
    Console.WriteLine($"Subscribed to channel {i} - {sub.SubscribedEndpoint(channel)}");
```

I see it connects to the same server for any "i" value

```
Subscribed to channel 4 - 10.1.37.157:6379
Subscribed to channel 0 - 10.1.37.157:6379
Subscribed to channel 2 - 10.1.37.157:6379
Subscribed to channel 3 - 10.1.37.157:6379
Subscribed to channel 5 - 10.1.37.157:6379
Subscribed to channel 1 - 10.1.37.157:6379
```

And then I receive only ~ 33% of notifications.

if I subscribe using redis-cli on each server - I receive all notifications too.

What should I do to receive all notifications in C# program?

atakavci:
Closes/Fixes #2846 

This approach essentially uses the same methodology as pub/sub subscriptions but runs on all primary nodes in the cluster, attempting to combine the results into a single output.

While some parts of it align with existing pub/sub functionality, making them reusable compromises readability. Therefore, I'm not sure it's preferable to make them reusable.

Another point is that I went back and forth on API naming, trying to decide whether to reference the keyspace or not. To provide more clarity on what is happening, I chose a name that refers to the primaries, and I’ve kept the keyspace-related explanations in the documentation.

### DIFF
--- a/src/StackExchange.Redis/Interfaces/ISubscriber.cs
+++ b/src/StackExchange.Redis/Interfaces/ISubscriber.cs
@@ -62,6 +62,39 @@ namespace StackExchange.Redis
         Task SubscribeAsync(RedisChannel channel, Action<RedisChannel, RedisValue> handler, CommandFlags flags = CommandFlags.None);
 
         /// <summary>
+        /// Subscribe to perform some operation when a message to any primary node is broadcast, without any guarantee of ordered handling.
+        /// This is most useful when addressing key space notifications. For user controlled pub/sub channels, it should be used with caution and further it is not advised.
+        /// </summary>
+        /// <param name="channel">The channel to subscribe to.</param>
+        /// <param name="handler">The handler to invoke when a message is received on <paramref name="channel"/>.</param>
+        /// <param name="flags">The command flags to use.</param>
+        /// <remarks>
+        /// See
+        /// <seealso href="https://redis.io/docs/latest/develop/use/keyspace-notifications/"/>,
+        /// <seealso href="https://redis.io/commands/subscribe"/>,
+        /// <seealso href="https://redis.io/commands/psubscribe"/>.
+        /// </remarks>
+        Task SubscribeAllPrimariesAsync(RedisChannel channel, Action<RedisChannel, RedisValue> handler, CommandFlags flags = CommandFlags.None);
+
+        /// <summary>
+        /// Unsubscribe from a specified message channel on all primary nodes.
+        /// Note: if no handler is specified, the subscription is canceled regardless of the subscribers.
+        /// If a handler is specified, the subscription is only canceled if this handler is the last handler remaining against the channel.
+        /// This is used in combination with <see cref="SubscribeAllPrimariesAsync"/> and as mentioned there,
+        /// it is intended to use with key space notitications and not advised for user controlled pub/sub channels.
+        /// </summary>
+        /// <param name="channel">The channel that was subscribed to.</param>
+        /// <param name="handler">The handler to no longer invoke when a message is received on <paramref name="channel"/>.</param>
+        /// <param name="flags">The command flags to use.</param>
+        /// <remarks>
+        /// See
+        /// <seealso href="https://redis.io/docs/latest/develop/use/keyspace-notifications/"/>,
+        /// <seealso href="https://redis.io/commands/unsubscribe"/>,
+        /// <seealso href="https://redis.io/commands/punsubscribe"/>.
+        /// </remarks>
+        Task UnsubscribeAllPrimariesAsync(RedisChannel channel, Action<RedisChannel, RedisValue>? handler = null, CommandFlags flags = CommandFlags.None);
+
+        /// <summary>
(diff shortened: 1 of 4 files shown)


src/StackExchange.Redis/Interfaces/ISubscriber.cs
```diff
@@ -61,6 +61,39 @@ public interface ISubscriber : IRedis
...
+        /// <param name="channel">The channel to subscribe to.</param>
+        /// <param name="handler">The handler to invoke when a message is received on <paramref name="channel"/>.</param>
+        /// <param name="flags">The command flags to use.</param>
+        /// <remarks>
+        /// See
+        /// <seealso href="https://redis.io/docs/latest/develop/use/keyspace-notifications/"/>,
+        /// <seealso href="https://redis.io/commands/subscribe"/>,
+        /// <seealso href="https://redis.io/commands/psubscribe"/>.
+        /// </remarks>
+        Task SubscribeAllPrimariesAsync(RedisChannel channel, Action<RedisChannel, RedisValue> handler, CommandFlags flags = CommandFlags.None);
```
mgravell, line 77: I wonder if we should infer "all primaries" (vs secondaries) from `flags`, like we would with regular subscribe, so this becomes `SubscribeAll`; I also wonder whether we should internalize this entirely, specifically against the keyspace events, because for regular keys:

- non-'s' variants will get broadcast duplicates that will be hard to untangle
- `s` variants *don't make sense* to attach to the entire cluster, since they are routed using shard rules

That would also have the advantage of not needing a separate API. I'm not certain on this - mostly thinking aloud here.

src/StackExchange.Redis/RedisSubscriber.cs
```diff
@@ -432,6 +434,86 @@ public Task<bool> SubscribeAsync(RedisChannel channel, Action<RedisChannel, Redi
...
+                }
+            }
+
+            if (tasks.Count == 0)
+            {
+                return CompletedTask<bool>.Default(false);
+            }
+
+            // Create a new task that will collect all results and observe errors
+            return Task.Run(async () =>
```
mgravell, line 510: I *suspect* we can unwrap a little of the extra task layer here, but I'm not too concerned - this is low throughput

atakavci:
@mgravell ,

>   -  non-'s' variants will get broadcast duplicates that will be hard to untangle
>   -  s variants don't make sense to attach to the entire cluster, since they are routed using shard rules

this makes total sense and my attempt is solely for the sake of `key-space-notifications`.
alternatively i was also considering about a special kind of `RedisChannel` which is only targeting `key-space-notifications` by design.

src/StackExchange.Redis/RedisSubscriber.cs
```diff
@@ -432,6 +434,86 @@ public Task<bool> SubscribeAsync(RedisChannel channel, Action<RedisChannel, Redi
...
+        {
+            ThrowIfNull(channel);
+            if (handler == null && queue == null) { return CompletedTask<bool>.Default(null); }
+
+            var sub = multiplexer.GetOrAddSubscription(channel, flags);
+            sub.Add(handler, queue);
+            return EnsureSubscribedToPrimariesAsync(sub, channel, flags, false);
+        }
+
+        private Task<bool> EnsureSubscribedToPrimariesAsync(Subscription sub, RedisChannel channel, CommandFlags flags, bool internalCall)
```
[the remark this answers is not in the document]
atakavci: oww, now i see.. 
we are missing the cases with failovers and etc..
i ll try to explore the cases. thanks.

src/StackExchange.Redis/Interfaces/ISubscriber.cs
JoelHdzF: I'll like the idea to having just a single Subscribe call that will identify if the Channel is a keyspace event or a regular Pub/Sub channel, that way the user doesn't have to handle the possible duplication of "regular" events and also doesn't have to manage to Connect to every single Node, which also agree we could say via flags whether to connect to master or replicas, or all if the user still wants that.

PR #2873, opened 2025-04-12
Make it easier to use user certificate files

mgravell:
Historically, certs only worked with PFX files, which made it a pain to work with the cert pairs typically downloaded from hosts like Redis Ltd; however, on more recent .NET versions PEM is fully available. Here, we:

- add new methods on `ConfigurationOptions` (akin to the existing `TrustIssuer` method) to configure certificates from files
- support PEM in the ambient setup

There are no tests added here, due to the inherent problems of CI talking to such servers; however:

``` c#
 var options = ConfigurationOptions.Parse("redis-redacted.redislabs.com:4242"); // port also redacted
 options.TrustIssuer(@"C:\Code\RedisKeys\redis_ca.pem");
 options.SetUserPemCertificate(@"C:\Code\RedisKeys\redis-redacted.crt", @"C:\Code\RedisKeys\redis-redacted.key");
 using var conn = await ConnectionMultiplexer.ConnectAsync(options);
 for (int i = 0; i < 10; i++)
 {
     var ttl = await conn.GetDatabase().PingAsync();
     Console.WriteLine(ttl);
 }
```

works with the files downloaded from my hosted Redis Ltd endpoint:

image

To @philon-msft : open question: can we check Azure Redis with similar?

^^^ update: Philo assures me that Azure Redis doesn't have any relevant client-cert scenarios to consider

### DIFF
--- a/src/StackExchange.Redis/PhysicalConnection.cs
+++ b/src/StackExchange.Redis/PhysicalConnection.cs
@@ -1524,10 +1524,7 @@ namespace StackExchange.Redis
                 if (!string.IsNullOrEmpty(certificatePath) && File.Exists(certificatePath))
                 {
                     var passwordPath = Environment.GetEnvironmentVariable("SERedis_ClientCertPasswordPath");
-                    if (!string.IsNullOrEmpty(passwordPath) && File.Exists(passwordPath))
-                    {
-                        return ConfigurationOptions.CreatePemUserCertificateCallback(certificatePath, passwordPath);
-                    }
+                    return ConfigurationOptions.CreatePemUserCertificateCallback(certificatePath, passwordPath);
                 }
 #endif
             }
--- a/src/StackExchange.Redis/PhysicalConnection.cs
+++ b/src/StackExchange.Redis/PhysicalConnection.cs
@@ -1504,21 +1504,32 @@ namespace StackExchange.Redis
         {
             try
             {
-                var pfxPath = Environment.GetEnvironmentVariable("SERedis_ClientCertPfxPath");
-                var pfxPassword = Environment.GetEnvironmentVariable("SERedis_ClientCertPassword");
-                var pfxStorageFlags = Environment.GetEnvironmentVariable("SERedis_ClientCertStorageFlags");
-
-                X509KeyStorageFlags? flags = null;
-                if (!string.IsNullOrEmpty(pfxStorageFlags))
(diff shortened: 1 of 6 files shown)


mgravell:
In particular: contrast this to the recommended approach currently on my hosted Redis Ltd instance:

```
 configurationOptions.CertificateSelection += delegate
        {
            return new X509Certificate2(
                "<path_to_redis.pfx_file>",
                "<pfx_file_password>"
            );
        };
```

Which a: requires additional manual steps to export the pem/key file to pfx, which *isn't trivial*, and b: is more additional work. Sure: if you *want* more control, it is great to have the full options, but most people will just have the file pair.

mgravell:
also /cc @atakavci to a: see if there's any feedback from the Redis side, and b: (future, obvs) see if we can get updating the cloud console updated to suggest this approach.

src/StackExchange.Redis/PhysicalConnection.cs
```diff
@@ -1504,21 +1504,29 @@ public ConnectionStatus GetStatus()
...
+                if (!string.IsNullOrEmpty(certificatePath) && File.Exists(certificatePath))
                 {
-                    flags = Enum.Parse(typeof(X509KeyStorageFlags), pfxStorageFlags) as X509KeyStorageFlags?;
+                    var password = Environment.GetEnvironmentVariable("SERedis_ClientCertPassword");
+                    var pfxStorageFlags = Environment.GetEnvironmentVariable("SERedis_ClientCertStorageFlags");
+                    X509KeyStorageFlags storageFlags = X509KeyStorageFlags.DefaultKeySet;
+                    if (!string.IsNullOrEmpty(pfxStorageFlags))
+                    {
+                        var tmp = Enum.Parse(typeof(X509KeyStorageFlags), pfxStorageFlags) as X509KeyStorageFlags?;
+                        if (tmp is not null) storageFlags = tmp.GetValueOrDefault();
```
NickCraver, line 1516: Could use .TryParse a bit simpler here

NickCraver (approved):
One quick comment - TODOs: get test server setup and docs here as follow-ups

src/StackExchange.Redis/PhysicalConnection.cs
mgravell: done: https://github.com/StackExchange/StackExchange.Redis/pull/2873/commits/7a9208f432841c5f6e9380a007e41c6554ae5098

PR #2887, opened 2025-05-20
Support sharded pubsub commands

mgravell:
Continuation of https://github.com/StackExchange/StackExchange.Redis/pull/2498/

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -9,6 +9,8 @@ Current package versions:
 ## Unreleased
 No pending unreleased changes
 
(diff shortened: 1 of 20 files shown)


mgravell:
important TODOs before final merge:

- [ ] investigate break/reconnect logic
- [ ] investigate cluster shard migration logic
- [ ] investigate unsubscribe-all logic

atakavci:
hey @mgravell , i think [this branch](https://github.com/atakavci/StackExchange.Redis/tree/ali/ssubscribe) fixes and tests reconnect logic as well as testing for hashslot migrations.

the only item remaining is unsubscribe-all logic; what should be the expected behaviour with unsubscribe-all??

- [x]  investigate break/reconnect logic
- [x]   investigate cluster shard migration logic 
- [ ]  investigate unsubscribe-all logic

Issue #2899, opened 2025-06-23
Use logging best practices for core logging messages


Meir017:
To be more memory / performance efficient we should use https://learn.microsoft.com/en-us/dotnet/core/extensions/logger-message-generator when logging from the redis-client internals.

similar to https://github.com/dotnet/orleans/issues/9273

I'll be happy to contribute this

PR #2903, opened 2025-06-29
Replaces inline logging with source-generated LoggerMessage

Meir017:
related #2899

Adopts source-generated LoggerMessage for all logging calls, replacing direct string interpolation and formatting in log messages. Introduces strongly-typed helper structs for efficient log value formatting and factors out complex ToString() calls.

Improves logging performance, ensures consistency across log entries, and reduces unnecessary allocations by deferring message formatting until required. Lays groundwork for easier log message maintenance and better structured logging.

### DIFF
--- a/src/StackExchange.Redis/ConnectionMultiplexer.cs
+++ b/src/StackExchange.Redis/ConnectionMultiplexer.cs
@@ -170,10 +170,10 @@ namespace StackExchange.Redis
                         lock (log) // Keep the outer and any inner errors contiguous
                         {
                             var ex = a.Exception;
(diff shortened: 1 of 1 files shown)


Meir017:
a have a question here:

The `ILogger` in the `ConnectionMultiplexer` class is nullable, this doesn't play well with the logging source-generator, would it make sense to make it none-null (default to `NullLogger<ConnectionMultiplexer>.Instance` or modify the `configuration.LoggerFactory` to default to the null factory)

mgravell:
Additional change / observation: the generated code is (for me) generating errors because of a missing null-check:

image

I'll work around that

PR #2909, opened 2025-07-16
Fix CLIENT ID error during handshake

mgravell:
This is odd; no idea how this hasn't already flagged, but: the `CLIENT ID` issued in handshake *breaks connections* - the auto-config handler didn't report success correctly, leading it to manifest as a protocol failure.

This is probably urgent.


philon-msft (approved):

philon-msft:
Maybe this wasn't flagged before because we execute CLEINT ID as fire-and-forget during handshake? Absence of ConnectionId should definitely cause tests to fail, but I don't see any library code that has a hard dependency on it. Are we TOO tolerant of protocol failures?

mgravell:
>  Are we TOO tolerant of protocol failures?

We're not even slightly tolerant of them. In my local testing, it was burning the connection constantly. Protocol failures are deemed fatal.

PR #2915, opened 2025-07-20
Make private & internal stuff sealed

Henr1k80:


Henr1k80:
It results in less IL, as the virtual dispatch is no longer needed, and makes the functions candidates for inlining:
https://github.com/dotnet/runtime/issues/49944

The exception is .NET Framework, where it has no effect.

I admit that I have not done any performance analysis of this, but the motivation is correctness and presumed that any performance gains would be welcome, albeit probably hardly measurable.  
The `sealed` keyword is already in heavy use elsewhere, so it can as well be streamlined, maybe even consider adding [CA1852](https://learn.microsoft.com/en-gb/dotnet/fundamentals/code-analysis/quality-rules/ca1852)

Henr1k80:
One more thing, it removes the `jmp`, where the branch predictor has to guess where to load the next instructions from.  
If there is a mispredict, the speculative loaded code & data must be discarded and correct code & data loaded, a pipeline stall.

Branch predictors are very good, but by removing the `jmp`, there is one less branch to track.  
Branches to track is not only in this code, but in the entire application using this library.  
The more the branch predictor has to keep track of, the worse the performance: https://blog.cloudflare.com/branch-predictor/

mgravell:
I'm very aware of the *potential* advantages of devirtualization; however, whether such things are useful is *hugely* contextual. Yes, there are some tight loop CPU bound scenarios that might be meaningful *in hot code paths*, but unless I'm very mistaken, none of that is remotely relevant to any of the situations presented. When suggested arbitrarily without concrete *applicable* reasons, this ... just seems ... [cargo cult programming](https://en.wikipedia.org/wiki/Cargo_cult_programming). I don't mean to be dismissive or disparaging - but I'm also not a fan of changes without demonstrable reasons.

Henr1k80:
I think the comprehensional cost of adding the `sealed` keyword is minimal.  
So little, that it is not worth proving that it makes a difference.  
The IL is smaller, has no branches and makes it possible to inline. Do it always, no discussion needed.  
It is not like I am introducing mediator pattern or something, there is not added any new code.

If just a single virtual dispatch can be avoided for a call to redis, it is worth it, to avoid polluting the branch predictors & caches with unneeded stuff.  
Scale it with the popularity of this library worldwide.  
I expect the libraries I use to be as optimal as can be.

mgravell (approved):

WeihanLi:
maybe better to enable the analyzer on editorconfig also

PR #2922, opened 2025-07-21
Make ConfigurationOptions IDisposable

philon-msft:
Creating this PR for discussion. 

The main advantage of IDisposable ConfigurationOptions is it enables graceful shutdown of IDisposable default options providers like AzureCacheOptionsProviderWithToken, which has a token refresh Timer that should be stopped.

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -8,7 +8,7 @@ Current package versions:
 
 ## Unreleased
 
-- nothing yet
+- Make ConfigurationOptions IDisposable ([#2922 by philon-msft](https://github.com/StackExchange/StackExchange.Redis/pull/2922))
 
 ## 2.8.58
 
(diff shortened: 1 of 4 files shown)


mgravell:
The fact that the options and defaults provider often/usually have very different lifetimes makes this... awkward. It is *not* usually the cast that the two are linked, and *most* of the time this would be unnecessary and/or incorrect - making busy-work without need. I wonder whether there's some new wrapper API that we'd need instead, i.e.

```
public readonly struct NewWrapperNamingIsHard : IDisposable
{
    // wraps options+defaults provider

    public ConfigurationOptions Options => ...
    public static implicit operator ConfigurationOptions(NamingIsHard value) => value.Options;
}
```

so some API would return a `NewWrapperNamingIsHard`

philon-msft:
Closing this. We'll use a different approach that relies on a new `AfterDisconnectAsync()` callback on `DefaultOptionsProvider` to shut down custom DefaultOptionsProviders when zero connections are using their `ConfigurationOptions`.

PR #2924, opened 2025-07-23
chore: Replaces string interpolation with structured logging - ServerEndPoint

Meir017:
related #2899 

Migrates Redis server endpoint logging from string interpolation to LoggerMessage attributes for better performance and structured logging support.

Adds 14 new LoggerMessage methods for server connection lifecycle events including handshake, authentication, configuration, and buffer operations.

Uses ServerEndPointLogValue wrapper to ensure consistent server representation in logs while maintaining type safety.

github-copilot prompt used:
image

### DIFF
--- a/src/StackExchange.Redis/ServerEndPoint.cs
+++ b/src/StackExchange.Redis/ServerEndPoint.cs
@@ -458,7 +458,7 @@ namespace StackExchange.Redis
             // But if GETs are disabled on this, do not fail the connection - we just don't get tiebreaker benefits
             if (Multiplexer.RawConfig.TryGetTieBreaker(out var tieBreakerKey) && Multiplexer.CommandMap.IsAvailable(RedisCommand.GET))
             {
-                log?.LogInformationRequestingTieBreak(new LoggerExtensions.EndPointLogValue(EndPoint), tieBreakerKey.ToString());
+                log?.LogInformationRequestingTieBreak(new(EndPoint), tieBreakerKey);
                 msg = Message.Create(0, flags, RedisCommand.GET, tieBreakerKey);
                 msg.SetInternalCall();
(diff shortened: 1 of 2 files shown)


src/StackExchange.Redis/ServerEndPoint.cs
```diff
@@ -458,7 +458,7 @@ internal async Task AutoConfigureAsync(PhysicalConnection? connection, ILogger?
             // But if GETs are disabled on this, do not fail the connection - we just don't get tiebreaker benefits
             if (Multiplexer.RawConfig.TryGetTieBreaker(out var tieBreakerKey) && Multiplexer.CommandMap.IsAvailable(RedisCommand.GET))
             {
-                log?.LogInformation($"{Format.ToString(EndPoint)}: Requesting tie-break (Key=\"{tieBreakerKey}\")...");
+                log?.LogInformationRequestingTieBreak(new LoggerExtensions.EndPointLogValue(EndPoint), tieBreakerKey.ToString());
```
mgravell, line 461: can this not be target-typed, i.e. `new(EndPoint)` ? also: might need `tieBreakerKey?.ToString()` and `string?` in the receiver

mgravell (changes requested):
tie-breaker NRE?
target-type `new(...)`?

if neither make sense: let me know

src/StackExchange.Redis/ServerEndPoint.cs
Meir017: we enter this code block only if `Multiplexer.RawConfig.TryGetTieBreaker(out var tieBreakerKey)` is true, I think the `tieBreakerKey` is not null here

Meir017:
@mgravell ready

mgravell (approved):

PR #2925, opened 2025-07-23
chore: Replaces string interpolation with structured logging - ResultProcessor

Meir017:
related to #2899

Replaces manual string interpolation with compile-time generated LoggerMessage methods for better performance and structured logging.

Adds 16 new LoggerMessage extension methods covering response logging and auto-configuration scenarios for Redis server connections.

Improves logging efficiency by eliminating runtime string formatting overhead and enables better log analysis through consistent message templates.

satisfying!

https://github.com/user-attachments/assets/788c563f-9f63-430c-a88a-40b0f7dc812a

image

### DIFF
--- a/src/StackExchange.Redis/ResultProcessor.cs
+++ b/src/StackExchange.Redis/ResultProcessor.cs
@@ -960,12 +960,12 @@ namespace StackExchange.Redis
                                     if (val.IsEqual(CommonReplies.yes))
                                     {
                                         server.ReplicaReadOnly = true;
-                                        Log?.LogInformationAutoConfiguredConfigReadOnlyReplicaTrue(new(server));
+                                        Log?.LogInformationAutoConfiguredConfigReadOnlyReplica(new(server), true);
                                     }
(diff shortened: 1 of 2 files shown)


src/StackExchange.Redis/ResultProcessor.cs
```diff
@@ -960,12 +960,12 @@ protected override bool SetResultCore(PhysicalConnection connection, Message mes
                                     if (val.IsEqual(CommonReplies.yes))
                                     {
                                         server.ReplicaReadOnly = true;
-                                        Log?.LogInformation($"{Format.ToString(server)}: Auto-configured (CONFIG) read-only replica: true");
+                                        Log?.LogInformationAutoConfiguredConfigReadOnlyReplicaTrue(new(server));
```
mgravell, line 963: Can we merge these two and make the true/false a `bool` parameter? I'm fine with the case changing if we use default logging

mgravell (approved):

PR #2928, opened 2025-07-24
Improve `double` formatting performance on net8+ and fix equality incorrectness re special doubles

mgravell:
by using `double : IUtf8Formattable`

open question: this is net8+; however, for anything netcoreapp2.1+ we could actually use the `Span<char>` overload to a scratch buffer, which would avoid a `string` alloc (and copy down if needed, like we do today from `string`). However, this would only really impact net6 which is EOL, so I do *not* propose to introduce extra code for that.

This PR also fixes a pre-existing oddity where, (as `RedisValue`):

- `"na" != "NA"`
- `"nan" == "NAN"`
- `"nand" != "NAND"`

(and similar for `inf`)


NickCraver:
@mgravell I made you a present to use here: #2931

NickCraver (approved):

PR #2929, opened 2025-07-24
chore: Replaces string interpolation with structured logging - PhysicalConnection & PhysicalBridge

Meir017:
related #2899

Introduces dedicated LoggerMessage attributes for PhysicalConnection operations to improve logging performance and consistency.

Replaces string interpolation and manual log calls with strongly-typed logging methods that provide better compile-time safety and runtime performance through source generation.

Covers connection lifecycle events including endpoint validation, connection attempts, TLS configuration, socket operations, and connection establishment.


mgravell (approved):
you're on a roll

PR #2930, opened 2025-07-24
chore: Replaces string interpolation with structured logging - left overs

### LINKED ISSUE #2899: Use logging best practices for core logging messages

Meir017:
closes #2899

Converts manual log message formatting to LoggerMessage source generator methods for better performance and structured logging support.

Improves logging efficiency by eliminating string allocations when logging is disabled and provides consistent structured data for log analysis tools.

### DIFF
--- a/src/StackExchange.Redis/ConnectionMultiplexer.Sentinel.cs
+++ b/src/StackExchange.Redis/ConnectionMultiplexer.Sentinel.cs
@@ -396,7 +396,7 @@ public partial class ConnectionMultiplexer
         var logger = Logger.With(writer);
         if (connection.RawConfig.ServiceName is not string serviceName)
         {
-            logger?.LogInformation("Service name not defined.");
+            logger?.LogInformationServiceNameNotDefined();
             return;
         }
 
--- a/src/StackExchange.Redis/ConnectionMultiplexer.cs
+++ b/src/StackExchange.Redis/ConnectionMultiplexer.cs
@@ -1335,14 +1335,16 @@ namespace StackExchange.Redis
             if (log == null) return;
 
             var tmp = GetServerSnapshot();
-            log?.LogInformation("Endpoint Summary:");
+            log.LogInformationEndpointSummaryHeader();
             foreach (var server in tmp)
             {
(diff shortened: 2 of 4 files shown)


mgravell (approved):

mgravell:
@NickCraver that depends on the definition of "disabled"; if there exists a non-null logger isn't enabled *for the specific log-level being used* (generally "information" in this case), then this will indeed prevent those allocations, rather than the old code which creates the `string` *before* getting that far

mgravell:
examples from the codegen:

``` c#
        [global::System.CodeDom.Compiler.GeneratedCodeAttribute("Microsoft.Extensions.Logging.Generators", "8.0.9.3103")]
        internal static partial void LogInformationCheckingServerAvailable(this global::Microsoft.Extensions.Logging.ILogger logger, global::StackExchange.Redis.LoggerExtensions.EndPointLogValue endPoint)
        {
            if (logger.IsEnabled(global::Microsoft.Extensions.Logging.LogLevel.Information))
            {
                __LogInformationCheckingServerAvailableCallback(logger, endPoint, null);
            }
        }
       // ...     
        [global::System.CodeDom.Compiler.GeneratedCodeAttribute("Microsoft.Extensions.Logging.Generators", "8.0.9.3103")]
        internal static partial void LogErrorOperationFailedOnServer(this global::Microsoft.Extensions.Logging.ILogger logger, global::System.Exception exception, global::StackExchange.Redis.LoggerExtensions.EndPointLogValue endPoint, global::System.String errorMessage)
        {
            if (logger.IsEnabled(global::Microsoft.Extensions.Logging.LogLevel.Error))
            {
                __LogErrorOperationFailedOnServerCallback(logger, endPoint, errorMessage, exception);
            }
        }
```

NickCraver:
@mgravell Yeah that's fair, when someone says disabled generally that means 100% but agree they could mean a few things.

PR #2934, opened 2025-07-29
Tests: Improve stability for AbortOnConnectFailTests

NickCraver:
These had a lot timeout for runtime due to a shared method for faster test execution but in the day but the success case _should_ be fast but worst case it should not fail - we'd much rather it occasionally take a second than fail the test suite and add noise to the process.


mgravell (approved):

NickCraver:
@mgravell Did you see comment above?

PR #2939, opened 2025-08-11
Propose vector-set API

mgravell:
as per https://redis.io/docs/latest/develop/data-types/vector-sets/

- all methods/types start `VectorSet...`
- all core methods implemented
- the only "unusual" one is `VLINKS`; the server returns this as nested data, but the nesting is not meaningful to the caller (instead being related to the server core), so I've flattened the result
- usage is shown in `VectorSetIntegrationTests`, in particular `VectorSetSimilaritySearch_WithFilter` is useful for overview - since the main two primary APIs are `VADD` and `VSIM`
- since vector set data can be non-trivial, all return types lean on `Lease<T>` rather than `T[]`. Inputs are `ReadOnlyMemory<T>`; the exception to this is `string` data for JSON and filters; these are explicitly text, never blobs, so `RedisValue` seems inappropriate. I think forcing `string` is OK here
- in acknowledgement that some of our files are too large, I've started splitting the `VectorSet*` bits out via `partial` files; I will follow this up with "everything else" (.Strings.cs, .Hashes.cs) in a separate PR

This PR also introduces the start of a new literal-matching API, ala `FastHash`. It is intended that this will be extended at a later date.

CI: may be dependent on the `vectorset` module; if it fails, I'll add suitable validation. All `VectorSetTests` pass locally:

image


src/StackExchange.Redis/VectorSetSimilaritySearchMessage.cs
```diff
@@ -0,0 +1,322 @@
...
+        if (HasFlag(VsimFlags.SearchExplorationFactor))
+        {
+            physical.WriteBulkString("EF"u8);
+            physical.WriteBulkString(_searchExplorationFactor);
+        }
+
+        if (filterExpression is not null)
+        {
+            physical.WriteBulkString("FILTER"u8);
+            physical.WriteBulkString(filterExpression);
```
NickCraver, line 217: Let's check in tests for empty string here to be sure that behaves correctly

src/StackExchange.Redis/VectorSetSimilaritySearchMessage.cs
```diff
@@ -0,0 +1,322 @@
...
+         if (maxConnections.HasValue)
+         {
+             physical.WriteBulkString("M"u8);
+             physical.WriteBulkString(maxConnections.GetValueOrDefault());
+         }
+     }
+
+     public override int GetHashSlot(ServerSelectionStrategy serverSelectionStrategy)
+         => serverSelectionStrategy.HashSlot(key);
+         */
```
NickCraver, line 321: Mean to leave to leave in for save for next writer PR?

tests/StackExchange.Redis.Tests/VectorSetIntegrationTests.cs
```diff
@@ -0,0 +1,592 @@
...
+    }
+
+    [Fact]
+    public async Task VectorSetAdd_WithAttributes()
+    {
+        await using var conn = Create(require: RedisFeatures.v8_0_0_M04);
+        var db = conn.GetDatabase();
+        var key = Me();
+
+        await db.KeyDeleteAsync(key);
```
NickCraver, line 47: Can not await any of these deletes

src/StackExchange.Redis/VectorSetSimilaritySearchMessage.cs
mgravell: applying a null-or-whitespace filter instead, with integration tests for null, empty, whitespace, and non-whitespace values

src/StackExchange.Redis/VectorSetSimilaritySearchMessage.cs
mgravell: removed detritus

tests/StackExchange.Redis.Tests/VectorSetIntegrationTests.cs
mgravell: liberally sprinkled with F+F

mgravell:
image

kevin-montrose:
Lemme see if I follow.

Current API for VADD is:
```csharp
bool VectorSetAdd(
        RedisKey key,
        RedisValue element,
        ReadOnlyMemory<float> values,
        int? reducedDimensions = null,
        VectorSetQuantization quantization = VectorSetQuantization.Int8,
        int? buildExplorationFactor = null,
        int? maxConnections = null,
        bool useCheckAndSet = false,
        string? attributesJson = null,
        CommandFlags flags = CommandFlags.None);

public enum VectorSetQuantization
{
    Unknown,
    None,
    Int8,
    Binary,
}
```

So we'd change (or add as a new overload) that to something like:
```csharp
bool VectorSetAdd<TVectorData>(
        RedisKey key,
        RedisValue element,
        TVectorData values,
        int? reducedDimensions = null,
        VectorSetQuantization quantization = VectorSetQuantization.Int8,
        int? buildExplorationFactor = null,
        int? maxConnections = null,
        bool useCheckAndSet = false,
        string? attributesJson = null,
        CommandFlags flags = CommandFlags.None)
where TVectorData: IVectorData;

public readonly struct VectorSetQuantization(ReadOnlyMemory<byte> Name)
{
  public static readonly VectorSetQuantization Unknown = new(default);
  public static readonly VectorSetQuantization None = new("NOQUANT"u8);
  public static readonly VectorSetQuantization Int8 = new("Q8"u8);
  public static readonly VectorSetQuantization Binary = new("BIN"u8);
}

public interface IVectorData
{
  // Some sort of write callback here?
}
```

That makes sense to me.

mgravell:
Sorry, I was only thinking about VSIM - my bad. It would seem desirable to unify the approach between VSIM and VADD, and maybe marry the data and quant encoding (understanding that there is some overlap). I wonder if we should take a leaf from System.Text.Encoding here. Suggestion:

```
// Interprets a single T as N elements. T is most likely
// ROM-float, ROM-double, or similar; but is determined
// by the implementation
public abstract class VectorEncoding<TVector>
{
    public abstract ReadOnlySpan<byte> DataEncoding {get;} // "FP32"u8
    public abstract ReadOnlySpan<byte> Quantisation {get;} // "INT8"u8
    // note: if -ve reply from TryGetByteCount, VALUES is used instead
    public abstract int TryGetByteCount(in TVector vector); // 4x.Length
    public abstract void GetBytes(in TVector vector, Span<byte> buffer); // Unsafe.Cast ... CopyTo

    // used for VALUES encoding
    public abstract int GetElementCount(in TVector vector); // .Length
    public abstract double GetElement(in TVector vector, int index); // [index]
}
```

Where both VSIM and VADD could take a TVector and a VectorEncoding-TVector

Thoughts?

mgravell:
Or are the quantization and encoding entirely independent? They don't *feel* entirely independent....

mgravell:
@kevin-montrose I looked at this this morning. I *think* there's a very strong likelihood of over-designing if we don't have concrete requirements, and in any event: we'd probably still want the default and most obvious API / overload to be the existing one. I think, for now, we can do a minimal change here that keeps the door open:

- make `VectorSetSimilaritySearchRequest` abstract
- create two internal versions - one for "by member", one for "by vector, FP32"
- create two static methods on VSSSR

so:

``` diff
- [SER001]StackExchange.Redis.VectorSetSimilaritySearchRequest.Member.get -> StackExchange.Redis.RedisValue
- [SER001]StackExchange.Redis.VectorSetSimilaritySearchRequest.Member.set -> void
- [SER001]StackExchange.Redis.VectorSetSimilaritySearchRequest.Vector.get -> System.ReadOnlyMemory<float>
- [SER001]StackExchange.Redis.VectorSetSimilaritySearchRequest.Vector.set -> void
- [SER001]StackExchange.Redis.VectorSetSimilaritySearchRequest.VectorSetSimilaritySearchRequest() -> void
+ [SER001]static StackExchange.Redis.VectorSetSimilaritySearchRequest.ByMember(StackExchange.Redis.RedisValue member) -> StackExchange.Redis.VectorSetSimilaritySearchRequest!
+ [SER001]static StackExchange.Redis.VectorSetSimilaritySearchRequest.ByVector(System.ReadOnlyMemory<float> vector) -> StackExchange.Redis.VectorSetSimilaritySearchRequest!
```

This means that in the future, we can add whatever other factory methods we need to support additional encodings, and we haven't forced `Vector` to be any particular scale.

On the "add" side: we can just overload trivially.

I'm adding this as a commit - feedback encouraged before I merge ;p

kevin-montrose:
I'm fine being conservative until we have the final Garnet (or others) implementation to compare against (and obviously, consult about), and that seems reasonable.

Though this still leaves the question of `VADD` extensions unaddressed, yes?

NickCraver (approved):
Discussed doing an object for `VectorSetAdd` for future additions, otherwise looking good 👍

mgravell:
@kevin-montrose see Nick's comment above: working exactly that

PR #2953, opened 2025-09-16
Add overridable DefaultOptionsProvider.AfterDisconnectAsync() callback

philon-msft:
Enables graceful release and shutdown in custom DefaultOptionsProviders. For example, Microsoft.Azure.StackExchangeRedis uses a token refresh Timer that should be stopped when the connection is closed. 

This approach using a callback was suggested as an alternative to previous PR #2922

### DIFF
--- a/docs/ReleaseNotes.md
+++ b/docs/ReleaseNotes.md
@@ -8,7 +8,7 @@ Current package versions:
 
 ## Unreleased
 
-- (none)
+- Add overrideable `AfterDisconnectAsync()` callback on `DefaultOptionsProvider` ([#2952 by philon-msft](https://github.com/StackExchange/StackExchange.Redis/pull/2952))
 
 ## 2.9.17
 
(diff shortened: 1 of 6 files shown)


NickCraver:
In the refresh timer, why don't we check if any of the weak refs to multiplexers are still alive instead? And if not, stop the timer. This would be a lot less race-prone and ambiguous/confusing (e.g. disconnects happen all the time, per-server). Happy to chat more when back next week!

philon-msft:
> In the refresh timer, why don't we check if any of the weak refs to multiplexers are still alive instead? And if not, stop the timer. This would be a lot less race-prone and ambiguous/confusing (e.g. disconnects happen all the time, per-server). Happy to chat more when back next week!

Yes that works better - I'll remove a muxer from my list if its weak ref goes bad or it throws an ObjectDisposedException, and if the list is empty I'll stop the timer.

Issue #2920, opened 2025-07-21
HGETALL Hash not found without error on high memory pressure and multiple threads


jcaspes:
Hello,

While testing the StackExchange.Redis library (specifically the HGETALL api) under high memory pressure to induce frequent OutOfMemory exceptions, I encountered cases where received hashentry[] array is null or empty. No error is returned, so the calling program assumes the key does not exist, when in fact it should throw an exception.

Tested with this program: https://github.com/jcaspes/StackExchange.Redis/tree/MemoryTester/toys/MemoryTester

In this program, each thread only performs infinite HGETALL operations on a key with its thread ID, so it should only retrieve data that it has written itself.

here is a sample output; we can see that thread 27 (Th27hT) have retrieved nothing, and no exception was thrown:

```
27 : Key 'MemoryTester:Th27hT' not found
````

mgravell:
Thanks for the report. StackExchange.Redis **3.0** ships a substantial rewrite of the low-level IO core — a new RESPite-based stack that replaces `Pipelines.Sockets.Unofficial` and the old backlog/write loop — and we're confident this issue is addressed by that work.

Closing as resolved in 3.0. If you can still reproduce on 3.0.x, please comment and we'll reopen and investigate with fresh diagnostics.

Issue #2919, opened 2025-07-21
Receiving other thread HGETALL request result on high memory pressure and multiple threads


jcaspes:
Hello,

While testing the StackExchange.Redis library (on HGETALL api) under high memory pressure to induce frequent OutOfMemory exceptions, I encountered cases where received data is valid but not expected and appears to be the query result from an other thread.

Tested with this program: https://github.com/jcaspes/StackExchange.Redis/tree/MemoryTester/toys/MemoryTester

In this program, each thread only performs infinite HGETALL operations on a key with its thread ID, so it should only retrieve data that it has written itself.

Here a sample output; we can see that thread 218 (Th218hT) retrieved valid data that was queried by thread 58 (Th58hT)
```
218 : Error in key, expected:Th218hT, actual:Th58hT, value is a valid but from an other query: True
````

frobion:
Hello,

I reproduced deterministically the issue in [this test program](https://github.com/jcaspes/StackExchange.Redis/tree/Reproduce_wrong_Redis_response_between_threads/toys/RedisResponseTester). For this corruption to happen, we need at least:
- 2 threads (if we use only synchrone commands)
- 2 exceptions thrown at specific location in the library StackExchange.Redis

## Explanation step by step
1. A Redis command is sent (eg. "GET key1") and proceeds to `PhysicalBridge.WriteMessageToServerInsideWriteLock`. This method is responsible to first enqueue the command to `PhysicalConnection._writtenAwaitingResponse` then to send the command to the Redis server. However, a first exception is thrown between these two operations, and the command is enqueued in `PhysicalConnection._writtenAwaitingResponse` but not sent to the Redis server.
2. The catch block in `PhysicalBridge.WriteMessageToServerInsideWriteLock` should kill the connection, which should lead to a new `PhysicalConnection` being created, with a new (clean) queue. However, a second exception occurs in this catch block, preventing this cleanup.
3. The catch block in `PhysicalBridge.WriteMessageTakingWriteLockSync` triggers and mark the command in error.
4. The command having a `WriteResult` in error, it does not wait in `ConnectionMultiplexer.ExecuteSyncImpl` and return immediately in error.
5. Another command is sent (eg. "GET key2") and performs all the usual operations. Notably, it enqueues itself to `PhysicalConnection._writtenAwaitingResponse` and is sent to the Redis server. Then, it waits the response in `ConnectionMultiplexer.ExecuteSyncImpl`.
[...]

PR #2966, opened 2025-10-06
Make PhysicalBridge? physical member threadSafe

jcaspes:
By reading PhysicalBridge code i see that the phycical connexion is hosted by a volatile class member to be "protected".
This have been done to securize concurrent accesses to the physical connection because of volatile keyword i think.. but iot's not really safe.
This PR try to move physical connection in a threadSafe accessor class so no conflicts can happen.

Some details:

Seeing this kind of unsafe codes:
image
or another example
image
and more example where we only test the nullity of the memeber and then work on it... but can be set to null, and /or shutdown / disposed at any time by an other thread

And reading the microsoft documentation for the volatile keyword here: https://learn.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/volatile?devlangs=csharp
image

This code protection + HighIntegrity mode seem to fix the two issue: #2919 and #2920

### DIFF
--- a/src/StackExchange.Redis/PhysicalBridge.cs
+++ b/src/StackExchange.Redis/PhysicalBridge.cs
@@ -52,7 +52,8 @@ namespace StackExchange.Redis
 
         // private volatile int missedHeartbeats;
         private long operationCount, socketCount;
-        private volatile PhysicalConnection? physical;
+
+        private readonly ThreadSafePhysicalConnectionAccessor threadSafePhysicalConnectionAccessor = new ThreadSafePhysicalConnectionAccessor();
 
         private long profileLastLog;
         private int profileLogIndex;
@@ -61,7 +62,7 @@ namespace StackExchange.Redis
 
         private volatile int state = (int)State.Disconnected;
 
-        internal long? ConnectionId => physical?.ConnectionId;
+        internal long? ConnectionId => threadSafePhysicalConnectionAccessor.GetConnectionId();
 
 #if NETCOREAPP
         private readonly SemaphoreSlim _singleWriterMutex = new(1, 1);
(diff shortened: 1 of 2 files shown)


jcaspes:
An other example, in the TryEnqueue bridge Method:
image

`physical` is tested for nullity before the Messages loop, but during the loop `physical` can have been set to null or disposed

same in the tow TryWriteSync methods

NickCraver:
Just noting here that we can't take this route - it adds a _ton_ of locking in a critical path :)

jcaspes:
Last commit lower locking.

I've enhanced locks by only using WriteLock (blocking) when modifying the physical member like Dispose(), shutdown(), set to null...

Read lock do not lock other threads calls with read lock, only calls with write locks will need to wait for all readlocks to end, so during normal run (no connection timeout / recreation), only readlock are used, so theorically no overload, and no "ton of locking in a critical path" ;-)

I've done a quick benchmark with 200 threads on localhost with only HGETS. i cannot see performances downgrade for now.

image

Now try to find why these locks make the issue disappears

jcaspes:
> re
> 
> > physical is tested for nullity before the Messages loop, but during the loop physical can have been set to null or disposed
> 
> that's why it is snapshotted into a local, rather than using the field
> 
> Overall, however, I'm not seeing anything in this discussion that suggests a possible, let alone likely, cause of a problem here. Before we even consider changing code: we need to agree on what the change is trying to achieve. I'm not seeing that yet.

Hello,
You're absolutely right — this PR was just a quick attempt to help isolate the source of the leak. It’s not intended to be merged 😊

At the moment, I have less time to investigate further, as we're preparing to test the ServiceStack.Redis library with HighIntegrity in our production environment. It seems to address the issue, but it does introduce additional load on the servers.

I'll continue exploring and experimenting on my side whenever time permits.

Thanks again for your valuable feedback!

PR #2986, opened 2025-12-09
Options: Split out AMR to its own options provider

NickCraver:
This breaks out AMR into its own default options provider to specify default versions but also allow us to do things like specify default RESP later on.

Tagging @philon-msft for test run on both.


philon-msft:
@NickCraver I pushed a couple changes. I tested OSS and AMR caches including those changes, and all looks good

mgravell:
@philon-msft  are you happy for this to merge? I feel you're the code-owner of this area...

PR #3002, opened 2026-02-03
Handle Backlog Processor Startup Failures

nathan-miller23:
## Behavior

Whenever our client host experiences transient memory contention, we note that a portion of our `ConnectionMultiplexer` clients get 'stuck' in state where backlog writer is in `Activating` state. Example error message below:

image
<br></br>

Further analysis showed that client returned `IsConnected=true` with PhysicalBridge in `ConnectedEstablished` state. ProcDump revealed that no `StackExchange.Redis Backlog` thread was present

## Bug
When `StartBacklogProcessor` fails to start the backlog processor thread (say, when `thread.Start` fails with `System.OutOFMemoryException`) system gets stuck in state where `_backlogProcessorIsRunning` is true but no backlog thread is doing work. 

## Fix
`try`-`catch`-`finally` pattern to ensure we reset `_backlogPRocessorIsRunning` state when we fail to start thread

### DIFF
--- a/src/StackExchange.Redis/PhysicalBridge.cs
+++ b/src/StackExchange.Redis/PhysicalBridge.cs
@@ -934,20 +934,40 @@ namespace StackExchange.Redis
         {
             if (Interlocked.CompareExchange(ref _backlogProcessorIsRunning, 1, 0) == 0)
             {
-                _backlogStatus = BacklogStatus.Activating;
-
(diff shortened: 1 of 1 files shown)


mgravell:
[2.0.14](https://www.nuget.org/packages/StackExchange.Redis/2.10.14)

PR #3025, opened 2026-03-05
3.1 - Client-side geographic failover

mgravell:
- [x] basic connect
- [x] automatic ongoing active group detection
- [x] command routing
- [x] pubsub routing
- [x] improved configuration for down-detection via circuit-breaker (not just health-check)
- [x] align config API with other libraries (slightly complicated because of multiplexer vs pool, i.e. there are some fundamental differences between the cores)
- [x] command retry
  - [x] transactions ~and batches~
- [x] review config APIs - there are 4 in here (retry, circuit-breaker, health-check, multi-group), and they're not 100% in alignment

discussion topics / follow up items:

- [x] naming; prefer "active active" over "multi group" - clearer?
- [ ] can we support it at the single connection string level?
- [ ] pubsub *unsubscribe*


src/StackExchange.Redis/Availability/HealthCheck.cs
```diff
@@ -0,0 +1,125 @@
...
+        Probe = Probe,
+        ProbePolicy = ProbePolicy,
+    };
+
+    object ICloneable.Clone() => Clone();
+
+    /// <summary>
+    /// Create a new health check instance.
+    /// </summary>
+    public HealthCheck()
```
uglide: You have a deviation for default values compared to the design doc:

```                                                                                                                                                                                                                           
  ┌───────────────┬────────────┬────────────────┐                                                                                                                                                                          
  │    Setting    │    Code    │ Design default │                                                                                                                                                                          
  ├───────────────┼────────────┼────────────────┤                                                                                                                                                                          
  │ Interval      │ 10s        │ 5s             │                                                                                                                                                                          
  ├───────────────┼────────────┼────────────────┤                                                                                                                                                                          
  │ ProbeTimeout  │ 2s         │ 3s             │                                                                                                                                                                          
  ├───────────────┼────────────┼────────────────┤                                                                                                                                                                          
  │ ProbeInterval │ 1s         │ 500ms          │                                                                                                                                                                          
  ├───────────────┼────────────┼────────────────┤                                                                                                                                                                          
  │ ProbePolicy   │ AnySuccess │ HEALTHY_ALL    │                                                                                                                                                                          
  └───────────────┴────────────┴────────────────┘
```

src/StackExchange.Redis/Availability/MultiGroupOptions.cs
```diff
@@ -0,0 +1,49 @@
...
+using System.Threading;
+using RESPite;
+
+namespace StackExchange.Redis;
+
+/// <summary>
+/// Configuration options for controlling connections to multiple groups.
+/// </summary>
+[Experimental(Experiments.ActiveActive, UrlFormat = Experiments.UrlFormat)]
+public sealed class MultiGroupOptions()
```
uglide: Not sure if this is the correct file to mention this in, but we also need a way to disable Auto-fallback. In some cases, the operators might not be very confident about letting the client library decide whether the original endpoint is healthy.

src/StackExchange.Redis/Interfaces/IConnectionGroup.cs
```diff
@@ -0,0 +1,124 @@
...
+{
+    /// <summary>
+    /// A change occured to one of the connection groups.
+    /// </summary>
+    event EventHandler<GroupConnectionChangedEventArgs>? ConnectionChanged;
+
+    /// <summary>
+    /// Adds a new member to the group.
+    /// </summary>
+    Task AddAsync(ConnectionGroupMember member, TextWriter? log = null);
```
uglide, line 27: Right now, there is no way to add a member who is not yet healthy. In the design, we have a property called `skip_initial_health_check` to allow this.

uglide (commented):
The main missing peaces at this stage:
- Circuit breaker/Failure detector on the organic load
- Retries
- Per-member health checks
- Limited configuratbility and deviations in the default values (see inline comments)

src/StackExchange.Redis/Availability/MultiGroupOptions.cs
mgravell: https://github.com/StackExchange/StackExchange.Redis/blob/marc/aa2/docs/ActiveActive.md#unhealthy-state-and-failback

src/StackExchange.Redis/Availability/HealthCheck.cs
mgravell: updated

src/StackExchange.Redis/Interfaces/IConnectionGroup.cs
mgravell: done

Issue #3055, opened 2026-04-14
System.Memory binding


jcaspes:
Hello @mgravell 

After upgrading to Release version 2.12.14, a binding error appear when including StackExchange.Redis in a .net 4.8 project.
Indeed, by looking dependencies, we can see that two versions of System.Memory is now used:

<img width="441" height="158" alt="Image" src="https://github.com/user-attachments/assets/73731f10-adde-4e77-9c0a-4b2c0114db03" />

The new reference to System.IO.Hashing in this commit (https://github.com/StackExchange/StackExchange.Redis/commit/94e0090f2db03f6941547e839f1af40f5935ee21 and https://github.com/StackExchange/StackExchange.Redis/commit/c979c6b8714f19a823b5cd1b1cadbcafbc63e5d6), have introduced system.memory V 4:0:1:2 but System.IO.pipeline.dll 2.2.8 from your unoficial pipeline package reference system.memory 4:0:1:1.

I made a PR that fix this issue, your pipeline package version 2.2.16 fix the dependency, but also needed to upgrade  Microsoft.Bcl.AsyncInterfaces and Microsoft.Extensions.Logging.Abstractions

NickCraver:
You're in need of a proper binding redirect, these are generally generated for you automatically. The fix is never for every package to upgrade (that would require all package combinations in the world to agree). 

This comes up often enough we made a post for it: https://nickcraver.com/blog/2020/02/11/binding-redirects/

jcaspes:
Yes, I know, but as I explained, we have a lot of tools that didn't have a .config file or a bindingredirect until now. 

That's why I tried to figure out where the problem came from after upgrading StackExchange.Redis from 2.9.32 to 2.12.14 to avoid starting with redirect, to stay "clean" ;-).

Also i think that in this case it's not a "all package combinations in the world to agree" but just StackExchange.Redis that is not agree with it's own dependencies :-) like you can see in my screenshot in this issue first post or with the simple 4.8 project that only reference StackExchange.Redis 2.12.14 here : https://github.com/StackExchange/StackExchange.Redis/pull/3056#issuecomment-4246645278

And this is due to the recent inclusion of System.IO.Hashing of SDK 9 here https://github.com/StackExchange/StackExchange.Redis/commit/94e0090f2db03f6941547e839f1af40f5935ee21 so there is a mix of SDK versions.

mgravell:
(just for clarity: many of my thoughts are shown on the PR)

[...]

PR #3056, opened 2026-04-14
Upgrade version of some packages needing system.memory

jcaspes:
so they use the same version (4:0:1:2) than System.Hashing.IO

System.Hashing.IO can not be downgraded for old .net because first version with XxHash3 use system.memory 4:0:1:2

see issue https://github.com/StackExchange/StackExchange.Redis/issues/3055 for more informations

My tests from a simple .net 4.8 cmmandline thant include Stackexchange.redis seem to be OK without needing to add ugly runtime assemblyBinding in .config file

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -1,9 +1,9 @@
 <Project>
   <ItemGroup>
     <!-- Packages we depend on for StackExchange.Redis, upgrades can create binding redirect pain! -->
-    <PackageVersion Include="Microsoft.Bcl.AsyncInterfaces" Version="6.0.0" />
-    <PackageVersion Include="Microsoft.Extensions.Logging.Abstractions" Version="6.0.0" />
-    <PackageVersion Include="Pipelines.Sockets.Unofficial" Version="2.2.8" />
+    <PackageVersion Include="Microsoft.Bcl.AsyncInterfaces" Version="9.0.14" />
+    <PackageVersion Include="Microsoft.Extensions.Logging.Abstractions" Version="8.0.3" />
+    <PackageVersion Include="Pipelines.Sockets.Unofficial" Version="2.2.16" />
     <PackageVersion Include="System.Diagnostics.PerformanceCounter" Version="5.0.0" />
     <PackageVersion Include="System.Threading.Channels" Version="5.0.0" />
     <PackageVersion Include="System.Runtime.InteropServices.RuntimeInformation" Version="4.3.0" />
--- a/src/StackExchange.Redis/TextWriterLogger.cs
+++ b/src/StackExchange.Redis/TextWriterLogger.cs
@@ -17,11 +17,7 @@ internal sealed class TextWriterLogger : ILogger
         _wrapped = wrapped;
     }
 
-#if NET8_0_OR_GREATER
     public IDisposable? BeginScope<TState>(TState state) where TState : notnull => NothingDisposable.Instance;
-#else
-    public IDisposable BeginScope<TState>(TState state) => NothingDisposable.Instance;
-#endif
 
     public bool IsEnabled(LogLevel logLevel) => _writer is not null || _wrapped is not null;
     public void Log<TState>(LogLevel logLevel, EventId eventId, TState state, Exception? exception, Func<TState, Exception?, string> formatter)
--- a/tests/StackExchange.Redis.Tests/LoggerTests.cs
+++ b/tests/StackExchange.Redis.Tests/LoggerTests.cs
@@ -62,11 +62,7 @@ public class LoggerTests(ITestOutputHelper output) : TestBase(output)
         public int LogCount = 0;
         private ILogger Inner { get; } = toWrap;
 
-#if NET8_0_OR_GREATER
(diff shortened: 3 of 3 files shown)


mgravell:
For clarity, can I ask: were you seeing a specific issue that this addresses? Note that we've already applied this in the v3 branch (currently in preview, see [here](https://github.com/StackExchange/StackExchange.Redis/blob/v3/Directory.Packages.props)), so I'm torn between this vs "leave v2 alone". Also, since .NET 10 is GA and LTS, I wonder if the SDK 10 packages may be preferable to SDK 9.

jcaspes:
Hello! yes you can ask ;-)

I've explained the "problem" here: https://github.com/StackExchange/StackExchange.Redis/issues/3055

A simple app in .Net 4.8 including stackexchange.redis last release stop with exception with an Asssembly reference error from System.IO.Pipelines.dll coming from you unofficial io.pipeline package 2.2.8.

System.IO.Pipelines.dll need system.memory 4:0:1:1, but the one delivered is 4:0:1:2 due to System.IO.Hashing.dll.

At our side, we have a lot of commandlines that include our own redis dll including stackexchange.redis and can not run due to this.

A solution can be for us to add app.config to all these cmdlines with an assemblyBinding redirection to 4:0:1:2, but it's not a very good solution.

Not sure that the SDK 10 for .net 10 will be usable for our projects in .net 4.8 that we can't migrate to .net 10 for now :-(

jcaspes:
A simple .net framework 4.8 program like in this screenshot + no .config file

image

reproduce the problem easily.

returning to version 2.9.32 in same project run perfectly ;-)

mgravell:
Out of curiosity only, and purely a "if you have tim we" convenience (I don't expect you to update to a preview): if you have time to try the pre-release V3 build, is that "clean"?

jcaspes:
here a simple project to reproduce.

[StackExchange.test.net.4.8.zip](https://github.com/user-attachments/files/26726218/StackExchange.test.net.4.8.zip)

when building the binding problem is visible:

image

jcaspes:
> Out of curiosity only, and purely a "if you have tim we" convenience (I don't expect you to update to a preview): if you have time to try the pre-release V3 build, is that "clean"?

Hum, why not, i can try this. I can build this v3 branch and copy/past binaries to try it, but it will be tomorrow, i'm out of office now ;-)

NickCraver:
Your issue looks to be some version pinning of transitive dependencies - we should definitely not update SE.Redis here.

jcaspes:
Hello, it seem a problem because of a mix of microsoft sdk versions in packages used by SE.Redis by recent add of System.IO.Hashing from the SDK 9.
So now, SE.Redis on .net 4.x can not work any more alone without a bindingredirect, I think that's sad.

What is the reel reason to not update, Are you worried about performance issues or anything else?

mgravell:
> we should definitely not update SE.Redis here.

@NickCraver if we are taking the dep via "hashing", I'm fine with formalizing it

NickCraver:
@mgravell To clarify: I mean we shouldn't bump all the package versions, no objections to hashing specifically with minimal pain.

@jcaspes To answer your question:
> What is the reel reason to not update, Are you worried about performance issues or anything else?

Because it causes a lot of pain, and breaks people. It breaks more people, for no reason. What we have in your case is a misunderstanding of what the problem is. Binding redirects are a fundamental part of .NET Framework and they're _generated by default_. Otherwise think about it: every package in the world would have to agree at all times. That's just _never_ the case. Fixing you breaks other people - none of that is a win. We go with maximum compatibility and lowest requirements always - everything else is per-application to redirect and has been that way for decades.

jcaspes:
@nickCraver, first thank to read my messages :-).

I totally get this whole redirection thing—I’ve been working on a lot of C# projects for a long time and run into these issues all the time.

But in this case, as I show with the simple little project above, redirection is mandatory because otherwise SE.Redis doesn’t work. The issue is internal to SE.Redis; it can’t function correctly natively… I find that weird, I've never encountered this problem with a NuGet package.

Also, for your information, I'm working on this "issue" because we have, for example, a project in Managed C++/CLI DLL, and in this case, redirects aren't generated automatically or detected by MSBuild; so the problem have been detected only at runtime (thanks to the automated tests ;-) ), but it's dangerous.

Also, this PR doesn’t bump all packages, but only the minimum necessary to ensure that SE.Redis is clean with everyone linking to System.Memory 4:0:1:2

jcaspes:
@mgravell, today i found time to test the V3 with the Managed C++/CLI cmdline project in .net framework 4.8 we have.

The test was successful, no binding redirection to define ;-).
This CL do not do complex things but some HSET/HGET but worked perfectly.

I will test this V3 with our more complex project that is a web app in asp .net 4.8 also.

Thank for your work !

mgravell:
We discussed this in our weekly; as an update and further discussion: @NickCraver: please note that this PR *does not* add any explicit new System.Memory deps or update that; all it does is rev some secondary packages which in turn unify the transitive System.Memory. To me: this makes total sense, and also aligns nicely with the fair practice of going into nuget and "take latest". I'm fine with it, to be honest. Does this change your thoughts at all?

mgravell (approved):

PR #3063, opened 2026-04-23
Fix #2173: Honor AbortOnConnectFail on sentinel path

### LINKED ISSUE #2173: Setting AbortConnectOnFail doesn't work for Redis Sentinel setups
**Issue**
Setting AbortConnectOnFail doesn't work for Redis Sentinel setups.

**Details**
Setting `AbortConnectOnFail=false` works perfectly in Redis Master Slave setups. For example, this code:

```
// Startup.cs
var redisOptions = ConfigurationOptions.Parse("127.0.0.1:6379");
redisOptions.Password = "dev";
redisOptions.AbortOnConnectFail = false;
var redisConnection = ConnectionMultiplexer.Connect(redisOptions);
```

does not throw an exception when Redis is unavailable.

However, the setting does not work in Redis Sentinel setups. For example, this code:

```
// Startup.cs
var redisOptions = ConfigurationOptions.Parse("127.0.0.1:26379,serviceName=mymaster");
redisOptions.Password = "dev";
redisOptions.AbortOnConnectFail = false;
var redisConnection = ConnectionMultiplexer.Connect(redisOptions);
```

throws the following exception when Redis Sentinel is unavailable:

> Unhandled exception. StackExchange.Redis.RedisConnectionException: Sentinel: The ConnectionMultiplexer is not a Sentinel connection. Detected as: Standalone
   at StackExchange.Redis.ConnectionMultiplexer.GetSentinelMasterConnection(ConfigurationOptions config, TextWriter log) in /_/src/StackExchange.Redis/ConnectionMultiplexer.Sentinel.cs:line 163
   at StackExchange.Redis.ConnectionMultiplexer.Connect(ConfigurationOptions configuration, TextWriter log) in /_/src/StackExchange.Redis/ConnectionMultiplexer.cs:line 675

**Context**
I'm trying to set up the Redis configuration in Startup.cs to register Singleton IConnectionMultiplexer and to configure .NET's data protection library to persist keys in Redis:

```
services.AddSingleton<IConnectionMultiplexer>(_ => ConnectionMultiplexer.Connect(redisOptions));
services.AddDataProtection()
    .PersistKeysToStackExchangeRedis(ConnectionMultiplexer.Connect(redisOptions), "DataProtection-Keys");
```

When method `ConnectionMultiplexer.Connect(redisOptions)` throws an exception in the Startup.cs, the .NET application will abort. As a result, it's not possible to properly implement liveness and readiness (i.e. live = app is running, ready = app is running and all dependencies are available).

**Analysis**
I've tried to resolve this error myself, but unfortunately I didn't succeed... I think that the fix should be to move method `GetSentinelMasterConnection()` (see: [ConnectionMultiplexer.Sentinel.cs#L133-L135)](https://github.com/StackExchange/StackExchange.Redis/blob/main/src/StackExchange.Redis/ConnectionMultiplexer.Sentinel.cs#L133-L135) into method `ReconfigureAsync()` (see: [ConnectionMultiplexer.cs#L1178-L1506](https://github.com/StackExchange/StackExchange.Redis/blob/main/src/StackExchange.Redis/ConnectionMultiplexer.cs#L1178-L1506)). So, if the master node and/or the sentinel node are not available during startup, the connections will be regenerated during reconfiguration. Please let me know if I can assist in solving the issue!

IvanWennberg:
When `Connect` is called with `ServiceName` and `AbortOnConnectFail=false`,
the non-sentinel path returns a disconnected multiplexer that reconnects in
the background. The sentinel path ignores this flag and throws unconditionally
with "Sentinel: The ConnectionMultiplexer is not a Sentinel connection.
Detected as: Standalone".

This is because `SentinelConnect` returns a disconnected multiplexer
(ServerType defaults to Standalone when the handshake hasn't completed),
and `GetSentinelMasterConnection` throws when it sees the wrong ServerType
— without checking AbortOnConnectFail.

Changes in `GetSentinelMasterConnection`:

- Skip the ServerType check when AbortOnConnectFail is false and the
  connection hasn't completed handshake.
- When primary discovery returns null and AbortOnConnectFail is false,
  return a disconnected multiplexer with OnManagedConnectionFailed /
  OnManagedConnectionRestored wired up instead of throwing. Self-healing
  works via the existing SwitchPrimary retry timer.

While Sentinel isn't the future direction for Redis HA, it's still widely
used — especially with Valkey, AWS ElastiCache, and Kubernetes deployments
that haven't moved to cluster mode.

Tests in `SentinelAbortOnConnectTests.cs` — no running Redis/Sentinel
instance required. Self-healing reuses the existing SwitchPrimary mechanism
tested by `ManagedPrimaryConnectionEndToEndWithFailoverTest`.

### DIFF
--- a/src/StackExchange.Redis/ConnectionMultiplexer.Sentinel.cs
+++ b/src/StackExchange.Redis/ConnectionMultiplexer.Sentinel.cs
@@ -206,7 +206,7 @@ public partial class ConnectionMultiplexer
             {
                 if (!config.AbortOnConnectFail)
                 {
-                    connection = ConnectImpl(config, log, endpoints: config.EndPoints);
+                    connection = ConnectImpl(config, log, endpoints: config.EndPoints.Clone());
                     connection.ConnectionRestored += OnManagedConnectionRestored;
                     connection.ConnectionFailed += OnManagedConnectionFailed;
                     lock (sentinelConnectionChildren)
--- a/src/StackExchange.Redis/ConnectionMultiplexer.Sentinel.cs
+++ b/src/StackExchange.Redis/ConnectionMultiplexer.Sentinel.cs
@@ -166,7 +166,8 @@ public partial class ConnectionMultiplexer
     /// <param name="log">The writer to log to, if any.</param>
     public ConnectionMultiplexer GetSentinelMasterConnection(ConfigurationOptions config, TextWriter? log = null)
     {
-        if (ServerSelectionStrategy.ServerType != ServerType.Sentinel)
+        if (ServerSelectionStrategy.ServerType != ServerType.Sentinel
+            && (RawConfig.AbortOnConnectFail || IsConnected))
         {
             throw new RedisConnectionException(
                 ConnectionFailureType.UnableToConnect,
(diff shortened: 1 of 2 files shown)


mgravell:
Will try to look today

IvanWennberg:
This fixed the first hurdle, but after more extensive testing new issues keeps arising where the mux get stuck in a state it won't recover from depending on the code path it goes through.

I'll look into migrating to cluster instead of using sentinel. 
Closing this one for now.

PR #3065, opened 2026-04-28
8.8 add support for INCREX

mgravell:
INCREX is pre-checked INCRBY[FLOAT] with built in expiration support and range limiting (with choice of overflow behaviour)

Notes:

- there is a new `ENX` expiration concept; this is added into the existing `Expiration` type via a new `ExpirationFlags`
- new `ENX` is rejected by default on pre-existing `Expiration` scenarios; opt-in only
- core API is `StringIncrement` with a key, a value that is *either* `long` or `double`, `Expiration` (non-optional, but default is fine), and optional upper bound/lower-bound (in the same measure as the value). The result is a new `StringIncrementResult<T>` where `T` is the value kind (`long`/`double`).
- the new API is marked `[Experimental]` against 8.8, as are the new expiration flags scenario
- implemented with custom message / result processor
- unit tests implemented via toy-server
- integration tests use 8.8RC1


src/StackExchange.Redis/Interfaces/IDatabase.cs
```diff
@@ -3437,6 +3437,36 @@ IEnumerable<SortedSetEntry> SortedSetScan(
...
+        /// Atomically increments the integer value stored at key, optionally constraining the result and applying expiration semantics.
+        /// </summary>
+        /// <param name="key">The key of the string.</param>
+        /// <param name="value">The amount to increment by.</param>
+        /// <param name="expiry">The expiration to apply. Use <see cref="Expiration.Default"/> to retain the existing TTL.</param>
+        /// <param name="lowerBound">The optional lower bound for the resulting value.</param>
+        /// <param name="upperBound">The optional upper bound for the resulting value.</param>
+        /// <param name="options">The options to use for this operation.</param>
+        /// <param name="flags">The flags to use for this operation.</param>
+        /// <returns>The resulting value and the increment actually applied.</returns>
```
uglide, line 3450: I'd add here a link to the actual Redis command doc, so it's easy to figure out to which Redis command this method is mapped, like for the other methods:

```
/// <remarks><seealso href="https://redis.io/docs/latest/commands/increx/"/></remarks>
```

uglide (approved):

PR #3067, opened 2026-04-28
AMR: use RESP3 and disable the default config channel by default

mgravell:
@philon-msft 

Specific to AMR endpoints for now (planned more generally for v3)

- prefer RESP3
- don't use the old booksleeve pub/sub channel
- don't run the post-connect maintenance listener hook

Unit test validates behaviour. As a side change: the toy server now supports TLS so we can emulate AMR as closely as possible; we use a self-signed cert for this.

Outstanding:

- pub/sub connections are too eager - need to make lazier, so that even on RESP2 with booksleeve channel disabled:
  1. pub/sub doesn't connect initially
  2. but it *does* connect *when needed*

update: the attempt to make pub/sub connections "lazy" hit snags and was super unstable in CI; it needs more effort. Proposal: merge "as is", and consider that as a later PR. However, if the intent is to make RESP3 the default everywhere, and since NRedisStack now supports RESP3: it may also be acceptable to just live with the status quo: "RESP2 is suboptimal; oh well, then: don't use it!".


src/StackExchange.Redis/Configuration/DefaultOptionsProvider.cs
```diff
@@ -259,6 +259,11 @@ protected virtual string GetDefaultClientName() =>
         /// </summary>
         public virtual bool SetClientLibrary => true;
 
+        /// <summary>
+        /// Gets the preferred protocol to use for the connection.
+        /// </summary>
+        public virtual RedisProtocol? Protocol => null;
```
philon-msft, line 265:
```suggestion
        public virtual RedisProtocol? Protocol => RedisProtocol.Resp2;
```
...if we want to explicitly default to RESP2 here for clarity?

src/StackExchange.Redis/PublicAPI/PublicAPI.Unshipped.txt
```diff
@@ -1 +1,4 @@
 ﻿#nullable enable
+override StackExchange.Redis.Configuration.AzureManagedRedisOptionsProvider.ConfigurationChannel.get -> string!
```
philon-msft, line 2: at the risk of pointing out the obvious: don't forget to move these to PublicAPI._Shipped_.txt

src/StackExchange.Redis/Configuration/AzureManagedRedisOptionsProvider.cs
```diff
@@ -54,9 +53,15 @@ private bool IsHostInDomains(string hostName, string[] domains)
...
+            => Task.CompletedTask;
 
         /// <inheritdoc/>
         public override bool GetDefaultSsl(EndPointCollection endPoints) => true;
+
+        /// <inheritdoc/>
+        public override RedisProtocol? Protocol => RedisProtocol.Resp3; // prefer RESP3 on AMR
+
+        /// <inheritdoc/>
+        public override string ConfigurationChannel => ""; // disable on AMR
```
philon-msft, line 67: Does this empty channel name disable the default subscription connection entirely?

src/StackExchange.Redis/Configuration/DefaultOptionsProvider.cs
mgravell: IMO we should leave this alone until V3

src/StackExchange.Redis/Configuration/AzureManagedRedisOptionsProvider.cs
mgravell: in combination with not using the post-connect hook, yes; I'm adding a second test that verifies the same when explicitly choosing to use RESP2 (for example, for compat with modules that do odd things)

philon-msft (approved):

src/StackExchange.Redis/PublicAPI/PublicAPI.Unshipped.txt
mgravell: yep, I tend to leave that to the end when I know the PR is complete-ish

PR #3082, opened 2026-05-19
Don't serialize (ToString() or clone Protocol if it comes from the defaults-provider

mgravell:
The AMR default RESP3 change is accidentally manifesting in `options.ToString()` and `options.Clone()`; fix that.


philon-msft (approved):
Release note?

PR #3084, opened 2026-05-19
default RESP3 everywhere

mgravell:


tests/StackExchange.Redis.Tests/RespProtocolTests.cs
```diff
@@ -30,9 +30,9 @@ public async Task ConnectWithTiming()
     [InlineData("someserver,protocol=2,$HELLO=", false, "resp2")]
     [InlineData("someserver,protocol=2,$HELLO=BONJOUR", false, "resp2")]
     // specify a pre-6 version - only used if protocol specified
-    [InlineData("someserver,version=5.9", false)]
+    [InlineData("someserver,version=5.9", true)]
     [InlineData("someserver,version=5.9,$HELLO=", false)]
-    [InlineData("someserver,version=5.9,$HELLO=BONJOUR", false)]
+    [InlineData("someserver,version=5.9,$HELLO=BONJOUR", true)]
```
philon-msft: RESP3 wasn't supported until 6.0... if connection string or ConfigOptions specifies a server version < 6.0, should we not bother trying RESP3?

mgravell: Good idea

mgravell:
@philon-msft  you're right about the v6 thing; I've refactored how this works - I've put the defaults back in place, and changed the logic so that it'll try RESP3 if:

- if the protocol is set (explicitly or via the defaults-provider): that wins
- if the default-version is >= 6, try RESP3
- otherwise, don't
- the *default* for default-version: is now 6
- in all cases, `HELLO` is still needed; if disabled: nope

This seems a reasonable update to me; if the default-version is explicitly specified as something low (and the protocol is not specified): that will disable RESP3

This seems fine as a "major"

PR #3090, opened 2026-05-26
Fix platform-specific TCP keep-alive issues

### LINKED ISSUE #3086: TCP keep-alive failing on DNS endpoints pre .NET 10
using VALKEY 

Unhandled exception. StackExchange.Redis.RedisConnectionException: It was not possible to connect to the redis server(s). Error connecting right now. To allow this multiplexer to continue retrying until it's able to connect, use abortConnect=false in your connection string or AbortOnConnectFail=false; in your code.

   at StackExchange.Redis.ConnectionMultiplexer.ConnectImpl(ConfigurationOptions configuration, TextWriter writer, Nullable`1 serverType, EndPointCollection endpoints) in /_/src/StackExchange.Redis/ConnectionMultiplexer.cs:line 691

   at StackExchange.Redis.ConnectionMultiplexer.Connect(ConfigurationOptions configuration, TextWriter log) in /_/src/StackExchange.Redis/ConnectionMultiplexer.cs:line 663

   at StackExchange.Redis.ConnectionMultiplexer.Connect(String configuration, TextWriter log) in /_/src/StackExchange.Redis/ConnectionMultiplexer.cs:line 641

   at StocksRssFeed.Startup.ConfigureServices(IServiceCollection services)

   at System.RuntimeMethodHandle.InvokeMethod(Object target, Void** arguments, Signature sig, Boolean isConstructor)

   at System.Reflection.MethodBaseInvoker.InvokeDirectByRefWithFewArgs(Object obj, Span`1 copyOfArgs, BindingFlags invokeAttr)

   at System.Reflection.MethodBaseInvoker.InvokeWithOneArg(Object obj, BindingFlags invokeAttr, Binder binder, Object[] parameters, CultureInfo culture)

   at Microsoft.AspNetCore.Hosting.ConfigureServicesBuilder.InvokeCore(Object instance, IServiceCollection services)

   at Microsoft.AspNetCore.Hosting.GenericWebHostBuilder.UseStartup(Type startupType, HostBuilderContext context, IServiceCollection services, Object instance)

mgravell:
Fix https://github.com/StackExchange/StackExchange.Redis/issues/3086

before .NET 10, there are problems with TCP keep-alive on DNS endpoint on non-Windows; here we compensate:

1. make TCP keep-alive an explicit option (defaults to enabled, usual setup i.e. defaults-provider etc)
2. restrict TCP keep-alives to IP and DNS endpoints
3. if not .NET10+, disable for DNS endpoints if not on Windows

(also updates CI image to Redis 8.8)

### DIFF
--- a/src/StackExchange.Redis/ConfigurationOptions.cs
+++ b/src/StackExchange.Redis/ConfigurationOptions.cs
@@ -111,7 +111,8 @@ namespace StackExchange.Redis
                 Tunnel = "tunnel",
                 SetClientLibrary = "setlib",
                 Protocol = "protocol",
-                HighIntegrity = "highIntegrity";
+                HighIntegrity = "highIntegrity",
+                TcpKeepAlive = "tcpkeepalive";
 
             private static readonly Dictionary<string, string> normalizedOptions = new[]
             {
@@ -169,6 +170,7 @@ namespace StackExchange.Redis
         private Version? defaultVersion;
 
         private int? keepAlive, asyncTimeout, syncTimeout, connectTimeout, responseTimeout, connectRetry, configCheckSeconds;
+        private bool? tcpKeepAlive;
 
         private Proxy? proxy;
 
@@ -597,6 +599,15 @@ namespace StackExchange.Redis
         }
 
         /// <summary>
+        /// Gets or sets whether to enable TCP keep-alive when appropriate (endpoint- and platform-dependent).
+        /// </summary>
+        public bool TcpKeepAlive
+        {
+            get => tcpKeepAlive ?? Defaults.TcpKeepAlive;
+            set => tcpKeepAlive = value;
+        }
+
+        /// <summary>
         /// The <see cref="ILoggerFactory"/> to get loggers for connection events.
         /// Note: changes here only affect <see cref="ConnectionMultiplexer"/>s created after.
         /// </summary>
@@ -847,6 +858,7 @@ namespace StackExchange.Redis
             heartbeatInterval = heartbeatInterval,
             heartbeatConsistencyChecks = heartbeatConsistencyChecks,
             highIntegrity = highIntegrity,
+            tcpKeepAlive = tcpKeepAlive,
         };
 
         /// <summary>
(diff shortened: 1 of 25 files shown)


src/StackExchange.Redis/SocketManager.cs
ndyakov, line 79: can users provide their own SocketManager and/or are they expected to use it and access the `CreateSocket`? I am asking since this PR is changing the signature of the method and if so - this would be a breaking change.

ndyakov (approved):
Other than the question I added this looks good to me.

src/StackExchange.Redis/SocketManager.cs
mgravell: No; or rather: it doesn't matter - the method being changed is `internal` and `static`; it is basically just a library-private utility method

mgravell: It is a great question, though; for info: if you see on the left, PublicAPI.*.txt: that is a build-time-verified list of the APIs. Any changes to the publicly visible / callable API gets exposed there. So you can see that the public API change is limited to (abusing diff syntax):

``` diff
  ConfigurationOptions
+    bool TcpKeepAlive {get;set;}
  DefaultOptionsProvider
+    bool TcpKeepAlive {get;}
```

ndyakov: Thank you, consider this resolved.

tests/StackExchange.Redis.Tests/TcpKeepAliveTests.cs
```diff
@@ -0,0 +1,135 @@
...
+{
+    private TcpListener server;
+    private CancellationTokenSource cts = new();
+    public IPEndPoint IP { get; }
+    public DnsEndPoint Dns { get; }
+
+    public TcpTestFixture()
+    {
+        int port = 18000;
+#if NET10_OR_GREATER
```
uglide: Looks like a typo, shouldn't it be `NET10_0_OR_GREATER`?

src/StackExchange.Redis/ConfigurationOptions.cs
uglide, line 149: If I understand the code correctly, the `TcpKeepAlive` option should be added to `normalizedOptions` so users can use any case, and the parser will still correctly recognize the option.

uglide (changes requested):
I believe I found 2 small issues, worth double-checking.

tests/StackExchange.Redis.Tests/TcpKeepAliveTests.cs
mgravell: Yes, it should, thanks. Fixing. Just test isolation (against parallelism), but: fixed.

src/StackExchange.Redis/ConfigurationOptions.cs
mgravell: Adding a reflection-based test to prevent any future recurrence.

liukonen:
Thank you again for the vast turnaround on this!  You rock

Issue #3091, opened 2026-05-26
Write-side queue desync: a throw inside `Message.WriteTo` leaves the response queue out-of-sync with the wire


Pranish-Pantha:
## Summary

When `Message.WriteImpl` throws partway through writing a frame (most commonly under memory pressure), `Message.WriteTo` swallows the exception and returns normally. The outer write path in `PhysicalBridge` therefore considers the write successful, the message stays at the head of the in-flight queue, and the connection stays open. The next response the server sends — which is a reply to *some other* in-flight message, not the one whose write was aborted — gets paired with the wrong caller.

This is the same symptom that has been reported (and tentatively attributed to memory pressure) in:

- #2883 — *Under low-memory conditions, StackExchange.Redis can return incorrect results to simple `StringGetAsync` calls*
- #2804 — *"Unexpected response to HMGET: Integer: 1"* under OOM, including wrong values on GETs
- #2919 — *Receiving other thread's HGETALL request result on high memory pressure and multiple threads*
- #2913 — *Receiving rendered redis blob instead of string content on high memory pressure*

`HighIntegrity` mode mitigates the symptom (the per-message echo checksum catches the mismatch after the fact), but it does not prevent the desync — and as #2992 / #2993 show, `HighIntegrity` itself has rough edges in some configurations. The reports above all emphasize *memory pressure*, but the same code path can be tripped by any exception out of `WriteImpl`: a buffer-pool exhaustion, a serializer bug, an internal invariant violation, anything.

## Root cause

`src/StackExchange.Redis/Message.cs`, `WriteTo`:

```csharp
internal void WriteTo(PhysicalConnection physical)
{
    try
    {
        WriteImpl(physical);
    }
    catch (Exception ex) when (ex is not RedisCommandException)
    {
        physical?.OnInternalError(ex);
        Fail(ConnectionFailureType.InternalFailure, ex, null, physical?.BridgeCouldBeNull?.Multiplexer);
        // exception is swallowed — caller has no idea anything went wrong
    }
}
```

If `WriteImpl` had already pushed a partial frame into the output pipe before throwing (e.g. it wrote the multi-bulk header `*3\r\n` and the first argument, then OOM'd while encoding the second), those bytes are now on the wire. But because `WriteTo` returns normally:

1. `PhysicalBridge.WriteMessageInsideLock` returns `WriteResult.Success`.
2. The flush completes and the message remains in the response queue.
3. The connection is *not* torn down.
[...]

Issue #2993, opened 2026-01-13
Error when enabling highIntegrity configuration with Redis Sentinel


freezonezk:
I have deployed a Redis Sentinel cluster and integrated it with my web application as a distributed cache. During usage, I encountered a situation where at certain times, when reading cache from Redis with key A, it returns data from key B, accompanied by exceptions like StackExchange.Redis.RedisConnectionException: Unexpected response to HMGET: Integer: 1. After investigation, I discovered this issue is documented in https://github.com/StackExchange/StackExchange.Redis/issues/2804

I found that the recommended solution is to add highIntegrity=true to the Redis connection string to resolve the issue.

However, when I tried this configuration, I encountered a new error:

`
Unhandled exception. StackExchange.Redis.RedisConnectionException: Sentinel: The ConnectionMultiplexer is not a Sentinel connection. Detected as: Standalone
   at StackExchange.Redis.ConnectionMultiplexer.GetSentinelMasterConnection(ConfigurationOptions config, TextWriter log) in D:\Temp\StackExchange.Redis-2.8.58\src\StackExchange.Redis\ConnectionMultiplexer.Sentinel.cs:line 173
   at StackExchange.Redis.ConnectionMultiplexer.SentinelPrimaryConnect(ConfigurationOptions configuration, TextWriter log) in D:\Temp\StackExchange.Redis-2.8.58\src\StackExchange.Redis\ConnectionMultiplexer.Sentinel.cs:line 140
   at StackExchange.Redis.ConnectionMultiplexer.Connect(ConfigurationOptions configuration, TextWriter log) in D:\Temp\StackExchange.Redis-2.8.58\src\StackExchange.Redis\ConnectionMultiplexer.cs:line 661
   at StackExchange.Redis.ConnectionMultiplexer.Connect(String configuration, TextWriter log) in D:\Temp\StackExchange.Redis-2.8.58\src\StackExchange.Redis\ConnectionMultiplexer.cs:line 639
   at Program.<Main>(String[] args) in D:\ConsoleApp1\Program.cs:line 6
`

Below is my test code sample. I need assistance on how to enable the highIntegrity=true configuration when using Redis Sentinel:
`
var redis = ConnectionMultiplexer.Connect("192.168.41.180:26379,192.168.41.181:26379,192.168.41.182:26379,serviceName=mymaster,password=123321@Xyz,abortConnect=false,highIntegrity=true");
`
I am using the StackExchange.Redis library version 2.8.58.
[...]

Issue #2992, opened 2026-01-12
Sentinel connection error when using the highIntegrity = true configuration


thangpv212:
I am using Redis with the sentinel model.
Environment information:

Redis version: 7.0.5
StackExchange.Redis version: 2.8.58
My connection string configuration is as follows:
“10.0.0.1:26379,10.0.0.2:26379,10.0.0.3:26379,serviceName=mycluster,abortConnect=false,highIntegrity=true”

I encountered the following error:
```
StackExchange.Redis.RedisConnectionException: Sentinel: The ConnectionMultiplexer is not a Sentinel connection. Detected as: Standalone
   at StackExchange.Redis.ConnectionMultiplexer.GetSentinelMasterConnection(ConfigurationOptions config, TextWriter log) in /_/src/StackExchange.Redis/ConnectionMultiplexer.Sentinel.cs:line 173
   at StackExchange.Redis.ConnectionMultiplexer.SentinelPrimaryConnectAsync(ConfigurationOptions configuration, TextWriter writer) in /_/src/StackExchange.Redis/ConnectionMultiplexer.Sentinel.cs:line 157
   at Microsoft.Extensions.Caching.StackExchangeRedis.RedisCache.ConnectSlowAsync(CancellationToken token)
   at Microsoft.Extensions.Caching.StackExchangeRedis.RedisCache.GetAndRefreshAsync(String key, Boolean getData, CancellationToken token)
   at Microsoft.Extensions.Caching.StackExchangeRedis.RedisCache.GetAsync(String key, CancellationToken token)
   at Microsoft.Extensions.Caching.Distributed.DistributedCacheExtensions.GetStringAsync(IDistributedCache cache, String key, CancellationToken token)
```

Before setting highIntegrity=true, I did not encounter this error.

Please help me find the root cause of the error.
Thank you!

Issue #2913, opened 2025-07-18
Receiving rendered redis blob instead of string content on high memory pressure


jcaspes:
Hello,

While testing the StackExchange.Redis library (on HGETALL api) under high memory pressure to induce frequent OutOfMemory exceptions, I encountered cases where the returned string value was a hexadecimal representation of the Redis response bytes (e.g., "4C-E8-3A...") instead of valid data. Notably, the client API did not report any error in these scenarios.

This behavior originates from the RedisValue string conversion operator: https://github.com/StackExchange/StackExchange.Redis/blob/8a527830fd45b5eaef54e9b3b96f3f6ca7af3813/src/StackExchange.Redis/RedisValue.cs#L789

The conversion uses a try/catch block that, on any exception, returns the hex string. Is there a specific reason or intended purpose for this approach?

This forces us to check the validity of the values returned by the library, which makes things complicated.
Why not let the exception be thrown?

mgravell:
Thanks, I've pulled that down; no luck with a repro yet, but I guess it may take some time...

<img width="536" height="82" alt="Image" src="https://github.com/user-attachments/assets/4ecc8d98-3f37-4d1c-b53f-175a2c2c37c4" />

jcaspes:
Yes some time it take long other time not... 
For this issue with blob rendered, you must give the argument checkContent (donet run checkContent) to reproduce.

<img width="980" height="140" alt="Image" src="https://github.com/user-attachments/assets/bcaaf953-9f6b-4d1c-9a60-bc34d2a49738" />

jcaspes:
I'm in an Hyper-V VM, the hoster is a windows 2011 enterprise, and the VM where the test run is windows 2022 server (VM with 50G memory and 16 logical processors).
It need to run quick, for example when i enable logs, it's more difficult to reproduce, all depends on where the random OutOfMemory occure...

mgravell:
Well, the "good" news is that it seems to fairly reliably break on Windows very quickly (repeatedly under 30s); that gives me something to work with, at least; thanks.

jcaspes:
OK, cool.
You can also add the log parameter (MemoryTester.exe checkcontent log) it will trace all different stack and exception type found, it can help

jcaspes:
For this issue only in RedisValue class where you try/catch the GetString call, i can see that the GetString only call the Encoding.UTF8.GetString api.

This api can only throw:

[...]

PR #3092, opened 2026-05-26
Tear down connection on write failure to prevent queue desync

Pranish-Pantha:
## Summary

- `Message.WriteTo` now re-throws after calling `Fail()`, so `PhysicalBridge`'s outer write path can act on the failure instead of seeing a phantom success.
- `PhysicalBridge.HandleWriteException` takes the `PhysicalConnection` and calls `RecordConnectionFailed(ConnectionFailureType.InternalFailure, ...)`, tearing the connection down and draining the response queue with failures.
- All callers of `HandleWriteException` (sync, async, backlog, post-flush completion) thread the connection through.

Fixes the underlying cause behind the symptoms reported in #2883, #2804, #2919, #2913 — without depending on `HighIntegrity` mode (which catches the desync after the fact rather than preventing it).

See accompanying issue for the full root-cause writeup.

## Why

If `WriteImpl` throws partway through a frame (OOM during serialization is the most-reported trigger, but any exception out of `WriteImpl` does it), the old `WriteTo` swallowed the exception. The bridge then saw `WriteResult.Success`, kept the message at the head of the in-flight queue, and left the socket connected. Any subsequent reply from the server matched against the wrong message.

`Fail()` signals the broken message's awaiter, but it does nothing about the wire state or the sibling messages whose ordering has now been corrupted. Tearing down the physical connection completes all in-flight awaiters with `RedisConnectionException(InternalFailure)` and forces a clean reconnect.

## Test plan

- [x] New `WriteTo_PropagatesWriteImplException` — fails on `main`, passes here. Asserts `WriteTo` rethrows non-`RedisCommandException` from `WriteImpl`.
- [x] New `WriteTo_DoesNotWrapRedisCommandException` — `RedisCommandException` continues to surface unwrapped (it's excluded from the catch filter).
- [x] New `WriteFailure_TearsDownPhysicalConnection` — end-to-end: a `Message` whose `WriteImpl` throws faults the awaiter with `RedisConnectionException(InternalFailure)` *and* raises a `ConnectionFailed` event with `InternalFailure`.
- [x] Verified the new tests fail on the unfixed source and pass with the fix applied (stash/unstash cycle).
- [x] `dotnet build` clean for `src/StackExchange.Redis/StackExchange.Redis.csproj` and `tests/StackExchange.Redis.Tests/StackExchange.Redis.Tests.csproj` on `net10.0`.

## Risk

The change widens the conditions under which a connection is torn down — specifically, an in-process exception out of `WriteImpl` now triggers a reconnect where it previously left the connection up. That's the intended behavior (the connection was already in an undefined state), but it may surface as additional `ConnectionFailed` events for users who were silently hitting `Fail()` paths. In every case observed in the issues above, that reconnect is what users *wanted* to happen.

## Related

- Issue: #3091
- Existing reports: #2883, #2804, #2919, #2913
- Adjacent: #2992, #2993 (rough edges around `HighIntegrity` itself, which is the current workaround)

### DIFF
--- a/src/StackExchange.Redis/Message.cs
+++ b/src/StackExchange.Redis/Message.cs
@@ -797,6 +797,11 @@ namespace StackExchange.Redis
             {
                 physical?.OnInternalError(ex);
                 Fail(ConnectionFailureType.InternalFailure, ex, null, physical?.BridgeCouldBeNull?.Multiplexer);
+                // Re-throw so the outer write path (PhysicalBridge.HandleWriteException) can tear down the
+                // connection. A partial write would otherwise leave bytes on the wire while the response
+                // queue still considers the slot healthy, allowing a subsequent reply to match the wrong
+                // in-flight message.
+                throw;
             }
         }
 
--- a/src/StackExchange.Redis/PhysicalBridge.cs
+++ b/src/StackExchange.Redis/PhysicalBridge.cs
@@ -865,7 +865,7 @@ namespace StackExchange.Redis
                 physical.SetIdle();
                 return result;
             }
-            catch (Exception ex) { return HandleWriteException(message, ex); }
+            catch (Exception ex) { return HandleWriteException(physical, message, ex); }
             finally
             {
                 UnmarkActiveMessage(message);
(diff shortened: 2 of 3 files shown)


Pranish-Pantha:
Are some tests flaky and need rerun of workflow to pass?

mgravell:
Yes, the CI can be flakey, which is not ideal. There is a plan to help thatz but it needs a lot of effort. I'll take a look.

mgravell (approved):
LGTM

mgravell:
This looks like a great find, thanks; merged

Issue #3085, opened 2026-05-20
Race condition in ReadFromPipe after heartbeat-driven pipe thread restart (2.12.14+)


PhbsSmn:
# Race condition in ReadFromPipe after heartbeat-driven pipe thread restart (2.12.14+)

## Summary

After upgrading from StackExchange.Redis 2.12.4 to 2.12.14 (and subsequently 2.13.1), all Redis operations consistently fail with either `InvalidOperationException: Reading is not allowed after reader was completed` or `TaskCanceledException`. The issue was introduced in 2.12.14 by PR #2965 ("Allow heartbeat to restart pipe thread with sync commands").

## Environment

- **StackExchange.Redis**: 2.13.1 (also reproduced on 2.12.14 and 2.12.15)
- **Last working version**: 2.12.4
- **Runtime**: .NET 8.0 on Linux (x64)
- **Hosting**: AWS Lambda (512 MB, 30s timeout)
- **Redis**: AWS Valkey

## Symptoms

Every Redis operation fails consistently. Two distinct exceptions are observed:

### 1. InvalidOperationException (the root cause)

```
System.InvalidOperationException: Reading is not allowed after reader was completed.
   at System.IO.Pipelines.ThrowHelper.ThrowInvalidOperationException_NoReadingAllowed()
   at System.IO.Pipelines.Pipe.AdvanceReader(SequencePosition& consumed, SequencePosition& examined)
   at System.IO.Pipelines.Pipe.DefaultPipeReader.AdvanceTo(SequencePosition consumed, SequencePosition examined)
   at StackExchange.Redis.PhysicalConnection.ReadFromPipe() in /_/src/StackExchange.Redis/PhysicalConnection.cs:line 2098
```

### 2. TaskCanceledException (consequence)

```
System.Threading.Tasks.TaskCanceledException: A task was canceled.
   at RedisExtensions.TransactAsync(IDatabase db, Action`1 addCommands)
```

When the `ReadFromPipe()` thread crashes, all pending command Tasks are canceled — this is the more commonly observed symptom.

## Root Cause Analysis

The issue is a race condition likely introduced by changes in 2.12.14, specifically PR #2965 ("Allow heartbeat to restart pipe thread with sync commands") and possibly PR #3040 ("Support multi-DB cluster usage on valkey"). Our Redis server is **AWS Valkey**, so the product variant detection introduced in #3040 may interact with the heartbeat restart logic in #2965 — both landed in the same release.

[...]

PR #3110, opened 2026-06-23
fix(bridge): enforce TaskCompletionSource handshake gate to prevent connection race

### LINKED ISSUE #2989: Messages timing out while Backlog is SpinningDown
In which: I notice a lot of issues with (intermittent?) timeouts that say bw: SpinningDown and wonder if it is a significant pattern. 

Client reports timeouts from a slow EVAL command. But there was no specific symptom of timeout or slow EVAL command on the server side.

It occurs to me that there are a lot of issues mentioning "SpinningDown", and maybe they arise from a race condition related to the backlog, around the _backlogProcessorIsRunning and _backlogAutoReset state. So, let's see..

_The theoretical analysis_

The job of `StartBacklogProcessor` is to ensure a backlog processing thread is running and handling any backlog work. It handles this either by starting a new backlog processor, or by trying to wake up an existing one by setting the `_backlogAutoReset` event, if it sees that _backlogProcessorIsRunning.

The event is observed by doing `WaitOne(_backlogAutoReset, 5 seconds)`, but there is a race as to:

a) when anybody *sets* the event, and whether WaitOne randomly returns after 5 seconds, or the event is set first.
b) if multiple people set the event how many times it is observed - in theory though, being observed once should be enough, so I'm not convinced this is an actual problem

If `WaitOne` sees the event, great, the backlog processing thread wakes up and loops to handle the work.
If `WaitOne` doesn't see the event, the event is set, and the backlog processing thread exits, releasing its hold on _backlogProcessorIsRunning, and leaving the _backlogStatus equal to SpinningDown. No new backlog thread got started up, since the event was triggered... just not in time.

So yes, we can get to the following undesirable state:

a) we have an item in the backlog
b) backlog status is spinning down
c) backlog processor is no long running, and released `_backlogProcessorIsRunning`
d) somebody might have tried to restart the backlog processor, but backlog thread didn't observe that

_Example repro error message that I feel matches this scenario and motivated me to look deeper_

(from 2025)
>StackExchange.Redis.RedisTimeoutException: Timeout awaiting response (outbound=0KiB, inbound=1KiB, 3168ms elapsed, timeout is 3000ms), command=EVAL, next: EVAL, inst: 0, qu: 0, qs: 0, aw: False, bw: SpinningDown, rs: ReadAsync, ws: Idle, in: 0, last-in: 0, cur-in: 0, sync-ops: 0, async-ops: 63738, serverEndpoint: x.x.x.x:15011, conn-sec: 2617.06, aoc: 0, mc: 1/1/0, mgr: 10 of 10 available, clientName: xxx(SE.Redis-v2.10.1.65101), PerfCounterHelperkeyHashSlot: 13549, IOCP: (Busy=0,Free=1000,Min=1,Max=1000), WORKER: (Busy=3,Free=32764,Min=10,Max=32767), POOL: (Threads=13,QueuedItems=0,CompletedItems=8851526,Timers=1526), v: 2.10.1.6510
 
Now I just wonder if there's an actual decent way to repro this to prove the theory...

HarshalPatel1972:
### Description

Fixes #2989 / #3085

This PR introduces an explicit, asynchronous `TaskCompletionSource<bool>` handshake coordination gate within the `PhysicalBridge` processing loop. This completely eliminates an intermittent protocol corruption race condition that occurs when the .NET ThreadPool experiences heavy starvation during rapid socket reconnection cycles.

### Technical Root Cause
When a multiplexed socket connection drops under thread strain, the `PhysicalBridge` correctly initiates an asynchronous restart. However, the subsequent server handshake commands (`HELLO`, `AUTH`, `CLIENT SETNAME`) rely on parsing asynchronous callbacks handled downstream by the `PipeReader` running on the shared ThreadPool.

Under severe ThreadPool exhaustion, these parsing continuations can be heavily delayed. Prior to this patch, the connection state was flagged as available prematurely relative to the actual protocol acknowledgment. As a result, concurrent user application threads checking the connection status would immediately start dispatching application-level data payloads down the socket. Because these payloads interleave into the raw socket pipe *before* the server completely processes and acknowledges the initial `AUTH` handshake, the protocol boundaries become completely misaligned, throwing deceptive `NOAUTH` or `Unauthorized` exceptions and locking the bridge into an unstable retry loop.

### Resolution Strategy
* **Handshake Completion Boundary:** Introduced a private `TaskCompletionSource<bool>` (`_handshakeCompletion`) assigned with `TaskCreationOptions.RunContinuationsAsynchronously` to isolate the lifetime of the socket connection sequence.
* **Non-Blocking Await Fence:** Refactored the core outbound write paths inside `PhysicalBridge.cs`. If a data payload attempts a transmission while the state is `ConnectedEstablishing`, the calling task cleanly executes an `await _handshakeCompletion.Task.ConfigureAwait(false);`.
* **Atomic Pipeline Unlocking:** The gate is held closed throughout the execution of all validation setup steps (`AUTH`, `HELLO`, `CLIENT SETNAME`). Once `OnFullyEstablished()` is safely invoked following the final protocol validation, `_handshakeCompletion.TrySetResult(true)` is fired, atomically releasing any queued application tasks to send their data down a perfectly authenticated stream.

### Verification
* Validated locally with simulated thread pool constraints (`ThreadPool.SetMinThreads(1, 1)`) to ensure high-concurrency requests cleanly queue behind the establishing gate without dropping or triggering thread lockups.

### DIFF
--- a/tests/StackExchange.Redis.Tests/ConnectionRaceTests.cs
+++ b/tests/StackExchange.Redis.Tests/ConnectionRaceTests.cs
@@ -10,6 +10,9 @@ public class ConnectionRaceTests(ITestOutputHelper output) : TestBase(output)
     [Fact]
     public async Task HandshakeCompletionGatePreventsUnauthenticatedPayloads()
     {
+        #if RELEASE
+        Assert.Skip("Ignoring in CI, due to threading");
+        #endif
         // Simulate severe thread pool exhaustion using ThreadPool.SetMinThreads(1, 1);
         ThreadPool.GetMinThreads(out int workerThreads, out int completionPortThreads);
         ThreadPool.SetMinThreads(1, 1);
@@ -17,15 +20,15 @@ public class ConnectionRaceTests(ITestOutputHelper output) : TestBase(output)
         {
             var options = new ConfigurationOptions
             {
-                EndPoints = { { TestConfig.Current.MasterServer, TestConfig.Current.MasterPort } },
-                Password = TestConfig.Current.MasterPassword,
+                EndPoints = { { TestConfig.Current.PrimaryServer, TestConfig.Current.PrimaryPort } },
+                Password = TestConfig.Current.PrimaryPassword,
                 AbortOnConnectFail = false,
                 AllowAdmin = true
             };
 
             await using var conn = await ConnectionMultiplexer.ConnectAsync(options);
             var db = conn.GetDatabase();
-            var server = conn.GetServer(TestConfig.Current.MasterServer, TestConfig.Current.MasterPort);
+            var server = conn.GetServer(TestConfig.Current.PrimaryServer, TestConfig.Current.PrimaryPort);
 
             // Trigger an asynchronous connection reset loop
             var resetLoop = Task.Run(async () =>
--- a/tests/StackExchange.Redis.Tests/ConnectionRaceTests.cs
+++ b/tests/StackExchange.Redis.Tests/ConnectionRaceTests.cs
@@ -1,0 +1,72 @@
+using System;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+using Xunit.Abstractions;
+
+namespace StackExchange.Redis.Tests;
+
+public class ConnectionRaceTests(ITestOutputHelper output) : TestBase(output)
+{
+    [Fact]
+    public async Task HandshakeCompletionGatePreventsUnauthenticatedPayloads()
+    {
+        // Simulate severe thread pool exhaustion using ThreadPool.SetMinThreads(1, 1);
+        ThreadPool.GetMinThreads(out int workerThreads, out int completionPortThreads);
+        ThreadPool.SetMinThreads(1, 1);
+        try
+        {
+            var options = new ConfigurationOptions
+            {
+                EndPoints = { { TestConfig.Current.MasterServer, TestConfig.Current.MasterPort } },
+                Password = TestConfig.Current.MasterPassword,
+                AbortOnConnectFail = false,
+                AllowAdmin = true
+            };
+
+            await using var conn = await ConnectionMultiplexer.ConnectAsync(options);
+            var db = conn.GetDatabase();
+            var server = conn.GetServer(TestConfig.Current.MasterServer, TestConfig.Current.MasterPort);
+
+            // Trigger an asynchronous connection reset loop
+            var resetLoop = Task.Run(async () =>
+            {
+                for (int i = 0; i < 5; i++)
+                {
+                    server.SimulateConnectionFailure(SimulatedFailureType.AuthenticationFailure);
+                    await Task.Delay(10);
+                }
+            });
+
+            // Concurrently launch 100 parallel Tasks trying to dispatch rapid PingAsync() operations.
+            int taskCount = 100;
+            var tasks = new Task[taskCount];
+            for (int i = 0; i < taskCount; i++)
+            {
+                tasks[i] = Task.Run(async () =>
+                {
+                    for (int j = 0; j < 50; j++)
+                    {
+                        try
+                        {
+                            await db.PingAsync().ConfigureAwait(false);
+                        }
+                        catch (Exception ex)
+                        {
+                            // Assert that no commands throw a NOAUTH or protocol corruption exception
+                            Assert.DoesNotContain("NOAUTH", ex.Message);
+                            Assert.DoesNotContain("protocol", ex.Message.ToLowerInvariant());
+                        }
+                    }
+                });
+            }
+
+            await Task.WhenAll(tasks);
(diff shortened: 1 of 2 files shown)


mgravell:
Will twill try to look tomorrow before the next 3.0.x release

Note that since this targets "main" which is now V3, there is no more pipe-reader, but that doesn't mean the async goes away. However, there *is* now technically support for running the read loop on dedicated sync IO threads - I have not exposed this on the public API yet, however.

tests/StackExchange.Redis.Tests/ConnectionRaceTests.cs
```diff
@@ -0,0 +1,71 @@
...
+    public async Task HandshakeCompletionGatePreventsUnauthenticatedPayloads()
+    {
+        // Simulate severe thread pool exhaustion using ThreadPool.SetMinThreads(1, 1);
+        ThreadPool.GetMinThreads(out int workerThreads, out int completionPortThreads);
+        ThreadPool.SetMinThreads(1, 1);
+        try
+        {
+            var options = new ConfigurationOptions
+            {
+                EndPoints = { { TestConfig.Current.MasterServer, TestConfig.Current.MasterPort } },
```
mgravell: I'm kinda confused here... this hasn't existed for 4+ years... was this PR based on a *really* old fork?

tests/StackExchange.Redis.Tests/ConnectionRaceTests.cs
```diff
@@ -0,0 +1,71 @@
...
+namespace StackExchange.Redis.Tests;
+
+public class ConnectionRaceTests(ITestOutputHelper output) : TestBase(output)
+{
+    [Fact]
+    public async Task HandshakeCompletionGatePreventsUnauthenticatedPayloads()
+    {
+        // Simulate severe thread pool exhaustion using ThreadPool.SetMinThreads(1, 1);
+        ThreadPool.GetMinThreads(out int workerThreads, out int completionPortThreads);
+        ThreadPool.SetMinThreads(1, 1);
```
mgravell, line 18: this isn't a great idea in a CI setup; I understand the intent, but... this might be something we simply can't do in this context due to parallelism concerns, plus the impact on the overall runtime

mgravell:
The CI seems to *hate* this; I suspect it is causing a sync-over-async that is blowing up the thread-pool; if it doesn't turn out to be a false positive, I might take the premise and start from scratch in terms of fixes.

mgravell:
Additionally: given that this PR seems to have targeted 4+ year old code, I'm suspicious that it may be based on out-of-date logic. Either way: I don't think we can take the PR as is. I'm going to close the PR, but I'd love to discuss the topic in an *issue*, where the first question will be "With what library version are you seeing this?"

HarshalPatel1972:
> Additionally: given that this PR seems to have targeted 4+ year old code, I'm suspicious that it may be based on out-of-date logic. Either way: I don't think we can take the PR as is. I'm going to close the PR, but I'd love to discuss the topic in an _issue_, where the first question will be "With what library version are you seeing this?"

@mgravell Ah, my apologies for the confusion! I accidentally targeted the PR against main (V3), but my branch was actually based off the 2.13.17 release tag where the PipeReader architecture still exists. That explains why the code looked 4+ years out of date to you!

We were seeing this race condition on 2.13.17 under heavy load when the thread pool gets temporarily starved. The test case with ThreadPool.SetMinThreads(1, 1) was definitely a sledgehammer approach just to reliably reproduce the starvation locally—I completely understand why CI hated it and why it can't be merged as-is.

Since main has moved on to V3 and the internal async architecture has changed so much, I agree this specific patch doesn't fit anymore.

If you think this race condition might still exist in the new V3 architecture, I'd love to help test any new approaches or dedicated sync IO threads you're experimenting with. Otherwise, is there any appetite for a backported (and safer/sync-over-async free) version of this handshake gate for the 2.x track, or are you primarily focused on V3 for these types of core connection pipeline fixes?

mgravell:
>  but my branch was actually based off the 2.13.17 release tag where the PipeReader architecture still exists. That explains why the code looked 4+ years out of date to you!

2.13.17 is very recent - no problem there; however, I was talking about the test usage of `MasterPort` etc - that went in cf3e30bfb263734c124e007a500b7792e724a778, which was 4 years ago. Not a big problem, just: confusing!

I'm definitely open to working through the issue, both in main and 2.x - however, introducing a sync-over-async into core logic is probably a little on the dangerous side, especially if that could cause a slam during connection. I'd quite like to take a step back and make sure I understand what we're fixing; if the problem is:

> Prior to this patch, the connection state was flagged as available prematurely relative to the actual protocol acknowledgment.

then rather than introducing an extra gate so that there's a second hold point, I'd lean towards "don't flag it as available prematurely" - then there's no need for the gate!

PR #3121, opened 2026-07-06
Build tools: centralize code-gen indent handling in a CodeWriter

zhengyazhao:
Implements #3033 using the second sketch from the issue: a small builder-style `CodeWriter` wrapping the existing `StringBuilder`, replacing the per-method `int indent` + local `NewLine()` pattern in `AsciiHashGenerator` (4 occurrences).

On the `IndentedTextWriter` option: its indent-on-next-write semantics would subtly change emitted whitespace. Since there was no functional win to justify that, I kept the `StringBuilder` core so the change is provably inert. (The preamble in `Generate` intentionally still writes to the raw `StringBuilder` to keep the diff minimal.)

**Verification**: captured generated output (`EmitCompilerGeneratedFiles`) before and after — `AsciiHash.generated.cs` (1,835 lines) is **byte-for-byte identical**; full solution builds with zero warnings; RESPite.Tests pass (1717/1717).

Closes #3033

### DIFF
--- a/eng/StackExchange.Redis.Build/AsciiHashGenerator.cs
+++ b/eng/StackExchange.Redis.Build/AsciiHashGenerator.cs
@@ -441,29 +441,27 @@ public class AsciiHashGenerator : IIncrementalGenerator
         sb.AppendLine("using StackExchange.Redis;");
         sb.AppendLine("#pragma warning disable CS8981, SER004");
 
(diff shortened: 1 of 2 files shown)


mgravell:
Seems fine.. similar to some related code in my other libs... I can't think of a reason *not* to...

mgravell (approved):

PR #3122, opened 2026-07-07
Throw a clear error when an Execute command contains whitespace

### LINKED ISSUE #2689: Execute should check for whitespace (was: ACL Management)
I am trying to manage ACLs programmatically, but can't seem to figure out how to do it. Is that possible at this point? I'm aware that ACLs aren't supported yet, but I'm not sure why executing raw commands are an issue.

Something as simple as this:

```
database.Execute("ACL SETUSER test");
```
Results in this error:
```
StackExchange.Redis.RedisServerException: 'ERR unknown command 'ACL SETUSER TEST', with args beginning with: '
```

All the ACL commands seem to do this.

Longer commands like
```
$"ACL SETUSER {username} on >{password} +@subscribe +@psubscribe &mychannels*"
```
results in 
```
System.ArgumentOutOfRangeException: 'Specified argument was out of the range of valid values. (Parameter 'Command 'ACL SETUSER  on > +@subscribe +@psubscribe &mychannels*' exceeds library limit of 31 bytes')'
```

If ACLs worked at all, I could get around that latter error by making multiple, shorter commands, though I'd prefer running a single command.

arpitjain099:
Fixes #2689, which you reopened as a usability bug. Calling something like `db.Execute("ACL SETUSER x")` passes the whole line as a single command token, so it goes out as one unknown command and the only feedback is an opaque `ERR unknown command` from the server. Since a redis command token never has internal whitespace, that call is always a mistake, so the `ExecuteMessage` constructor now checks for it up front (right alongside the existing too-many-args and command-disabled guards) and throws a `RedisCommandException` that names the command and points at the correct token-per-argument form.

I couldn't run the integration suite locally without a server, but the guard fires purely client-side in the constructor, so I added unit tests to `AdhocMessageRoundTrip` that build the message directly (matching the existing disabled-command tests): a few whitespace inputs throw, and `Execute("ACL", "SETUSER", "x")` still constructs fine.

Happy to switch to auto-splitting instead if you'd prefer that direction, but your reopen comment read as leaning toward the throw, so I went with that.

### DIFF
--- a/src/StackExchange.Redis/ExceptionFactory.cs
+++ b/src/StackExchange.Redis/ExceptionFactory.cs
@@ -30,6 +30,9 @@ namespace StackExchange.Redis
         internal static Exception TooManyArgs(string command, int argCount)
             => new RedisCommandException($"This operation would involve too many arguments ({argCount + 1} vs the redis limit of {MessageWriter.REDIS_MAX_ARGS}): {command}");
 
+        internal static Exception CommandHasWhitespace(string command)
+            => new RedisCommandException($"The command '{command}' contains whitespace and would be sent as a single unknown token; pass each word as a separate argument, for example Execute(\"ACL\", \"SETUSER\", \"x\") rather than Execute(\"ACL SETUSER x\").");
+
         internal static Exception ConnectionFailure(bool includeDetail, ConnectionFailureType failureType, string message, ServerEndPoint? server)
         {
             var ex = new RedisConnectionException(failureType, message);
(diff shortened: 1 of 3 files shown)


mgravell (approved):

mgravell:
Intent is great, but IMO we should restrict ourselves to the "obvious" case, `' '` - looping char-by-char testing for every possible white-space condition is more expensive than we really want, with a non-trivial impact for libs like NRedisStack; I've tweaked to limit it to simple space.

PR #3125, opened 2026-07-10
Support overriding library version

philon-msft:
Allow customizing the reported library version via a DefaultOptionsProvider. Intended for use by the Microsoft.Azure.StackExchangeRedis extension package because knowing the extension's version is usually more helpful than the SE.Redis version when diagnosing issues customers are having. 

This PR also aligns behavior of LibraryName with other ConfigurationOptions properties in terms of defaulting to a value from the current DefaultOptionsProvider.

### DIFF
--- a/src/StackExchange.Redis/PublicAPI/PublicAPI.Shipped.txt
+++ b/src/StackExchange.Redis/PublicAPI/PublicAPI.Shipped.txt
@@ -1693,7 +1693,6 @@ static StackExchange.Redis.Configuration.DefaultOptionsProvider.AddProvider(Stac
 static StackExchange.Redis.Configuration.DefaultOptionsProvider.ComputerName.get -> string!
 static StackExchange.Redis.Configuration.DefaultOptionsProvider.GetProvider(StackExchange.Redis.EndPointCollection! endpoints) -> StackExchange.Redis.Configuration.DefaultOptionsProvider!
 static StackExchange.Redis.Configuration.DefaultOptionsProvider.GetProvider(System.Net.EndPoint! endpoint) -> StackExchange.Redis.Configuration.DefaultOptionsProvider!
-static StackExchange.Redis.Configuration.DefaultOptionsProvider.LibraryVersion.get -> string!
 static StackExchange.Redis.ConfigurationOptions.Parse(string! configuration) -> StackExchange.Redis.ConfigurationOptions!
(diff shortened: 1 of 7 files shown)


src/StackExchange.Redis/PublicAPI/PublicAPI.Shipped.txt
```diff
@@ -1693,7 +1693,6 @@ static StackExchange.Redis.Configuration.DefaultOptionsProvider.AddProvider(Stac
 static StackExchange.Redis.Configuration.DefaultOptionsProvider.ComputerName.get -> string!
 static StackExchange.Redis.Configuration.DefaultOptionsProvider.GetProvider(StackExchange.Redis.EndPointCollection! endpoints) -> StackExchange.Redis.Configuration.DefaultOptionsProvider!
 static StackExchange.Redis.Configuration.DefaultOptionsProvider.GetProvider(System.Net.EndPoint! endpoint) -> StackExchange.Redis.Configuration.DefaultOptionsProvider!
-static StackExchange.Redis.Configuration.DefaultOptionsProvider.LibraryVersion.get -> string!
```
philon-msft, line 1696: @mgravell is there any chance this change is "okay"? I really want to change `LibraryVersion` from a `protected static` to `public virtual`, but that means removing this line from the shipped API. 

If this is too much of a break, I can either a) leave the static and add a new instance property with a different name, or b) have the extension package set `ConfigurationOptions.LibraryVersion` directly rather than bubbling it up form its custom DefaultOptionsProvider. But that's a weird divergence from the other options and risks interacting with customers' use of `ConfigurationOptions.LibraryVersion`

mgravell:
This gets awkward when multiple libraries think they're king. For example, NRedisStack also tries hard to mess with this. We discussed this about a year ago: https://github.com/StackExchange/StackExchange.Redis/discussions/2937 - is it perhaps worth us resurrecting that sort of approach, perhaps with better defaults-provider support?

src/StackExchange.Redis/PublicAPI/PublicAPI.Shipped.txt
mgravell: The problem isn't changing the shipped file - the problem is the `MissingMethodException` etc at runtime when someone builds against one version and gets given another - this applies doubly for libraries that package SE.Redis such as azure. I think maybe the right thing here is to accept a sucky name for a new member.

Issue #1891, opened 2021-10-21
Problems with IP change of Redis instances in failure scenarios


ThomasTosik:
Hi,

we are testing failure scenarios and have some issues that we can not resolve. We are killing a Redis pod and some seconds later a new pod is spawned with a new IP. The Sentinel and Redis conglomerate has no issue with this scenario as far as we can tell from the logs. The newly joined Redis pod will be detected. Also a simulated Sentinel outage seems not problematic. Our application is also running without issues as long as Redis stays up.
The StackExchange.Redis initially notices that the old Redis pod is not available and reports correctly "StackExchange.Redis.RedisConnectionException: No connection is active/available to service this operation". But there is no switch to one of the other running Redis. He insists on using the old IP of the not anymore existing Redis pod. Even after the new Redis pod is available (with a new IP).
We can reliably produce this situation on all of our Kuberntes clusters. The only thing that currently works is restarting the affected application pod.
It seems we are missing something either in our Sentinel/Redis Cluster config or some secret in StackExchange.Redis ;).

Architecture:
3 Sentinels and 3 Redis in Kubernetes

Redis operator:
https://github.com/spotahome/redis-operator
redis alpine image 5.0.14

StackExchange.Redis version
2.2.79

We tried some client related stuff from other issues with no effect:
- abortConnect and resolveDns had no effect
- Register on ConnectionFailed events and try Configure or PublishReconfigure

Log from starting sequence:
`[DEBUG] 2021-10-20 02:49:40 StartupBase (null) (null) (null) (null) (null)      Configuring Redis Sentinel usage @[rfs-redisfailover-persistent.redis.svc.cluster.local:26379,abortConnect=false,resolveDns=true]
Connecting (sync) on .NET 5.0.11
[...]

PR #3130, opened 2026-07-13
Sentinel: re-elect primary on failover between already-known nodes

rebaseandpanic:
On an in-place Sentinel failover (the current primary is demoted to a replica
without changing endpoint, and a known replica is promoted), the client never
switches to the new primary and keeps writing to the demoted node until restart.
Seen in production on a Sentinel-managed Valkey deployment on Kubernetes (pods
not recreated): a Redis Streams consumer stayed pinned to the demoted node for
hours until the process was restarted.

`SwitchPrimary` only reconfigured when the new primary was an endpoint it had
never seen (`!servers.Contains(...)`). When the promoted node is already known,
that guard is false and it returns without re-electing. A plain reconfigure also
isn't enough: with `reconfigureAll: false` an already-connected node doesn't
re-read its role, so the election just re-picks the stale node.

Fix: when a known primary's cached view is stale, reconfigure with
`reconfigureAll: true` so connected nodes re-read their role first (the guard is
self-clearing — no reconfigure storm). Also run that reconfigure outside the
`sentinelConnectionChildren` lock so one service's failover doesn't block others.

Refs #1891 — the failure logs there show this known-endpoint failover, despite
the title mentioning IP changes; the separate new-primary-IP path is untouched.

### DIFF
--- a/src/StackExchange.Redis/ConnectionMultiplexer.Sentinel.cs
+++ b/src/StackExchange.Redis/ConnectionMultiplexer.Sentinel.cs
@@ -40,6 +40,13 @@ public partial class ConnectionMultiplexer
                     // We don't care about the result of this - we're just trying
                     _ = Format.TryParseEndPoint(string.Format("{0}:{1}", messageParts[1], messageParts[2]), out var switchBlame);
 
+                    // Resolve the child under the lock, but run the (potentially multi-second,
+                    // blocking) SwitchPrimary reconfigure OUTSIDE it: otherwise a failover for one
+                    // service would block +switch-master handling and GetSentinelMasterConnection
+                    // for every other service sharing this sentinel connection. SwitchPrimary is
+                    // already invoked without this lock from the reconnect timer and the restore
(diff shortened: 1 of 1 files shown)


mgravell:
any preference between #3130 and #3131?

rebaseandpanic:
I noticed you keep both a v2 and a v3/main line, so I figured a PR into each made sense (#3130 → main, #3131 → v2) — it's the exact same commit, cherry-picked, so they're identical.

I opened the v2 one specifically because we're running the 2.x line in production and hit this bug in the wild there, so having the fix land in v2 matters to us directly — not just main.

That said, I see contributions here normally target main only, so if you'd prefer to handle the v2 backport yourself, I'm happy to close #3131 and keep just #3130. Whatever's easiest for you.

mgravell:
ah, I hadn't noticed that branch detail! sorry, the cost of reading things in a hurry; I'll try to review, thanks

mgravell (approved):
LGTM, thanks

PR #3131, opened 2026-07-13
Sentinel: re-elect primary on failover between already-known nodes

rebaseandpanic:
On an in-place Sentinel failover (the current primary is demoted to a replica
without changing endpoint, and a known replica is promoted), the client never
switches to the new primary and keeps writing to the demoted node until restart.
Seen in production on a Sentinel-managed Valkey deployment on Kubernetes (pods
not recreated): a Redis Streams consumer stayed pinned to the demoted node for
hours until the process was restarted.

`SwitchPrimary` only reconfigured when the new primary was an endpoint it had
never seen (`!servers.Contains(...)`). When the promoted node is already known,
that guard is false and it returns without re-electing. A plain reconfigure also
isn't enough: with `reconfigureAll: false` an already-connected node doesn't
re-read its role, so the election just re-picks the stale node.

Fix: when a known primary's cached view is stale, reconfigure with
`reconfigureAll: true` so connected nodes re-read their role first (the guard is
self-clearing — no reconfigure storm). Also run that reconfigure outside the
`sentinelConnectionChildren` lock so one service's failover doesn't block others.

Refs #1891 — the failure logs there show this known-endpoint failover, despite
the title mentioning IP changes; the separate new-primary-IP path is untouched.

### DIFF
--- a/src/StackExchange.Redis/ConnectionMultiplexer.Sentinel.cs
+++ b/src/StackExchange.Redis/ConnectionMultiplexer.Sentinel.cs
@@ -41,6 +41,13 @@ public partial class ConnectionMultiplexer
                     // We don't care about the result of this - we're just trying
                     _ = Format.TryParseEndPoint(string.Format("{0}:{1}", messageParts[1], messageParts[2]), out var switchBlame);
 
+                    // Resolve the child under the lock, but run the (potentially multi-second,
+                    // blocking) SwitchPrimary reconfigure OUTSIDE it: otherwise a failover for one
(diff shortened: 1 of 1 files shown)


mgravell:
any preference between #3130 and #3131?

rebaseandpanic:
Replied in #3130 to keep it in one thread — short version: same commit as this, targeting v2 because we run 2.x in production; happy to close this one if you'd rather backport yourself.

mgravell (approved):
LGTM, thanks

Issue #2990, opened 2025-12-22
MOVED redirects to the same endpoint are not retried


barshaul:
Hey Marc,

I came across an issue while working with SE.Redis and wanted to share it with you.

In some managed Redis environments, clients connect via a stable DNS endpoint that fronts multiple backend nodes or proxies. That same DNS address is also used as the node address returned in the cluster topology (e.g., via CLUSTER SLOTS). Importantly, while the DNS name stays constant, the actual backend instance a client connects to may change over time.

During certain maintenance or backend lifecycle events, the system may intentionally respond to a client request with a MOVED error pointing back to the same DNS address, and then close the connection. The intention is to signal to the client that it’s safe to retry: upon retry, the client reconnects via DNS and establishes a new connection to a healthy backend, allowing the operation to succeed.

The issue is that SE.Redis suppresses retries for MOVED errors when the redirect target matches the address the error came from (see [code link](https://github.com/StackExchange/StackExchange.Redis/blob/main/src/StackExchange.Redis/ResultProcessor.cs#L262)). In this scenario, the MOVED error is immediately propagated back to the application. However, if the client were to retry, it would discover that the existing connection is no longer valid, reconnect via DNS, and complete the operation successfully. This retry-based recovery is the behavior I’ve observed in all other major Redis clients I tested.

I understand that the current behavior is likely intended to prevent circular MOVED loops, which is a reasonable safeguard. However, it also blocks legitimate recovery scenarios where a MOVED redirect to the “same” address (from the client’s perspective) is expected and necessary, because the address is a DNS alias rather than a single fixed node.

To support this use case while still avoiding infinite redirect loops, I see a couple of possible approaches:
	•	Remove the current restriction and introduce a configurable limit on redirect attempts.
	•	Add a configuration option that explicitly allows retries on “self-pointing” MOVED redirects.

I’d be happy to help contribute a fix or configuration option if this is something you’d be open to considering. Before going further, I wanted to check whether this direction makes sense to you.

Thanks!

PR #3134, opened 2026-07-21
fix: Queue batch commands to backlog during MOVED reconnection

advaMosh:
## Problem

When a MOVED-to-same-endpoint triggers reconnection (as implemented in PR #3003), batch commands (`IBatch`) issued during the ~50-100ms reconnection window fail immediately with `RedisConnectionException: NoConnectionAvailable`. This is because:

1. `RedisBatch.Execute()` calls `SelectServer()` which returns null for disconnected/reconnecting servers.
2. `PhysicalBridge.TryEnqueue()` returns false when `!IsConnected`, unlike `TryWriteAsync`/`TryWriteSync` which queue to the backlog via `QueueOrFailMessage()` → `BacklogEnqueue()`.

This creates an inconsistency: individual async commands (e.g. `db.StringSetAsync()`) are queued and succeed after reconnection, but batch commands throw immediately during the same reconnection window.

## Impact

- Applications using `IBatch` see 1-3 `RedisConnectionException` errors per MOVED redirect, while individual async commands see zero errors in the same scenario.
- This affects deployments where Redis is behind DNS records, load balancers, or proxies that use MOVED-to-same-endpoint.

## Fix

**`RedisBatch.Execute()`**: When `SelectServer()` returns null and `BacklogPolicy.QueueWhileDisconnected` is enabled, retry with `allowDisconnected: true` to locate the server endpoint. This mirrors the pattern already used in `PrepareToPushMessageToBridge()`.

**`PhysicalBridge.TryEnqueue()`**: When `!IsConnected || NeedsReconnect` and `BacklogPolicy.QueueWhileDisconnected` is enabled, queue messages to the backlog instead of returning false. This matches the existing behavior of `TryWriteSync()` and `TryWriteAsync()` for individual commands.

## Testing

Tested with a proxy that sends MOVED-to-same-endpoint followed by connection close:
- **Before fix**: 1-3 `NoConnectionAvailable` errors per redirect event (batch.Execute throws during ~50-100ms reconnection window)
- **After fix**: 0 errors — batch commands are queued to backlog and sent after reconnection completes

The fix only activates when `BacklogPolicy.QueueWhileDisconnected` is true (the default), preserving existing behavior for users who have disabled backlog queuing.

## Related

- #3003: Handle MOVED error pointing to same endpoint (the original MOVED reconnect fix)
- #2990: Original issue report for MOVED-to-same-endpoint

### DIFF
--- a/src/StackExchange.Redis/PhysicalBridge.cs
+++ b/src/StackExchange.Redis/PhysicalBridge.cs
@@ -730,13 +730,39 @@ namespace StackExchange.Redis
 
             if (isDisposed) throw new ObjectDisposedException(Name);
 
-            if (!IsConnected)
+            if (!IsConnected || NeedsReconnect)
             {
+                // During reconnection (e.g. MOVED-to-same-endpoint), queue messages to the backlog
+                // so they can be sent once the connection is re-established.
+                // This matches the behavior of TryWriteAsync/TryWriteSync for individual commands.
+                if (Multiplexer.RawConfig.BacklogPolicy.QueueWhileDisconnected)
+                {
+                    foreach (var message in messages)
+                    {
+                        message.SetEnqueued(null);
+                        BacklogEnqueue(message);
+                    }
+                    StartBacklogProcessor();
+                    return true;
+                }
                 return false;
             }
 
             var physical = this.physical;
-            if (physical == null) return false;
+            if (physical == null)
+            {
+                if (Multiplexer.RawConfig.BacklogPolicy.QueueWhileDisconnected)
+                {
+                    foreach (var message in messages)
+                    {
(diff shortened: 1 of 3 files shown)


mgravell (approved):

advaMosh:
hey @mgravell , thank you for the quick response !
I see the failing tests in the failing check are flaky regardless to this PR. can / will it be merged in that state? 
also - Do you have an estimate for when this PR is expected to be merged, and which SE.R versions it’s likely to be included in?

mgravell:
The CI has a bit of noise, sadly. I've poked it. I'll do another eyeball tomorrow, but there's no reason to delay this any - we tend to build fairly regularly when there's reason to.

PR #3140, opened 2026-07-29
Add possibility to use different passwords for Sentinel and Redis host (#1698)

doosterkamp:
Adding the possibility to use different passwords for Sentinel and Redis host.
Fixing the long standing issue #1698 

## Checklist

- [X] I fully and freely contribute this code in accordance with the [project license](../LICENSE) (and am legally able to do so)  
- [X] I take responsibility for this contribution's quality and correctness, including any portions produced with AI assistance (see [CONTRIBUTING.md](../CONTRIBUTING.md)).


src/StackExchange.Redis/ConfigurationOptions.cs
```diff
@@ -209,7 +210,7 @@ private enum OptionFlags : ulong
 
         private OptionFlags optionFlags;
 
-        private string? tieBreaker, sslHost, configChannel, user, password;
+        private string? tieBreaker, sslHost, configChannel, user, sentinelUser, password, sentinelPassword;
```
mgravell, line 224: note there's a fields unit test that checks ConfigurationOptions; the idea is to remind me to check things like Clone() and Clear(), but it'll fail until added explicitly

src/StackExchange.Redis/ConfigurationOptions.cs
```diff
@@ -941,7 +962,9 @@ public static ConfigurationOptions Parse(string configuration, bool ignoreUnknow
             defaultVersion = defaultVersion,
             connectTimeout = connectTimeout,
             user = user,
+            SentinelUser = SentinelUser,
```
mgravell: this should be reading and assigning the fields, not the properties - otherwise it will appear that they *actively* have values, rather than *passively* (if you see what I mean)

mgravell: also, please wipe the fields in Clear(); I regret that Clear() exists, but it does...

src/StackExchange.Redis/ConfigurationOptions.cs
```diff
@@ -1040,7 +1063,9 @@ public string ToString(bool includePassword)
             Append(sb, OptionKeys.Version, defaultVersion);
             Append(sb, OptionKeys.ConnectTimeout, OptionFlags.ConnectTimeoutHasValue, in connectTimeout);
             Append(sb, OptionKeys.User, user);
+            Append(sb, OptionKeys.SentinelUser, sentinelUser);
             Append(sb, OptionKeys.Password, (includePassword || string.IsNullOrEmpty(password)) ? password : "*****");
+            Append(sb, OptionKeys.SentinelPassword, (includePassword || string.IsNullOrEmpty(SentinelPassword)) ? sentinelPassword : "*****");
```
mgravell: this should be checking the field, not the property; i.e. if I assign just .Password, I *do not* expect it to start showing in the config-string as the SentinelPassword=... value

mgravell: (actually, this is only used for the empty check, but the point remains - the check should be consistent in the IsNullOrEmpty and value part)

src/StackExchange.Redis/ConnectionMultiplexer.Sentinel.cs
```diff
@@ -144,7 +144,12 @@ public static Task<ConnectionMultiplexer> SentinelConnectAsync(ConfigurationOpti
     /// <param name="log">The <see cref="TextWriter"/> to log to.</param>
     private static ConnectionMultiplexer SentinelPrimaryConnect(ConfigurationOptions configuration, TextWriter? log = null)
     {
-        var sentinelConnection = SentinelConnect(configuration, log);
+        // Use separate sentinel credentials when provided on the configuration
+        var sentinelConfig = configuration.Clone();
```
mgravell, line 148: this feels ... weird; I wonder whether what we *should* be doing is making sure that ConnectionMultiplexer stores the `serverType` (which is already passed in), or (preferable) just a `private readonly bool _isSentinel`, and using *that* during handshake to *read* from the appropriate ~fields~ *properties* of the non-mutated options; thoughts?

mgravell (commented):
Concept is great, +1; a few nits in the `ConfigurationOptions` piece just because that type is super weird (not your fault), and I'm not a huge fan of the clone usage - I welcome your thoughts there.

src/StackExchange.Redis/ConfigurationOptions.cs
mgravell:
```
    Failed ConfigTests.ExpectedFields (RESP2) [22 ms]
    Error Message:
     Assert.Equal() Failure: Collections differ at index 28
             ↓ (pos 0)
  Expected: "ServiceName"
  Actual:   "sentinelPassword"
             ↑ (pos 0)
    Stack Trace:
       at StackExchange.Redis.Tests.ConfigTests.ExpectedFields() in /_/tests/StackExchange.Redis.Tests/ConfigTests.cs:line 62
```

doosterkamp:
> Concept is great, +1; a few nits in the `ConfigurationOptions` piece just because that type is super weird (not your fault), and I'm not a huge fan of the clone usage - I welcome your thoughts there.

Thanks!
I will try to fix the 'nits' today! We really need this fix and are testing with these changes as we speak.
I'm not a huge fan of the clone usage either, but it does the job for now, without a 'big' rewrite of the under-laying code ( I started trying to change that, but it got a lot bigger, very fast, so I stepped back for a moment).

src/StackExchange.Redis/ConfigurationOptions.cs
doosterkamp: I missed that one ... sorry
I added the sentinel credential fields to the test.

src/StackExchange.Redis/ConfigurationOptions.cs
doosterkamp: Sometimes I want to go to fast ...
I corrected and double-checked the usage of fields vs. properties. These should be correct now.
And also wiping the Sentinel user/pw fields in the Clear method now,

src/StackExchange.Redis/ConfigurationOptions.cs
doosterkamp: Again, wanted to go to fast ... sorry.
Has been corrected.

By the way; is there a reason you use IsNullOrEmpty and not IsNullOrWhitespace? 
I always liked IsNullOrWhitpace a little more, as it also checks for (unwanted?) spaces, tabs, etc.

src/StackExchange.Redis/ConnectionMultiplexer.Sentinel.cs
doosterkamp: I like the thought of ```private readonly bool _isSentinel```!
It makes the intend more clear and it should make it more robust and threadsafe, avoiding issues with the handshake.

It looked like a small change, so I took the liberty to apply this (I've put it in a separate commit, so it's an easy roll-back if you don't like it)

Also, I have rolled back the cloning part and took a better look at how and where the connection was set up.
It turns out to be a small change also, so I moved the assignment of the (Sentinel) user/pw to the HandshakeAsync method.

doosterkamp:
@mgravell is there a nice/correct way to clear the CodeQL security failure or do I need to add an inline suppression message?

src/StackExchange.Redis/ConfigurationOptions.cs
mgravell: Is `"\t \u00A0"` a valid password? And is that answer on Redis? Garnet? Valkey? Something else more exotic? I *have no clue*, honestly, and I'm going to consider it "not my problem" ;p

mgravell:
I have dismissed the CodeQL alert; the alert is "right" in that this should be a considered decision, but this is an entirely consistent and appropriateusage.

src/StackExchange.Redis/PublicAPI/PublicAPI.SentinelUnshipped.txt
```diff
@@ -0,0 +1,5 @@
+#nullable enable
```
mgravell: this should all just be in PublicAPI.Unshipped.txt - it isn't feature-specific

src/StackExchange.Redis/ConfigurationOptions.cs
```diff
@@ -941,7 +962,9 @@ public static ConfigurationOptions Parse(string configuration, bool ignoreUnknow
             defaultVersion = defaultVersion,
             connectTimeout = connectTimeout,
             user = user,
+            SentinelUser = sentinelUser,
```
mgravell:
```suggestion
            sentinelUser = sentinelUser,
```

src/StackExchange.Redis/ConfigurationOptions.cs
```diff
@@ -941,7 +962,9 @@ public static ConfigurationOptions Parse(string configuration, bool ignoreUnknow
...
             password = password,
+            SentinelPassword = sentinelPassword,
```
mgravell:
```suggestion
            sentinelPassword = sentinelPassword,
```

src/StackExchange.Redis/ServerEndPoint.cs
```diff
@@ -957,13 +957,15 @@ private async Task HandshakeAsync(PhysicalConnection connection, ILogger? log)
                 Multiplexer.Trace("No connection!?");
                 return;
             }
+
             Message msg;
-            // Note that we need "" (not null) for password in the case of 'nopass' logins
             var config = Multiplexer.RawConfig;
-            string? user = config.User;
-            string password = config.Password ?? "";
+            var isSentinelConnection = Multiplexer.ServerSelectionStrategy.ServerType == ServerType.Sentinel;
```
mgravell: should this not be checking something like `Multiplexer.IsSentinel`, i.e. that `bool` we just added? Alternatively: if the code shown *is* valid: we possibly don't need the explicit bool, and a property like

```
internal bool IsSential => ServerSelectionStrategy.ServerType is ServerType.Sentinel;
```

is fine, without the explicit `bool` ? In either case: we should be consistent in the decision process.

mgravell (changes requested):
super close now; sorry to be pedantic on some of these - I'm not trying to frustrate you

mgravell:
(note Windows CI may stall due to a packaging bug in 8.10 - server folks are looking into it; I've added a workaround in `main` (#3145), but I don't think this branch includes that yet; basically: don't panic if Windows CI fails in truly epic ways)

doosterkamp:
> super close now; 

Does not feel like it :P
Testing in our setup fails during handshake (sees serverType == ServerType.Standalone), trying to fix this now 

> sorry to be pedantic on some of these - I'm not trying to frustrate you

Don't worry about that! Better safe than sorry ...

src/StackExchange.Redis/PublicAPI/PublicAPI.SentinelUnshipped.txt
doosterkamp: Okay .. Removed this file

doosterkamp:
I moved back to overwriting the user/pw with the Sentinel credentials in the `SentinelPrimaryConnect(Async)` method in the `ConnectionMultiplexer(.Sentinel)`. Even though I'm not a big fan of the use of `Clone()`, this is simply working well in my environment.

I tried to implement the usage of the Sentinel credentials in the `HandshakeAsync` method (in `ServerEndPoint`), but I lose track of the origin of the call in spaghetti chain of calls leading to the handshake.
Yes, I can see if the configuration is a Sentinel setup (using the created `IsSentinel` property), but not if it's originated from `sentinelPrimaryConnect` (which needs the Sentinel credentials in the handshake) of `GetSentinelMasterConnection` (which needs the Redis credentials in the handshake). After a few days of trying and debugging I gave up and reverted back to my first solution.

Maybe this can be accepted as a stable and working solution (tested this build thorough in our corporate OTA environment), as we really need this solution in the short term.

src/StackExchange.Redis/ServerEndPoint.cs
doosterkamp: See my comment, later on in this PR, about reverting back to setting the user/pw in de ConnectionMultiplexer.

mgravell (approved):

src/StackExchange.Redis/ConfigurationOptions.cs
```diff
@@ -2,7 +2,6 @@
 using System.Buffers;
 using System.Collections.Generic;
 using System.ComponentModel;
-using System.Diagnostics.CodeAnalysis;
```
mgravell, line 5: this is needed

mgravell:
OK; we *do* need the `using` directive that got culled, for [`Experimental`](https://learn.microsoft.com/dotnet/api/system.diagnostics.codeanalysis.experimentalattribute) - let me see if I can push to the branch...

src/StackExchange.Redis/PublicAPI/PublicAPI.Shipped.txt
```diff
@@ -287,6 +287,10 @@ StackExchange.Redis.ConfigurationOptions.ResolveDns.get -> bool
 StackExchange.Redis.ConfigurationOptions.ResolveDns.set -> void
 StackExchange.Redis.ConfigurationOptions.ResponseTimeout.get -> int
 StackExchange.Redis.ConfigurationOptions.ResponseTimeout.set -> void
+StackExchange.Redis.ConfigurationOptions.SentinelUser.get -> string?
```
mgravell: not trying to be a pain, but these should really be in `PublicAPI.Unshipped.txt` (because it is currently: not shipped); if I manage to push to your remote, I'll fix it; I'm not going to be a tyrant and insist that we fix this - just trying to offer helpful feedback

mgravell:
(sorry for force-push; repairing after a git identity failure)

mgravell:
https://github.com/StackExchange/StackExchange.Redis/releases/tag/3.1.13

src/StackExchange.Redis/PublicAPI/PublicAPI.Shipped.txt
doosterkamp: Don't worry about it ...
I'm not so familiar (yet) with Ms CodeAnalysis, so my apologies for the fixes you have to apply

doosterkamp:
Thank you for the quick responses, feedback and release.

PR #3174, opened 2026-08-13
Preparation work for smart-client-handoffs

mgravell:
Smart -client-handoffs needs prep work:

- [x] make sure unknown PUSH messages don't break anything
- [x] ensure `-MOVED` with *unknown* routing: works (as in: fails correctly, and doesn't attempt to route to `?`) - new `RedisErrorKind`
- [x] (nice-to-have) expose CLUSTER SLOTS on the public API
- [x] (nice-to-have) expose CLUSTER SLOTS via the export API (noted in intellisense for existing Cluster option)
- [x] (nice-to-have) handle the extra up-level details in CLUSTER NODES that we currently discard
- [x] toy-server: support CLUSTER SLOTS and CLUSTER NODES in all naming configurations, with per-server configuration
- [x] toy-server: allow forced unknown `-MOVED` targets
- [x] toy-server: allow connect by the different naming conventions, including TLS auth

Moving to a follow-up more invasive PR:

- [ ] support richer CLUSTER topology states, primarily via enabling CLUSTER SLOTS for our own topology building
  - [ ] (fetch code present and inert, needs reactivating)
  - [ ] not *strictly* required, but thematically related: implement endpoint pruning of things *discovered* via CLUSTER NODES/SLOTS (as opposed to declared via `ConfigurationOptions`), which then migrate out of use
- [ ] allow routing to nodes via any known identity
- [ ] ensure client TLS authenticates correctly via IP and DNS
- [ ] retire (drain etc) duplicated connections when identity rules resolve two connections to be the same

## Checklist


mgravell:
Follow-up raised as #3177 - the invasive half: `CLUSTER SLOTS` driving the slot map and endpoint discovery, identity resolution, and the endpoint lifecycle (retirement with drain semantics, pruning, duplicate merge).

PR #3178, opened 2026-08-17
Add SER305/SER306: waiting for a queued transaction result

mgravell:
Commands queued on an `ITransaction` or `IBatch` are buffered, not sent. The task comes back incomplete and stays that way until `Execute[Async]`, so awaiting one at the point of queueing waits for something that can only happen further down the method:

```c#
var tran = db.CreateTransaction();
var value = await tran.StringGetAsync(key);   // never completes - nothing has been sent yet
await tran.ExecuteAsync();                    // never reached
```

This is the mistake "always await your tasks" actively teaches, and it fails silently rather than loudly, so **SER305 is an error** rather than a suggestion. It can be one without hedging because the shape is certain by construction: the wait sits at the queueing site, so no arrangement of the surrounding code rescues it — even a prior `Execute` is no help, since *this* command was queued after it. That is also why the flow-sensitive shapes (a task captured to a local, or gathered by `Task.WhenAll`, then awaited) are deliberately **not** reported — awaiting a captured task *after* `Execute` is the recommended fix, and an error that gets the order wrong condemns correct code.

**SER306** is the one case that survives the wait. Fire-and-forget hands back an already-completed task carrying the default value, so awaiting it is legal — it just never carries the server's answer, whenever it is read. Its own ID, and a warning, so that deliberate fire-and-forget code can silence it without silencing the rule that says a transaction cannot work. The only guidance there is "discard it".

Covers `await`, `.Result`, `.Wait()` and `.GetAwaiter().GetResult()`, on both transactions and batches. Two code fixes are offered — discard the result, or capture it and await after `Execute` — in a new `StackExchange.Redis.CodeFixes` assembly packed alongside the analyzer (a fix needs Workspaces, which csc does not ship and the IDE does).

### Two decisions that were not obvious

- **Terminators are recognised by declaring interface, never by name.** `ITransaction[Async].ExecuteAsync(CommandFlags)` must be awaited, while `IDatabaseAsync.ExecuteAsync(string, ...)` — the raw-command escape hatch — is an ordinary queued command that must still be reported. A name match on `ExecuteAsync` gets both exactly the wrong way round. The receiver is not always the interface either: a decorator class implementing it resolves to the *class's* member, which this repo's own build caught on `KeyPrefixedTransaction`.
- **A non-constant flags expression is read only far enough to prove `FireAndForget` present** — `flags | CommandFlags.FireAndForget`, since `|` only sets bits. There is deliberately no matching treatment of `&`/`~` to prove it *absent*: that would feed SER305, and an error must not rest on an expression we only partly understood.

### A retry bug this turned up

Writing the tests for SER306 exposed a real divergence, fixed in the first commit. `RetryTransaction` records each queued call and hands back a durable proxy task, and did so regardless of the flags — which it could not see, since they are captured inside the generated per-command state. So a fire-and-forget operation stayed incomplete until `ExecuteAsync` forwarded it, where a plain `RedisTransaction` hands back an already-completed task. Identical caller code returned instantly on one and blocked for good on the other, which defeats the stated goal of letting `ITransaction` code move onto a retrying database unchanged.

The generated captured-arguments struct now exposes its `CommandFlags` through `IFlaggedRedisArgs` — following `IMappableRedisArgs`, which exists for the same reason — and the proxy is settled up front when the flag is set. The operation is still recorded and replayed like any other: fire-and-forget declines the reply, not the command.

### Verification

- `QueuedResultTests` pins the underlying behaviour against a live server, since both rules rest on it: a queued task does not complete before `Execute`, a fire-and-forget one is already complete and *still executes*, for transactions, batches and retrying transactions.
- 135 analyzer/code-fix tests green, including the negative cases — which matter more than usual here, an error being a broken build.
- Dogfooded: the whole repo builds clean with the rule live, and the one thing it did flag was a genuine false positive in the rule, now fixed and covered.
- `dotnet pack` verified to ship both `StackExchange.Redis.Build.dll` and `StackExchange.Redis.CodeFixes.dll` under `analyzers/dotnet/cs`, with no Workspaces dll carried along.

Docs: new `docs/rules/SER305.md` and `SER306.md`, `docs/rules/index.md` updated (the family is no longer uniformly warnings-about-working-code, so the wording around severity and the shared quiet-list needed adjusting), and a "Don't await inside the transaction" section in `docs/Transactions.md`.

### DIFF
--- a/Directory.Packages.props
+++ b/Directory.Packages.props
@@ -33,6 +33,9 @@
     <!-- analyzer test harness; deliberately newer Roslyn than the shipped floor above (the harness needs it,
          and we cannot test at 4.3 without pinning the whole world) -->
     <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.Analyzer.Testing" Version="1.1.2" />
+    <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.CodeFix.Testing" Version="1.1.2" />
+    <!-- StackExchange.Redis.CodeFixes overrides this down to the shipped floor (4.3.0); the version here is
+         the one the test harness needs, exactly as for the analyzer reference above -->
     <PackageVersion Include="Microsoft.CodeAnalysis.CSharp.Workspaces" Version="4.14.0" />
     <PackageVersion Include="Microsoft.CodeAnalysis.PublicApiAnalyzers" Version="5.6.0" />
     <PackageVersion Include="Microsoft.NETFramework.ReferenceAssemblies" Version="1.0.3" />
--- a/StackExchange.Redis.slnx
+++ b/StackExchange.Redis.slnx
@@ -1,6 +1,7 @@
 <Solution>
   <Folder Name="/eng/">
     <Project Path="eng/StackExchange.Redis.Build/StackExchange.Redis.Build.csproj" />
+    <Project Path="eng/StackExchange.Redis.CodeFixes/StackExchange.Redis.CodeFixes.csproj" />
   </Folder>
   <Folder Name="/Solution Items/">
     <File Path=".editorconfig" />
--- a/docs/Transactions.md
+++ b/docs/Transactions.md
@@ -117,6 +117,38 @@ var wasSet = (bool) db.ScriptEvaluate(@"if redis.call('hexists', KEYS[1], 'Uniqu
 
 (note that the response from `ScriptEvaluate` and `ScriptEvaluateAsync` is variable depending on your exact script; the response can be interpreted by casting - in this case as a `bool`)
 
+Don't await inside the transaction
+---
+
+This is the most common mistake with transactions, and it does not fail loudly - it hangs.
+
+Commands queued on an `ITransaction` (or an `IBatch`) are *buffered*, not sent. They go to the server only when
+you call `Execute`/`ExecuteAsync`, so the `Task` handed back at the point of queueing cannot complete until
+after that call. Awaiting it there waits for something that can only happen further down the method:
+
+```csharp
+var tran = db.CreateTransaction();
+var value = await tran.StringGetAsync(key);   // never completes - nothing has been sent yet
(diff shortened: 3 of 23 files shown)


mgravell:
Cost check, since this registers on `Invocation`/`PropertyReference`. `ReportAnalyzer`, Release rebuild of `src/StackExchange.Redis` — the largest compilation here, and the worst case, since its own types resolve so the gate opens:

| Time (s) | % | Analyzer |
|---|---|---|
| 7.544 | — | total analyzer execution |
| 3.996 | 52% | StyleCop.Analyzers |
| 0.019 | <1% | TransactionAnalyzer (existing) |
| 0.003 | <1% | QueuedResultAnalyzer (SER305/306) |

On RESPite, which doesn't reference the library, it is <0.001s: the compilation-start action resolves `IBatch`/`ITransactionAsync` once and registers nothing when they are absent. Where it does register, each callback starts with an ordinal name compare (`Result`/`Wait`/`GetResult`) and all the symbol work sits behind that — which is why it costs ~6x less than the block-walking analyzer beside it.

mgravell:
Allocation check on `IsFireAndForget`, since it tests a generic value against an interface (which boxes in IL) on every queued operation. Bytes allocated per call, net10.0:

| variant | implements `IFlaggedRedisArgs` | does not |
|---|---|---|
| `is` pattern | 24.00 | 0 |
| cast instead of `is` | 24.00 | 0 |
| `is` + `AggressiveOptimization` | **0.00** | 0 |

Value-type generics are not shared, so the JIT knows the exact `TState` and compiles this to a field load with no allocator call — 11 bytes of machine code — but only once optimized, hence the attribute. Rewriting the pattern as a cast does *not* avoid the box; both spellings emit the same one. A `TState` that captured no flags costs nothing even in tier-0, since the test folds to a constant and the dead box is removed. `#if NET`, as the enum member is absent on the netfx targets.

mgravell (commented):
👍

PR #3180, opened 2026-08-18
Add SER308: flag the library's own Wait/WaitAll/TryWait helpers

mgravell:
Follow-up to #3179, **reworked after review feedback** — the original approach used `[Obsolete]`; this one does not. The old commits are gone from the branch; the discussion above is the record of why.

`Wait`, `WaitAll` and `TryWait` are what SER307 discourages everywhere else, reached through the API the library itself offers for it: the calling thread is held for the round-trip while the reply needs a thread of its own to be processed. Shipping SER307 while providing a blessed way to do the same thing was only ever half a position.

### Why an analyzer rule and not `[Obsolete]`

The feedback was right: `[Obsolete]` reports `CS0618`, which every obsoletion from every source shares. A consumer who wants to silence *this* has to silence *all* of them, including deprecations in their own code and in unrelated packages. That is far too blunt for an opinion we are choosing to have.

Options considered and rejected:

- **`ObsoleteAttribute.DiagnosticId`** — exists for precisely this, and would be the natural answer. It is net5+, and this library still targets netstandard2.0, so it would give a granular ID on some target frameworks and `CS0618` on others.
- **A hybrid** (attribute where available, analyzer elsewhere) — the attribute would need `#if` on the interface *and* on `ConnectionMultiplexer`'s own members, since marking only the interface does not warn through the class; and on modern targets both mechanisms would report at the same location.
- **`[Experimental]`** — granular, but these APIs are long-standing rather than preview, so it would be saying something untrue.

An ID of our own behaves identically on every target framework, sits with the rest of the `SER3xx` family, and has a docs page. Recorded in the descriptor as *worth revisiting rather than settled*: if netstandard2.0 and net4x are ever dropped, the attribute becomes strictly better — metadata, no analyzer needed, works in every tool — and SER308 can retire in its favour.

### Two interfaces, not one

`IRedisAsync` declares these for database/server/subscriber calls; `IConnectionMultiplexer` declares its own, unrelated set. Both are gathered — testing only one would have left half the surface unguarded, which is exactly how the earlier attempt missed the multiplexer entirely.

Matching goes through `FindImplementationForInterfaceMember` as well as directly, because `ConnectionMultiplexer` is what `Connect` returns, making `conn.Wait(task)` the common shape with the *class* as the containing type.

### Suppressions in this repo

Split into its own commit so the rule is reviewable on its own. Two kinds, distinguished in the comments rather than left to the reader:

- **Forwarding** — `KeyPrefixed`, `RedisBase`, the `Retry*`/`MultiGroup*` wrappers, the test fixture. They must implement the interfaces in full; forwarding a flagged member is not using it, and the implementation cannot be dropped while the interface declares it.
- **One real one** — `CursorEnumerable`'s synchronous scan path, reached only from a caller who asked for the sync API, where `TryWait` is what applies the configured timeout.

Test call sites are suppressed rather than rewritten to `await`, deliberately: the `Wait` helpers apply the multiplexer's timeout and a bare `await` does not, so rewriting would quietly change what those tests exercise. Happy to convert them if preferred.

### Verification

- 154 analyzer tests green, including negative cases: an unrelated `Wait(Task)` on someone else's type, `Task.Wait()`, and `ManualResetEventSlim.Wait()` are all left alone — the rule matches the interface members, not the name.
- Whole repo builds clean on all TFMs.
- Library test results unchanged vs `main` on the touched files.

### DIFF
--- a/docs/SyncOverAsync.md
+++ b/docs/SyncOverAsync.md
@@ -1,8 +1,8 @@
 ﻿# Sync over async, and thread-pool starvation
 
-If you are here from a `SER307` warning or a support conversation: this page explains why blocking on an
-asynchronous redis call is worse than it looks, why the symptom shows up somewhere else entirely, and what to
-do about it.
+If you are here from a `SER307` or `SER308` warning, or from a support conversation: this page explains why
+blocking on an asynchronous redis call is worse than it looks, why the symptom shows up somewhere else
+entirely, and what to do about it.
 
 ## The short version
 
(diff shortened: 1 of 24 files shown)


mgravell:
For @NickCraver and @philon-msft's consideration: I'm "done" with sync-over-async; thoughts on just saying no, but gently? See also https://seredis.dev/SyncOverAsync

PR #3187, opened 2026-08-18
Fix Sentinel connection leak and AbortOnConnectFail=false handling

### LINKED ISSUE #2980: Bug: Sentinel connection with AbortOnConnectFail=false fails to return multiplexer and leaks objects on startup failure
**Version:** StackExchange.Redis 2.9.32

**Description:**

When calling `ConnectionMultiplexer.ConnectAsync()` with Sentinel configuration & `AbortOnConnectFail = false`, if sentinels are unavailable at startup:

1. Library creates a Sentinel `ConnectionMultiplexer` internally (via `SentinelConnectAsync()`)
2. Defaults to `ServerType.Standalone`  
3. Calls `GetSentinelMasterConnection()` which throws: `RedisConnectionException: "Sentinel: The ConnectionMultiplexer is not a Sentinel connection. Detected as: Standalone"`
4. Never returns the ConnectionMultiplexer - making it inaccessible to caller
5. Leaks the ConnectionMultiplexer - when `LoggerFactory` is attached, logs continue being written from the orphaned multiplexer showing ongoing retry attempts

**Reproduction:**
```csharp
{
    EndPoints = { "sentinel1:26379", "sentinel2:26379", "sentinel3:26379" },
    ServiceName = "mymaster",
    AbortOnConnectFail = false,
    LoggerFactory = myLoggerFactory
};

// Start application BEFORE Redis/Sentinel servers are running
var multiplexer = await ConnectionMultiplexer.ConnectAsync(options); 
// Throws exception but multiplexer continues retrying in background (leaked)
```

**Expected Behavior:**

With `AbortOnConnectFail = false`, `ConnectAsync()` should return the `ConnectionMultiplexer` even when initial connection fails, allowing event handler registration and automatic reconnection.

**Actual Behavior:**

- Exception thrown, no multiplexer returned
- Multiplexer created internally but inaccessible (leaked)
- Manual retry loop required, likely creating additional leaked objects on each attempt

**Why This Bug Matters**

Redis Sentinel is the **actively maintained, official solution** for high availability in non-clustered deployments. This bug prevents proper usage of StackExchange.Redis with standard Redis HA deployments, particularly in orchestrated environments (Kubernetes, Docker Compose) where applications commonly start before dependencies.

**Related Issues:** #2173, #2649, #2808, #2893

HarnageaGabriel:
## Summary

Fixes #2980.

When connecting via Sentinel (`ConnectionMultiplexer.Connect`/`ConnectAsync`/`SentinelConnect`) with `AbortOnConnectFail=false`, a failed initial connect — sentinels unreachable, or no primary discovered within `ConnectTimeout` — threw a `RedisConnectionException` instead of returning a multiplexer to the caller, and leaked the internally-created `ConnectionMultiplexer` instances (the sentinel monitor connection, and per-retry primary candidates in `GetSentinelMasterConnection`'s loop), since nothing disposed them on the failure path.

This didn't match the non-Sentinel `Connect`/`ConnectImpl` behavior, which honors `AbortOnConnectFail=false` by returning the muxer (with `LastException` set) instead of throwing.

Changes in `ConnectionMultiplexer.Sentinel.cs`:
- `AbortOnConnectFail=false`: no longer throws. Returns a usable multiplexer with `LastException` set, the same event wiring (`ConnectionRestored`/`ConnectionFailed`) and background retry (`SwitchPrimary`/reconnect timer) as the success path, so callers can attach handlers and let it reconnect once sentinels/primary become reachable — mirroring the non-Sentinel path.
- `AbortOnConnectFail=true`: preserves today's throwing behavior and exception messages (no breaking change), but now disposes the sentinel connection and any superseded/failed candidate connections before the exception propagates, instead of leaking them.
- Per-iteration candidate connections in the primary-discovery loop are now disposed before being superseded, regardless of `AbortOnConnectFail`.

No public API changes.

## Test plan

- [x] `dotnet build` (Release) — 0 errors, 0 warnings
- [x] New tests in `SentinelConfigTests.cs` covering `AbortOnConnectFail=false` (sync `Connect`, async `ConnectAsync`, and `SentinelConnect`) returning a non-null, non-throwing, disconnected multiplexer with `LastException` set against unreachable sentinel endpoints, plus a regression guard that `AbortOnConnectFail=true` still throws — all against closed loopback ports, no live topology required
- [x] Full `StackExchange.Redis.Tests` suite (net10.0, Release) against the docker-compose Redis/Sentinel/Cluster topology: 5880 passed, 0 failed (the live Sentinel integration tests are gated `[SkipOnWindows]`, unaffected by this change)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mgravell:
I think the sentinel side of this is reasonable and captures the key scenarios (not quite full test coverage, but probably enough), however the slot stuff shares the same feedback from #3185 (and for good reason: they share a commit)

mgravell (changes requested):
(hash-tag / slot feedback common with #3185)

HarnageaGabriel:
Re: "hash-tag / slot feedback common with #3185" - confirmed, this branch carries #3185's commit (`1c2a4725`) in its history, so it inherited the same bugs in `ServerSelectionStrategy`'s hash-tag handling, `InventKey`, `GetClusterNode`, and `ClusterNode.Parent`. This PR's own commit doesn't touch that code directly.

Pushed the same structural fix as a separate commit (`a98b42a0`) on top of this branch - see [#3185's review thread](https://github.com/StackExchange/StackExchange.Redis/pull/3185#issuecomment-5393825664) for the point-by-point breakdown of what changed. Sentinel-specific code (`ConnectionMultiplexer.Sentinel.cs`, `SentinelConfigTests.cs`) is untouched.

mgravell (approved):

PR #3191, opened 2026-08-24
Server-native maintenance notifications: opt in, react, hand off, and recover when nothing is announced

mgravell:
Server-native maintenance notifications ("smart client handoffs") for Redis Enterprise and Redis Cloud: we ask
to be told when a deployment is about to disrupt us, and then act on it - relaxing timeouts, re-reading the
cluster topology, recovering stranded sharded subscriptions, and moving off an endpoint before it is taken
away. Plus the recovery path for when *nothing* is announced, which is where the worst field failure came from.

Validated end to end against real Redis Enterprise deployments driven by the fault injector, which is where
most of the design below comes from: the specifications are prose, the payloads were never published, and
nearly every assumption I started with was wrong in some way that mattered. The measurements are the
interesting part of this PR.

## What a user gets

Configuration - `maintNotifications`, `maintMovingEndpointType`, plus `maintRelaxedTimeout`,
`maintRelaxedWindowMax`, `maintPostEventRelaxed`, and a `defaults` key for naming an options provider:

```csharp
// usually nothing: the Redis Cloud and Enterprise providers turn it on where it is supported
var muxer = await ConnectionMultiplexer.ConnectAsync("db.example.cloud.redislabs.com:6379");

// or explicitly
var options = ConfigurationOptions.Parse("host:6379,maintNotifications=Enabled,maintRelaxedTimeout=15");
```

`MaintenanceNotificationMode` is tri-state. `Auto` asks and tolerates refusal - the default where a provider
enables it. **`Enabled` rejects the connection** if the server will not deliver them, including when we end up
on RESP2, because a caller who asked for guarantees should not silently get none; the IntelliSense says so in
those words.

Observation is a first-class use: `ConnectionMultiplexer.ServerMaintenanceEvent` raises `PushMaintenanceEvent`
carrying the notification type, sequence id, endpoint, announced time, replacement endpoint and parsed
`ClusterSlotMigration`s, so an operator can watch what their deployment announces before any behaviour depends
on it.

## What we do about each notification

| notification | reaction |
|---|---|
| `MIGRATING`, `FAILING_OVER`, `SMIGRATING` | relax timeouts on that server for a clamped window |
| `MIGRATED`, `FAILED_OVER` | close the window, keep a post-event tail while things settle |
| `SMIGRATED` | as above, plus a jittered, coalesced topology re-read and recovery of stranded sharded subscriptions |
| `MOVING` | ask for a replacement address, or poll DNS until it stops naming ours; drain in-flight work, then replace the connections |

Timeout relaxation is per-server and read at sweep time, and any timeout or connection failure raised inside a
window carries `MaintenanceType`, so a caller can tell "the deployment was moving" from "your query is slow".
The trailing edge is deliberately generous: timeouts are raised by a once-a-second sweep, and a command that
timed out had already been waiting for its whole timeout before that, so a window covering the command's
entire life can have closed before the exception is built. A window that closed no longer ago than the command
could have been waiting still counts, which is the tightest bound that catches every genuine case.

## Measured behaviour that shaped the design

Each of these contradicted a reasonable-sounding assumption:

- **The announced window is a floor with slack, not a deadline.** Closes arrived 1.6s to 4.1s *after* a declared
  15s, across four runs. So we act *within* the window and never assume the socket survives to it.
- **DNS trails the notification, and can lose the race.** The endpoint moves server-side at +8.6s, DNS follows
  anywhere from +4.4s to +18.7s, and one cluster closed the socket at +15.7s with the record still stale. So the
  handoff polls, and "expired, still stale" is a normal outcome that does nothing rather than guessing.
- **The successor field was never unreachable - we were never asking.** Every observed `MOVING` carried an
  explicit null *because a bare opt-in means "no endpoint type"*. Request one and the server names an address,
  which turns the handoff from a DNS race into a direct move. `maintMovingEndpointType` defaults to `Auto`,
  deriving the scope from the address we are connected to (private or public) and the form from whether the
  connection is encrypted, since a DNS-only certificate cannot validate an IP.
- **`MOVING` fires when the connection's own proxy leaves the address set *and* the set gains a member.** Both
  halves matter: pure reductions and pure widenings are both silent. Thirteen observations, no exception.
- **The two families have opposite scoping.** `MOVING` reaches only the doomed proxy's connections;
  `MIGRATING`/`MIGRATED` are broadcast to every proxy. Hence relaxation is per-server while the public event is
  collapsed on `(type, sequence)` - one callback per logical event, whichever node told us first.
- **`MOVING` is re-sent to a connection that opts in mid-window.** Since the handoff replaces the connection,
  acting on the repeat is a feedback loop - it produced twelve recycles from one event before being gated on the
  sequence dedup.
- **The retained completion never expires.** The server keeps the most recent shard-scoped *completion* and
  replays it to whoever opts in next - and it is still doing so **three hours later**, to every fresh
  connection, with no time field to date it by. So a completion that arrives during the handshake is history:
  it opens no window and raises no event, and is recorded in the log instead. Starters and `SMIGRATED` are not
  retained, so one arriving mid-handshake is news and is treated as such.
- **Only completions are retained, never `MOVING`.** That is why the handoff needs no staleness guard: the one
  notification that demands action can never arrive as catch-up.

## Recovery when nothing is announced

From a support ticket: across a Redis Cloud node replacement, a 2.12.14 client dialled three endpoints that no
longer existed **for around 37 hours**, and recovered only when something unrelated finally provoked a
topology re-read. Every existing refresh path needs somebody *else* to notice first - a redirect from a
reachable node, a peer's announcement, a notification - and the flag that drives refresh-on-failure is only set
once a connection has been *established*, so an endpoint that never connected had nobody to speak for it.

Three consecutive failed connects to an endpoint now provoke the same jittered, coalesced refresh, rate-limited
to `configCheckSeconds`. The restraint is the point: the gate it replaces existed to stop a stampede, so
trading a stuck client for a thundering herd would be no improvement. Note `configCheckSeconds` was never a
rebuttal to this - it drives an `INFO replication` on an established connection, not a topology read. The
counter is also not limited to refused connections: it counts any failure of a connection that never
established, so accepting TCP and then failing the handshake counts too.

That covers an endpoint that cannot be reached, and pruning covers one that has left the topology. The case
left over is an endpoint that is **reachable, answers a handshake, and is no longer part of the deployment** -
a re-bound port now serving something else, which produces no failure, no redirect and nothing announced. New
`topologyRefreshSeconds` re-reads the topology anyway, every 30 minutes by default, jittered by up to 30
seconds per cycle so a fleet started together does not stay in step, and `0` disables it. It is deliberately a
backstop: the event-driven paths react in seconds where this reacts in minutes.

Note that one is a new default for every consumer rather than only for maintenance-notification users, and it
is not behind the experiment gate, because it is ordinary topology hygiene rather than part of this feature.

## Diagnostics

An `ILoggerFactory` now gets the whole story: the opt-in and its outcome per server, every notification
received (marked `(catch-up)` when it is a replay), each handoff outcome, and a **warning when a handoff fails
to establish a replacement inside the announced window** - which is otherwise invisible, since commands succeed
either way and a handoff that took three times its budget looks exactly like one that worked.

## Testing

Three tiers. Unit tests for anything decidable without a server, including the DNS decision logic, which takes
an injected resolver because no in-process fake can move a record. The in-process toy server for the protocol -
it speaks the opt-in, sends all seven notification types, models the retention/replay channel, and closes
sockets after `MOVING` scoped to the node, which is how the fixtures stopped being fiction. Then
`tests/StackExchange.Redis.FaultInjector.Tests`, which drives a real deployment through the same fault-injector
surface go-redis, redis-py and node-redis are tested against.

```bash
cd <environment dir> && docker compose up -d
export SER_FI_CONFIG_DIR=$PWD E2E_SCENARIO_TESTS=true
dotnet test tests/StackExchange.Redis.FaultInjector.Tests
```

Without that configuration every test in the tier skips; with it present but broken they **fail**, because a
suite that skips on a broken environment reports success for tests that never ran. CI names the main test
project explicitly, so CI is unaffected either way.

Live results: the opt-in accepted and active on standalone and `oss_cluster` databases; a real shard migration
with `SMIGRATING`/`SMIGRATED` parsed source-to-target across 1440 reads and zero failures; 16/16 sharded
channels recovering unaided; `FAILING_OVER`/`FAILED_OVER` end to end; `MOVING` acted on ~16 seconds before the
server closed the socket; and the whole path again over TLS and mTLS with certificate validation on.

The destructive scenarios were run supervised at the end of a cluster's life, and one of them is the most
useful result in here: **removing a node validated endpoint retirement against a real deployment**, where
before it had only ever been tested against the in-process fake, in which "the node left" is a method call.

```
cluster nodes:     1=master@a.b.c.1, 2=slave@a.b.c.2, 3=slave@a.b.c.3
endpoints before:  <hostname>, a.b.c.1, a.b.c.2
removing node 2 (we are served by node 1)
  +35.1s restored | +40.0s SocketClosed x2 | +40.1s restored | +54.1s action finished
endpoints after:   <hostname>, a.b.c.1, a.b.c.3
```

Killing the node that actually serves the client was survived too (`SocketClosed` at +6.8s, restored at
+50.1s, recovered unaided) - which needed the node *identified* first, because killing an arbitrary node
produced zero drops: the deployment absorbed it and the client never noticed. A `shard_failure` on a
replicated database is likewise invisible. Worth stating plainly, because a green run of those proves the
deployment coped, not that we did.

Also here: a soak (`toys/MaintenanceSoak`) that ran 5000 handoff cycles - 25.2M commands, 200 recycles - flat
at about 700KB and one client, which is what discharges the no-regression requirement that a unit test cannot.

## Not in this PR

`CLIENT MAINT_NOTIFICATIONS_INFO` is documented but deliberately unimplemented: it reports what the *client*
asked for rather than whether delivery works, so it cannot distinguish genuine support from a stub `+OK`, which
was the only reason we wanted it.

AMR is set to `Auto` pre-emptively and cannot be validated until their fleet emits. Relaxed timeouts are proven
to rescue a command against the fake (3000ms succeeds where 200ms times out, with the fault attributed to
`FailingOver`) but have no *live* evidence, because every scenario run had zero failures; `network_latency` is
node-wide netem and turned out to be the wrong tool for producing one.

Two things left open on purpose: whether health-check probes should drive detection on single-group
deployments, and when this surface stops being `[Experimental]`.

## Public API

128 lines added to `PublicAPI.Unshipped.txt`, most of them gated behind `[Experimental(SER010)]` - the
notification types, `PushMaintenanceEvent`, `ClusterSlotMigration`, `MaintenanceNotificationMode`,
`MaintenanceEndpointType`, the `maint*` configuration properties, the provider types, and `MaintenanceType` on
the timeout and connection exceptions. The ungated addition worth noting is
`ConnectionFailureType.MaintenanceHandoff`, so a deliberate handoff is visible rather than appearing as an
unexplained reconnect.


mgravell:
@philon-msft  my plan is to turn on SCH for AMR (in the options-provider) pre-emptively - i.e. ahead of AMR actually supporting it; this would be in "auto" mode, which means "only try in RESP3 mode; if it gets rejected: meh, no problem" - however, before I do this, I'd want to validate against a real AMR endpoint that this doesn't break anything. Would you be able to lend me something to test this against, or otherwise help me validate this? If we need to do "turn it on in AMR if the server reports version > XXX, otherwise don't because it breaks the proxy/whatever" then that's totally fine too: we'll make it work - I just want to not *break* your users.

Specifically, this would issue an additional command: `CLIENT MAINT_NOTIFICATIONS ON` pipelined during handshake *for RESP3 connections only*.

- if this causes the connection to become severed today (which seems unlikely), that's a bad thing, and would mean we can't use blind auto - we'd need to version check *or* use opt-in for now, deferring auto until AMR confirm the fleet is updated
- if it just reports `-ERR something` that's today fine, we'll ignore it
   - however, if this would impact your reporting, *you* might also want to pre-emptively suppress that at your end
- if it reports success and just never sends us anything: that's fine - it costs us nothing at the client (although it does make it a little confusing for the client, since there's also an "enabled" mode, which effectively *demands* support; I don't think we'll cry about it, but in this case, "supported" would be indistinguishable from "silently not supported")

Issue #1972, opened 2022-01-31
Exception when calling SortedSetRemoveAsync on IDatabase with empty values array


olizarevichroman:
I'm getting an exception (**ERR wrong number of arguments for 'zrem' command**) when calling database.SortedSetRemoveAsync(setKey, values) where "values" is empty array.

As a result the same call within a transaction also throws an exception, but with less descriptive message (**EXECABORT Transaction discarded because of previous errors.**)

It seems like such command might be considered as always successful.

Package version "2.2.88"

hankovich:
I've inspected all IDatabase's methods and found several which don't handle empty collections gracefully.

Here's the complete list (and don't forget about all `Async` versions):
```
GeoAdd
GeoHash
GeoPosition

HyperLogLogLength
HyperLogLogMerge

KeyExists

SetCombine
SetCombineAndStore

SortedSetCombineAndStore
SortedSetRemove

StreamAcknowledge (*)
StreamAdd (*)
StreamClaim (*)
StreamClaimIdsOnly (*)
StreamDelete
StreamRead (*)
StreamReadGroup (*)
```

Methods which are marked with `(*)` have explicit `throw new ArgumentOutOfRangeException(...)` for empty collections, so they differ from the others. 

However, I think that they can be converted to act as no-op to be consistent with other methods which handle empty arrays gracefully. At least it doesn't make sense to throw `ArgumentOutOfRangeException`, because there's no range. It should have been just `ArgumentException`. I understand that it's not always possible not to throw exceptions (or throw exception of another type) because of compatibility reasons. Just thoughts.

NickCraver:
I have some large PRs in pipe around normalizing docs and such, so this would wait until after, but I think it's reasonable to put a similar check in here for `.Length == 0 => default` like we do in other places. That should be a net win without breaking anyone. Thoughts?

mgravell:
Closed by #3192

PR #3192, opened 2026-08-25
Return 0 from SortedSetRemove(Async) for an empty member array

AbhinavMir:
#1972


mgravell (approved):
Yep, we do this short-circuit in some other places, so this is not unreasonable; LGTM

PR #3194, opened 2026-08-25
Maintenance notifications: relax timeouts, and read the cluster delta (D4, part of D5)

mgravell:
Stacked on #3193, which is stacked on #3191. Draft. This is the first part of the feature that **changes behaviour**.

## Relaxation (`6b76faf3`)

An opening notification (`MOVING`, `MIGRATING`, `FAILING_OVER`, `SMIGRATING`) raises command timeouts for that server; a closing one stands them back down. Three settings, of which only the first is prescribed cross-client - the other two say "our invention" in their own XML docs:

- `maintRelaxedTimeout` (10s): what timeouts are relaxed *to*, as a floor. Effective timeout is `max(configured, this)`, so a generous caller keeps their value.
- `maintRelaxedWindowMax` (3x): a backstop for a closing notification that never arrives.
- `maintPostEventRelaxed` (2x, matching go-redis): a closing notification means the server-side *operation* finished, not that the server is back to normal latency - and completion is exactly when every other client that got the same notification re-engages. It does **not** apply after a cap expiry, where nothing told us the event finished and extending past the backstop would defeat it.

These are in **seconds**, unlike every other timeout in `ConfigurationOptions`, because that is the unit the cross-client contract names. The mistake is silent in one direction, so `maintRelaxedTimeout=30000` is rejected with a message that names the unit rather than quietly configuring an eight-hour timeout.

Three things worth a reviewer's eye:

- **Relaxation is server-scoped state read at sweep time, not stamped per message.** Both timeout sweeps rely on head-of-line ordering and stop at the first message that has not timed out; per-message timeouts would invalidate that short-circuit and turn every heartbeat into a full scan of everything outstanding.
- **`CheckBacklogForTimeouts` was measuring message age against `_singleWriter.TimeoutMilliseconds`** - the write-*lock* acquisition timeout, which is about contention between writers rather than server latency. It now has its own expression, so relaxation cannot leak into lock acquisition. Pre-existing conflation that only became visible because relaxation forced the question.
- **The sync path grows a re-wait loop**, because `Monitor.Wait` commits to a duration when it parks: without it a sync caller already waiting when a notification lands would time out at the strict timeout while its async neighbour was relaxed. The loop also handles the way back - if a window closes mid-wait, the revised timeout drops and the wait ends.

Also seqID dedup, per notification type, which lands here because a replayed opening notification extending a window is the first place a replay does damage. The sequence numbers are not defined by any specification, so it is deliberately conservative.

**What is not relaxed, as a rule:** keep-alive, the heartbeat, and connection-failure detection - otherwise a server that died mid-maintenance would linger for the whole window, turning a latency mitigation into an availability regression. There is a test that kills a connection inside an absurdly generous window and requires the failure to still be noticed. The one refinement is the dead-socket heuristic in `PhysicalBridge`, which is *derived from* the command timeout and now tracks the effective one: at a 60s relaxed timeout, firing at 4x the strict 5s would tear down precisely the connection relaxation was protecting.

## Fault surface (`2116f7f3`)

`MaintenanceType` on `RedisTimeoutException`, `RedisConnectionException` and `FaultContext`, defaulting to `None`. "Timeout" and "timeout during an announced failover" call for very different reactions from whoever reads the log.

On `FaultContext` it is deliberately **reported and not acted on**: a fault during announced maintenance is expected and transient, and counting it towards a circuit-breaker trip that then withdraws a whole server is the opposite of what the notification was for - but "ignore faults during maintenance" is a judgement about a deployment, not about the protocol, so it belongs in policy. Happy to change `DefaultIsFailure` instead if that is the preference; it would change existing behaviour for AA users, so I left it.

## Cluster delta, read only (`70c07835`)

Cross-checking against the shipped clients (go-redis, redis-py) says `SMIGRATED` is nested - `["SMIGRATED", seq, [[source, target, slots], ...]]` - which two independent implementations agree on, and which means **we were dropping it**: the parser rejected any non-scalar element after the type. Now read and exposed as `ClusterSlotMigration` on the event. Still nothing applies the delta.

Two incidental fixes the cross-check produced, both worth knowing independently:

- We required a readable sequence id and dropped the frame without one; go-redis length-checks these at two elements and reads no sequence number at all for the shard notifications, so that was stricter than a client that demonstrably works. A missing id now costs only dedup.
- **`SlotRange.TryParseInt16` was `checked`**, so an out-of-range slot number threw `OverflowException` from a `Try*` method - pre-existing, reachable today from the public `SlotRange.TryParse` and from `CLUSTER NODES` parsing. It rejects now.

## Verification

Full `Build.csproj` clean across all TFMs (`TimeSpan * int` is netstandard2.1+, so the arithmetic is on `.Ticks`). Full net10.0 suite at 6249 tests with the switch at `Auto`; the only failure is `FailoverTests.SubscriptionsSurvivePrimarySwitchAsync`, which fails identically with the feature switched off and passes in isolation - an order/state artefact of the local failover pair, not from this work.

### DIFF
--- a/docs/Configuration.md
+++ b/docs/Configuration.md
@@ -100,6 +100,11 @@ The `ConfigurationOptions` object has a wide range of properties, all of which a
 | setlib={bool}          | `SetClientLibrary`     | `true`                       | Whether to attempt to use `CLIENT SETINFO` to set the library name/version on the connection              |
 | protocol={string}      | `Protocol`             | `null`                       | Redis protocol to use; see section below                                                                  |
 | highIntegrity={bool}   | `HighIntegrity`        | `false`                      | High integrity (incurs overhead) sequence checking on every command; see section below                    |
+| defaults={string}      | `Defaults`             | `null`                       | Selects a named defaults provider; see section below                                                      |
+| maintNotifications={string} | `MaintenanceNotifications` | `disabled`           | Whether to ask servers for maintenance notifications; see section below                                    |
+| maintRelaxedTimeout={int}   | `MaintenanceRelaxedTimeout` | `10`                | **Seconds** that command timeouts are relaxed to during announced maintenance; see section below           |
+| maintRelaxedWindowMax={int} | `MaintenanceRelaxedWindowMax` | `30`              | **Seconds** a relaxed window may last at most; see section below                                          |
+| maintPostEventRelaxed={int} | `MaintenancePostEventRelaxedDuration` | `20`      | **Seconds** timeouts stay relaxed after maintenance completes; see section below                           |
 
 Additional code-only options:
 - LoggerFactory (`ILoggerFactory`) - Default: `null`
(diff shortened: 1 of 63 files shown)


mgravell:
Question from @philon-msft : do we invoke the ReconfigureAsync path, like we used to for the previous azure notification events?

answer: not yet, but that is something we can look at in some of our later pieces; MOVING is not a completion, so it probably isn't the right time - but we have *other* notifications for that; using my own internal deliverable tracking, I think that comes under D6

PR #3199, opened 2026-08-29
add Dragonfly to ProductVariant

pairbit:
dragonfly does not support shebangs in lua scripts, so an exception must be made for this product.

### DIFF
--- a/src/StackExchange.Redis/AutoConfigureInfoField.cs
+++ b/src/StackExchange.Redis/AutoConfigureInfoField.cs
@@ -41,6 +41,12 @@ internal enum AutoConfigureInfoField
     [AsciiHash("dragonfly_version")]
     DragonflyVersion,
 
+    [AsciiHash("memurai_version")]
+    MemuraiVersion,
+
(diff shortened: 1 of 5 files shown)


mgravell:
The PR doesn't really touch on Lua, so I assume the key thing here is "let's make it identifiable so that callers can identify it" - if *that* is the goal, I don't have a problem with it, but I might do a bit more of an audit of other variants to add at the same time, for example Memurai.

pairbit:
> let's make it identifiable so that callers can identify it

Yes, the purpose is only for identification.

As part of testing my product, we tested the following options: garnet, valkey, redict, keydb, memuria, dragonfly.
I have added all the product options known to me.
The difficulty in identifying KeyDB was its only product, which doesn't have a version number.
I had to use the value of the executable: /data/keydb-server field if you don't mind.

mgravell (approved):

mgravell:
Thanks; tidied some minutiae (shipped vs unshipped, combined a `case`+`when`), merging.

Issue #3167, opened 2026-08-06
InternalFailure: "Output pipe not initialized" when a write races connection teardown


mgravell:
### Summary

A write that reaches `PhysicalBridge.WriteMessageToServerInsideWriteLock` while the underlying
connection is being torn down and re-established fails with an `InvalidOperationException`
("Output pipe not initialized"), surfaced to the caller as
`RedisConnectionException: InternalFailure`. Because it is an `InternalFailure` rather than a
recognised connection failure, it is not subject to the usual backlog/retry handling.

Seen intermittently; @mgravell reports having hit it before, so this is not a one-off.

### Trace

```
StackExchange.Redis.RedisConnectionException : InternalFailure on [0]:UNWATCH (TransactionProcessor)
---- System.InvalidOperationException : Output pipe not initialized

----- Inner Stack Trace -----
   at StackExchange.Redis.PhysicalConnection.<get_Output>g__Throw|172_0() in /_/src/StackExchange.Redis/PhysicalConnection.Write.cs:line 18
   at StackExchange.Redis.PhysicalConnection.get_Output() in /_/src/StackExchange.Redis/PhysicalConnection.Write.cs:line 17
   at StackExchange.Redis.Message.WriteTo(PhysicalConnection physical) in /_/src/StackExchange.Redis/Message.cs:line 950
   at StackExchange.Redis.PhysicalBridge.WriteMessageToServerInsideWriteLock(PhysicalConnection connection, Message message) in /_/src/StackExchange.Redis/PhysicalBridge.cs:line 1612
```

### What provoked it

Observed during a full test-suite run, in `TransactionTests.BasicTranWithStreamLengthCondition`.
The relevant detail is what was happening on the *shared* connection at that moment: a parallel
`ClientKillTests.ClientKillWithMaxAge` had just issued a `CLIENT KILL ... MAXAGE`, killing the
interactive connection, and the bridge was mid-backlog-drain when the reconnect happened.

From the same test's output, immediately before the failure:

```
Interactive/127.0.0.1:6379: MessagesSinceLastHeartbeat: 317, ConnectedAt: 2026-08-06 14:30:14Z,
  Writer: Active, BacklogStatus: WritingMessage,
  BacklogMessagesPending: (Queue: 193, Counter: 193), TotalBacklogMessagesQueued: 369,
  Connection: (SentAwaitingResponse: 0, AvailableOnSocket: 0 byte(s), InReadPipe: 0 byte(s),
[...]

Issue #2883, opened 2025-05-15
Under low-memory conditions, StackExchange.Redis can return incorrect results to simple StringGetAsync calls


burns47:
This is another report of #2804 

Environment details:

- We're seeing this on StackExchange.Redis 2.8.16, but I've reproduced on 2.8.37.
- We're running dockerized .NET 8 on linux `FROM mcr.microsoft.com/dotnet/aspnet:8.0`
- I'm running against an AWS Elasticache Redis cluster with these options:  `clustercfg.mycluster.amazonaws.com:6379,password=REDACTED,ssl=True,abortConnect=False`

I'm hoping I can help identify this issue because I've attached a reproduction to this issue.  It's *not a perfect repro - you may have to run it several times (15+ for at least 30 seconds each) before seeing the bug.  It may also OOM and you may have to restart it*.

The TLDR of the repro is:

- I'm using 1000 async `Tasks` to call `StringGetAsync` and `StringSetAsync` with a GUID key and the same GUID as the value
- I'm using a separate thread to consume as much process memory as I can by stuffing large `int[]`s into a static list.
- There's one additional thread that's attempting to read and write large values (100KB) to and from redis
- The process memory is limited to 2GB

When the bug occurs, most or all calls to `StringGetAsync` return a value from a different call.   In some cases the connection heals itself, but in other cases (including in production) we've seen the connection stay broken despite the memory pressure going away, and the calls continue to be mixed up.  In our production scenario, this resulted in mostly deserialization errors (since the data returned did not match the format we were expecting), but in some cases resulted in data being returned to clients that did not own that data (for mixups that happened to have the correct format)

The program will report exceptions from the GUID-writing processes and the large value writer.  It will also report when the GUID-writing processes get values they do not expect.  When this happens, it happens in big bursts:

<img width="454" alt="Image" src="https://github.com/user-attachments/assets/1cf0a1ed-3b2b-4d04-93b0-e45b2f6bc795" />

Any help with this issue would be appreciated!  From my experimentation it seems that `HighIntegrity` mode helps, but...why?  How can memory pressure lead to what seems to be a protocol desync?

[RedisRepro.zip](https://github.com/user-attachments/files/20233775/RedisRepro.zip)

mgravell:
That's very weird. I seem to have secured some dedicated library time, so I'll see what I can find. Interesting!

mgravell:
[...]

PR #3200, opened 2026-08-30
Prevent response queue writes during connection teardown

jetersen:
## Summary

Close a failed physical connection's response queue to new writers before detaching the bridge and draining outstanding messages.

A writer can already hold a reference to the old `PhysicalConnection` when `RecordConnectionFailed` starts. Today teardown detaches the bridge, drains `_writtenAwaitingResponse`, and only then calls `Shutdown()`. In that interval the writer can enqueue a new response expectation against the old socket. A reply already in the socket/read buffer can then be matched to that later command, shifting response ownership.

This complements #3189 / #3167. That change correctly treats a writer losing its output pipe as a normal closure, but it does not close the response queue while teardown is draining it.

The change:

- marks shutdown under the response-queue lock at the start of `RecordConnectionFailed`
- rejects commands that captured the old physical connection with `SocketClosed`
- drops trailing replies after the dead connection's response queue has been drained
- extends the existing teardown stress test to fail on response-integrity symptoms

## Reproduction

I reproduced this against Valkey 9.1.0 by running concurrent `SET`/`GET`, hash, list, and optional pub/sub traffic while repeatedly issuing `CLIENT KILL` for the active client sockets.

The failure occurs on:

- StackExchange.Redis 3.1.31 and current `main` (`322688fa`)
- RESP2 and RESP3
- direct Redis traffic with no FusionCache or pub/sub operations

Observed errors include:

```text
Received Interactive/Resp3 response with no message waiting: BulkString
Unexpected response to PUBLISH: BulkString
Unexpected response to HMSET: Integer
Invalid format parsing BulkString as Boolean
```

That makes this protocol-independent and rules out RESP3 push framing and FusionCache as requirements. The server is closing sockets on request; the corruption is the client-side race between response-queue draining and a writer that captured the old physical connection.

Patched stress runs:

- RESP3, direct-only: 18,274 expected disconnect errors, 0 protocol faults
- RESP2, direct-only: 15,505 expected disconnect errors, 0 protocol faults
- RESP3, shared direct + FusionCache/pub-sub: 43,343 expected disconnect errors, 0 protocol faults

No wrong non-null values were observed in the patched runs.

## Validation

- focused `WriteFailureTeardownTests`: pass
- deterministic #3167 test: pass
- explicit `WritesRacingTeardownAreNeverInternalFailures`: pass with response-integrity assertions
- multi-target Release build: pass
- complete net10 suite: 6,037 passed, 150 skipped, 1 unrelated failure

The single suite failure is `ConnectionFailureErrorsTests.SocketFailureError`, where DNS resolution produced an outer exception without the expected inner exception. It reproduces unchanged from a clean `322688fa` checkout in this environment.

Related: #3091, #3167, #3189

### DIFF
--- a/src/StackExchange.Redis/PhysicalConnection.Read.cs
+++ b/src/StackExchange.Redis/PhysicalConnection.Read.cs
@@ -702,6 +702,15 @@ internal sealed partial class PhysicalConnection
 
             if (!_writtenAwaitingResponse.TryDequeue(out msg))
             {
+                // A failure can race bytes that were already in the socket or read buffer. Once shutdown has
+                // started, no new command can enter this queue, so these are replies to commands the failure
+                // path has already completed. Drop them with the dead connection instead of reporting a fresh
+                // protocol failure.
+                if (_isShutdown)
+                {
+                    return;
+                }
+
                 Throw(frame, connectionType, _protocol);
 
                 [DoesNotReturn]
--- a/src/StackExchange.Redis/PhysicalConnection.Write.cs
+++ b/src/StackExchange.Redis/PhysicalConnection.Write.cs
@@ -13,8 +13,9 @@ internal partial class PhysicalConnection
     private BufferedStreamWriter? _output;
 
     /// <summary>
-    /// Set by <see cref="Shutdown"/> before it discards <see cref="_output"/>, so a writer that finds the
-    /// output gone can tell the two cases apart; see <see cref="ThrowOutputUnavailable"/>.
+    /// Set when connection shutdown starts, before responses are failed or <see cref="_output"/> is discarded,
+    /// so writers cannot add new response expectations to a connection being torn down; see
+    /// <see cref="ThrowOutputUnavailable"/>.
     /// </summary>
     private volatile bool _isShutdown;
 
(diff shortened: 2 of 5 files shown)


jetersen:
Existing-issue/PR cross-check:

- #3092 fixes a different desync source: an exception while serializing/writing a command left the connection alive.
- #3189 fixes the adjacent teardown race where a writer loses `_output`, plus stale `physical` snapshots and failure classification.
- #3189's PR description explicitly records the remaining symptom this PR addresses: during teardown storms, both `main` and that branch logged `Received Interactive/Resp3 response with no message waiting`; it was called pre-existing and left for follow-up.
- The currently open PRs do not contain another response-queue teardown fix. #3194 touches `PhysicalConnection.cs` for maintenance notification state but does not change `_writtenAwaitingResponse`, `RecordConnectionFailed`, or response matching.
- #2883/#2804 are the older wrong-response reports, closed on the expectation that v3's IO rewrite resolved them. This reproduction is on current v3 `main` and is protocol-independent.

So #3200 is a follow-up to #3189 rather than a duplicate: it closes the response expectation queue atomically with teardown and handles already-buffered trailing replies.

mgravell:
Reviewing now, thanks

mgravell (approved):

PR #3205, opened 2026-09-03
Never match responses on a connection whose write failed mid-flight

### LINKED ISSUE #2919: Receiving other thread HGETALL request result on high memory pressure and multiple threads

mgravell:
Fix #2919

A write that dies after its message has been queued in _writtenAwaitingResponse must either complete, or take the connection out of service. Otherwise the queue is permanently one message out of step with the replies on the wire, and every caller on that connection silently receives somebody else's answer.

Both recovery paths allocate before they reach RecordConnectionFailed: message.Fail goes via ResultProcessor.ConnectionFail (StringBuilder plus message.ToString()), and HandleWriteException builds a RedisConnectionException. That code only ever runs *because* something already failed, so under memory exhaustion it fails again and teardown is skipped. v3 adding the second teardown attempt in HandleWriteException narrowed this from two faults to three; it did not close it.

So make the decision to stop trusting the connection allocation-free, infallible, and first:

- PhysicalConnection.PoisonWrite sets a volatile bool, nothing more
- both write-failure catches call it before anything that can allocate
- MatchNextResult refuses to match once poisoned, throwing a pre-allocated sentinel so the read loop runs the normal teardown and fails everything queued
- the write path refuses to queue onto a poisoned connection

Worst case this degrades to timeouts, which is the right trade against handing back silently wrong data.

Adds a regression test, plus the internal seam it needs to simulate a teardown that cannot allocate. On the previous code that test fails with GET k0 returning v1 - the exact symptom reported in #2919.

## Checklist

### DIFF
--- a/src/StackExchange.Redis/ConnectionMultiplexer.Debug.cs
+++ b/src/StackExchange.Redis/ConnectionMultiplexer.Debug.cs
@@ -1,4 +1,5 @@
-﻿using System.Threading;
+﻿using System;
+using System.Threading;
 
 namespace StackExchange.Redis;
 
@@ -15,6 +16,15 @@ public partial class ConnectionMultiplexer
         finalized = Volatile.Read(ref _collectedWithoutDispose);
         return created - (disposed + finalized);
     }
+
+    /// <summary>
+    /// Test seam: invoked immediately before a failed write attempts to tear the connection down.
+    /// Lets tests simulate the case that motivated <see cref="PhysicalConnection.PoisonWrite"/> - an
+    /// OutOfMemoryException landing in the failure path itself, so teardown never completes (#2919).
+    /// </summary>
+    internal volatile Action? BeforeWriteTeardown;
+
+    internal void OnBeforeWriteTeardown() => BeforeWriteTeardown?.Invoke();
 
     /// <summary>
     /// Invoked by the garbage collector.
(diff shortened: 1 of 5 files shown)


jcaspes:
Hi, thanks for your quick response. I'll run my tests again on this pull request.

mgravell:
That would be fantastic, thanks

jcaspes (commented):
Hello,
After numerous attempts to reproduce the issue with different launch parameters for the tool https://github.com/jcaspes/StackExchange.Redis/tree/MemoryTester/toys/MemoryTester on your branch https://github.com/jcaspes/StackExchange.Redis/tree/marc/issue-2919-write-poison, I really think your fix is OK because I can no longer reproduce the problem.

As soon as I remove your commit, the problem returns almost instantly.

Marking the connection as non-reusable after a failure does indeed seem like a good idea!

with version 3.1.34:
image

with version 3.1.40:
image

mgravell:
@jcaspes awesome; I have Monday out-of-office, and I need to align a few merges: expect release Tuesday

PR #3227, opened 2026-09-16
A tracer redirect proves liveness, so it completes the handshake

mgravell:
Closes the remaining scope of #2970 — the part that turned out not to be cosmetic.

## The bug

The keyed `EXISTS` tracer, reached when `ECHO`, `PING` and `TIME` are all unavailable, cannot target a slot the
node owns during its first handshake: the `CLUSTER NODES` reply that would say which slots those are is still
in flight in the same pipeline batch, and `HELLO`'s `mode` field with it, so `ServerType` is still the seeded
`Standalone` when the key is chosen. On an OSS cluster the probe earns a MOVED.

`TracerProcessor` classified that as a protocol failure:

```csharp
else
{
    connection.RecordConnectionFailed(ConnectionFailureType.ProtocolFailure, ...);  // <- MOVED landed here
}
```

...which tore the connection down. The node never established, and the connect burned its **whole
`ConnectTimeout`** first. Measured against the local six-node cluster, seeding one endpoint:

```
connected in 10050ms — 127.0.0.1:7002 serves slots but did not establish
```

That is the default `AbortOnConnectFail=true` path. It self-heals on a later attempt, once `ServerType` is
`Cluster` and the tracer key is slotted — but only after a ten-second stall and a failed connect.

## The fix

A redirect answers the only question a tracer asks — *is this server up, speaking RESP, and talking to us* — so
`Moved` and `Ask` complete the handshake rather than failing it. No topology needed, no extra round-trip, and
standalone is untouched. The log then shows `OnConnectedAsync completed (From command: EXISTS)`.

## Testing

The new test seeds a **single** endpoint so the rest are discovered from topology, each having to complete a
handshake whose tracer key is chosen before it knows its own slots. Seeding every node lets enough of them
settle to mask the failure. Asserted without polling, because the point is the *first* pass:

| | Runs passed (both protocols) |
|---|---|
| Before | **0 / 6** |
| After | **6 / 6** |

Full suite against the docker-compose topology: 6362 passed, 0 failed. All TFMs build with 0 warnings.

## A separate bug found on the way — not fixed here

The test deliberately ignores nodes serving no slots. Those are registered inert at
`ConnectionMultiplexer.cs:2101`:

```csharp
// Registered *inert*: known and addressable via GetServer, but not dialled...
GetServerEndPoint(node.EndPoint, ServerProvenance.ClusterTopology, activate: false);
```

...but they still land in `_serverSnapshot`, and `ReconfigureAsync` awaits `OnConnectedAsync` on every server
in it — so a connection deliberately never opened is waited on for the full `ConnectTimeout`:

```
Registering 127.0.0.1:7005 without connecting: serves no slots
127.0.0.1:7005: OnConnectedAsync init (State=(null))
... 10s later ...
127.0.0.1:7005: Did not respond (Task.Status: WaitingForActivation)
```

It is a race on whether the inert registration beats the task-gathering, and it happens with the **default
command map** too — I measured 1/12 single-seed cluster connects, unrelated to this PR's probes. Worth its own
issue and its own fix.


mgravell (commented):
👍

PR #3229, opened 2026-09-16
Hash RedisValue with seeded XxHash3 rather than Marvin

mgravell:
Stacked on #3228 — **take it or leave it independently**; the equality work does not depend on this.

Base is `marc/redisvalue-equality-alloc`, so the diff shows only the hashing change. Retarget to `main` if
#3228 lands first.

## Why

Benchmarking hashing while working on #3228 turned up something larger than the allocation that PR removes.
`HashBlob` and `HashString` sit within ~4% of each other at every size, which means the UTF8 decode is *not*
the cost — the hash function is, at roughly 2 GB/s. At 64KB that is ~31µs to produce a single `int`.

## The change

`XxHash3` over the decoded characters, seeded once per process. No new dependency: `System.IO.Hashing` is
already referenced by the main library and `XxHash3` is already used by `ValueCondition`. It also removes the
`#if NET` split, since the down-level targets no longer need to fall back to building a string.

| Size | `HashBlob` | `HashString` |
|---|---|---|
| 16 | 30.4 → **27.3 ns** (1.1×) | 33.7 → **29.7 ns** (1.1×) |
| 1024 | 503.7 → **85.4 ns** (5.9×) | 489.6 → **69.1 ns** (7.1×) |
| 65536 | 31,026 → **2,970 ns** (10.4×) | 29,885 → **1,762 ns** (**17.0×**) |

Long *strings* benefit more than blobs, so this is not a blob-specific fix.

## What is given up — the part that needs a decision

The whole content is still hashed. Nothing is sampled, so equal values land in the same bucket by
construction, whatever representation they arrived in, and the hash contract is safe.

What changes is the *kind* of hash. Marvin is **keyed**: collisions cannot be constructed without the key.
xxHash's seed is not a key — it gives per-process unpredictability, not resistance, and published
collision-finding techniques for xxHash do not depend on the seed. So this trades hash-flooding resistance for
speed, on the basis that a `RedisValue` is rarely a dictionary key and practically never an attacker-chosen
one.

That is worth weighing against the June discussion on #3116, where the leaning was the other way — towards
bringing Marvin *to* `RedisKey`, which today uses the hand-rolled `728271210`-seeded hash and has no Marvin at
all. This PR moves `RedisValue` toward `RedisKey` rather than the reverse. Defensible, but a deliberate choice
rather than a drift.

Alternatives considered and rejected:

- **Sample the two ends and keep Marvin** — ~1000× at 64KB, but it abandons the unread middle entirely, and
  Redis data routinely shares long prefixes, so common-prefix collisions become free. Strictly worse on the
  security axis than this.
- **A keyed-but-faster hash (SipHash-1-3)** — keeps genuine resistance, but only ~2–3×, and means carrying an
  implementation.
- **Hash the UTF8 bytes directly** — would remove the remaining decode, but reintroduces the string/blob
  divergence that #3228 exists to avoid.

## Note for reviewers

`((RedisValue)"abc").GetHashCode()` previously equalled `"abc".GetHashCode()`, because the string case
forwarded straight to it. It no longer does. Hash codes have always been per-process and must never be
persisted, so nothing may legitimately depend on the value — but the coincidence is gone, and a test somewhere
could be asserting it.

The decode is now the visible remainder: it was under 4% of the old cost, and at 64KB it is ~40% of the new
one (2,970 ns blob against 1,762 ns string). I would stop here rather than chase it.

## Testing

Full suite: 6365 passed, 0 failed — including `EqualStringAndBlobShareHashCodes` from #3228, which is the
invariant that matters. All six TFMs build with 0 warnings.


mgravell:
Closing — this is a breaking change to shipped behaviour, and it does not belong in a minor.

The specific trap, verified rather than theorised. `RedisValue.Equals(object)` accepts a `string` (via
`TryParse`), and `RedisValue.GetHashCode()` currently agrees with `string.GetHashCode()` for the same text
because the string case forwards straight to it. Those two facts together make a real interop path work today:

```
                                          main          this PR
string.GetHashCode()                  22895958      -291348134
RedisValue.GetHashCode()              22895958      1353807393
same hash                                 True           False
redisValue.Equals(string)                 True            True    <-- unchanged
Hashtable keyed by RedisValue,
  probed with a string            value-via-...          (null)   <-- silently lost
```

So after this change `redisValue.Equals(someString)` still reports `true` while their hash codes differ: a
hash-contract violation across the `object` boundary, which is precisely what I objected to in #3116, just at
a different seam. Anyone with a `Hashtable`/`Dictionary<object,…>` keyed by `RedisValue` and probed by string
loses lookups silently, with no compile error and no exception.

The measured win was real (10-17x, and more for long strings than blobs), so the finding stands even though
the change does not: hashing a `RedisValue` runs at roughly 2 GB/s and the UTF8 decode alongside it is under
4% of the cost. It is just not worth a silent break outside a major.

Superseded by the opt-in approach: a `RedisValue.EqualityComparer` abstract base with `Default` and `Binary`,
so callers who want byte-domain speed ask for it and the shared type keeps its behaviour. That can go further
than this PR anyway - `Binary` never decodes at all, so it reaches ~900ns for a 64KB comparison against the
5,397ns in #3228, and hashes raw bytes.

#3228 is unaffected and still stands on its own: it changes only *how* `operator ==` computes the same
relation, leaves `GetHashCode` alone, and the interop above still works on that branch (verified: `same hash =
True`).

Issue #1044, opened 2019-01-16
Allow to register global profiler


SergeyKanzhelev:
Allowing to specify global profiler for all connections will help simplify enablement of profiling auto-magically. Ideally registering of a global profiler should work retroactively for the already created connections. So there will be no application startup race between the connection creation and profiler registration.

It may be a static method that will register the factory of creating ProfilerSession factories. It will create a session factory based on profiler registered for the specific connection using `RegisterProfiler` method.

```
public static void RegisterDefaultProfiler(Func<Func<ProfilingSession>, Func<ProfilingSession>> profilingSessionProviderFactory);
```

This way default profiler will not be simply overridden by connection-specific profiler and there will not be an incentive to call `connection.RegisterProfiler(() => null);` for every new connection.

NickCraver:
The problem with a global profiler for most applications is it would be an unbounded memory leak by nature. I think referencing a default profiler as a wrapper is a better bet (leaving it to the user here). For example, if we set this up on Stack Overflow we'd probably OOM within the hour...that's a pretty dangerous and easy pit to fall into. People can register a global profiler on the shared connection already in a few lines of code, so I don't think we need an API addition here to support the use case +if you so choose_ to do something so global. Does that make snese?

SergeyKanzhelev:
@NickCraver will memory leak because of the hanging sessions? Or some other reason?

Also, what is the default pattern configuring Redis connection with the DI? So we can recommend the best practice on how to configure profiler in a single place

NickCraver:
@SergeyKanzhelev missed this (saw in the other issue) - yes a profiler that keeps running would be an unbounded memory leak. I'd question the goals to profile _everything an application ever does_ - that is not a goal of our design in the profiler. Our goal is to profile things in some context or some period of time (e.g. for a web request or a window). I want to sanity check goals here on what we're _trying_ to do.

For context: I'd be profiling many billions of Redis hits a day, but I also get that's not the normal use case. Scale slants my view as "...that'd be crazy expensive", where I imagine you're viewing a more general use case with a lot less impact of scale. Thoughts? Am I way off there?

SergeyKanzhelev:
> I want to sanity check goals here on what we're trying to do.

The end goal is to enable continuous monitoring of Redis calls. For low load apps - we can end up recording information on every call. For high-throughput apps - sampling will be applied per distributed trace (e.g. web request) so many calls will not be recorded.

[...]

PR #3241, opened 2026-09-17
[exploration, not a plan] Notes: what in-box OpenTelemetry support would actually involve

mgravell:
> [!WARNING]
> **Exploration only. This is not a plan, and nothing here has been agreed.**
>
> No product code is changed by this PR — it adds three markdown files under a new `notes/`
> directory and nothing else. The document called `plan.md` is a *proposal written to be argued
> with*, not a commitment, a roadmap, or a statement of intent from the maintainers. Several of its
> central questions are explicitly unresolved, including the strongest argument against its own
> direction.
>
> Please read it as "here is what we found out, and here is one coherent thing we could do about
> it" — and please push back.

## What this is

[#1044](https://github.com/StackExchange/StackExchange.Redis/issues/1044) was revived recently, and
the underlying subject — whether StackExchange.Redis should emit telemetry itself instead of being
instrumented from outside via reflection — has come up roughly every two years since 2019 without
ever being written down. This PR writes down the research so the next round starts from facts
rather than from memory.

It also starts a `notes/` convention: `docs/` is the published Jekyll site (so anything placed
there lands in `sitemap.xml`), `notes/` is internal handover material that is not built, not
published and not indexed. Topic per subdirectory; files named for their content.

| file | what it is |
| --- | --- |
| `notes/readme.md` | the convention and the `docs/` vs `notes/` split |
| `notes/otel/findings.md` | research — evidence, with sources |
| `notes/otel/plan.md` | one possible response to it — **proposed, not agreed** |

## Findings worth surfacing

> Throughout, **"the contrib package"** means
> [`OpenTelemetry.Instrumentation.StackExchangeRedis`](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Instrumentation.StackExchangeRedis),
> which lives in `opentelemetry-dotnet-contrib` — the OpenTelemetry .NET SIG's repository for
> instrumenting libraries outside the core distribution. It is what instruments StackExchange.Redis
> today, from the outside, via `RegisterProfiler` and reflection. Where the *people* are meant, they
> are named as its maintainers.

- **What actually broke in 3.2.0.** Exactly one reflection site in the contrib package, reached
  only under the opt-in
  `SetVerboseDatabaseStatements`: a `DynamicMethod` field getter for `ProfiledCommand.Message`, a
  `PropertyFetcher` for `Message.CommandAndKey`, and the private `script` field on the EVAL message
  type that #3211 renamed. The one thing we do not already expose publicly is that statement
  string.
- **The bridge's cost is architectural, not incidental.** Because `RegisterProfiler` is the only
  hook, the instrumentation runs a background drain thread per multiplexer, keeps a
  `ConcurrentDictionary` keyed by `(TraceId, SpanId)` with a `ProfilingSession` allocated per parent
  span, manually saves and restores `Baggage.Current` across threads, and builds spans
  retroactively up to 10s late. Most of that would disappear if the data were emitted where it is
  produced.
- **There is a known-good `System.*` answer.** `ActivitySource` and `Meter` *are* the .NET tracing
  and metrics API; OpenTelemetry .NET only attaches listeners to them. Emitting them couples us to
  the BCL, not to a vendor. On `net8.0`/`net10.0` this is **zero new package references** —
  `System.Diagnostics.DiagnosticSource` is in the shared framework. `net461` is the only target it
  cannot reach (the package dropped its `net461` asset after 6.0.1).
- **…which turned up a question bigger than telemetry.** net461 went **out of support on 27 April
  2022**, and *every* Microsoft package we already depend on — logging abstractions, hashing,
  pipelines, async interfaces, channels — bottoms out at `net462`. Our net461 build is already
  resolving all of them through `netstandard2.0`, the combination the .NET Standard docs explicitly
  advise against. So the real question is not how to compile telemetry out of net461 but why net461
  is still a target; and if the floor moves it should be `net472`, not `net462`, since 4.6.2 itself
  loses support on 13 January 2027 and 4.7.2 is both unexpiring and already shipped. That is a
  repo-level call, noted in the plan but deliberately not decided there.
- **Failures are currently invisible.** The existing profiling API carries nothing about errors, so
  no instrumentation built on it can set `error.type`, `db.response.status_code` or
  `ActivityStatusCode`. A failed command produces a span indistinguishable from a successful one.
- **Cross-language conventions have not converged.** go-redis, Lettuce and redis-py were surveyed:
  go-redis ships an official OTel module in-repo but is still on `db.statement`; Lettuce defines a
  vendor-neutral tracing SPI with Brave/Micrometer adapters; redis-py monkeypatches. There is no
  shared vocabulary beyond the semantic conventions —
  [`db/redis.md`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/db/redis.md)
  over
  [`db/database-spans.md`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/db/database-spans.md),
  the ones linked from #1044, plus
  [`db/database-metrics.md`](https://github.com/open-telemetry/semantic-conventions/blob/main/docs/db/database-metrics.md)
  — and adoption of those lags unevenly.
- **The attribute names are not the ones you would guess.** The old/new split is driven by the
  standard `OTEL_SEMCONV_STABILITY_OPT_IN` environment variable, whose default is **`Old`** — so
  unopted users are seeing `db.system`/`db.statement` today, not `db.system.name`/`db.query.text`.
  Emitting only the current names would look like the modern, correct choice and would silently
  break every existing consumer. This is the trap most likely to catch an implementation.

## Deliberately unresolved

These are in the notes as open questions, not as decisions:

1. **Who carries semantic-convention churn?** The contrib package has never shipped a stable
   version in 48 releases, and its README says why: the database conventions are still
   Experimental. StackExchange.Redis does not have the option of shipping perpetual betas, so going
   native means either marking the telemetry `[Experimental]` or accepting that a stable package
   emits attributes that may be renamed under it. This is the strongest argument *against* the
   direction `plan.md` leans, and it deserves to be argued properly.
2. **In-box, or better hooks for someone else to build on?** @martincostello framed these as the
   "utopian ideal" and the "middle ground" respectively. The notes lean toward the former; Lettuce
   is a working existence proof of the latter. The second is a strict subset of the work either
   way, because the data that has to become reachable is identical — only the caller of
   `StartActivity` differs.
3. **Do `Filter`/`Enrich` belong in-box?** `ActivityListener.Sample` arguably makes them
   unnecessary, but Npgsql and the contrib package both ship them anyway — two independent
   implementations landing on the opposite answer. Left open.
4. **Should the semconv choice be an explicit enum, not just an environment variable?**
   `OTEL_SEMCONV_STABILITY_OPT_IN` is ambient and process-wide, and a library quietly changing the
   attribute names it emits because of an environment variable is how you get a baffled bug report.
   An explicit `options.SemanticConventions = ...` — with a `Default` that still follows the
   variable, so a service-wide setting keeps every instrumentation agreeing — means nobody is
   surprised. Costs a handful of public API lines; probably worth it for the one setting most
   likely to cause "why did my dashboard go blank".

   This also defuses question 1. `ExperimentalAttribute` permits `AttributeTargets.Field`, and enum
   members are fields, so the *members* that track still-experimental conventions can be gated
   individually while `Default`/`Old` stay plain stable API — the churn risk lands only on callers
   who opted into churn, instead of a blanket `[Experimental]` in front of everyone. When the
   database conventions stabilise, deleting the attribute is neither a source nor a binary break.
5. **Does any of this wait for v4?** The IO core rewrite moves every hook site involved.

## On replacing the contrib package

`plan.md` sketches taking over from `OpenTelemetry.Instrumentation.StackExchangeRedis`. To be
explicit, since this is a public repository and that is someone else's project: **this has not been
agreed with its maintainers**, it is a conversation to be had with them (code owner
@matt-hensley, with @martincostello active on it), and nothing will be built on that assumption
before that conversation happens.

Two things the notes flag as needing their explicit input: their test suite is the natural
compatibility oracle and is Apache-2.0 going into an MIT repository, and any transition would
change the `ActivitySource` name, which breaks hand-written `AddSource(...)` calls even where the
emitted data is identical.

## Reviewing this

The useful feedback is on the facts and on the open questions, not on prose. Specifically:

- anything in `findings.md` that is wrong, out of date, or missing a source;
- whether the compatibility contract in `plan.md` actually covers what people have built
  dashboards on;
- whether the phase ordering (statement text → metrics → traces) is the right risk order;
- the semconv-ownership question above, which is the one that decides the whole direction.

Draft, and staying draft until those are answered.


notes/otel/findings.md
```diff
@@ -0,0 +1,461 @@
...
+| 4.6.2 | **13 January 2027** |
+| 4.7, 4.7.1, 4.7.2, 4.8, 4.8.1 | none |
+
+4.5.2, 4.6 and 4.6.1 were retired specifically because they were signed with SHA-1 certificates.
+
+Note the middle row: **net462 is supported for about four more months** from these notes. Bumping a
+floor onto a version that expires within the year is not a bump worth doing — which is what points
+at `net472`, the oldest .NET Framework with no end-of-support date, the version Microsoft's own
+footnote names ("upgrade the project to target .NET Framework 4.7.2 or higher"), and a target
+`StackExchange.Redis` already ships.
```
martincostello, line 164: We're already planning to drop `net462` as a TFM in the .NET OTel SDK in January 2027. The current replacement for us is going to be `net47` as it doesn't seem that `net47` and `net471` are going out of support any time soon. We could maybe instead go to `net472`, but we'd likely only do that if EoL dates appear for 4.7.0 and 4.7.1 or unforeseen pain arises from trying `net47` as the new floor.

Personally it doesn't bother me if Redis itself moves to `net472` and pushes a firmer floor on version support, as if you changed that without adding any telemetry as proposed here that's your decision on end-users and independent of OTel anyway.

notes/otel/findings.md
```diff
@@ -0,0 +1,461 @@
...
+spans are the half that needs sampling.
+
+### The migration tax is real
+
+The contrib package currently declares **three** `ActivitySource`s — one at semconv 1.23.0 emitting
+`db.system`/`db.statement`, one at 1.42.0 emitting `db.system.name`/`db.operation.name`/
+`db.query.text`, and one emitting both — driven by `EmitOldAttributes`/`EmitNewAttributes`. If we
+own the source, we own that kind of migration. The counter-argument is that the new names are now
+the stable ones and the dual-emit window is closing; but it is the strongest single argument for
+leaving span production to someone whose job is tracking semconv.
```
martincostello, line 222: Going straight to "new" (v1.44.0 right now) is fine as built-in support would be net-new here.

The database semantic conventions are now stable so that's the best thing to write any support on top of - it's just the Redis-specific ones are still in development, so that could be treated as more advisory to avoid the need to adapt to any possible breaking change in that part.

The [AWS instrumentation](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Instrumentation.AWS#readme) (which is stable, because they didn't want to have pre-release versions) [has an escape hatch](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/blob/aace0ab84517e2269d79adc9783836f0d6b67abf/src/OpenTelemetry.Instrumentation.AWS/AWSClientInstrumentationOptions.cs#L19) to allow semantic convention naming to adapt over time and be opt-in on the part of the user.

This means you could just drop `OTEL_SEMCONV_STABILITY_OPT_IN` entirely and not worry about it, but keeping it helps regarding future change as old would be frozen in time (v1.23.0 of databases SemConv IIRC) and not change, and opted-in users would just get whatever the current state is any time this library ships any changes.

notes/otel/findings.md
```diff
@@ -0,0 +1,461 @@
...
+
+> In the *ideal* case the OpenTelemetry instrumentation would actually become obsolete because
+> it's entirely built-in to the library here. A middle ground where we have the hooks to wrap
+> things and produce the telemetry ourselves is also fine (just not the *utopian ideal*).
+
+So, named:
+
+- **Utopian** — we emit `Activity` and `Meter` ourselves. Consumers write
+  `AddSource("StackExchange.Redis")`. The contrib package collapses to a one-line convenience
+  extension or is retired. #1044 stops existing, because there is nothing to register.
```
martincostello, line 314: In an ideal world all the instrumentation would move here, and we'd eventually drop our instrumentation library entirely. We're not precious to it becoming obsolete in lieu of better long-term support. We've made a similar ask for EFCore: https://github.com/dotnet/efcore/issues/29281#issuecomment-3497968401. That would take us down the retire path.

For [HTTP Instrumentation](https://github.com/open-telemetry/opentelemetry-dotnet-contrib/tree/main/src/OpenTelemetry.Instrumentation.Http#readme) we're already in a middle ground - the majority of the implementation is now delegated to .NET itself since .NET 9. Our instrumentation adds some sugar on top like enrichment and filtering, but we don't (apart from for earlier versions of .NET) actually produce the metrics and spans anymore.

notes/otel/findings.md
```diff
@@ -0,0 +1,461 @@
...
+maintainer, so the wrapper need not be ours at all.
+
+## 8. What the contrib package emits today — the compatibility baseline
+
+If we intend to replace the package (see `plan.md`), this is the inventory we have to match. Read
+off contrib `main` as of 2026-09-17, package version 1.18.0-beta.3.
+
+**Release history:** 48 versions since 0.3.0-beta.1, and **not one stable release**. Their README
+attributes this to the database semantic conventions still being Experimental, not to the code
+being immature.
```
martincostello, line 341: Yep, this is the only reason we haven't reached stable. Historically we were waiting on the _Database_ semantic conventions to stabilise, which has now happened. Now we're just waiting on Redis'.

It's also worth noting that because our version is prerelease, [Aspire ships its own vendored copy of the OTel Redis instrumentation](https://github.com/microsoft/aspire/tree/main/src/Vendoring/OpenTelemetry.Instrumentation.StackExchangeRedis), so that's another reason native OTel support here would be beneficial as that's another wider simplification that could be made.

notes/otel/findings.md
```diff
@@ -0,0 +1,461 @@
...
+
+**Activity identity**
+
+| | value |
+| --- | --- |
+| `ActivitySource.Name` | `OpenTelemetry.Instrumentation.StackExchangeRedis` (the assembly name, via `ActivitySourceFactory`) |
+| `ActivitySource.Version` | the package version |
+| `TelemetrySchemaUrl` | derived from the semconv version — 1.23.0 for the old source, 1.42.0 for the new, unset when emitting both |
```
martincostello, line 349: If telemetry goes native here, I would go with whatever the current version is at that point. Today it's [1.44.0](https://github.com/open-telemetry/semantic-conventions/releases/tag/v1.44.0).

notes/otel/findings.md
```diff
@@ -0,0 +1,461 @@
...
+and the script text. This is the reflection payload from §1.
+
+**Not emitted, at all:**
+
+- any error or failure information — no `error.type`, no `db.response.status_code`, and
+  `ActivityStatusCode` is never set. A Redis command that fails produces a span indistinguishable
+  from one that succeeded.
+- `db.operation.batch.size`
+- anything about retransmission (`// TODO: deal with the re-transmission`)
+- metrics of any kind — the package is traces only
```
martincostello, line 400: Having support for metrics here would definitely be a net win. I think we've avoided even trying to add it to the OTel instrumentation as it's probably a tonne of work, if we can even reliably generate the required metrics somehow. I imagine it would likely involve some less than ideal workarounds.

notes/otel/findings.md
```diff
@@ -0,0 +1,461 @@
...
+That is the entire package. Note the `options` parameter on the metrics one is accepted and
+ignored. The package exists for exactly one reason: those two extension methods need types from
+`OpenTelemetry`, and the core driver must not depend on `OpenTelemetry`.
+
+Two further observations worth carrying into our own design:
+
+- **They shipped `Filter` and `Enrich` in-box after all** — filters, enrichment callbacks *and*
+  span-name providers, for each of commands, batches and copy operations. That is direct
+  counter-evidence to the position that `ActivityListener.Sample` makes in-box filtering
+  unnecessary. Someone who has run this in production for years concluded otherwise.
```
martincostello, line 457: My _assumption_ is that the Filter and Enrich pattern is something that evolved for convenience over time, particularly when less instrumentation was built into .NET itself in previous releases (like with HTTP instrumentation).

notes/otel/findings.md
```diff
@@ -0,0 +1,461 @@
...
+- **`EnablePhysicalOpenTracing` and `EnableFirstResponseEvent` are opt-in, defaulting off** —
+  connection-establishment spans and extra timing events are not free enough to be on by default.
+  Contrib, by contrast, has its timing events **on** by default, which is a compatibility
+  constraint for us (§8) and not an endorsement.
```
martincostello, line 461: Typically we enable things marked as Required and Recommended (sometimes able to be opt-ed out of) by default. Opt-In functionality is not necessarily implemented at all, usually because it's difficult to generate the actual data in question.

notes/otel/plan.md
```diff
@@ -0,0 +1,513 @@
...
+are the .NET tracing and metrics API, and every collector (OpenTelemetry, Datadog, Application
+Insights, Elastic, New Relic) already listens to them.
+
+Consumer side, in full:
+
+```csharp
+builder.Services.AddOpenTelemetry()
+    .WithTracing(t => t.AddSource("StackExchange.Redis"))
+    .WithMetrics(m => m.AddMeter("StackExchange.Redis"));
+```
```
martincostello, line 21: It doesn't bother me, but we've had a non-zero number of issues over the years from people asking for an API (field, constant, property, whatever) to get this name from because they don't like hard-coding "magic" strings. That said, it's easy enough to do with `typeof(ConnectionMultiplexer).Assembly.Name` if that's your naming convention, so...

notes/otel/plan.md
```diff
@@ -0,0 +1,513 @@
...
+
+- `EmitQueryText` — whether to populate `db.query.text` / `db.statement`. Off by default because
+  keys can carry PII. Command **and key**; never argument values. Matches contrib's
+  `SetVerboseDatabaseStatements` and Lettuce's `includeCommandArgsInSpanTags`.
+
+## Subsuming the contrib package
+
+The intent is to **replace** `OpenTelemetry.Instrumentation.StackExchangeRedis`, not to sit
+alongside it. That needs agreement from its owners — code owner is @matt-hensley, with
+@martincostello active on it — and mgravell is raising it with them.
```
martincostello, line 69: As noted in a previous comment, I have no problem with our instrumentation being replaced by a native alternative.

notes/otel/plan.md
```diff
@@ -0,0 +1,513 @@
...
+
+- **Honour `OTEL_SEMCONV_STABILITY_OPT_IN`**, with the same three modes and the same default. This
+  is the single most important item, and the easiest to get wrong: the default today is `Old`, so
+  unopted users are seeing `db.system` and `db.statement`, *not* `db.system.name` and
+  `db.query.text`. Emitting only the current names would look like the modern, correct choice and
+  would silently break every existing consumer. Honouring the env var is also simply the correct
+  behaviour for any .NET library emitting database telemetry, and it gives a clean exit: when the
+  conventions go stable, our default flips in lockstep with the rest of the ecosystem rather than
+  on our own schedule. Honouring it does not mean it has to be the *only* way to choose — see open
+  question 5 on making the selection an explicit enum as well.
```
martincostello, line 98: As noted above, unless we want a very deliberate minimal-as-possible migration, I would avoid the tech debt you'd inherit by supporting this and go straight to implementing SemConv as of 1.44.0. That's however dependent on how you want to tackle Database-Stable vs. Redis-Development.

People would need to opt-in to the native instrumentation anyway, unless we go down the route of the instrumentation updating to be aware of what version is in use like we do in the HTTP instrumentation and automatically light-up the new native support on their behalf automatically. In that case we probably would want to keep the opt-in functionality.

notes/otel/plan.md
```diff
@@ -0,0 +1,513 @@
...
+  the reflection knows about.
+
+### Two things that cannot be preserved
+
+1. **The `ActivitySource` name changes.** Theirs is the assembly name,
+   `OpenTelemetry.Instrumentation.StackExchangeRedis`; ours will be `StackExchange.Redis`. We are
+   not squatting their name. Anyone calling `AddRedisInstrumentation()` sees nothing — the
+   extension just changes which source it adds — but anyone who hand-wrote
+   `AddSource("OpenTelemetry.Instrumentation.StackExchangeRedis")` has to edit one string. Document
+   it prominently; consider asking its maintainers to add *both* names for a transition window.
```
martincostello, line 131: As noted above, we can easily do version detection to light-up new functionality for this like we already do for HTTP and ASP.NET Core instrumentation.

notes/otel/plan.md
```diff
@@ -0,0 +1,513 @@
...
+2. **`Filter` and `Enrich`.** Callbacks on their options object with no obvious in-box equivalent.
+   `ActivityListener.Sample` is the better home for filtering — see open question 4.
```
martincostello, line 133: See previous comment.

notes/otel/plan.md
```diff
@@ -0,0 +1,513 @@
...
+Their code is Apache-2.0; this repo is MIT. Apache-2.0 is permissive and can be incorporated, but
+§4 has real obligations (retain notices, state changes, carry the license text for the derived
+portions) — so this is a decision to make on purpose, with their blessing, not a quiet copy-paste.
+
+The practical recommendation is to **use it as a specification rather than a source**: most of the
+implementation is machinery for the drain thread, the session cache and the baggage restoration,
+all of which we are deleting, and the rest is written against internals we will not have. What is
+genuinely worth having is (a) the exact attribute/event/default inventory, which is behaviour and
+is captured in findings §8, and (b) their **test suite**, which is the compatibility oracle and is
+the part where the licensing question actually bites. Ask about the tests explicitly.
```
martincostello, line 146: I'm not a lawyer, but I don't see there being a problem in using our test corpus/code.

notes/otel/plan.md
```diff
@@ -0,0 +1,513 @@
...
+2. **`ActivitySource` and `Meter` already are the indirection.** That is the entire point of them
+   living in the BCL. A second assembly binding over internals would be re-implementing
+   `ActivityListener`, worse and privately.
+3. **Cross-package IVT is lockstep with a runtime failure mode.** `StackExchange.Redis` 3.5 plus
+   `StackExchange.Redis.Telemetry` 3.2 is a perfectly resolvable NuGet graph that throws
+   `MissingMethodException` on first use. That is the same failure class as the reflection this
+   whole exercise is meant to retire — moved behind a compiler check, but not removed.
+4. **The reference is free where it matters.** `System.Diagnostics.DiagnosticSource` is in-box on
+   `net8.0` and `net10.0`; only `net472` and `netstandard2.0` gain an actual package reference, and
+   `net461` gains nothing because it compiles out.
```
martincostello, line 235: Just noting for completeness, in the core OpenTelemetry SDK we always reference the latest stable version of `System.Diagnostics.DiagnosticSource` because it contains the core metrics and tracing APIs (and bug fixes etc) we need, so for the end-user unless they're running .NET vLatest they'll always have a copy of the assembly with their app.

There's no need for you to do that here as you're not using the APIs the same way we are and really just need the API surface, but worth me mentioning anyway.

notes/otel/plan.md
```diff
@@ -0,0 +1,513 @@
...
+
+**This is a repo-level decision that telemetry merely surfaced, not a telemetry decision** — it
+affects the whole package and wants deciding on its own terms, probably alongside v4. Recorded here
+because this is where the evidence turned up. If net461 stays for reasons outside this topic, the
+`TELEMETRY` symbol is the fallback and costs net461 users nothing they have today; the profiling
+API stays on every target regardless.
+
+## Pay-for-play
+
+The cost when nobody is listening must be a predictable, tiny constant:
```
martincostello, line 301: We've started an initiative in OTel to start benchmarking scenarios officially, but it's early days: [OTEP-5109](https://github.com/open-telemetry/opentelemetry-specification/pull/5118).

These live in [open-telemetry/opentelemetry-benchmarks](https://github.com/open-telemetry/opentelemetry-benchmarks).

The only scenario we have implemented today is the _"near zero overhead to do nothing"_ scenario for metrics.

notes/otel/plan.md
```diff
@@ -0,0 +1,513 @@
...
+
+## Sequencing
+
+Phases are ordered by risk and by how much of the design they commit us to. Each is independently
+shippable.
+
+### Phase 1 — unblock the reflection, now
+
+Expose the statement string so the contrib package can stop emitting `DynamicMethod` field getters
+against our privates, on 3.x, today. This is worth doing *whatever* we decide about the rest:
```
martincostello, line 341: Phrasing 😄

notes/otel/plan.md
```diff
@@ -0,0 +1,513 @@
...
+the contrib package has to support 3.x for years regardless.
```
martincostello, line 342: I'm not sure if we would support 3.x for _years_ if a new version produces better/faster instrumentation.

notes/otel/plan.md
```diff
@@ -0,0 +1,513 @@
...
+
+1. **Who owns semconv churn?** Contrib ships three `ActivitySource`s to straddle 1.23 vs 1.42, and
+   has never shipped a stable version in six years *because* the conventions are experimental. If
+   we emit natively, that becomes our problem — and StackExchange.Redis does not have the option of
+   shipping perpetual betas. Do we mark the telemetry `[Experimental]` (we already have the
+   `SER00x` machinery in `src/RESPite/Shared/Experiments.cs`), or do we accept that our stable
+   package emits attributes that may be renamed under us? This is the strongest argument for the
+   middle ground and we should hear it argued before dismissing it. **See question 5 for a way to
+   make this much less binary** — gating individual enum members rather than the whole feature, so
+   only the callers who opted into unstable conventions carry the diagnostic.
```
martincostello, line 425: See previous comments regarding keeping/dropping the opt-in and the AWS precedent.

notes/otel/plan.md
```diff
@@ -0,0 +1,513 @@
...
+2. **Do its maintainers want the package to become a one-liner, or to keep producing spans?** If we
+   go native, does it retire, or keep an `AddRedisInstrumentation()` that calls `AddSource` plus the old
+   `Filter`/`Enrich` knobs? Those two callbacks are the only features that do not obviously survive
+   the transition. Either way, ask them to add our source name — and ideally to keep adding their
+   own for a window, so existing hand-written `AddSource` calls do not break.
```
martincostello, line 430: I think we would want one of:

1. The package becomes entirely obsolete
2. We configure the new source for users on `vSupportsOTel` of StackExchange.Redis and keep the enrich/filter sugar on top. We keep the tracing infrastructure for users not using the newer version and we continue to compile against v3.
3. We update our dependency to `vSupportsOTel` and then just have the sugar and source configuration convenience.

notes/otel/plan.md
```diff
@@ -0,0 +1,513 @@
...
+3. **Can we have the tests?** Their test suite is the compatibility oracle, and it is Apache-2.0
+   going into an MIT repo. Explicit blessing, an agreed attribution form, or a clean-room rewrite
+   from the behaviour inventory — but decided up front, not discovered in review.
```
martincostello, line 433: I don't have any specific objection - as long as whatever is done doesn't violate whatever the two licenses would require as-written.

notes/otel/plan.md
```diff
@@ -0,0 +1,513 @@
...
+4. **`Filter` and `Enrich` equivalents.** The case against is that `ActivityListener.Sample` already
+   does this at the collector — which answers Nick's 2022 objection that *deciding not to profile*
+   costs something per command. The case for is that both existing implementations shipped them
+   anyway: Npgsql has filters, enrichment callbacks and span-name providers for commands, batches
+   and copy operations, which is the bulk of their public surface (findings §9), and the contrib
+   package has `Filter`/`Enrich`. If we adopt them, what gets handed to the callback has to be
+   settled first — see "Public API impact".
```
martincostello, line 440: Some answers to this already made, but the filter/enrich functionality is typically to do additional dynamic filtering/enrichment if telemetry has been turned on in the first place. It isn't intended to be _the_ mechanism, so if you never enabled tracing the code that would call these callbacks would never be invoked anyway.

notes/otel/plan.md
```diff
@@ -0,0 +1,513 @@
...
+5. **Should the semconv choice be an explicit enum rather than only an environment variable?**
```
martincostello, line 441: I would avoid the opt-in mechanism (as that's intended to cover the pre-stable database changes since v1.23.0) and instead go with a mechanism like AWS'.

notes/otel/plan.md
```diff
@@ -0,0 +1,513 @@
...
+   `OTEL_SEMCONV_STABILITY_OPT_IN` is ambient, process-wide, and read once at construction — a
```
martincostello, line 442: If you _did_ want to support old vs. new, then `OTEL_SEMCONV_STABILITY_OPT_IN` needs to be honoured even if there's a first-class public API surface to configure it, as that's a requirement of the Semantic Conventions.

notes/otel/plan.md
```diff
@@ -0,0 +1,513 @@
...
+   }
+   ```
+
+   The stable half — `Default` and `Old`, which is what everyone gets today — stays plain stable
+   API. Only the members that actually track still-experimental conventions carry the gate, so the
+   churn risk is scoped to the people who opted into churn. Marking the *whole* telemetry feature
+   `[Experimental]` is all-or-nothing and puts a diagnostic in front of people who only ever wanted
+   the default behaviour.
+
+   And the exit is clean: when the database conventions go stable, delete the attribute from the
```
martincostello, line 488: Nit: database conventions are stable - it's Redis' that aren't.

notes/otel/plan.md
```diff
@@ -0,0 +1,513 @@
...
+   member. That is neither a source nor a binary break — the diagnostic simply stops firing.
+
+   Repo convention is `[Experimental(Experiments.X, UrlFormat = Experiments.UrlFormat)]` with a
+   `docs/exp/SERxxx.md` page; `SER001`–`SER009` are taken (retired IDs stay reserved), so this would
+   be `SER010`.
+6. **Version/schema pinning.** Should the `ActivitySource` version track the package version, or a
+   semconv schema version (contrib uses the latter via `ActivitySourceFactory.Create<T>(version)`)?
```
martincostello, line 495: Should track SemConv as that's what you're implementing - it would be your version if you had custom extra conventions just for StackExchange.Redis.

notes/otel/plan.md
```diff
@@ -0,0 +1,513 @@
...
+7. **Does this need to wait for v4?** The IO core rewrite moves all five hook sites. Hooks placed
+   now survive as *concepts* but not as code. Landing metrics on 3.x and traces on 4.x is a
+   defensible split; so is landing both on 3.x and accepting the port.
```
martincostello, line 498: v3 vs. v4 is probably dependent on the ultimate agreed plan. v4 gives a clear cut-over, but incremental/gradual makes it more available to end-users.

martincostello (commented):
Thanks for putting this together ❤️

notes/otel/plan.md
matt-hensley: I'd be happy to see instrumentation moving into the driver, and more than willing to contribute to the implementation. I only took over ownership because the package wasn't being actively maintained and had become a candidate to stop supporting.

Moving to some form of native instrumentation is only a positive IMO.

notes/otel/plan.md
```diff
@@ -0,0 +1,513 @@
...
+- **Retransmission.** `-MOVED` / `-ASK` resends, which the contrib converter carries a literal
+  `// TODO: deal with the re-transmission` for. `PrepareToResend` is ours; a child span or an
+  `ActivityLink` back to the original is straightforward from the inside.
+- **Batches and transactions.** One span with `db.operation.batch.size`, name prefixed `MULTI` or
+  `PIPELINE` per semconv. Both go-redis (`redis.pipeline`, `db.redis.num_cmd`) and redis-py
+  (`db.redis.pipeline_length`) collapse a pipeline to one span; we should too.
+- **Connection establishment.** go-redis traces `redis.dial`; we have far richer material —
+  handshake, discovery, failover, sentinel — and it is low-volume enough to trace unconditionally.
+
+### Phase 5 — pub/sub
```
matt-hensley, line 407: Messaging semconv has been mostly dormant since 2024, currently working on reviving it, this can be deferred for a long while.

notes/otel/plan.md
matt-hensley: If 4.x is basically drop-in plus the new features outlined in #3221, I think at least a pseudo-gradual path would be appreciated by end-users. A lot of the SREs I speak with struggle to get dev time from product teams - a minor bump of a package can take several quarters. If we can make the upgrade smooth it has a much better chance of being adopted.

notes/otel/findings.md
matt-hensley: Often stabilization is slow because of a lack of subject matter experts. Working to implement instrumentation directly in a driver would be a very compelling reason to finalize Redis-specific attributes. 

Removing the need for vendoring would be good for the community overall.

PR #3246, opened 2026-09-18
Retire SER001: vector sets are not experimental any more

mgravell:
`SER001` was already dead — no constant in `Experiments.cs`, no `[Experimental]` applying it anywhere in `src`/`tests`/`toys`, and none of the **94 public API entries** for vector sets gated by it. But it was still suppressed in `NoWarn` and its docs page still described the feature in the present tense as subject to change.

## Changes

- **`Directory.Build.props`** — `SER001` dropped from `NoWarn`; the neighbouring comment now lists it with the other retired IDs.
- **`docs/exp/SER001.md`** — kept, so the diagnostic's help link and any existing bookmarks still resolve. Rewritten in the past tense: it leads with the fact that the diagnostic is no longer produced and that an existing suppression is inert and can be removed whenever convenient, then retains what it used to mean below, since somebody arriving from an older build's warning still needs that.
- **`Experiments.cs`** — `SER001 = vector sets` added to the reserved-IDs list beside SER002/3/6.

## Why the reserved list matters

That comment is what stops a number being recycled. A recycled `SER001` would be silently suppressed by every stale `NoWarn` still carrying it — a new experimental API shipping with its warning invisible, which is precisely the failure the convention exists to prevent. This was the only retired ID missing from the list.

## Verification

Removing the suppression is its own check: if anything still emitted `SER001`, the build would now fail, since warnings are errors here. Clean full-solution build, plus `/p:RunAnalyzers=true` on the library so the analyzers run on every target framework rather than just the newest. 0 warnings, 0 errors.

## Not changed

Vector set APIs themselves are untouched — they ship ungated today and this does not change that. If they were ever meant to stay gated, this PR is the wrong fix and the right one is to apply `[Experimental]` to them instead.


mgravell (commented):
👍

PR #3247, opened 2026-09-18
Promote geo-redundant failover and the Redis 8.10 bindings out of preview

mgravell:
Retires **SER007** (client-side geographic failover) and **SER008** (Redis 8.10 features). SER007's own docs page promised exactly this — *"the marker is expected to be short-lived… the diagnostic simply stops being emitted, with no code change needed on your side"* — so this delivers it.

## What changed

| | |
|---|---|
| **38** `[Experimental]` attributes removed | across 21 files, all the identical single-line form |
| **247** public API entries de-annotated | 240 `[SER007]`, 7 `[SER008]` |
| `Experiments.cs` | both constants removed, both added to the reserved-IDs list beside SER002/3/6 |
| `Directory.Build.props` | both IDs out of `NoWarn` |
| `docs/exp/SER007.md`, `SER008.md` | kept, rewritten in the past tense |
| `docs/Configuration.md` | circuit-breaker and health-check options no longer labelled Experimental |

The API-file churn is the bulk of the diff and is **purely prefix removal** — verified by stripping the annotation from the removed lines and diffing against the added ones: identical. The files stay sorted, because `eng/public-api.py` sorts on a key that already ignores the annotation.

Doc pages are kept rather than deleted, because the diagnostics' help URLs point at them and older builds still emit the IDs. Each now leads with "no longer produced, an existing suppression is inert", and retains what it used to mean below for anyone arriving from an old build's warning.

## Verification

Removing the suppressions is its own check: warnings are errors here, so anything still emitting either ID would now fail the build.

- clean full-solution build
- `/p:RunAnalyzers=true` on the library, so analyzers run on every target framework rather than just the newest
- 499 tests green across the config, health-check, multi-group and retry areas

## Two things worth your eye

1. **SER008's premise was "Redis 8.10 is currently in preview".** Promoting it asserts that is no longer true. I have taken that from the request rather than verified it — if 8.10 is still in preview, this half should wait.
2. **This conflicts with #3191**, which uses `Experiments.GeoRedundantFailover` in its own new code and carries `[SER007]` API entries. After this merges, that branch needs the same treatment for the sites it adds — happy to do that pass once this lands.

Stacked on the same reasoning as #3246 (SER001); that one can merge first or independently, they touch different IDs.


mgravell (commented):
👍

Issue #2826, opened 2024-12-18
Cluster: support host names when available, via CLUSTER SLOTS


mgravell:
Situation:

- especially from v7, cluster shards may be routed via hosts instead of IPs
- thus, the advertised IPs may sometimes (with known examples) may not be routable
- SLOTS and SHARDS (v7+) provide the hosts; NODES does not, as far as known

Proposal:

- prefer SLOTS to NODES (but not SHARDS due to v7 dependency)
- route via hostname when available, IP otherwise
- possibly with option to use fallback old behaviour?

---

Was: "Lookup by endpoint should check IP and host"

Currently, only exact endpoint matches are considered. However, specially in the case of CLUSTER, a node may have both host and IP evidently. Further, `-MOVED` may return the unexpected option, leading to additional connections.

We should:

- record both identities as declared, for example from the `CLUSTER NODES` response
- if no equality endpoint match is found, check also using the best data available (after type-testing the endpoint)

Context: https://github.com/dotnet/aspnetcore/issues/59211

Auxiliary consideration: should cluster routing be limited to IP, as current? Should there be a "use host routing" option? Or should we try IP and use host if that fails?

CaringDev:
Additional complexity for bonus points: (part of) the cluster might be hidden behind a proxy / load balancer requiring SNI, i.e. all (some) hostnames resolve to the same IP.

mgravell:
Side note: some of this may already be shimmable by individual callers, if they already know how they want things to connect: the `Tunnel` API allows callers to override how connections are established. This would, in theory, allow custom client-side code to take a requested endpoint and do *whatever it needs*, as long as it can create a `Stream`. I've seen this use proxy servers, custom DNS-like things, you name it.

CaringDev:
I got a response from the team providing "redis as a service"... the summary is as follows:
> The advertised hosts when receiving a MOVED response are correct [...]
> the library should not use IPs from “CLUSTER NODES” responses for its own routing. We explicitly configure internal IPs for the Redis routing so that the sharding and replication does not go out the OpenShift network. We then expose different routing for client libraries as these need to be routable from outside the network. [...]
> The documentation [...] states “Note that normally clients willing to fetch the map between Cluster hash slots and node addresses should use CLUSTER SLOTS instead.” The output of that command provides the hostname as first routing priority, and returns correct hostnames [...]
> We will not be able to make the IPs routable from outside the OpenShift cluster

[...]

PR #3250, opened 2026-09-22
Distinguish explicit and inferred TLS host names useful when cluster shards exposed using TLS/SNI

sandeepkunusoth:
Fixes TLS SNI selection for hostname-routed clusters from a single DNS endpoint.

For clusters behind a shared load balancer that routes by SNI, every discovered node consequently reused the initial endpoint’s SNI name and could be routed to the wrong backend.

this change distinguishes an explicitly configured `SslHost` from an inferred fallback and applies this precedence:

1. An explicitly configured `SslHost` always wins.
2. Otherwise, a `DnsEndPoint` uses its own `Host`.
3. Otherwise, the default provider’s `GetSslHostFromEndpoints(EndPoints)` value is used.
4. If none is available, the endpoint address is used.

This preserves existing behavior for callers that deliberately configure `SslHost`, while allowing each discovered DNS cluster endpoint to present its own SNI name.

The resolution logic is centralized and applied consistently to:

- The standard physical connection TLS path
- `LoggingTunnel`
- `TlsOptions.ResolveHost`

`ConfigurationOptions.SslHost` now reports only an explicitly configured value rather than an inferred one.

Related to #2826, specifically:
https://github.com/StackExchange/StackExchange.Redis/issues/2826#issuecomment-5772196088

## Testing

Added and updated coverage for:

- Explicit `SslHost` overriding a DNS endpoint
- A DNS endpoint taking precedence over an inferred hostname
- An IP endpoint using an explicit `SslHost`
- An IP endpoint using the inferred fallback
- An IP endpoint falling back to its address
- Matching behavior through `TlsOptions.ResolveHost`
- TLS cluster connections using hostname-only certificates

## Testing

Validated against a hostname-routed Valkey cluster using a single seed hostname, shared VIP, and per-node SNI routing.

## Checklist

- [X] I fully and freely contribute this code in accordance with the [project license](../LICENSE) (and am legally able to do so)
- [X] I take responsibility for this contribution's quality and correctness, including any portions produced with AI assistance (see [CONTRIBUTING.md](../CONTRIBUTING.md)).


mgravell:
I'm concerned that as written this is a break to many scenarios - it *stops* using the sslhost as an override in some DNS scenarios that work today.

Question: are you actually specifying the sslhost in your scenario?

mgravell:
I wonder if the real fix here is for the library to never infer sslhost on your behalf, so it only applies if the caller means it, I.e. change this:

```
get => sslHost ?? Defaults.GetSslHostFromEndpoints(EndPoints);
```

So:

- SslHost always wins if specified
- otherwise, for DNS endpoint: use .Host
- otherwise, use `Defaults.GetSslHostFromEndpoints(EndPoints)`

sandeepkunusoth:
Thanks for quick reply.
> I'm concerned that as written this is a break to many scenarios - it _stops_ using the sslhost as an override in some DNS scenarios that work today.
> 
> Question: are you actually specifying the sslhost in your scenario?

No, in our case we are not explicitly setting SslHost. It is being inferred from the single seed endpoint via:
`get => sslHost ?? Defaults.GetSslHostFromEndpoints(EndPoints);`
That inferred value is then effectively treated like an explicit override for the DNS endpoints discovered from CLUSTER SLOTS, which is what causes the wrong SNI to be reused across shards.


> I wonder if the real fix here is for the library to never infer sslhost on your behalf, so it only applies if the caller means it, I.e. change this:

I agree your suggested direction is safer and preserves the existing override behavior:

explicit SslHost always wins
otherwise, for a DnsEndPoint, use DnsEndPoint.Host
otherwise, fall back to Defaults.GetSslHostFromEndpoints(EndPoints)

That should solve our single-dns cluster case without changing behavior for callers that intentionally set SslHost.

I can update the PR in that direction and adjust the tests accordingly.

mgravell:
> Retrigger CI

I was about to do that from this side ;p Yes, there's some annoying flakes, but if you let me deal with that: I can do it more granularly

Changes look good, thanks

sandeepkunusoth:
> > Retrigger CI
> 
> I was about to do that from this side ;p Yes, there's some annoying flakes, but if you let me deal with that: I can do it more granularly
> 
> Changes look good, thanks

please feel freee to update anything if u want.

mgravell (approved):

sandeepkunusoth:
Hi @mgravell, thanks for merging this! Could you let me know approximately when the next version containing this fix is expected to be released?

Also, is there a way to use a build/tag from main in a .NET project in the meantime, so we can start using and testing the fix before the official release?

mgravell:
I don't object to doing a build for it, which would be 3.3.1 - however, GitHub is failing to load /releases for me, which makes that hard

PR #3253, opened 2026-09-23
Make RedisValue numeric comparison consistent across storage types

### LINKED ISSUE #3233: RedisValue equality treats "1,000", "(5)" and " 5 " as numbers, via NumberStyles.Any
Raised during review of #3230, where the numeric side of `RedisValue` equality came up. Pre-existing
behaviour, not caused by that PR.

## What happens

`operator ==` runs `Simplify()` on both sides, so text that parses as a number is compared as a number.
`Format.cs:178` does that parse with `NumberStyles.Any`:

```csharp
return double.TryParse(s, NumberStyles.Any, NumberFormatInfo.InvariantInfo, out value);
```

`NumberStyles.Any` allows thousands separators, parentheses for negatives, leading and trailing whitespace, a
currency symbol, and exponents. So:

| comparison | result | storage |
|---|---|---|
| `"1,000" == "1000"` | **True** | String / String |
| `"(5)" == "-5"` | **True** | String / String |
| `" 5 " == "5"` | **True** | String / String |
| `"1,0,0,0" == "1000"` | **True** | String / String |
| `"5." == "5"` | True | String / String |
| `"+5" == "5"` | True | String / String |
| `"1e2" == "100"` | True | String / String |

Both sides are `String`-typed throughout — these are ordinary string values, not numbers that happen to be
spelled differently.

Every integer path in `Format.cs` uses the much tighter `NumberStyles.Integer`; `TryParseDouble` is the only
one using `Any`, which makes it look more like an oversight than a decision.

## Why it matters

These are **distinct values on the server** — distinct set members, distinct hash fields, distinct keys — that
compare equal in the client. Anything deduplicating `RedisValue`s in a `HashSet<RedisValue>`, or keying a
`Dictionary<RedisValue, T>`, silently merges them. `" 5 "` and `"5"` are plainly different values by any
reading.

It is at least *self-consistent*: `GetHashCode` simplifies too, so equal values do share a hash and no hash
container is corrupted. The problem is that the equality is much wider than "numeric text".

## Possible direction

`NumberStyles.Float` (leading/trailing whitespace, leading sign, decimal point, exponent) would drop the
separators, parentheses and currency symbol while keeping what Redis actually emits for numbers. Even that
leaves `" 5 " == "5"` true and `"1e2" == "100"` true, so the whitespace and exponent allowances are worth a
separate decision.

Any change here alters shipped comparison behaviour, so it wants its own discussion rather than being folded
into unrelated work — hence this issue rather than a PR.

banlor:
`RedisValue` treats strings such as `1,000` as numbers but keeps the same UTF-8 bytes as text. This uses the existing byte parser for equality, hashing and ordering across storage types, without changing explicit numeric conversions or the binary comparer.

The tests cover storage parity, dictionary/set lookups, long numeric strings and conversion compatibility. Nonnumeric characters are rejected before renting a buffer for long strings or fragmented sequences; valid numeric text still has no length cutoff.

`dotnet build Build.csproj -c Release /p:CI=true` passed for all five library targets. The selected net10.0 value/parser tests passed (163 tests). Tests on net8.0/full framework and Redis server integration were not run.

Fixes #3233.

## Checklist


mgravell:
In the PR body (so: here) can we please be very clear and explicit exactly what scenario(s) this is addressing? Specifically, any change is a potential break to *someone*, so it would be good to be very clear to see what we're looking to change here. In particular, i wonder whether what we *actually* want here is an equality comparer instance, like `Binary`, but with some preferred and well-defined semantics.

mgravell:
If the point is to defer any *lossy* conversions until compare time: that's great, I just think we should be super clear about what actually changes in the PR body.

src/StackExchange.Redis/RedisValue.cs
```diff
@@ -1593,15 +1624,41 @@ internal RedisValue Simplify()
...
+                        {
+                            foreach (byte b in segment.Span)
+                            {
+                                if (!CouldBeNumberCharacter((char)b)) return this;
+                            }
+                        }
+                    }
+                    if (forComparison || seq.Length <= Format.MaxDoubleTextLen)
+                    {
+                        int len = checked((int)seq.Length);
```
mgravell: I think we need to be mindful of what scenarios we need to allow/disallow here, and what means for the inputs; for example, if we can't simplify something that is over-long, it seems unlikely that we'd ever need the lease here - we can presumably discount such values immediately. So: what `forComparison` scenario forces this lease? Something with pathological numbers of leading zeros... probably isn't a useful case. I also wonder about `[+-]inf` etc.

banlor: The lease preserves numeric comparison for overlong inputs, including leading zeroes; it isn't needed for inf, +inf, -inf or nan, which remain text for equality even though double casts can parse them. I tested removing it with a uniform 40-byte cutoff, but that also breaks existing long byte-array comparisons, so I haven't pushed that change. The body now gives the round-trip mismatch and breaking changes explicitly: would you prefer the bounded default behavior or an opt-in comparer?

mgravell: Again, it is hard for me to opine because the PR body doesn't really make it clear what it *intends* to change, behaviour-wise, in concrete terms / examples.

mgravell:
I wonder if we can simplify the intent here. For example, what if the non-comparing simplify *only* handled strict integers (no commas etc) and floating points that parse and round-trip back to the exact same bytes/chars, so it is never lossy, and the compare parse does what it has always done. Does that solve the underlying issue more simply? Is that what this PR is trying to do?

